PR #247, opened 2014-02-11
Added check for parameter name in GuardClauseAssertion.

valmaev:
This pull request added new logic to NullReferenceBehaviorExpectation's Verify method. Now it verifies that command throws ArgumentNullException with proper parameter name. Inspired by [this]( http://stackoverflow.com/questions/15078400/autofixture-test-for-invalid-constructor-parameter/15083283]) question on StackOverflow.

This pull request contains breaking change for IGuardClauseCommand interface.


ploeh:
Opening in order to see if I can pull the commits into the *v4* branch.

ploeh:
Thank you for your contribution. It's now merged into the *v4* branch of AutoFixture.

When AutoFixture 4 is going to be be released, I can't currently say, but at least this means that you don't have to keep your fork around.

valmaev:
That's great! Thanks!

Issue #17, opened 2012-10-12
Ensuring Disposable of objects created from Customizations


JohannesRudolph:
I am writing tests for a Web API project. I want AutoFixture to provide a pre-configured HttpConfiguration fixture to my tests and properly dispose of it when my test has executed (with either outcome).

I'm using the following code:

``` c#
        [AttributeUsage( AttributeTargets.All, Inherited = false, AllowMultiple = true )]
        sealed class AutoHttpConfigurationAttribute : AutoDataAttribute
        {
            public AutoHttpConfigurationAttribute()
            {
                Fixture.Customize( new DisposableTrackingCustomization() );
                Fixture.Customize( new HttpConfigurationCustomization() );
            }

            private class HttpConfigurationCustomization : ICustomization
            {
                public void Customize( IFixture fixture )
                {
                    fixture.Register<HttpConfiguration>( CreateInitializedConfiguration );

                }

                private HttpConfiguration CreateInitializedConfiguration()
                {
                    var configuration = new HttpConfiguration();
 ...
                    return configuration;
                }
            }
        }
```

However, it seems that the HttpConfiguration does not get disposed properly when using the attribute with a test case like this

``` c#
            [Theory, AutoHttpConfiguration]
            public void TestWithConfiguration( HttpConfiguration sut )
```

Am I doing something wrong? How do I achieve automatic disposal of a fixture supplied by a customization? Please mark this issue as a "question".

PR #294, opened 2014-06-15
Xunit2

TeaDrivenDev:
Although/because xUnit 2 is still in beta, I'll put this up for review and discussion.

I created a copy of the `AutoFixture.Xunit.net` project (and the test project) and modified it to work with the current beta version of xUnit 2. It has exactly the same capabilities as the existing xUnit integration. The build process has been adjusted to allow for running xUnit 2 tests.

Although it cannot currently be merged, this is fully armed and operational with all tests passing, and has been verified on a real-world project that has used `AutoFixture.Xunit` for over a year (but does not use every feature).

### Points of note: ###

* To my best knowledge, I have only made the minimum necessary changes to the original code. The purpose was solely to allow using the existing functionality of the xUnit integration with v2. The copy operation and the subsequent adjustments are in separate commits, so the changes can be verified.

* As this has a dependency on a pre-release package, NuGet refuses to build a "stable" .nupkg. Therefore, the build script currently adds "-beta2" to the version of all created NuGet packages.

* Because xUnit 2 uses .NET 4.5, both AutoFixture.Xunit.net2 and the test project had to be set to 4.5 as well.

* As there is a NuGet package for the xUnit 2 runners, `Lib\XunitRunnerMSBuild` folder has been removed and the `xunit.runners` package added via NuGet. I have opted to leave out the version from the package directory name to allow for updating the runner without having to adjust the build script. The xUnit 2 runner is used to run both the existing v1 as well as the new v2 tests.

* `ClassDataAttribute` has been removed from xUnit v2; to keep the few `AutoFixture.Xunit` tests that used it intact, the respective xUnit 1 class has been copied to the test project.

* `InlineAutoDataAttribute` made use of `InlineDataAttribute` as the first element of a `CompositeDataAttribute`; this no longer works, as xUnit 2 completely changes the way values are retrieved from `InlineDataAttribute`. I have tried to change `CompositeDataAttribute` to follow the new pattern, but didn't find a way to accomplish that. **As a crutch**, I have introduced `ProxyInlineDataAttribute` as a copy of the original `InlineDataAttribute` from xUnit 1, so that I only needed to slightly modify `InlineAutoDataAttribute` to use it with xUnit 2. The type is public because a few tests need it, but has no public constructor. This solution currently works, but is certainly not very idiomatic in regard to the xUnit 2 API. Maybe @bradwilson can point to a more elegant way to handle this.


ploeh:
@TeaDrivenDev, prompted by the discussion we had regarding #335, I was wondering if it makes sense to move on with this. I must admit that I have yet to look at what's going on with xUnit.net 2. In what state is this PR? Does it target the most recent version of xUnit.net 2?

Would it be worthwhile to start publishing an AutoFixture.Xunit2 NuGet package, with a dependency on a specific xUnit.net 2 beta version?

shiftkey:
@ploeh FWIW I've worked with upgrading a couple of non-trivial test suites over to `2.0.0-beta5-build2785` and am happy with the perf boosts :grin:

TeaDrivenDev:
@ploeh The overall state is the same as described in my original post - AutoFixture.Xunit2 is AutoFixture.Xunit with the minimum changes required to support the new version. The only updates that have happened since then have been basing it on AutoFixture 3.19.2 instead of 3.19.0, and targeting xUnit 2 beta 4 instead of beta 2. All the "points of note" stated in the original post are still relevant; I never got around to specifically bugging @bradwilson about that one API question.

To my best knowledge, this is functionally complete, and I had no issues on the one (small-ish) project I tested it on. It has not been updated to the most recent xUnit release, which was beta 5, and it has not been brought up-to-date with AutoFixture development since August.

Releasing this (after doing the aforementioned updates) at this point would require it to be some kind of "pre-release" or "beta" package; this, as mentioned, is mandated by NuGet which does not allow creating "stable" packages that depend on pre-release ones. This, of course, should only apply to AutoFixture.Xunit2, not the "established" packages, and as far as I remember, the build process is currently not capable of accommodating that.

@shiftkey What changes did you have to make to be able to use beta 5?

As for xUnit 2, a release candidate was announced on Twitter in mid-December for "early January", but as far as I could see, that has not shown up yet.

shiftkey:
@TeaDrivenDev most of the work is covered in this guide http://xunit.github.io/docs/test-migration.html - depends on how complex your tests are

shiftkey:
And they just announced an RC http://xunit.github.io/release-notes/2015-01-20.html :tada:

bradwilson:
Let us know if there are any questions or issues.

TeaDrivenDev:
As the official release of xUnit 2 appears to be imminent, I have just taken the time to update the pull request to the current RC 4.

Changes made:
- Updated NuGet.exe (required to install xUnit MSBuild runner package)
- Updated xUnit 2 from Beta 4 to RC 4, including replacing the xunit.runners package with xunit.runner.msbuild

I was going to also merge the current master branch, but I am now stuck in NuGet/MSBuild hell and cannot get anything to build anymore. That means this PR has the AutoFixture projects somewhere around v3.19.

adamchester:
What sort of issues did you have with "NuGet/MSBuild hell"?

I often have to manually go into the .csproj files and edit the hint path when upgrading packages. Drives me insane.

TeaDrivenDev:
Apparently the runner package wasn't correctly installed at some point; reinstalling fixed it.

xUnit 2 went "gold" yesterday; I have updated the pull request to the final release. I have also merged master and bumped the version number to 3.25 (although in hindsight, that wasn't quite my call to make).

sergeyshushlyapin:
+1 for merging the PR.  
This will solve the ["Ensuring Disposable of objects created from Customizations #17"](https://github.com/AutoFixture/AutoFixture/issues/17) issue,

moodmosaic:
IMO, we could consider dividing the work into separate, smaller, PRs. – For example:

- [ ] Add a new solution similar to *AutoFixture.xUnit.net.sln* but for xUnit.net 2.0.
- [ ] Install xUnit.net 2.0 via NuGet. – The diff output will be big since the packages are checked-in.
- [ ] Add a very basic test in order to add the new project to the *BuildRelease.msbuild* script.
- [ ] Add AutoDataAttribute.
- [ ] Add CompositeDataAttribute.
- [ ] Add CustomizeAttribute.
- [ ] Add FavorArraysAttribute.
- [ ] Add FavorEnumerablesAttribute.
- [ ] Add FavorListsAttribute.
- [ ] Add FrozenAttribute.
- [ ] Add GreedyAttribute.
- [ ] Add InfiniteDataAttribute.
- [ ] Add InlineAutoDataAttribute.
- [ ] Add ModestAttribute.
- [ ] Add NoAutoPropertiesAttribute.

I haven't thoroughly analyzed this, but this should be probably faster than reviewing a big PR.

ploeh:
Now that xUnit.net 2.0 is finally out, like everyone else, I'm keen to be able to release *AutoFixture.Xunit2*  as soon as possible. Like @moodmosaic, I think that the *fastest* way to do that would be to split this into a sequence of smaller pull requests. [Here are some reasons for this](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests) - particularly the reasons outlined in point 1 and 2 of that article.

@TeaDrivenDev, if you don't want to split up your pull request, I understand, and since I still appreciate the effort you've already put into this, I'll get around to it eventually, but I'll have to set aside quite a large chunk of time to do that, and that's not so easy for me to do these days.

Another decision we should consider is this: should we upgrade *all* the AutoFixture unit tests to xUnit.net 2? If we do that, will we still be able to run the tests for *AutoFixture.Xunit*? Can we have tests with both xUnit.net 1 and 2 side-by-side? If @moodmosaic, @adamchester, and @ecampidoglio have comments to this question, I'd like to hear them :smile:

moodmosaic:
If we've upgraded the tests to xUnit.net 2.0 it wouldn't be a problem since AutoFixture unit tests are using xUnit.net 1.9.2. – According to [xUnit.net 2.0 release notes](http://xunit.github.io/release-notes/2015-03-16.html), it is backward compatible with version 1.9.2:

>we will continue to support 1.9.2 through our new 2.0 runners, so your upgrade path is incremental. We have no plans to release any fixes for 1.9.2 unless they are complete showstoppers, but given its age, we don't expect that to ever happen.

So, it looks like we can have tests using both xUnit.net 1.9.2 and xUnit.net 2.0 running side-by-side, as long as we upgrade the runner and the build automation script(s).

ploeh:
Does this mean that the first step should be to upgrade the runners?

That's how I read this.

TeaDrivenDev:
Yes. Upgrading the runner can be done completely independently, but is required as soon as xUnit 2 tests are introduced (which in turn is necessary for testing the xUnit 2 glue library).

To clarify, what @moodmosaic described is exactly the current state in the PR - none of the tests in any of the existing projects have been touched; all of those still use xUnit 1.9.2, but **all** xUnit tests in **all** tests projects are now handled by the v2 runner.
Upgrading the existing tests to v2 seems to be more of a maintenance task that can then be done at any later point.

As for splitting up the PR, I understand how that would make the review process easier, but I'd have to look at how I go about it, because it's entirely unlike the blanket "out with the old, in with the new, and fix what breaks" approach I used for doing the actual port.

TeaDrivenDev:
One more question on how this would work in practice - how would I make individual pull requests that are based on each other? I can't submit the basic xUnit 2 solution PR without the test runner PR, can't add code to the solution without the solution PR etc.

Waiting until each PR is merged and then branching again for the next one seems like an unreasonable amount of work.

ploeh:
@TeaDrivenDev, you don't have to split it up, if you don't want to. Just let me know which way you want to go. I *do* appreciate the work you've already put into this, and I don't want to see it go to waste for what may look like 'bureaucratic' reasons.

In that light, I can completely understand if you'd be most inclined to just let the current PR sit here, and wait until it gets pulled. One can get fed up with things one already considers done.

This is most likely the approach that'll involve the least amount of work for you, but it's not necessarily the strategy that will get this feature released *the fastest*.

The reasons for this I've [already outlined in some depth](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests), but to make it more concrete:

- This PR consists of *96 changed files, with 7,307 additions and 1,244 deletions*, and while I realise that some of those files are NuGet package files, that's still a lot to review. In order to do so, I'll need to set aside a *contiguous* block of time for at least a couple of hours. Since I also have a day job, that's not particularly easy for me to do, so that alone means that it may take days, or even weeks, before I can find the time to even start looking at this.
- "Waiting until each PR is merged and then branching again for the next one seems like an unreasonable amount of work." The implicit assumption here is that no-one is going to have any comments about the PR. While I haven't looked at it yet, I'd be surprised if that's going to be the case.

Why am I being to contrary about this? Couldn't I just appreciate that someone has done a lot of work, and take the code?

The reason that I'm being cautious about this is that I've already gotten burned in the past. People have contributed features to AutoFixture that I was uncomfortable with, and left me with the responsibility of supporting it, including answering questions about why it doesn't work in various circumstances.

Some contributors (particularly the AutoFixture core members) have proven themselves to be here for the long haul. Others have delivered high-quality contributions and moved on. Others again have been more hit-and-run.

These days, I want to make sure that when I take contributions into the AutoFixture code base, it's something I'd feel comfortable supporting. That's the reason the review bar seems high.

(I also believe it's the reason that AutoFixture has existed and evolved for more than five years.)

This is getting long, but ultimately, it's up to you, @TeaDrivenDev, to decide how you'd like to proceed, now that I've explained the trade-offs:

- Do you want me to review this PR as is?
- Do you want to split this PR into smaller PRs?

Let us know, and we'll take it from there.

TeaDrivenDev:
Sorry for the very late reply; I have only now found the time to properly think this through and figure out how this can be done.

I agree that it is best to have this reviewed in smaller parts; I will be largely following the plan set out by @moodmosaic. I am closing this pull request and will start sending new ones to rebuild xUnit 2 support incrementally.

ploeh:
No worries about the time.

Please ask if you have questions of any sort. We very much would like to get this feature done, but I also don't want to steal your thunder, which is why I'm currently assuming a somewhat passive, or reactive role. However, I'll be happy to help if warranted, and I'm sure others will as well :smile:

moodmosaic:
@TeaDrivenDev Please count me in, if I can be of help :smiley:

adamralph:
Thanks to all for making the effort on this. Over at scriptcs we're looking forward to xunit2 support, since it's blocking our wholesale move to xunit2 - https://github.com/scriptcs/scriptcs/issues/1001

jimmymain:
@TeaDrivenDev If you are able to send me a chunk of work, I am also happy to step in and help.

CreepyGnome:
Is there any status on when support for xUnit 2 will be released?  It looks like you were working on it but it seems to have stopped on April 6, 2015, and the current released AutoFixture.xUnit 3.37.3 still depends on xUnit.Extensions less than 2.0.

XUnit 2.0 was released March 16, 2015 and 2.1 was released September 27. 2015.

The reason I ask is because I am using xUnit 2.1 and went to add AutoFixture xUnit and well it added without giving me any complaints and then when I tried to use AutoData attributes the test runner throwns an InvalidOperationException stating "No data found for XXX".

So I can't go backwards on xUnit 2.x to 1.9.x and I am really interested in using AutoFixture, however this is a blocker for me to use it.

anders-salomonsen:
It is released, search for AutoFixture.Xunit2 in your nuget manager, it is a seperate package from AutoFixture.Xunit.

ploeh:
Indeed, [AutoFixture.Xunit2](https://www.nuget.org/packages/AutoFixture.Xunit2) is the package that targets xUnit.net 2. It was first released April 8, 2015.

Looking at this thread alone, I can see that this isn't easily discoverable - although now there's this very notice :smile: 

Is there anything else we can do in order to make this more discoverable?

TeaDrivenDev:
I have added a notice on this at the beginning of my original post in this thread to make it a bit clearer for passers-by that this is not the pull request they are looking for.

moodmosaic:
:+1:

ploeh:
@TeaDrivenDev :+1: Thank you!

CreepyGnome:
Thanks :+1: Guys!!  I guess I  missed it since it doesn't show up in the top list of the AutoFixture search ...  Didn't think to add Xunit duh, sorry about this and the fix to this issue makes it more clear too.

adamralph:
The good way to make AutoFixture.Xunit2 discoverable may be to put a note in the package description of AutoFixture.Xunit on nuget.org

ploeh:
@adamralph Great idea :+1: Thank you.

PR #317, opened 2014-10-15
Freeze on match customization

ecampidoglio:
Addresses #120 by introducing a new customization that freezes specimens based on a set of criteria.

### Usage

The `FreezeOnMatchCustomization` uses the `IRequestSpecification` abstraction as a way to match the incoming requests with the frozen specimen. For example:

```csharp
var fixture = new Fixture();
var freeze = new FreezeOnMatchCustomization(
    typeof(string),
    new PropertySpecification(typeof(string), "Foo"));
fixture.Customize(freeze);
```

This will assign the same string value to _all properties of type `string` named `"Foo"`_.

Multiple matching criteria can also be combined using the `OrRequestSpecification`. Here's an example:

```csharp
var fixture = new Fixture();
var freeze = new FreezeOnMatchCustomization(
    typeof(string),
    new OrRequestSpecification(
        new PropertySpecification(typeof(string), "Foo")
        new ParameterSpecification(typeof(string), "bar"))); 
fixture.Customize(freeze);
```

When no matching criteria are specified, the `FreezeOnMatchCustomization` behaves like the current `FreezingCustomization`, that is it will only match by _exact type_. More concretely:

```csharp
var freeze = new FreezeOnMatchCustomization(typeof(string));
```

is the same as saying:

```csharp
var freeze = new FreezeOnMatchCustomization(typeof(string), new ExactTypeSpecification(typeof(string));
```

### Follow-ups

The are a few things I deliberately left out until after I gather your feedback:

- Since the `FreezeOnMatchCustomization` effectively supersedes the existing `FreezingCustomization`, I propose we obsolete it and rename the `FreezeOnMatchCustomization` to simply be `FreezeCustomization` for consistency.
- I'll add the XML documentation once the feature is finalized.


ecampidoglio:
All of the existing feedback has been addressed.

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    public class FreezeOnMatchCustomization : ICustomization
+    {
+        private readonly Type targetType;
+        private readonly IRequestSpecification matcher;
+        private IFixture fixture;
```
ploeh: Could we make this field `readonly`, please?

ecampidoglio: > Could we make this field readonly, please?

Sure, the customization is immutable now.

ploeh:
It's looking good, but when I attempt to run `build-release.sh`, it gives me 290 build errors. It looks like XML documentation erros.

ecampidoglio:
> It's looking good, but when I attempt to run build-release.sh, it gives me 290 build errors. It looks like XML documentation erros.

Absolutely correct. I had saved that for last but never actually got around to doing it. I'm on it :+1:

ecampidoglio:
I added XML documentation and addressed a few remaining code analysis issues.

moodmosaic:
Both e6c29a7 and cafaac3 are looking good to me!

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.23.0](http://www.nuget.org/packages/AutoFixture/3.23.0).

ecampidoglio:
Nice :+1:

PR #318, opened 2014-10-15
Freeze on match data attribute

ecampidoglio:
This `CustomizeAttribute` leverages the `FreezeOnMatchCustomization` introduced in #317 by opening up its behavior to xUnit parameterized tests.

### Usage

The `FreezeAttribute` applies to the same scenarios as the current `FrozenAttribute`. Here's an example:

```csharp
[Theory, AutoData]
public void FreezeParameterForMatchingProperties(
    [Freeze(By = Matching.PropertyName, TargetName = "Property")]string p1,
    PropertyHolder<string> p2)
{
    Assert.Same(p1, p2.Property);
}
```

Following the same convention as the `FreezeOnMatchCustomization`, if no matching criteria are specified the attribute will mimic the behavior of the current `FrozenAttribute` matching only by _exact type_. In other words the following test:

```csharp
[Theory, AutoData]
public void FreezeParameterByExactType(
    [Freeze]string p1,
    string p2)
{
    Assert.Same(p1, p2);
}
```

is the same as:

```csharp
[Theory, AutoData]
public void FreezeParameterByExactType(
    [Freeze(By = Matching.ExactType)]string p1,
    string p2)
{
    Assert.Same(p1, p2);
}
```

The existing `As` functionality of the `FrozenAttribute` is also superseded by using the `Matching.ImplementedInterfaces` criteria.

For example, given `IFoo` and `Bar` where `Bar` implements `IFoo`, the following test:

```csharp
[Theory, AutoData]
public void FreezeParameterByExactType(
    [Frozen(As = typeof(IFoo))]Bar p1,
    IFoo p2)
{
    Assert.Same(p1, p2);
}
```

can now be expressed as:

```csharp
[Theory, AutoData]
public void FreezeParameterByExactType(
    [Freeze(By = Matching.ImplementedInterfaces)]Bar p1,
    IFoo p2)
{
    Assert.Same(p1, p2);
}
```

which will also match _any other interface_ implemented by `Bar`.

### Follow-ups

The are a few things I deliberately left out until after I gather your feedback:

- Since the `FreezeAttribute` effectively supersedes the existing `FrozenAttribute`, I propose we obsolete it and keep the "_Freeze_" prefix for consistency with the `FreezeCustomization`.
- I'll add the XML documentation once the feature is finalized.

### DIFF
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -58,7 +58,9 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         {
             // Fixture setup
             var registeredType = typeof(AbstractType);
+#pragma warning disable 0618
             var sut = new FrozenAttribute { As = registeredType };
+#pragma warning restore 0618
             var parameter = AParameter<ConcreteType>();
             // Exercise system
             var result = sut.GetCustomization(parameter);
@@ -73,7 +75,9 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         {
             // Fixture setup
             var registeredType = typeof(string);
+#pragma warning disable 0618
             var sut = new FrozenAttribute { As = registeredType };
+#pragma warning restore 0618
             var parameter = AParameter<ConcreteType>();
             // Exercise system and verify outcome
             Assert.Throws<ArgumentException>(() => sut.GetCustomization(parameter));
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -43,7 +43,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationFromNullParamterThrows()
+        public void GetCustomizationFromNullParameterThrows()
         {
             // Fixture setup
             var sut = new FrozenAttribute();
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -142,20 +142,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
-        {
-            // Fixture setup
-            var sut = new FrozenAttribute(Matching.PropertyName);
-            // Exercise system
-            var customization = sut.GetCustomization(AParameter<object>());
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            Assert.NotEmpty(matcher.Specifications.OfType<PropertySpecification>());
-            // Teardown
-        }
-
-        [Fact]
         public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
         {
             // Fixture setup
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -142,20 +142,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationWithMatchingByParameterNameShouldMatchByParameter()
-        {
-            // Fixture setup
-            var sut = new FrozenAttribute(Matching.ParameterName);
-            // Exercise system
-            var customization = sut.GetCustomization(AParameter<object>());
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            Assert.NotEmpty(matcher.Specifications.OfType<ParameterSpecification>());
-            // Teardown
-        }
-
-        [Fact]
         public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
         {
             // Fixture setup
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -141,20 +141,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
             // Teardown
         }
 
-        [Fact]
-        public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
-        {
-            // Fixture setup
-            var sut = new FrozenAttribute(Matching.FieldName);
-            // Exercise system
-            var customization = sut.GetCustomization(AParameter<object>());
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            Assert.NotEmpty(matcher.Specifications.OfType<FieldSpecification>());
-            // Teardown
-        }
-
         private static ParameterInfo AParameter<T>()
         {
             return typeof(SingleParameterType<T>)
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -111,13 +111,16 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByDirectBaseTypeShouldMatchByBaseType()
         {
             // Fixture setup
+            var parameter = AParameter<object>();
             var sut = new FrozenAttribute(Matching.DirectBaseType);
             // Exercise system
-            var customization = sut.GetCustomization(AParameter<object>());
+            var customization = sut.GetCustomization(parameter);
             // Verify outcome
             var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
             var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            Assert.NotEmpty(matcher.Specifications.OfType<DirectBaseTypeSpecification>());
+            var directBaseTypeMatcher = matcher.Specifications.OfType<DirectBaseTypeSpecification>().SingleOrDefault();
+            Assert.NotNull(directBaseTypeMatcher);
+            Assert.Equal(parameter.ParameterType, directBaseTypeMatcher.TargetType);
             // Teardown
         }
 
@@ -125,13 +128,16 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByImplementedInterfacesShouldMatchByImplementedInterfaces()
         {
             // Fixture setup
+            var parameter = AParameter<object>();
             var sut = new FrozenAttribute(Matching.ImplementedInterfaces);
             // Exercise system
-            var customization = sut.GetCustomization(AParameter<object>());
+            var customization = sut.GetCustomization(parameter);
             // Verify outcome
             var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
             var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            Assert.NotEmpty(matcher.Specifications.OfType<ImplementedInterfaceSpecification>());
+            var interfaceTypeMatcher = matcher.Specifications.OfType<ImplementedInterfaceSpecification>().SingleOrDefault();
+            Assert.NotNull(interfaceTypeMatcher);
+            Assert.Equal(parameter.ParameterType, interfaceTypeMatcher.TargetType);
             // Teardown
         }
 
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Linq;
+using System.Reflection;
+using Ploeh.AutoFixture.Kernel;
 using Ploeh.TestTypeFoundation;
 using Xunit;
 
@@ -15,6 +17,28 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
             var sut = new FrozenAttribute();
             // Verify outcome
             Assert.IsAssignableFrom<CustomizeAttribute>(sut);
+            // Teardown
+        }
+
+        [Fact]
+        public void InitializeShouldSetDefaultMatchingStrategy()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new FrozenAttribute();
+            // Verify outcome
+            Assert.Equal(Matching.ExactType, sut.By);
+            // Teardown
+        }
+
+        [Fact]
+        public void InitializeShouldSetDefaultTargetName()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new FrozenAttribute();
+            // Verify outcome
+            Assert.Null(sut.TargetName);
             // Teardown
         }
 
@@ -64,15 +88,12 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
+        public void GetCustomizationWithSpecificTypeShouldReturnCorrectResult()
         {
             // Fixture setup
             var registeredType = typeof(AbstractType);
             var sut = new FrozenAttribute { As = registeredType };
-            var parameter = typeof(TypeWithConcreteParameterMethod)
-                .GetMethod("DoSomething", new[] { typeof(ConcreteType) })
-                .GetParameters()
-                .Single();
+            var parameter = AParameter<ConcreteType>();
             // Exercise system
             var result = sut.GetCustomization(parameter);
             // Verify outcome
@@ -82,18 +103,127 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationWithIncompatibleRegisteredTypeThrowsArgumentException()
+        public void GetCustomizationWithIncompatibleSpecificTypeThrowsArgumentException()
         {
             // Fixture setup
             var registeredType = typeof(string);
             var sut = new FrozenAttribute { As = registeredType };
-            var parameter = typeof(TypeWithConcreteParameterMethod)
-                .GetMethod("DoSomething", new[] { typeof(ConcreteType) })
-                .GetParameters()
-                .Single();
+            var parameter = AParameter<ConcreteType>();
             // Exercise system and verify outcome
             Assert.Throws<ArgumentException>(() => sut.GetCustomization(parameter));
             // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationShouldMatchByExactParameterType()
+        {
+            // Fixture setup
+            var parameter = AParameter<object>();
+            var sut = new FrozenAttribute();
+            // Exercise system
+            var customization = sut.GetCustomization(parameter);
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            var exactTypeMatcher = matcher.Specifications.OfType<ExactTypeSpecification>().SingleOrDefault();
+            Assert.NotNull(exactTypeMatcher);
+            Assert.Equal(parameter.ParameterType, exactTypeMatcher.TargetType);
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationShouldMatchBySeedRequestForParameterType()
+        {
+            // Fixture setup
+            var parameter = AParameter<object>();
+            var sut = new FrozenAttribute();
+            // Exercise system
+            var customization = sut.GetCustomization(parameter);
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            var seedRequestMatcher = matcher.Specifications.OfType<SeedRequestSpecification>().SingleOrDefault();
+            Assert.NotNull(seedRequestMatcher);
+            Assert.Equal(parameter.ParameterType, seedRequestMatcher.TargetType);
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationWithMatchingByDirectBaseTypeShouldMatchByBaseType()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute { By = Matching.DirectBaseType };
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<DirectBaseTypeSpecification>());
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationWithMatchingByImplementedInterfacesShouldMatchByImplementedInterfaces()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute { By = Matching.ImplementedInterfaces };
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<ImplementedInterfaceSpecification>());
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationWithMatchingByParameterNameShouldMatchByParameter()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute { By = Matching.ParameterName, TargetName = "parameter" };
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<ParameterSpecification>());
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute { By = Matching.PropertyName, TargetName = "Property" };
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<PropertySpecification>());
+            // Teardown
+        }
+
+        [Fact]
+        public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute { By = Matching.FieldName, TargetName = "Field" };
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<FieldSpecification>());
+            // Teardown
+        }
+
+        private static ParameterInfo AParameter<T>()
+        {
+            return typeof(SingleParameterType<T>)
+                .GetConstructor(new[] { typeof(T) })
+                .GetParameters()
+                .Single(p => p.Name == "parameter");
         }
     }
 }
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Linq;
 using System.Reflection;
-using Ploeh.AutoFixture.Kernel;
 using Ploeh.TestTypeFoundation;
 using Xunit;
 
@@ -70,74 +69,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
             var parameter = AParameter<ConcreteType>();
             // Exercise system and verify outcome
             Assert.Throws<ArgumentException>(() => sut.GetCustomization(parameter));
-            // Teardown
-        }
-
-        [Fact]
-        public void GetCustomizationShouldMatchByExactParameterType()
-        {
-            // Fixture setup
-            var parameter = AParameter<object>();
-            var sut = new FrozenAttribute();
-            // Exercise system
-            var customization = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            var exactTypeMatcher = matcher.Specifications.OfType<ExactTypeSpecification>().SingleOrDefault();
-            Assert.NotNull(exactTypeMatcher);
-            Assert.Equal(parameter.ParameterType, exactTypeMatcher.TargetType);
-            // Teardown
-        }
-
-        [Fact]
-        public void GetCustomizationShouldMatchBySeedRequestForParameterType()
-        {
-            // Fixture setup
-            var parameter = AParameter<object>();
-            var sut = new FrozenAttribute();
-            // Exercise system
-            var customization = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            var seedRequestMatcher = matcher.Specifications.OfType<SeedRequestSpecification>().SingleOrDefault();
-            Assert.NotNull(seedRequestMatcher);
-            Assert.Equal(parameter.ParameterType, seedRequestMatcher.TargetType);
-            // Teardown
-        }
-
-        [Fact]
-        public void GetCustomizationWithMatchingByDirectBaseTypeShouldMatchByBaseType()
-        {
-            // Fixture setup
-            var parameter = AParameter<object>();
-            var sut = new FrozenAttribute(Matching.DirectBaseType);
-            // Exercise system
-            var customization = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            var directBaseTypeMatcher = matcher.Specifications.OfType<DirectBaseTypeSpecification>().SingleOrDefault();
-            Assert.NotNull(directBaseTypeMatcher);
-            Assert.Equal(parameter.ParameterType, directBaseTypeMatcher.TargetType);
-            // Teardown
-        }
-
-        [Fact]
-        public void GetCustomizationWithMatchingByImplementedInterfacesShouldMatchByImplementedInterfaces()
-        {
-            // Fixture setup
-            var parameter = AParameter<object>();
-            var sut = new FrozenAttribute(Matching.ImplementedInterfaces);
-            // Exercise system
-            var customization = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
-            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
-            var interfaceTypeMatcher = matcher.Specifications.OfType<ImplementedInterfaceSpecification>().SingleOrDefault();
-            Assert.NotNull(interfaceTypeMatcher);
-            Assert.Equal(parameter.ParameterType, interfaceTypeMatcher.TargetType);
             // Teardown
         }
 
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -54,40 +54,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void GetCustomizationReturnsCorrectResult()
-        {
-            // Fixture setup
-            var sut = new FrozenAttribute();
-            var parameter = typeof(TypeWithOverloadedMembers)
-                .GetMethod("DoSomething", new[] { typeof(object) })
-                .GetParameters()
-                .Single();
-            // Exercise system
-            var result = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezingCustomization>(result);
-            Assert.Equal(parameter.ParameterType, freezer.TargetType);
-            // Teardown
-        }
-
-        [Fact]
-        public void GetCustomizationReturnsTheRegisteredTypeEqualToTheParameterType()
-        {
-            // Fixture setup
-            var sut = new FrozenAttribute();
-            var parameter = typeof(TypeWithOverloadedMembers)
-                .GetMethod("DoSomething", new[] { typeof(object) })
-                .GetParameters()
-                .Single();
-            // Exercise system
-            var result = sut.GetCustomization(parameter);
-            // Verify outcome
-            var freezer = Assert.IsAssignableFrom<FreezingCustomization>(result);
-            Assert.Equal(parameter.ParameterType, freezer.RegisteredType);
-            // Teardown
-        }
-
-        [Fact]
         public void GetCustomizationWithSpecificTypeShouldReturnCorrectResult()
         {
             // Fixture setup
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -122,7 +122,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByDirectBaseTypeShouldMatchByBaseType()
         {
             // Fixture setup
-            var sut = new FrozenAttribute { By = Matching.DirectBaseType };
+            var sut = new FrozenAttribute(Matching.DirectBaseType);
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -136,7 +136,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByImplementedInterfacesShouldMatchByImplementedInterfaces()
         {
             // Fixture setup
-            var sut = new FrozenAttribute { By = Matching.ImplementedInterfaces };
+            var sut = new FrozenAttribute(Matching.ImplementedInterfaces);
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -150,7 +150,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByParameterNameShouldMatchByParameter()
         {
             // Fixture setup
-            var sut = new FrozenAttribute { By = Matching.ParameterName, TargetName = "parameter" };
+            var sut = new FrozenAttribute(Matching.ParameterName, "parameter");
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -164,7 +164,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
         {
             // Fixture setup
-            var sut = new FrozenAttribute { By = Matching.PropertyName, TargetName = "Property" };
+            var sut = new FrozenAttribute(Matching.PropertyName, "Property");
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -178,7 +178,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
         {
             // Fixture setup
-            var sut = new FrozenAttribute { By = Matching.FieldName, TargetName = "Field" };
+            var sut = new FrozenAttribute(Matching.FieldName, "Field");
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
--- a/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
@@ -32,17 +32,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Fact]
-        public void InitializeShouldSetDefaultTargetName()
-        {
-            // Fixture setup
-            // Exercise system
-            var sut = new FrozenAttribute();
-            // Verify outcome
-            Assert.Null(sut.TargetName);
-            // Teardown
-        }
-
-        [Fact]
         public void GetCustomizationFromNullParameterThrows()
         {
             // Fixture setup
@@ -150,7 +139,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByParameterNameShouldMatchByParameter()
         {
             // Fixture setup
-            var sut = new FrozenAttribute(Matching.ParameterName, "parameter");
+            var sut = new FrozenAttribute(Matching.ParameterName);
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -164,7 +153,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
         {
             // Fixture setup
-            var sut = new FrozenAttribute(Matching.PropertyName, "Property");
+            var sut = new FrozenAttribute(Matching.PropertyName);
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
@@ -178,7 +167,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
         {
             // Fixture setup
-            var sut = new FrozenAttribute(Matching.FieldName, "Field");
+            var sut = new FrozenAttribute(Matching.FieldName);
             // Exercise system
             var customization = sut.GetCustomization(AParameter<object>());
             // Verify outcome
--- a/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
@@ -129,7 +129,9 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
 
         [Theory, AutoData]
         public void FreezeFirstParameterAsBaseTypeAssignsSameInstanceToSecondParameterOfThatBaseType(
+#pragma warning disable 0618
             [Frozen(As = typeof(AbstractType))]ConcreteType p1,
+#pragma warning restore 0618
             AbstractType p2)
         {
             Assert.Equal(p1, p2);
@@ -137,7 +139,9 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
 
         [Theory, AutoData]
         public void FreezeFirstParameterAsNullTypeAssignsSameInstanceToSecondParameterOfSameType(
+#pragma warning disable 0618
             [Frozen(As = null)]ConcreteType p1,
+#pragma warning restore 0618
             ConcreteType p2)
         {
             Assert.Equal(p1, p2);
--- a/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
@@ -188,11 +188,27 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Theory, AutoData]
+        public void FreezeFirstParameterByParameterWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.ParameterName)]string parameter,
+            SingleParameterType<object> p2)
+        {
+            Assert.Equal(parameter, p2.Parameter);
+        }
+
+        [Theory, AutoData]
         public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
             [Frozen(Matching.PropertyName, "Property")]string p1,
             PropertyHolder<object> p2)
         {
             Assert.Equal(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyWithSameNameShouldAssignSameInstanceToSecondParameter(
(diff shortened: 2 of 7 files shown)


Src/AutoFixture.xUnit.net/FreezeAttribute.cs
```diff
@@ -0,0 +1,145 @@
...
+
+namespace Ploeh.AutoFixture.Xunit
+{
+    /// <summary>
+    /// An attribute that can be applied to parameters in an <see cref="AutoDataAttribute"/>-driven
+    /// Theory to indicate that the parameter value should be frozen and used to satisfy any requests
+    /// made through the same <see cref="IFixture"/> and that match a set of criteria.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
+    public sealed class FreezeAttribute : CustomizeAttribute
```
ploeh: Is it necessary to introduce a `[Freeze]` attribute? Isn't it possible to add the feature to the `[Frozen]` attribute?

If we add a new `[Freeze]` attribute, how will users know when to use `[Frozen]`, and when to use `[Freeze]`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -141,5 +141,85 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByInterfaceShouldAssignSameInstanceToSecondParameter(
+            [Freeze(By = Matching.ImplementedInterfaces)]NoopInterfaceImplementer p1,
+            IInterface p2)
+        {
+            Assert.Same(p1, p2);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByParameterNameShouldAssignSameInstanceToSecondParameter(
+            [Freeze(By = Matching.ParameterName, TargetName = "parameter")]string p1,
```
ploeh: Why do we have to supply `TargetName` here? Why can't we just write the following?

```C#
[Theory, AutoData]
public void FreezeFirstParameterByParameterNameShouldAssignSameInstanceToSecondParameter(
    [Freeze(By = Matching.Parameter)]string parameter,
    SingleParameterType<object> p2)
{
    Assert.Equal(parameter, p2.Parameter);
}
```

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -141,5 +141,85 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            SingleParameterType<object> p2)
+        {
+            Assert.Same(p1, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
+            [Freeze(By = Matching.PropertyName, TargetName = "Property")]string p1,
```
ploeh: Why do we have to supply `TargetName` here? Why can't we just write the following?

```C#
[Theory, AutoData]
public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
    [Freeze(By = Matching.Property)]string property,
    PropertyHolder<object> p2)
{
    Assert.Equal(property, p2.Property);
}
```

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -141,5 +141,85 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            PropertyHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
+            [Freeze(By = Matching.FieldName, TargetName = "Field")]string p1,
```
ploeh: Why do we have to supply `TargetName` here? Why can't we just write the following?

```C#
[Theory, AutoData]
public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
    Freeze(By = Matching.Field)]string field,
    FieldHolder<object> p2)
{
    Assert.Equal(field, p2.Field);
}
```

Src/AutoFixture.xUnit.net/FreezeAttribute.cs
ecampidoglio: My initial idea was have `[Freeze]` supersede the behavior of `[Frozen]`, which could then be made obsolete. However, I realized that the current set of matching possibilities don't include matching by a _specific type_ (like it's possible with the `[Frozen(As = typeof(Foo))]`) so they could probably be merged into the same attribute.

The question is whether the `As` and `By` matching behaviors should complement each other or be made mutually exclusive.

For example, it could be useful to be able to say:

```csharp
public void Test([Frozen(As = typeof(IFoo), By = Matching.PropertyName)]Bar p)
{
}
```

However, what happens if someone says:

```csharp
public void Test([Frozen(As = typeof(IFoo), By = Matching.ExactType)]Bar p)
{
}
```

Do we take into account either or both matching criteria? Does one override the other? How do we make it intuitive which behavior will apply?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ecampidoglio: We could certainly use the parameter's name as the default name to match, but what if we want to match using a different name?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ecampidoglio: See this [comment](https://github.com/AutoFixture/AutoFixture/pull/318#discussion_r23690407).

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ecampidoglio: See this [comment](https://github.com/AutoFixture/AutoFixture/pull/318#discussion_r23690407).

Src/AutoFixture.xUnit.net/FreezeAttribute.cs
ploeh: I'm not too happy combining `As` and `By`... Although I haven't fully thought it through in this particular case, this sounds like a recipe for a really confusing API.

My assumption about this feature is that most (90%) of the use cases for the `As` property could be replaced by `Matching.ImplementedInterfaces`, or am I mistaken?

There may be some rare situations where one interface derives from another interface, and I'm not sure `Type.GetInterfaces()` returns those, but apart from that situation, I wonder if it wouldn't work out if we deprecate the `As` property in favour of the `By` property.

What do you think?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ploeh: Why would we want to do that?

Src/AutoFixture.xUnit.net/FreezeAttribute.cs
ecampidoglio: I think you're right when you say that `By` is likely used for interfaces in the vast majority of the cases. On the other hand, if we deprecate `As`, we're left without a way to match a frozen object by a _specific type_ other than the type of the parameter.

That ability would come in handy in situations where a frozen object must be matched by a base type other than the direct parent, i.e. higher up in the inheritance hierarchy.

I recognize that such a necessity is likely a symptom of a problematic design that AutoFixture would _bring attention to_, in case we decided not to address that. However, it might as well be dictated by a pesky framework (I can think of a few) or a legacy system victim of _inheritance frenzy_.

I'm tempted to suggest that we keep `As` around in the `[Frozen]` attribute without obsoleting it and enable the new matching behavior in a new attribute. However, it shouldn't just be called `[Freeze]`, but rather have a more descriptive name like `[FrozenOnMatch]` or something along those lines.

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ecampidoglio: I have to admit that having a convention is _much better_ than having to specify a member name as a `string` (it's almost reminiscent of [NMock](http://nmock.sourceforge.net)).

However, such a convention would introduce a couple of somewhat unexpected side effects:

- Renaming the parameter would likely break the test
- Renaming the member, in most cases, won't automatically rename the parameter, which would also break the affected tests. In this scenario, having the name as a `string` (with matching case) is more likely to get picked up by a refactoring tool or by a simple _find & replace_ operation.

In spite of this, I still prefer the convention approach that you suggested. However, I think we should document these edge cases.

Src/AutoFixture.xUnit.net/FreezeAttribute.cs
ploeh: Valid points, all.

When I preach SOLID, I always say that the OCP means that you can't go back and change an existing type. If I should practice what I preach, then you've already pointed to the correct solution: leave `[Frozen]` as is, and introduce a new type with the desired behaviour. Possibly, sometime in that process, we could deprecate `[Frozen]`.

My main concern with that approach is that we'd confuse the hell out of users. Even if we put an `[Obsolete]` attribute on `[Frozen]`, with a nice message pointing to the new attribute, I'm afraid that many people would just give up at that point.

The only *consistent* complaint I hear about AutoFixture is that it's too complicated. Your proposed solution, while true to SOLID, doesn't make AutoFixture less complicated.

This is one of the times when it'd be really nice with some metrics about how people use AutoFixture... However, in lack of hard data, let's attempt to look at some incidental points.

The `[Frozen]` attribute has been around since August 2010(!). The `As` property has been around since December 2011, so that's also a long time.

However, [a Google search](https://www.google.com/search?q=AutoFixture%20Frozen%20As) from inside of my Filter Bubble only shows a single hit on the `As` keyword, among the many articles detailing the `[Frozen]` attribute, so I don't think this particular feature is well-known.

The [only hit I've found](http://stackoverflow.com/a/15447124/126014) is one where, as far as I can tell, `Matching.ImplementedInterfaces` would solve the problem nicely.

You are still correct that if we deprecate the `As` property, you *can* come up with examples we can no longer support with the `[Frozen]` attribute. The question is whether anyone use it that way at all.

Personally, I'd be ready to take the bet that no-one would notice if we pulled support for that particular corner case. Even if there's a single developer using it like that somewhere, it's still possible to freeze specimens like that imperatively using `Inject<T>` - you just can't do it declaratively any longer.

Now, if we knew that there are thousands of users using `As` to handle deep inheritance hierarchies, my position would be different, but although I don't have hard data, I don't think that's the case. Hence, my position is still that I'd prefer to deprecate the `As` property, and add the new feature to the existing `[Frozen]` attribute.

ploeh: FWIW, I just [asked on Twitter](http://www.twitter.com/ploeh/status/560555541525970944) for users to speak up if they use this feature :)

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ploeh: Certainly we should document how this work :)

While I admit that this approach isn't refactoring-safe, is the one that involves a `TargetName` property? That's a string as well.

Granted, some sophisticated refactoring tools may be able to detect that string as well, but I wouldn't bet on it.

In any case, since it sounds like you may agree that this is the direction to take, do consider where to put the main implementation of these less strict matching rules. Could they go into the AutoFixture kernel, in the new Specifications you recently added?

Src/AutoFixture.xUnit.net/FreezeAttribute.cs
ecampidoglio: Agreed. I've merged `FreezeAttribute` into the `FrozenAttribute` (6dc0e02) and deprecated `As` (700b7f7).

ploeh:
Did you move and rename many tests in `FrozenAttributeTest`?

When you look at the diff view, it's quite difficult to follow what was moved, what was removed, and what was added.

An alternative review technique I could employ, then, would be to look through each commit sequentially, but if I should do that, I'd appreciate if the commit history would abstain from adding code it deletes again three or four commits later.

ecampidoglio:
I deleted three tests (in 4b3c472) and renamed two (in 5e9e233). I realize that deleting or renaming tests is a delicate operation so I did that in separate commits to make them stand out.

I'm also aware that the final diff looks noisy but that's the result of a trade-off I made. When I merged the tests from `FreezeAttributeTests` into `FrozenAttributeTests`, I added those for the constructor at the top of the file. That effectively changed the order of the tests, making it _look like_ I had renamed the existing ones. My reasoning was that the final order of the tests in the file was more important than how they appeared in the diff. However, I wanted to make sure that those changes were very clear, so I did them in separate commits.

Regarding the lines that get added and removed, I assume you mean the ones where I [ignored a few superceded tests](https://github.com/AutoFixture/AutoFixture/blob/6dc0e02cef18da0311c9845a379bba2d702ac69e/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs#L54-97), which are then removed in a later commit (4b3c472).
The reason I ignored those tests in the first place was because they [asserted an implementation detail](https://github.com/AutoFixture/AutoFixture/blob/6dc0e02cef18da0311c9845a379bba2d702ac69e/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs#L77) that had changed when I merged in `By` from `FreezeAttribute` into `FrozenAttribute` and were therefore _failing_. The expected behavior, on the hand, was maintained in the new version of `FrozenAttribute`, as demonstrated by [a new test](https://github.com/AutoFixture/AutoFixture/blob/6dc0e02cef18da0311c9845a379bba2d702ac69e/Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs#L132-147). Instead of deleting the obsolete tests as part of the same commit, I decided to ignore them first and then delete them in the next commit for clarity.

I wanted to give the background of why the commits look like they do. However, we know that history is malleable so I can always go back and organize the changes in a different way if they don't convey the intent clearly.

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
         {
             Assert.Same(p1, p2);
         }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterShouldAssignSameInstanceToSecondParameter(
+            [Frozen]string p1,
+            string p2)
+        {
+            Assert.Same(p1, p2);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByExactTypeShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.ExactType)]ConcreteType p1,
+            ConcreteType p2)
+        {
+            Assert.Same(p1, p2);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByDirectBaseTypeShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.DirectBaseType)]ConcreteType p1,
+            AbstractType p2)
+        {
+            Assert.Same(p1, p2);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByInterfaceShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.ImplementedInterfaces)]NoopInterfaceImplementer p1,
+            IInterface p2)
+        {
+            Assert.Same(p1, p2);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByParameterNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.ParameterName, TargetName = "parameter")]string p1,
+            SingleParameterType<object> p2)
+        {
+            Assert.Same(p1, p2.Parameter);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.PropertyName, TargetName = "Property")]string p1,
+            PropertyHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Property);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.FieldName, TargetName = "Field")]string p1,
+            FieldHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Field);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(By = Matching.MemberName, TargetName = "parameter")]string p1,
+            SingleParameterType<object> p2)
+        {
+            Assert.Same(p1, p2.Parameter);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingProperty(
+            [Frozen(By = Matching.MemberName, TargetName = "Property")]string p1,
+            PropertyHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Property);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingField(
+            [Frozen(By = Matching.MemberName, TargetName = "Field")]string p1,
+            FieldHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Field);
```
ploeh: What do we really care about here? That these two values point to the same address in memory (an implementation detail of the platform), or that the values are equal to each other?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByInterfaceShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.ImplementedInterfaces)]NoopInterfaceImplementer p1,
+            IInterface p2)
+        {
+            Assert.Same(p1, p2);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByParameterNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.ParameterName, TargetName = "parameter")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `parameter`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            SingleParameterType<object> p2)
+        {
+            Assert.Same(p1, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.PropertyName, TargetName = "Property")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `property`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            PropertyHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(By = Matching.FieldName, TargetName = "Field")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `field`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            FieldHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(By = Matching.MemberName, TargetName = "parameter")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `parameter`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            SingleParameterType<object> p2)
+        {
+            Assert.Same(p1, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingProperty(
+            [Frozen(By = Matching.MemberName, TargetName = "Property")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `property`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            PropertyHolder<object> p2)
+        {
+            Assert.Same(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingField(
+            [Frozen(By = Matching.MemberName, TargetName = "Field")]string p1,
```
ploeh: Why have the extra bureaucracy of naming both `TargetName` and the parameter itself? Why not call the test method parameter `field`?

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -11,22 +12,50 @@ namespace Ploeh.AutoFixture.Xunit
     [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
     public sealed class FrozenAttribute : CustomizeAttribute
     {
+        private IRequestSpecification matcher;
+        private Type targetType;
```
ploeh: Would it be possible to implement this class without mutable fields?

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -11,22 +12,50 @@ namespace Ploeh.AutoFixture.Xunit
...
 
         /// <summary>
-        /// Gets a customization that freezes the <see cref="Type"/> of the parameter.
+        /// Gets or sets the <see cref="Matching"/> criteria used to determine
+        /// which requests will be satisfied by the frozen parameter value.
+        /// </summary>
+        /// <remarks>
+        /// If not specified, requests will be matched by exact type.
+        /// </remarks>
+        public Matching By { get; set; }
```
ploeh: Could this be a read-only property?

ecampidoglio:
I've addressed the feedback about read-only properties (3d71053) as well as matching by the parameter's name (af05c44).

Note that in order to enable the latter scenario, I've relaxed the member specifications (`ParameterSpecification`, `PropertySpecification` and `FieldSpecification`) to make a case-insensitive comparison on the member's name (a546a53).

Src/AutoFixture/Kernel/FieldSpecification.cs
```diff
@@ -95,7 +95,9 @@ private bool FieldIsCompatibleWithTargetType(object request)
 
         private bool FieldHasTargetName(object request)
         {
-            return ((FieldInfo)request).Name == this.targetName;
+            return ((FieldInfo)request)
+                   .Name
+                   .Equals(this.targetName, StringComparison.OrdinalIgnoreCase);
```
ploeh: Isn't this a breaking change?

What if someone wishes to use this class, but with a case-sensitive string comparison?

Src/AutoFixture/Kernel/ParameterSpecification.cs
```diff
@@ -95,7 +95,9 @@ private bool ParameterIsCompatibleWithTargetType(object request)
 
         private bool ParameterHasTargetName(object request)
         {
-            return ((ParameterInfo)request).Name == this.targetName;
+            return ((ParameterInfo)request)
+                   .Name
+                   .Equals(this.targetName, StringComparison.OrdinalIgnoreCase);
```
ploeh: Isn't this a breaking change?

What if someone wishes to use this class, but with a case-sensitive string comparison?

Src/AutoFixture/Kernel/PropertySpecification.cs
```diff
@@ -95,7 +95,9 @@ private bool PropertyIsCompatibleWithTargetType(object request)
 
         private bool PropertyHasTargetName(object request)
         {
-            return ((PropertyInfo)request).Name == this.targetName;
+            return ((PropertyInfo)request)
+                   .Name
+                   .Equals(this.targetName, StringComparison.OrdinalIgnoreCase);
```
ploeh: Isn't this a breaking change?

What if someone wishes to use this class, but with a case-sensitive string comparison?

Src/AutoFixture/Kernel/FieldSpecification.cs
ecampidoglio: > Isn't this a breaking change?

I was on the fence about making this change. My initial thought was to introduce a flag in the constructor that would determine whether the comparison should be CS or CI. In order to avoid breaking the class' public API, that flag would have to be an _optional_ parameter with a default value that indicated a CS comparison.

After giving it some more thought, however, I came to the conclusion that making the comparison CI by default wouldn't detract any previous matches but rather expand them. The resulting set would either be equal or a proper superset of the previous one. Is it a breaking change, then?

Well, the new behavior _could_ potentially be a problem if someone were to use AutoFixture to create an object model where multiple members of the same kind (i.e. properties, fields or parameters) have different semantic values based on the case of their names. For example:

```csharp
public class Foo
{
    public string Bar;
    public string BAR;
}
```

Assuming this model would even have a practical application, if the case is the only difference aren't `Bar` and `BAR` really the same thing?

So, to answer the question, this might be considered a breaking change _conceptually_, since it does change the existing behavior by including additional results, but I don't think it is a breaking change in practice.

Having said that, I realize we're still in the land of assumptions, so I'm going to go ahead and make the CI comparison optional like I meant to do initially. That would also make the specifications more flexible, as you pointed out.

Src/AutoFixture/Kernel/ParameterSpecification.cs
ecampidoglio: See [this comment](https://github.com/AutoFixture/AutoFixture/pull/318#discussion_r26196130)

Src/AutoFixture/Kernel/PropertySpecification.cs
ecampidoglio: See [this comment](https://github.com/AutoFixture/AutoFixture/pull/318#discussion_r26196130).

Src/AutoFixture/Kernel/FieldSpecification.cs
ploeh: [Clean Code](http://amzn.to/XCJi9X) explains why boolean flags are a bad idea.

Could you think of another way to vary comparison without introducing boolean flags?

ecampidoglio: I see what you mean. In this particular context, I considered the CS and CI comparisons to be cohesive enough not to break the SRP at the class level.

When it comes to [readability and clarity of intent](http://martinfowler.com/bliki/FlagArgument.html), however, I do agree that flags are easy to misinterpret from the caller's perspective. On top of that, they also bring higher cyclomatic complexity because of the unavoidable selection.

So, yes. I'll go down a different route.

ploeh: May I suggest that you put this particular pull request aside for a little while, and submit another pull request that expands on the comparison options for the affected Specifications in the kernel?

That would enable us to have a discussion about that, decoupled from the `[Freeze]` attribute.

ecampidoglio: I proposed the change in #398.

ecampidoglio:
I've addressed the feedback about read-only properties (b3992bb) as well as matching by the parameter's name (1216dd6).

Once `FrozenAttribute` is reviewed I'll make it available to the other test frameworks as well (i.e. xUnit2, NUnit).

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
```diff
@@ -82,18 +69,129 @@ public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
...
+        public void GetCustomizationWithMatchingByDirectBaseTypeShouldMatchByBaseType()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute(Matching.DirectBaseType);
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<DirectBaseTypeSpecification>());
```
ploeh: Doesn't it matter what's inside of the `DirectBaseTypeSpecification`?

ploeh: There may be other tests below that drive towards the appropriate behaviour, so forgive me if the above comment is irrelevant, but I'm commenting as I go.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
```diff
@@ -82,18 +69,129 @@ public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
...
+        public void GetCustomizationWithMatchingByImplementedInterfacesShouldMatchByImplementedInterfaces()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute(Matching.ImplementedInterfaces);
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<ImplementedInterfaceSpecification>());
```
ploeh: Doesn't it matter what's inside the `ImplementedInterfaceSpecification`?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
```diff
@@ -82,18 +69,129 @@ public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
...
+        public void GetCustomizationWithMatchingByParameterNameShouldMatchByParameter()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute(Matching.ParameterName, "parameter");
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<ParameterSpecification>());
```
ploeh: Doesn't it matter what's inside the `ParameterSpecification`?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
```diff
@@ -82,18 +69,129 @@ public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
...
+        public void GetCustomizationWithMatchingByPropertyNameShouldMatchByProperty()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute(Matching.PropertyName, "Property");
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<PropertySpecification>());
```
ploeh: Doesn't it matter what's inside the `PropertySpecification`?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
```diff
@@ -82,18 +69,129 @@ public void GetCustomizationWithSpecificRegisteredTypeReturnsCorrectResult()
...
+        public void GetCustomizationWithMatchingByFieldNameShouldMatchByField()
+        {
+            // Fixture setup
+            var sut = new FrozenAttribute(Matching.FieldName, "Field");
+            // Exercise system
+            var customization = sut.GetCustomization(AParameter<object>());
+            // Verify outcome
+            var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
+            var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
+            Assert.NotEmpty(matcher.Specifications.OfType<FieldSpecification>());
```
ploeh: Doesn't it matter what's inside of the `FieldSpecification`?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByInterfaceShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.ImplementedInterfaces)]NoopInterfaceImplementer p1,
+            IInterface p2)
+        {
+            Assert.Equal(p1, p2);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByParameterNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.ParameterName, "parameter")]string p1,
```
ploeh: Why do we need the string `"parameter"` here? Why can't we write the following?

```C#
[Frozen(Matching.ParameterName)]string parameter,
```

ploeh: Oh, I see now below that that's also possible :flushed: 

Should we have this overload?

We can't easily remove it once we've added it, but we can add it later if we start without it, and then realise that we need it.

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            SingleParameterType<object> p2)
+        {
+            Assert.Equal(p1, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByParameterWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.ParameterName)]string parameter,
```
ploeh: I like this :+1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.PropertyName, "Property")]string p1,
+            PropertyHolder<object> p2)
+        {
+            Assert.Equal(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.PropertyName)]string property,
```
ploeh: I also like this :+1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByParameterWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.ParameterName)]string parameter,
+            SingleParameterType<object> p2)
+        {
+            Assert.Equal(parameter, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByPropertyNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.PropertyName, "Property")]string p1,
```
ploeh: Again, do we need this overload to begin with? Can we do without it?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.FieldName, "Field")]string p1,
+            FieldHolder<object> p2)
+        {
+            Assert.Equal(p1, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByFieldWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.FieldName)]string field,
```
ploeh: ...aaand... surprise! I also like this :+1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByPropertyWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.PropertyName)]string property,
+            PropertyHolder<object> p2)
+        {
+            Assert.Equal(property, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByFieldNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.FieldName, "Field")]string p1,
```
ploeh: Do we need that overload?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(Matching.MemberName, "parameter")]string p1,
+            SingleParameterType<object> p2)
+        {
+            Assert.Equal(p1, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberWithSameNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(Matching.MemberName)]string parameter,
```
ploeh: :tada: :balloon: :+1: :fireworks:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByFieldWithSameNameShouldAssignSameInstanceToSecondParameter(
+            [Frozen(Matching.FieldName)]string field,
+            FieldHolder<object> p2)
+        {
+            Assert.Equal(field, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(Matching.MemberName, "parameter")]string p1,
```
ploeh: ... hmm...

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+        public void FreezeFirstParameterByMemberWithSameNameShouldAssignSameInstanceToMatchingParameter(
+            [Frozen(Matching.MemberName)]string parameter,
+            SingleParameterType<object> p2)
+        {
+            Assert.Equal(parameter, p2.Parameter);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingProperty(
+            [Frozen(Matching.MemberName, "Property")]string p1,
```
ploeh: :-1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            PropertyHolder<object> p2)
+        {
+            Assert.Equal(p1, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberWithSameNameShouldAssignSameInstanceToMatchingProperty(
+            [Frozen(Matching.MemberName)]string property,
```
ploeh: :+1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            PropertyHolder<object> p2)
+        {
+            Assert.Equal(property, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberNameShouldAssignSameInstanceToMatchingField(
+            [Frozen(Matching.MemberName, "Field")]string p1,
```
ploeh: :-1:

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -129,18 +129,150 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            FieldHolder<object> p2)
+        {
+            Assert.Equal(p1, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberWithSameNameShouldAssignSameInstanceToMatchingField(
+            [Frozen(Matching.MemberName)]string field,
```
ploeh: :+1:

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -11,22 +13,68 @@ namespace Ploeh.AutoFixture.Xunit
...
+        {
+            this.by = by;
+            this.targetName = targetName;
+        }
+
         /// <summary>
         /// Gets or sets the <see cref="Type"/> that the frozen parameter value
         /// should be mapped to in the <see cref="IFixture"/>.
         /// </summary>
+        [Obsolete("The ability to map a frozen object to a specific type is deprecated and will likely be removed in the future. If you wish to map a frozen type to its implemented interfaces, use [Frozen(By = Matching.ImplementedInterfaces)].")]
```
ploeh: This is a good and helpful message already. Should we also add a sentence about how to map to a base class, if that's the user's case?

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
ecampidoglio: I agree that the convention-based approach is the way to go. :ok_hand: 

I just thought that people might like to have a choice and not be forced to name their test parameters based on the names of the members they want to match.

But then again, if they need to match the values based on their names, it would make _more sense_ if the parameter and the member actually had _the same name_.

All this to say that I agree; we can start by only having the convention-based approach. AutoFixture _is_ an opinionated library, after all :wink:

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -11,22 +13,68 @@ namespace Ploeh.AutoFixture.Xunit
...
+        /// which requests will be satisfied by the frozen parameter value.
+        /// The default value is <see cref="F:Matching.ExactType"/>.
+        /// </param>
+        /// <param name="targetName">
+        /// The identifier used to determine which requests
+        /// for a class member will be satisfied by the frozen parameter value.
+        /// </param>
+        public FrozenAttribute(
+            Matching by = Matching.ExactType,
+            string targetName = null)
```
ploeh: Could we do without `targetName`?

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -34,8 +82,152 @@ public override ICustomization GetCustomization(ParameterInfo parameter)
...
+        private ICustomization FreezeAsType(Type type)
+        {
+            return new FreezingCustomization(
+                type,
+#pragma warning disable 0618
+                this.As ?? type);
+#pragma warning restore 0618
+        }
+
+        private ICustomization FreezeByCriteria(Type type, string name)
```
ploeh: Are there more than one *criteria?*

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
```diff
@@ -11,22 +13,68 @@ namespace Ploeh.AutoFixture.Xunit
     [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
     public sealed class FrozenAttribute : CustomizeAttribute
     {
+        private readonly Matching by;
+        private readonly string targetName;
+        private IRequestSpecification matcher;
```
ploeh: Could this field be `readonly`?

Src/AutoFixture.xUnit.net/FrozenAttribute.cs
ploeh: Yes, there is. Please ignore that previous comment.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: The `TargetType` and `TargetName` properties in `ParameterSpecification` were deprecated in e5d084a so we only look at the type.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: The `TargetType` and `TargetName` properties in `PropertySpecification` were deprecated in 088862d so we only look at the type.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: The `TargetType` and `TargetName` properties in `FieldSpecification` were deprecated in 1925183 so we only look at the type.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ploeh: That `ParameterSpecification` could contain anything, so what does this test prove?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ploeh: That `PropertySpecification` could contain anything, so what does this test prove?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ploeh: That `FieldSpecification` could contain anything, so what does this test prove?

ecampidoglio: The test wants to prove that the matching criteria (in this case `Matching.FieldName`) is converted into the right request specification. The intent could be clearer by asserting:

```csharp
var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
var fieldSpecification = matcher.Specifications.OfType<FieldSpecification>().SingleOrDefault();
Assert.NotNull(fieldSpecification);
```

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: The test wants to prove that the matching criteria (in this case `Matching.PropertyName`) is converted into the right request specification. The intent could be clearer by asserting:

```csharp
var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
var propertySpecification = matcher.Specifications.OfType<PropertySpecification>().SingleOrDefault();
Assert.NotNull(propertySpecification);
```

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: The test wants to prove that the matching criteria (in this case `Matching.ParameterName`) is converted into the right request specification. The intent could be clearer by asserting:

```csharp
var freezer = Assert.IsAssignableFrom<FreezeOnMatchCustomization>(customization);
var matcher = Assert.IsType<OrRequestSpecification>(freezer.Matcher);
var parameterSpecification = matcher.Specifications.OfType<ParameterSpecification>().SingleOrDefault();
Assert.NotNull(parameterSpecification);
```

ploeh: Couldn't `parameterSpecification` contain an `IEquatable<ParameterInfo>` that *always* returns `false`, and still pass this test?

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ploeh: Let's collapse this discussion into [this one](#r39746399).

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ploeh: Let's collapse this discussion into [this one](#r39746399).

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: Yes, the test doesn't exercise `IEquatable<ParameterInfo>` so it would definitely be the case.

Let's see where we are at. Since `ParameterSpecification` no longer contains any information about the target of the comparison – instead delegating the operation to an external function – we can't assert on the state of `parameterSpecification`.

At this point we have at least two options:

1. Exercise the `IEquatable<ParameterInfo>` used by `parameterSpecification` and assert that it returns `true` when passed a matching parameter. For example:
    
    ```csharp
    Assert.True(parameterSpecification.IsSatisfiedBy(parameter));
    ```

2. Delete the test and rely solely on the [end-to-end test](https://github.com/AutoFixture/AutoFixture/blob/freeze-attribute/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs#L182-L188) in `Scenario`.

Going with option **1** would increase the complexity of the test for `PropertySpecification` and `FieldSpecification` because we would have to create both a `ParameterInfo` _and_ a matching `PropertyInfo` and `FieldInfo`. Also, it would no longer be a _unit_ test since we would be acting on multiple objects, namely the `FrozenAttribute` and the implementation of `IEquatable<T>` used by the specification.

Based on that, I suggest we drop the test and simply rely on the end-to-end test in `Scenario`.

ploeh: > Based on that, I suggest we drop the test and simply rely on the end-to-end test in `Scenario`.

Agreed :+1:

ecampidoglio:
I added support for freezing an argument based on match in the xUnit2 (650ac3b) and NUnit2 (d0699d7) glue libraries.

Src/AutoFixture.xUnit.net.UnitTest/FrozenAttributeTest.cs
ecampidoglio: I deleted the relevant test involving `ParameterSpecification` (2068e34), `PropertySpecification` (ae53560) and `FieldSpecification` (6f70261).

Src/AutoFixture.NUnit2.UnitTest/FrozenAttributeTest.cs
```diff
@@ -31,12 +33,15 @@ public void GetCustomizationFromNullParamterThrows()
         }
 
         [Test]
-        public void GetCustomizationReturnsCorrectResult()
```
ploeh: There may be something that I'm unable to follow here, but I don't understand why this test is being deleted. Isn't this test exercising the default behaviour of the `[Frozen]` attribute, when no properties are set? Shouldn't that behaviour remain as it was before?

ploeh:
Did this pull request suddenly explode in size?

May I suggest splitting it up into three smaller pull requests? Perhaps start with the feature set for xUnit.net 2, and then back-port it to the other Glue Libraries if anyone cares.

ecampidoglio:
I could use this feature right now in two projects; one of them uses xUnit 1.9.2 and the other one NUnit 2.6.3.

I moved the commits for xUnit2 in #444 and for NUnit2 in #445. This pull request remains exclusively for xUnit.

I hope this makes things more manageable.

Src/AutoFixture.NUnit2.UnitTest/FrozenAttributeTest.cs
ecampidoglio: I moved the discussion to a new [pull request](https://github.com/AutoFixture/AutoFixture/commit/ae16349568323846c883e7a34c481714d692ce7a#commitcomment-13475877).

ecampidoglio:
Resolved the same issues raised during the code review of #444 and cleaned up the history.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.36.0](https://www.nuget.org/packages/AutoFixture.Xunit/3.36.0).

PR #319, opened 2014-10-15
The Matching DSL

ecampidoglio:
Built on top of the `IRequestSpecification` abstraction, this customization API makes it possible to associate specimen builders with different kinds of requests.

### Usage

Here's an example of how it can be used:

```csharp
fixture.Customize<Foo>(c => c
    .Without(s => s.AProperty)
    .Match()
        .ByBaseType()
        .Or.ByInterfaces()
        .Or.ByProperty("TheFoo"));
```

This effectually means:

> Create instances of `Foo` by omitting the `AProperty` property for every request for a type that is either a **direct base** of `Foo`, one of its **implemented interfaces** or for a **property named** `"TheFoo"` whose type is compatible (i.e. contravariant) to `Foo`.

A couple of things to point out:

- `Match<T>()` is implemented as an extension method to `IPostprocessComposer<T>`.
- The matching behavior is independent of the specimen builder used to satisfy the requests.

### Background

The idea behind the Matching DSL is to bring a simple and fairly common scenario like matching anonymous values based on different criteria up to a "user friendly" high level API, while in the past it used to require interaction with the lower level constructs.

Here's a concrete example taken from @ploeh's [post about building conventions in AutoFixture](http://blog.ploeh.dk/2010/10/19/Convention-basedCustomizationswithAutoFixture/).

In order to assign a valid currency code to all string parameters named "*currencyCode*", one would have to implement a custom `ISpecimenBuilder` and check for the right kind of requests:

```csharp
public class CurrencyCodeSpecimenBuilder : ISpecimenBuilder
{
    public object Create(object request, ISpecimenContext context)
    {
        var pi = request as ParameterInfo;

        if (pi == null)
        {
        }

        if (pi.ParameterType != typeof(string)
            || pi.Name != "currencyCode")
        {
        }
 
        return "DKK";
    }
}
```

Even if we're all very familiar with the concepts mentioned in that class (*specimen*, *request*, *specimen context*, *no specimen*) I'm pretty sure that they're completely foreign to most people who are approaching AutoFixture for the first time. And rightly so, since they belong to the realm of lower level abstractions.
In addition, this approach also requires you to deal with Reflection types directly.

Compare that with a customization that uses the Matching DSL:

```csharp
fixture.Customize<string>(c => c
       .FromFactory(() => "DKK")
       .Match().ByParameter("currencyCode"));
```

I can't count how many times I've wished I could simply write these kinds of customizations without having to first make custom builders, request specifications and so on.

### Follow-ups

The XML documentation is currently missing and I'll add it once the feature is finalized.

### DIFF
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -119,6 +119,10 @@
     <Compile Include="Kernel\DirectBaseTypeSpecification.cs" />
     <Compile Include="Kernel\ParameterSpecification.cs" />
     <Compile Include="Kernel\FieldSpecification.cs" />
+    <Compile Include="Dsl\IMatchComposer.cs" />
+    <Compile Include="Dsl\MatchComposer.cs" />
+    <Compile Include="Dsl\NullMatchComposer.cs" />
+    <Compile Include="Dsl\PostprocessComposer.cs" />
     <Compile Include="MapCreateManyToEnumerable.cs" />
     <Compile Include="Kernel\MultipleToEnumerableRelay.cs" />
     <Compile Include="Kernel\NoConstructorsSpecification.cs" />
--- a/Src/AutoFixture/Dsl/IMatchComposer.cs
+++ b/Src/AutoFixture/Dsl/IMatchComposer.cs
@@ -1,0 +1,21 @@
+﻿using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.Dsl
+{
+    public interface IMatchComposer<T> : ISpecimenBuilder
+    {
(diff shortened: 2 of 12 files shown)


ecampidoglio:
I'm closing this PR for the time being.

I still believe that this kind of customizations are general enough that they shouldn't require writing _ad-hoc_ specimen builders. However, I no longer believe that adding a whole new interface is the right answer, because it breaks the existing paradigm of customization we're familiar with in AutoFixture.

Instead, having some like this [`FilteringTypeRelay`](http://stackoverflow.com/a/32562980/26396) built-in, for example, would be a better option; it follows the idiom of adding customizations and builders to the `Fixture` without requiring custom code.

PR #336, opened 2015-01-14
Stub moq properties (v2)

dcastro:
This PR stems from #314.

It aims to improve the auto-configuration capabilities of `AutoConfiguredMoqCustomization` by:
* stubbing a mock object's overridable properties, such that:   
```csharp
var sut = fixture.Create<IInterface>();
sut.Property = "Hello";
Assert.Equal("Hello", sut.Property);
```
* populate overridable properties with values from a fixture

The first point is addressed by `StubPropertiesCommand`, similar to @ploeh's [`PropertiesPostprocessor`](http://blog.ploeh.dk/2013/04/05/how-to-configure-automoq-to-set-up-all-properties/).

The second point is addressed by `AutoMockPropertiesCommand`.
It wraps around `AutoPropertiesCommand` and excludes Castle's dynamic proxy's fields (`__interceptors` and `__target`).
Since this class populates both overridable and sealed properties, there's no longer any need for `MockSealedPropertiesCommand` (which populated sealed properties only), so I've marked it as Obsolete.


Src/AutoMoq/AutoMockPropertiesCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+using Moq;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.AutoMoq
+{
+    /// <summary>
+    /// A command that populates all public writable properties/fields of a mock object with anonymous values.
+    /// </summary>
+    public class AutoMockPropertiesCommand : ISpecimenCommand
```
ploeh: What does this class do that `AutoPropertiesCommand` doesn't already do?

Src/AutoMoq/StubPropertiesCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Stubs a mocked object's properties, giving them "property behavior".
+        /// Setting a property's value will cause it to be saved and later returned when the property is accessed.
+        /// </summary>
+        /// <param name="specimen">The mock to setup.</param>
+        /// <param name="context">The context of the mock.</param>
+        public void Execute(object specimen, ISpecimenContext context)
+        {
+            if (specimen == null) throw new ArgumentNullException("specimen");
```
ploeh: According to [Postel's Law](http://en.wikipedia.org/wiki/Robustness_principle), is throwing an exception here the correct thing to do?

Is `null` always an invalid *specimen*?

Src/AutoMoq/StubPropertiesCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            var mock = specimen as Mock;
+            if (mock == null)
+                return;
+
+            //disable generation of default values (if enabled), otherwise SetupAllProperties will hang if there's a circular dependency
+            var mockDefaultValueSetting = mock.DefaultValue;
```
ploeh: I don't think that we should implicitly change the behaviour of the mock object deep within the bowels of a Fixture instance. That could make things very hard to understand for users.

Src/AutoMoq/StubPropertiesCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+            mock.DefaultValue = DefaultValue.Empty;
+
+            //stub properties
+            mock.GetType()
+                .GetMethod("SetupAllProperties")
+                .Invoke(mock, new object[0]);
+
+            //restore setting
+            mock.DefaultValue = mockDefaultValueSetting;
```
ploeh: Oh, I now notice that it's being restored here. That makes my previous comment invalid, I guess...

ploeh: However, I don't understand how that makes any difference then. Does `SetupAllProperties` populate the properties?

Src/AutoMoq/StubPropertiesCommand.cs
dcastro: Good question. I never gave it much thought, but I was under the impression `null` was an invalid specimen, seeing as most implementations of `ISpecimenCommand` refuse it - namely, [`AutoPropertiesCommand`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/AutoPropertiesCommand.cs#L148), [`BindingCommand<T, TProperty>`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/BindingCommand.cs#L126) and [`DictionaryFiller`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DictionaryFiller.cs#L43).

Personally, I think it makes sense to throw an exception. Commands are executed on the specimen returned by a builder. The builder, as I understand it, should either return an actual value, an instance of `NoSpecimen` or an instance of `OmitSpecimen`. But I can't think of a situation where it would ever return `null`.

Src/AutoMoq/AutoMockPropertiesCommand.cs
dcastro: It's a thin layer that decorates/builds on top of `AutoPropertiesCommand`.

It "unwraps" the mock's `Object` property, and adds a stricter request specification to filter out dynamic proxy fields.

I think the alternative would be to:
 * Delete the `AutoMockPropertiesCommand` class
 * Make the `IgnoreProxyMembersSpecification` specification public and non-nested.
 * And change the composition root from this:

```csharp
            fixture.Customizations.Add(
                new Postprocessor(
                    builder: new MockPostprocessor(
                                new MethodInvoker(
                                    new MockConstructorQuery())),
                    command: new CompositeSpecimenCommand(
                                new MockVirtualMethodsCommand(),
                                new StubPropertiesCommand(),
                                new AutoMockPropertiesCommand())));

            fixture.ResidueCollectors.Add(Relay);
```

to this:

```csharp

            fixture.Customizations.Add(
                new Postprocessor(
                    builder: new MockPostprocessor(
                                new MethodInvoker(
                                    new MockConstructorQuery())),
                    command: new CompositeSpecimenCommand(
                                new MockVirtualMethodsCommand(),
                                new StubPropertiesCommand())));

            fixture.ResidueCollectors.Add(
                new Postprocessor(
                    Relay,
                    new AutoPropertiesCommand(
                        new IgnoreProxyMembersSpecification())));
```

Personally, I prefer the way it's currently implemented. `IgnoreProxyMembersSpecification` is not really intuitive, and is too specific to ever be reused (I think), so it's probably best to make it nested and private. `AutoMockPropertiesCommand` is an intuitive layer (albeit thin) and a well-defined responsibility. And the composition root looks cleaner and is easier to understand.

Src/AutoMoq/StubPropertiesCommand.cs
ploeh: Fair enough, let's make it consistent with the other implementations.

However, you [can actually Register null](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FixtureTest.cs#L4737-L4747), so `null` is a valid Specimen.

Src/AutoMoq/AutoMockPropertiesCommand.cs
ploeh: > It "unwraps" the mock's Object property

That it does, but it doesn't _use_ `mock` for anything other than calling `Execute` with `mock.Object`. Why does it need to do that 'unwrapping', then?

> and adds a stricter request specification to filter out dynamic proxy fields.

Couldn't that specification just be added to an `AutoPropertiesCommand` instance?

> IgnoreProxyMembersSpecification is not really intuitive, and is too specific to ever be reused (I think), so it's probably best to make it nested and private.

Couldn't you make it a private member of the `Customization` instead? Like [this](http://blog.ploeh.dk/2013/04/08/how-to-automatically-populate-properties-with-automoq)?

dcastro: > That it does, but it doesn't use mock for anything other than calling Execute with mock.Object. Why does it need to do that 'unwrapping', then?

The alternative to unwrapping `mock.Object` and passing it to an `AutoPropertiesCommand` would be to "decorate" the `Relay` with the `AutoPropertiesCommand` (since the `Relay` returns the type being mocked, not the `Mock<T>` type, and thus doesn't need unwrapping) as shown in my second code snippet. I'll explain why I'm not a big fan of wrapping the `Relay` below.

> Couldn't that specification just be added to an AutoPropertiesCommand instance?

It could. In the second code snippet above, for example, the customization wraps the `IRequestSpecification` in an `AutoPropertiesCommand`, which is then wrapped by a `PostProcessor`. But I personally believe avoiding all this wrapping and simply adding a new `ISpecimenCommand` to the existent post-processor looks cleaner.

I also think it's easier to understand what's going on in the first snippet ("the mock returned by the builder will be post-processed in 3 stages: first we'll setup its virtual methods, then stub its properties, and then populate them") than in the second snippet.

Since the first code snippet also keeps all the "mock configuration" stages in one place, it should also be easier to maintain, e.g. change the auto-configuration algorithm.

> Couldn't you make it a private member of the Customization instead?

If the `Customization` defined the `ISpecificationRequest` for the `AutoPropertiesCommand`, wouldn't that be against the Single Responsibility Principle?

Src/AutoMoq/StubPropertiesCommand.cs
dcastro: Ah, so it is! Interesting!

I can leave it as it is for now, to keep it consistent as you say. But would you like me to submit a PR changing these four implementations to accept a `null` specimen?

Src/AutoMoq/StubPropertiesCommand.cs
dcastro: I suppose it is, yes.
I'm guessing, if `Mock<T>.DefaultValue` is set to `DefaultValue.Mock`, then `SetupAllProperties` will setup properties to have normal property behaviour and then populate them with new mocks where applicable.

dcastro: I think [this](https://github.com/Moq/moq4/blob/master/Source/Mock.cs#L681) is the culprit.

Src/AutoMoq/StubPropertiesCommand.cs
ploeh: Perhaps we should just keep it as is for consistency... Apparently, it's not an issue.

Src/AutoMoq/StubPropertiesCommand.cs
ploeh: That's surprising, and, I believe, not correct behaviour from Moq (although it seems it's always done that), so I've created this issue: https://github.com/Moq/moq4/issues/149

Src/AutoMoq/AutoMockPropertiesCommand.cs
ploeh: I must admit that I can't follow this line of argument, or else I just don't agree with it.

AutoFixture already has perhaps 90% of the building blocks necessary to implement this feature, so I don't think it's warranted to add more code to the code base. That code would also be something that we'd have to maintain and support in the future.

We keep circling back to [my original proof of concept](http://blog.ploeh.dk/2013/04/08/how-to-automatically-populate-properties-with-automoq), which manages to implement the behaviour in question in less code. Why not use that approach?

dcastro: After rethinking this, I think I may be over complicating things. And I agree that we should avoid introducing new types unless they're actually a good compelling reason to do so. 

I'll go ahead and redesign this bit as per the POC

Src/AutoMoq/StubPropertiesCommand.cs
dcastro: Indeed, it does smell like a bug, I hadn't thought of it that way!

Src/AutoMoq/AutoMockPropertiesCommand.cs
dcastro: I just realized there's a problem with the [POC](http://blog.ploeh.dk/2013/04/08/how-to-automatically-populate-properties-with-automoq/). If this line in the test `fixture.Create<IHasProperties>()` is changed to `fixture.Create<Mock<IHasProperties>>().Object`, then the test fails.

That's because the `AutoPropertiesCommand` is applied to the residue collector, and not to the customization. When we ask for an instance of `Mock<IHasProperties>`, then the residue collector is skipped and the properties are never populated.

moodmosaic: But, wouldn't it be a POLA violation if the behavior of
```c#
new Mock<IHasProperties>().Object
```
is different from
```c#
new Fixture().Create<Mock<IHasProperties>>().Object
```
?

dcastro: What do you mean by "different behaviour" here?

moodmosaic: Indeed, perhaps value is a more appropriate word here. (See [updated](https://github.com/AutoFixture/AutoFixture/pull/336/files#r23910794) note.)

dcastro: Sorry but I still do not understand your point.

Yes, there's a number of ways in which `new Mock<IHasProperties>().Object` differs from `new Fixture().Create<Mock<IHasProperties>>().Object;`; one has its members setup and its properties populated, while the other one doesn't, for example. But that's the whole point of the `AutoConfiguredMoqCustomization`.

dcastro: @moodmosaic were you perhaps referring to my last comment where I pointed out the difference between `fixture.Create<IHasProperties>()` and `fixture.Create<Mock<IHasProperties>>().Object`? If so, yes, it would be a POLA violation.

moodmosaic: Yes, I was referring to [this](https://github.com/AutoFixture/AutoFixture/pull/336/files#r23900569) comment. Though, I must admit, I haven't followed all the previous Pull Requests, so there should be probably some context that I'm missing...

dcastro: You're completely right though. The two snippets should return the same object.

This means any post-processing needs to be done on the instance of `Mock<IHasProperties>` returned by the customization (which is executed in both cases), and not on the instance of `IHasProperties` returned by the residue collector (which is executed only in the second case).

As follows, it seems like we do need this `AutoMockPropertiesCommand` to unwrap `Mock<IHasProperties>.Object` and pass it to an `AutoPropertiesCommand` after all, even though the initial reasoning for doing this was flawed.

ploeh: FWIW, I don't consider it a POLA violation that `fixture.Create<Mock<IHasProperties>>()` returns a differently configured object than `new Mock<IHasProperties>()`... or, if it is, it's also a POLA violation that `fixture.Create<Foo>()` returns a differently populated object than `new Foo()` in all the cases when `Foo` has writable properties.

> While doing this change, a few integration tests started failing. I realized there's a problem with the POC. If this line in the test fixture.Create<IHasProperties>() is changed to fixture.Create<Mock<IHasProperties>>().Object, then the test fails.

This point is, I think, a completely valid reason for doing something more complex than the POC. I hadn't noticed that this was the case with the POC, but it's obviously a flaw in the POC.

dcastro: If there are no outstanding review notes, could this please be merged?

ploeh: Yes, it can!

Please accept my sincere apologies that I have been sitting on this for so long! That's a poor experience for you, as a contributor! :flushed:

Thank you for nudging me to look at this again!

dcastro: It's no problem at all! You posted somewhere else you've been swamped and we all know what that's like ^^

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoMoq](http://www.nuget.org/packages/AutoFixture.AutoMoq/3.24.2).

Issue #339, opened 2015-01-21
FormatException from fixture.Create<System.Net.Mail.MailAddress>()


cbowdon:
The following code fails with a FormatException:

```C#
new Fixture().Create<System.Net.Mail.MailAddress>()
```

Fairly obviously this is because the MailAddress constructor won't accept arbitrary strings. It's easy to work around with some customization. However, this feels like something that should just work, the same way that creating a System.Uri just works.

Looking at the AutoFixture source it looks like a fairly straightforward solution would be to add a MailAddressGenerator. Would that be acceptable? If so I will submit a PR.

ecampidoglio:
Since MailAddress is part of the BCL I don't see why not. 👍

ploeh:
Sounds reasonable.

However, I think it's important to ensure that the generated email addresses fall into a range that runs no risk of bothering people with real email addresses.

Although AutoFixture is intended to be used for unit testing, people use if for all sorts of funny purposes, so I wouldn't be surprised if, one day, someone would (accidentally) write a test suite that as a side effect attempts to actually send out emails.

One way to do that is to generate all addresses using [fictitious domain names](http://en.wikipedia.org/wiki/Example.com), e.g.

- 12a111d4-42d0-486e-b099-2be08789c923@example.com
- 3c38dd7d-3c42-4e59-9188-adbbcf879ae0@example.net
- a6be450b-3475-49c3-b550-eaabe338d0da@example.org

cbowdon:
That makes sense, I will ensure it only generates addresses for fictitious domains.

ploeh:
Resolved in d02f102301af95908e31dcb0e5c628abe48cbb8b.

PR #340, opened 2015-01-26
Added MailAddressGenerator and tests.

cbowdon:
This implements the enhancement in #339.

Let me know if there's any changes you need me to make to this. I'm sorry about the newline in AutoFixtureUnitTest.csproj. It's an artifact of manually editing the file to remove some cruft Xamarin Studio added (I'm not on Win at home). I did try and remove it, but gave up after an hour or so of fiddling around (life's too short, sorry).

### DIFF
--- a/Src/AutoFixture/MailAddressGenerator.cs
+++ b/Src/AutoFixture/MailAddressGenerator.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Globalization;
+using System.Net.Mail;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Creates new <see cref="MailAddress"/> instances.
+    /// </summary>
+    public class MailAddressGenerator : ISpecimenBuilder
+    {
+        private readonly string[] fictitiousDomains = 
+        { 
+            "example.com", 
+            "example.net", 
+            "example.org" 
+        };
+
+        /// <summary>
+        /// Creates a new MailAddress.
+        /// </summary>
+        /// <param name="request">The request that describes what to create.</param>
+        /// <param name="context">A context that can be used to create other specimens.</param>
+        /// <returns>
+        /// The requested specimen if possible; otherwise a <see cref="NoSpecimen"/> instance.
+        /// </returns>
+        /// <remarks>>
+        /// The generated MailAddress will have one of the reserved domains,
+        /// so as to avoid any possibility of tests bothering real email addresses
+        /// </remarks>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (!typeof(MailAddress).Equals(request))
+            {
+                return new NoSpecimen(request);
+            }
+
+            var name = Guid.NewGuid().ToString();
+            var host = fictitiousDomains[(uint)name.GetHashCode() % 3];
+
+            var email = string.Format(CultureInfo.InvariantCulture, "{0} <{0}@{1}>", name, host);
+            return new MailAddress(email);
+        }
+    }
+}
--- a/Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
+++ b/Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.Net.Mail;
+using System.Text.RegularExpressions;
+using Ploeh.AutoFixture;
+using Ploeh.AutoFixture.Kernel;
+using Ploeh.AutoFixtureUnitTest.Kernel;
+using Xunit;
+
+namespace Ploeh.AutoFixtureUnitTest
+{
+    public class MailAddressGeneratorTest
+    {
+        [Fact]
+        public void SutIsSpecimenBuilder()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new MailAddressGenerator();
+            // Verify outcome
+            Assert.IsAssignableFrom<ISpecimenBuilder>(sut);
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithNullRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var sut = new MailAddressGenerator();
+            var dummyContext = new DelegatingSpecimenContext();
+            // Exercise system
+            var result = sut.Create(null, dummyContext);
+            // Verify outcome
+            Assert.Equal(new NoSpecimen(), result);
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithNonMailAddressRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var request = new object();
+            var sut = new MailAddressGenerator();
+            var dummyContext = new DelegatingSpecimenContext();
+            // Exercise system
+            var result = sut.Create(request, dummyContext);
+            // Verify outcome
+            var expectedResult = new NoSpecimen(request);
+            Assert.Equal(expectedResult, result);
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithMailAddressRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var request = typeof(MailAddress);
+            var context = new DelegatingSpecimenContext();
+            var sut = new MailAddressGenerator();
+            // Exercise system
+            var result = (MailAddress)sut.Create(request, context);
+            // Verify outcome
+            Assert.True(Regex.IsMatch(result.Host, @"example\.(com|org|net)"));
+            // Teardown
+        }
+    }
+}
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -316,6 +316,7 @@
     <Compile Include="UriGenerator.cs" />
     <Compile Include="UriScheme.cs" />
     <Compile Include="UriSchemeGenerator.cs" />
+    <Compile Include="MailAddressGenerator.cs" />
   </ItemGroup>
   <ItemGroup>
     <CodeAnalysisDictionary Include="CodeAnalysisDictionary.xml">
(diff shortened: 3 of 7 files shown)


Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Net.Mail;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Creates new <see cref="Uri"/> instances.
```
moodmosaic: Does this class creates new *Uri* instances?

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -0,0 +1,44 @@
...
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Creates new <see cref="Uri"/> instances.
+    /// </summary>
+    public class MailAddressGenerator : ISpecimenBuilder
+    {
+        private readonly string[] fictitiousDomains =
```
moodmosaic: While these predefined domains correctly follow [this](https://github.com/AutoFixture/AutoFixture/issues/339#issuecomment-70984249) comment by @ploeh, wouldn't it be more flexible if we could also allow users to supply their own domains if they need to?

A similar example is the `UriGenerator` class, which [requests a `UriScheme`] (https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/UriGenerator.cs#L31) from the context.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        private readonly string[] fictitiousDomains = 
+        { 
+            "example.com", 
+            "example.net", 
+            "example.org" 
+        };
+
+        /// <summary>
+        /// Creates a new anonymous MailAddress.
```
moodmosaic: Is it really an *anonymous* MailAddress?

Src/AutoFixture/MailAddressGenerator.cs
cbowdon: Ah, shameful mistake. Fixed.

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithNullRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var sut = new MailAddressGenerator();
+            // Exercise system
+            var dummyContext = new DelegatingSpecimenContext();
```
moodmosaic: Is `dummyContext` part of the [Exercise SUT](http://xunitpatterns.com/exercise%20SUT.html) phase?

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        [Fact]
+        public void CreateWithNonMailAddressRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var request = new object();
+            var sut = new MailAddressGenerator();
+            // Exercise system
+            var dummyContext = new DelegatingSpecimenContext();
```
moodmosaic: Is `dummyContext` part of the [Exercise SUT](http://xunitpatterns.com/exercise%20SUT.html) phase?

moodmosaic:
Thank you for your contribution! I just left a couple of comments, but overall (IMO) this is looking good.

It would be nice to have a [Facade Test](http://blog.ploeh.dk/2012/06/27/FacadeTest/), demoing the feature like the other tests in the [FixtureTest](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FixtureTest.cs) class.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// so as to avoid any possibility of tests bothering real email addresses
+        /// </remarks>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (!typeof(MailAddress).Equals(request))
+            {
+                return new NoSpecimen(request);
+            }
+
+            var name = Guid.NewGuid().ToString();
```
ecampidoglio: If a builder needs to create a specimen it doesn't handle itself, it should usually delegate its creation to the `ISpecimenContext`. In this case, the name portion of the email address should be created by:

```csharp
var name = context.Resolve(typeof(string));
```

cbowdon:
Prompt review, thanks! I've pushed the most trivial fixes and will work on the facade test and user-supplied domains.

Regarding the UriScheme example - can you explain how it works if the user doesn't supply their own? It looks to me like the context should return null and hence the UriGenerator should return NoSpecimen.

Src/AutoFixture/MailAddressGenerator.cs
moodmosaic: Yes, similar to [this](https://github.com/AutoFixture/AutoFixture/pull/340#discussion_r23566830) comment.

moodmosaic:
The [UriGenerator](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/UriGenerator.cs) requests a [UriScheme](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/UriScheme.cs) from the context, so that users don't have to set a UriScheme, unless they want a custom one.

With the above pieces, it's possible to supply a custom URI scheme as shown below:
```c#
var fixture = new Fixture();
fixture.Inject(new UriScheme("ftp"));
var uri = fixture.Create<Uri>(); 
// Prints -> ftp://abc9f406-16f2-4e06-b6f9-0750dc115ac3/
```

This is handled by [UriSchemeGenerator](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/UriSchemeGenerator.cs) which is in turn [registered in DefaultPrimitiveBuilders iterator](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DefaultPrimitiveBuilders.cs#L30).

—

Hypothetically, with MailAddress it should be possible to supply a custom domain name as shown below:

```c#
var fixture = new Fixture();
fixture.Inject(new DomainName("bar.baz"));
var uri = fixture.Create<MailAddress>(); 
// Prints -> abc9f406-16f2-4e06-b6f9-0750dc115ac3@bar.baz
```

Please note that it'd be wise to have @ploeh's opinion before you make this additional effort.

HTH

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        [Fact]
+        public void CreateWithNullContextDoesNotThrow()
+        {
+            // Fixture setup
+            var sut = new MailAddressGenerator();
+            var dummyRequest = new object();
+            // Exercise system and verify outcome
+            Assert.DoesNotThrow(() => sut.Create(dummyRequest, null));
```
ploeh: Let's say that we take `MailAddressGenerator` as is (as I'm very close to thinking that we should), but later want to enhance it so that it'll request names and domain names from the context.

If that's the case, the `context` becomes a required parameter. If we keep this test around, it might be interpreted as a breaking change. What if we didn't have this test?

ploeh:
This Pull Request (PR) is looking good.

Both @ecampidoglio and @moodmosaic have great suggestions, and I think that both of these suggestions should indicate the direction in which this feature should move. However, let's assume that we take this PR as it is (preferably with one very small change that I've commented elsewhere), couldn't we add both of those features later, without making them breaking changes?

cbowdon:
@ecampidoglio - I will flatten commits once you're happy to accept this.

moodmosaic:
It's possible to take this PR as it is, and add, hypothetically, DomainName and DomainNameGenerator at some later point. AFAICT, if we do that there won't be any breaking changes.

cbowdon:
Thanks for the feedback all. As per @ploeh's suggestion I've removed the null context test, and I believe all the earlier issues around comments are now fixed.

I have not:
- tackled the DomainName enhancement
- added a facade test to the FixtureTest class

Is this acceptable without the FixtureTest? If so I'll flatten the commits.

moodmosaic:
Sounds good. – With this PR, what happens if you do

```c#
var fixture = new Fixture();
var sut = fixture.Create<MailAddress>();
```
?

Wouldn't it be nice to keep track of this new behavior by adding a test in the [FixtureTest](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FixtureTest.cs) class?

ploeh:
Running `build-release.sh` reports 20 build errors. Could you please address those?

ploeh:
If you don't want to run `build-release.sh` (which requires you to use Bash), you can also run `BuildRelease.ps1` from PowerShell; they do the same thing.

cbowdon:
I've pushed a couple of fixes, but please don't spend any further time on this yet. I need to clone this branch on a Windows machine and try running build-release for myself. (FWIW - trying it with xbuild on OS X resulted in a huge stream of errors unrelated to my changes.)

cbowdon:
I believe this is ready now. Should I update my branch though, I notice it's fallen behind master?

moodmosaic:
Running `build-release.sh` now completes successfully, reporting 0 warnings and 0 errors.

image

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.24.0](http://www.nuget.org/packages/AutoFixture/3.24.0).

Sorry it took some time before I was able to pick this up, but I've been quite snowed under these last weeks :grimacing:

cbowdon:
No problem, thank you all for your patience with my screw-ups!

ploeh:
No worries :smile: 

Are you [CBowdon](https://twitter.com/CBowdon) on Twitter? Usually, I tweet whenever there's a new minor release of AutoFixture, and drop the name of the contributor...

cbowdon:
Yes that's me, thanks :)

ploeh:
[Done](https://twitter.com/ploeh/status/568710595769462785) :smile:

PR #347, opened 2015-02-13
Supports returning abstract types setup calls to be overriden.

### LINKED ISSUE #346: Cannot override setup if return type is an abstract type, using AutoConfiguredNSubstituteCustomization
If I have a type with a property or method returning an abstract type, ``AutoConfiguredNSubstituteCustomization`` sets it up to return a substitute.

If I try to override the return value, it throws a ``CouldNotSetReturnDueToNoLastCallException``.

The following code illustrate the problem:

```c#
[Fact]
public void SetupCallsReturningAbstractTypesAreOverridable()
{
    var substitute = fixture.Create<IInterfaceWithCircularDependency>();
    var expected = fixture.Create<IInterfaceWithCircularDependency>();
    substitute.Component.Returns(expected);

    var result = substitute.Component;

    Assert.Equal(expected, result);
}
```

mrinaldi:
This PR fixes #346.


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoNSubstitute 3.23.2](http://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.23.2).

Sorry that it took so long, but I've been really snowed under lately :grimacing:

PR #351, opened 2015-02-25
Changed AutoConfiguredNSubstituteCustomization to return finite sequences for abstract types implementing IEnumerable<T>

### LINKED ISSUE #350: AutoConfiguredNSubstituteCustomization infinite enumeration on IReadOnlyList property
This test fails with timeout because Items property will infinitely generate new items.
The problem seems to not appear when using IList<> or IEnumeration<>.

    [TestClass]
    public class Tests
    {
        public class Data
        {
            public IReadOnlyList<string> Items { get; set; }
        }

        [TestMethod, Timeout (500)]
        public void Test ()
        {
            // arrange
            var fixture = new Fixture ();
            fixture.Customize (new AutoConfiguredNSubstituteCustomization ());

            var sut = fixture.Create<Data> ();

            // act
            var result = sut.Items.Count (x => true);

            // assert
            Assert.AreEqual (0, result);
        }
    }

mrinaldi:
This PR fixes #350.

It sets up auto-configured NSubstitute mocks that returns ``IEnumerator<T>`` and any abstract type implementing ``IEnumerable<T>`` to return a finite sequence.

Note that for interfaces like this:

```c#
public interface IMyList<T> : IEnumerable<T>
{
    int Count { get; }
}
```

``list.Count`` will not be equal to ``list.Count()``, since the former will be setup to return from AutoFixture while the latter will be processed by LINQ and will return the count of the ``GetEnumerator()`` items.

### DIFF
--- a/Src/AutoNSubstituteUnitTest/AutoConfiguredFixtureIntegrationTest.cs
+++ b/Src/AutoNSubstituteUnitTest/AutoConfiguredFixtureIntegrationTest.cs
@@ -415,5 +415,22 @@ namespace Ploeh.AutoFixture.AutoNSubstitute.UnitTest
             // Teardown
         }
 
+        [Fact]
+        public void InterfacesImplementingIEnumerableReturnFiniteSequence()
+        {
+            var fixture = new Fixture();
+            fixture.Customize(new AutoConfiguredNSubstituteCustomization());
+            var repeatCount = fixture.Create<int>();
+            fixture.RepeatCount = repeatCount;
+            var sut = fixture.Create<IMyList<string>>();
+
+            var result = sut.Take(repeatCount + 1).Count();
+
+            Assert.Equal(repeatCount, result);
+        }
+
+        public interface IMyList<out T> : IEnumerable<T>
+        {
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


ploeh:
Thank you for submitting this pull request.

Right now, I'm extraordinarily busy, so may first have time to look at this next week :sweat:

Src/AutoNSubstituteUnitTest/AutoConfiguredFixtureIntegrationTest.cs
```diff
@@ -415,5 +415,20 @@ public void InterfaceShadowedMethodsReturnValueFromFixture()
...
+        [Fact]
+        public void InterfacesImplementingIEnumerableReturnFiniteSequence()
+        {
+            var fixture = new Fixture();
+            fixture.Customize(new AutoConfiguredNSubstituteCustomization());
+            var sut = fixture.Create<IMyList<string>>();
+
+            var result = sut.Take(4).Count();
+
+            Assert.Equal(3, result);
```
ploeh: Why is exactly `3` expected?

mrinaldi: I should have derived the value. Done in 835ae9b.

ploeh:
Thank you for your contribution! It' now live as [AutoFixture.AutoNSubstitute 3.24.1](http://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.24.1), and [AutoFixture 3.24.1](http://www.nuget.org/packages/autofixture/3.24.1).

Issue #98, opened 2013-05-02
Idiomatic assertion for Value Object equality


ploeh:
Testing a Value Object for equality is important, but a rather tedious affair. It should be automated.

As an example, consider a Value Object like Money:

``` c#
public class Money
{
    private readonly string currency;
    private readonly decimal amount;

    // etc.
}
```

It's easy to write a Parameterized Test that verifies that if all fields are equal, then two instance of Money are also equal.

However, you also need to test that if just one of the fields holds different values, the Value Objects are not equal.

Now consider the case where the Value Object
- has constructor overloads
- takes non-primitive parameters
- implements `IEquatable<T>`

and there's a lot of permutations you need to cover before you have thoroughly tested Equals.

It would be nice with a new assertion as part of AutoFixture.Idioms that could test all of that:

``` c#
new ValueObjectEqualityAssertion(fixture).Verify(typeof(Money));
```

Galad:
Hello,
I did some work on this.
7c93cd93fedbbae33715e4de9a70efaa52b937a3

As requested, the assertion is performed for each `ConstructorInfo`, `PropertyInfo` and `FieldInfo`.
It uses the new MatchOnCustomization in order freeze values.

Here is how to use it : 

    var assertion = new EqualityAssertion(new Fixture());
    assertion.Verify(typeof(MyValueClass));

By default it uses Object.Equals, but you can also use a custom comparer :

    var assertion = new EqualityAssertion(new Fixture(), (x,y) => x == y);

When verifying the `ConstructorInfo` it does the following :
It creates 2 same instances with the given constructor, by freezing all the `ParameterInfo`, and verifies that the 2 instances are equal.
It iterates over all the `ParameterInfo` and freeze all except one. Then create 2 instances and verifies that they are not equal.

When verifiying the `PropertyInfo` and `FieldInfo`, it freezes all members of the `ReflectedType` except the given `PropertyInfo` or `FieldInfo`, creates 2 instances and verifies they are not equal.

What do you think of it ?

moodmosaic:
Nice! Could you please open a Pull Request? (It's not so easy to tell, without looking at a Pull Request.)

moodmosaic:
Closing due to inactivity. If you open a pull request, or if you need further help, then write here. We can always re-open the issue.

DixonDs:
[...]

PR #353, opened 2015-02-26
Adding EqualityAssertion.

Galad:
This PR fixes issue #98 

It assert that the given equality comparer behaves correctly.

The assertion is performed for each `ConstructorInfo`, `PropertyInfo` and `FieldInfo`.
It uses the new MatchOnCustomization in order freeze values.

Here is how to use it : 

    var assertion = new EqualityAssertion(new Fixture());
    assertion.Verify(typeof(MyValueClass));

By default it uses Object.Equals, but you can also use a custom comparer :

    var assertion = new EqualityAssertion(new Fixture(), (x,y) => x == y);

When verifying the `ConstructorInfo` it does the following :
It creates 2 same instances with the given constructor, by freezing all the `ParameterInfo`, and verifies that the 2 instances are equal.
It iterates over all the `ParameterInfo` and freeze all except one. Then create 2 instances and verifies that they are not equal.

When verifiying the `PropertyInfo` and `FieldInfo`, it freezes all members of the `ReflectedType` except the given `PropertyInfo` or `FieldInfo`, creates 2 instances and verifies they are not equal.

### DIFF
--- a/Src/Idioms/EqualityAssertion.cs
+++ b/Src/Idioms/EqualityAssertion.cs
@@ -1,0 +1,364 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.Idioms
+{
+	/// <summary>
+	/// Represent an assertion which verifies that an equality method returns the correct value.
+	/// </summary>
+	public class EqualityAssertion : IdiomaticAssertion
+	{
+		private readonly Func<object, object, bool> comparer;
+		private IFixture fixture;
+
+		/// <summary>
+		/// Creates a new instance of EqualityAssertion which verified the Object.Equals method
+		/// </summary>
+		/// <param name="fixture">A composer which can create instances required to implement the idiomatic unit test.</param>
+		public EqualityAssertion(IFixture fixture)
+			: this(fixture, (x, y) => x.Equals(y))
+		{
+		}
+
+		/// <summary>
+		/// Creates a new instance of EqualityAssertion which verified the given equality comparer
+		/// </summary>
+		/// <param name="fixture">A composer which can create instances required to implement the idiomatic unit test.</param>
+		/// <param name="comparer">The equality comparer</param>
+		public EqualityAssertion(IFixture fixture, Func<object, object, bool> comparer)
+		{
+			if (fixture == null) throw new ArgumentNullException("fixture");
+			if (comparer == null) throw new ArgumentNullException("comparer");
+			this.fixture = fixture;
+			this.comparer = comparer;
+		}
+
+		/// <summary>
+		/// Gets the fixture supplied by the constructor
+		/// </summary>
+		public IFixture Fixture
+		{
+			get { return this.fixture; }
+		}
+
+		/// <summary>
+		/// Gets the comparer supplied by the constructor
+		/// </summary>
+		public Func<object, object, bool> Comparer
+		{
+			get { return this.comparer; }
+		}
+
+		/// <summary>
+		/// Verifies that the equality comparer returns the correct value when comparing
+		/// 2 objects created from the given <see cref="ConstructorInfo"/>.
+		/// The comparer should return
+		/// - True when the constructor is invoked with the same arguments
+		/// - False when the constructor is invoked with different arguments
+		/// </summary>
+		/// <param name="constructorInfo">The constructor to verify</param>
+		public override void Verify(ConstructorInfo constructorInfo)
+		{
+			if (constructorInfo == null)
+			{
+				throw new ArgumentNullException("constructorInfo");
+			}
+			var parameters = constructorInfo.GetParameters();
+			var customizations = GetFreezingCustomizations(parameters);
+			var constructorInfoSpecimenBuilder =
+				new Postprocessor(
+					new ConstructorInfoSpecimenBuilder(),
+					new AutoPropertiesCommand()
+					);
+			using (new FixtureCustomizationsDisposable(this.fixture))
+			{
+				this.fixture.Customizations.Add(constructorInfoSpecimenBuilder);
+				this.fixture.Customize(
+					new CompositeCustomization(
+						GetFreezingCustomizations(constructorInfo.ReflectedType)));
+				VerifyEqualReturnsTrue(constructorInfo, customizations);
+				for (var i = 0; i < parameters.Length; ++i)
+				{
+					VerifyEqualReturnsFalse(constructorInfo, parameters, customizations, i);
+				}
+			}
+		}
+
+		/// <summary>
+		/// Verifies that the equality comparer returns the correct value when comparing
+		/// 2 objects where the given property has a different value
+		/// </summary>
+		/// <param name="propertyInfo">The PropertyInfo to verify</param>
+		public override void Verify(PropertyInfo propertyInfo)
+		{
+			if (propertyInfo == null)
+			{
+				throw new ArgumentNullException("propertyInfo");
+			}
+			if (!propertyInfo.CanWrite)
+			{
+				return;
+			}
+			var propertyInfos = GetWriteablePropertyInfo(propertyInfo.ReflectedType)
+											.Where(pi => pi != propertyInfo);
+			var fieldInfos = GetWriteableFieldsInfo(propertyInfo.ReflectedType);
+
+			VerifyMember(propertyInfo, propertyInfos, fieldInfos, propertyInfo.ReflectedType, "property");
+		}
+
+		/// <summary>
+		/// Verifies that the equality comparer returns the correct value when comparing
+		/// 2 objects where the given field has a different value
+		/// </summary>
+		/// <param name="fieldInfo">The FieldInfo to verify</param>
+		public override void Verify(FieldInfo fieldInfo)
+		{
+			if (fieldInfo == null)
+			{
+				throw new ArgumentNullException("fieldInfo");
+			}
+			if (fieldInfo.IsInitOnly)
+			{
+				return;
+			}
+			var propertyInfos = GetWriteablePropertyInfo(fieldInfo.ReflectedType);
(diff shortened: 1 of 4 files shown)


ploeh:
Thank you for submitting this pull request.

Right now, I'm extraordinarily busy, so may first have time to look at this next week :sweat:

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+	{
+		private readonly Func<object, object, bool> comparer;
+		private IFixture fixture;
+
+		/// <summary>
+		/// Creates a new instance of EqualityAssertion which verified the Object.Equals method
+		/// </summary>
+		/// <param name="fixture">A composer which can create instances required to implement the idiomatic unit test.</param>
+		public EqualityAssertion(IFixture fixture)
+			: this(fixture, (x, y) => x.Equals(y))
```
adamchester: It would be more clear if you used an `IEqualityComparer` here. Additionally, we have used this in some [other](https://github.com/AutoFixture/AutoFixture/blob/d02f102301af95908e31dcb0e5c628abe48cbb8b/Src/Idioms/ConstructorInitializedMemberAssertion.cs#L47) [places](https://github.com/AutoFixture/AutoFixture/blob/d02f102301af95908e31dcb0e5c628abe48cbb8b/Src/Idioms/CopyAndUpdateAssertion.cs#L59) already, therefore it would be more consistent.

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+		{
+		}
+
+		/// <summary>
+		/// Creates a new instance of EqualityAssertion which verified the given equality comparer
+		/// </summary>
+		/// <param name="fixture">A composer which can create instances required to implement the idiomatic unit test.</param>
+		/// <param name="comparer">The equality comparer</param>
+		public EqualityAssertion(IFixture fixture, Func<object, object, bool> comparer)
```
adamchester: It would be more clear and consistent if you used an `IEqualityComparer` here.

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+		/// </summary>
+		public IFixture Fixture
+		{
+			get { return this.fixture; }
+		}
+
+		/// <summary>
+		/// Gets the comparer supplied by the constructor
+		/// </summary>
+		public Func<object, object, bool> Comparer
```
adamchester: It would be more clear and consistent if you used an `IEqualityComparer` here.

Galad:
Sure, take your time :)

Src/Idioms/EqualityAssertion.cs
Galad: You are right, I thought there only was a generic version of IEqualityComparer, thanks :)

adamchester:
The fixture modifications and possibly the assumption that all other properties will be different (except the frozen member) troubles me a little.

Is there any chance you could explain this approach a little? Have I read it correctly?

Galad:
Well the assumption is that, when request 2 instance of the tested type, all the members will be the same, except one.

To do this I use the `ConstructorInfo` in order to build the type, with the `ConstructorInfoSpecimenBuilder`, which requests `ParameterInfo` from the fixture and then invokes the `ConstructorInfo` with it.
I also apply the `AutoPropertiesCommand` which will fill the fields and properties, by requesting `PropertyInfo` and `FieldInfo`. 

    new Postprocessor(
        new ConstructorInfoSpecimenBuilder(),
        new AutoPropertiesCommand()
    )

In order to freeze all those members, I create a `FreezeOnMatchCustomization` for each member, which will be statisfied if the request equal the given member (I do the same for the `ParameterInfo` of the constructor.

     new FreezeOnMatchCustomization(
		pi.PropertyType,
		new EqualsRequestSpecification(pi)
     )

Finally, when requesting a type from the `Fixture`, it will create an instance by only requesting `ParameterInfo`, `PropertyInfo` and `FieldInfo`. When creating the 2nd instance, the fixture will return the same values than the 1st instance since it requets frozen members (except for the member not frozen)

Here is an example. Let's say I have the following class.

    public class TestClass
	{
		private readonly int _x;
		public TestClass(int x, int y)
		{
			_x = x;
		}
		public int TestA { get; set; }
		public object TestB { get; set; }
		public string TestC;	
		public int TestX { get { return _x; } }
	}

Here is what I will have when testing the property `TestA`

    1st instance
        TestA: 10
        TestB: "foo"
        TestC: "bar"
        TestX: 40

    2nd instance
        TestA: 20
        TestB: "foo"
        TestC: "bar"
        TestX: 40

I can compare them and verify that the equality comparer return `false`.
This is repeated for each member.

______

Regarding the fixture being modified, that was something I was not sure about too.
After each comparison, the fixture customizations are "reset" to the previous state in order to "unfreeze" all the members. This is done thanks to the `FixtureCustomizationsDisposable`. 

Unfortunaly did not find another way to have this behavior. Maybe by decorating a specimen builder ? The FreezeOnMatchCustomization needs an IFixture to work, but I may have miss something here, and I am open to any suggestion

I hope I was able to answer your question.

adamchester:
Regarding the assumption that each _instance_ created by AutoFixture will be _different_, I can think of scenarios where that is not true. The most obvious scenario is where you have an even number of `bool` properties:

```csharp
class BoolProperties
{
    public bool Flag1 { get; set; }
    public bool Flag2 { get; set; }
}
```
Each time I create a specimen, I get the same values assigned to `Flag1` and `Flag2`.

Similarly, for the following class:
```csharp
class IntProperties
{
    [Range(1, 2)]
    public int Flag1 { get; set; }
    [Range(1, 2)]
    public int Flag2 { get; set; }
}
```
... I always get `1` and `2` values assigned to each instance generated.

adamchester:
Note that I'm not sure this behaviour I just described actually has a negative impact on your test at this stage, since you only need one member to freeze.

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.Idioms
+{
+	/// <summary>
+	/// Represent an assertion which verifies that an equality method returns the correct value.
```
ploeh: Are we hereby implying that the base implementation (reference equality) is incorrect?

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+	/// </summary>
+	public class EqualityAssertion : IdiomaticAssertion
```
ploeh: Is this name appropriate? It asserts equality... what sort of equality?

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+
+		/// <summary>
+		/// Verifies that the equality comparer returns the correct value when comparing
+		/// 2 objects created from the given <see cref="ConstructorInfo"/>.
+		/// The comparer should return
+		/// - True when the constructor is invoked with the same arguments
+		/// - False when the constructor is invoked with different arguments
+		/// </summary>
+		/// <param name="constructorInfo">The constructor to verify</param>
+		public override void Verify(ConstructorInfo constructorInfo)
```
ploeh: Why verify `ConstructorInfo`?

Src/Idioms/EqualityAssertion.cs
```diff
@@ -0,0 +1,364 @@
...
+		/// <param name="propertyInfo">The PropertyInfo to verify</param>
+		public override void Verify(PropertyInfo propertyInfo)
+		{
+			if (propertyInfo == null)
+			{
+				throw new ArgumentNullException("propertyInfo");
+			}
+			if (!propertyInfo.CanWrite)
+			{
+				return;
```
ploeh: Don't we want to verify writeable properties?

GrimaceOfDespair:
I'm sorry to poke in such an old closed issue, but this seemed the only place I could turn to.

Am I correct to state that the merge never took place? Are there any issues with such a general purpose equality assertion? Should we rely on the other idomatic equality assertions instead?

ploeh:
> Am I correct to state that the merge never took place?

That is correct. The pull request was closed by the author. 

#98 remains open, but as the discussion there testament, it's far from trivial.

GrimaceOfDespair:
I understand the peculiarities, but what about explicitly constraining it to a use case, e.g. by naming it EqualityByAllMembersAssertion?

Let me put it this way: I'm actually already using the code from this PR to cover this scenario. And while I understand that you need to maintain a high quality general-purpose lib (thumbs up!) that you cannot just inject any half-baked idea into, I also feel it's a pity that this piece of useful code will just be rotting away in GitHubs PR sink.

ploeh:
When this pull request was originally submitted, we started a review of it. The essential part of a review is that we ask questions about the code. (Sometimes, we may also flat-out state that we want something done in a particular way, but AFAICT, that's not the case here.)

That we ask questions is a perfectly normal part of any pull request submitted to AutoFixture (and many other open source code bases). It does, however, imply that there are issues (however minor) with the pull request.

As long-term maintainers, we want to feel confident that if we accept the pull request we can maintain the code in the future. That's the reason we ask questions.

Sometimes, a dialogue is all it takes, but often, the questions we ask trigger changes in the code.

What happened with this pull request is unusual, because @Galad closed it without comments. This has at least two implications:

- It's not clear to me whether @Galad actually wants us to take the contribution.
- There are still outstanding concerns.

For these reasons, I don't wish to pull the code as is. On the other hand, #98 is still open, and AutoFixture still takes contributions according to [the established procedure](https://github.com/AutoFixture/AutoFixture/blob/master/CONTRIBUTING.md).

GrimaceOfDespair:
Ok, clear. Thanks for your elaborate response.

I guess your story about code quality is exactly the same reason why my own adaptation hasn't made it to a pull request either.

Still, if I find the opportunity, I won't refrain from posting it. Until then: keep up the good work ;)

Issue #354, opened 2015-02-27
AutoFixture's AutoDataAttribute and xUnit's PropertyDataAttribute don't play nice together


alexfoxgill:
As described in [this StackOverflow answer](http://stackoverflow.com/a/18930198/1860652), the two attributes can't be combined in the usual ways.

I have shared my solution [here](http://stackoverflow.com/a/28767887/1860652) - if you're interested in implementing this in AutoFixture.xUnit.net, I could get it into shape for a PR

ploeh:
Thank you for writing, and for sharing your solution. As far as I can tell, that solution is explicitly about combining `[PropertyData]` with `[AutoData]` - is that correct?

It's not clear to me if `CompositeDataAttribute` behaves appropriately. From that Stack Overflow answer, I get the impression that it doesn't, so perhaps it'd be better to address the root cause of the problem...

alexfoxgill:
I think you're right. The body of [`CompositeAttribute.GetData`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/CompositeDataAttribute.cs#L53) looks like it could be adapted to lazily evaluate its children.

I noticed that the method body favours loops over LINQ - shall I stick to this the style or would you accept a more functional-looking solution?

alexfoxgill:
My proposed solution involves:

1. The aforementioned changes to `CompositeDataAttribute` to permit infinite sequences being provided
2. A `RepeatDataAttribute` that decorates a `DataAttribute`, repeating its data infinitely
3. A variation on `AutoDataAttribute` which allows a `Func<IFixture>` to be provided in the constructor. This means that the fixture can be reinitialized on every `GetData` invocation.

Does the above sound sensible?

moodmosaic:
Thank you for reporting this! It might be a bug – though, as @ploeh wrote, it'd be nice if we do a root cause analysis first.

Here's the algorithm of `CompositeDataAttribute`, as described [here](http://blog.nikosbaxevanis.com/2011/08/25/combining-data-theories-in-autofixture-xunit-extension/):

* Delegate the `GetData` method call to the first attribute.
* Loop through the results from the first `DataAttribute` and check if the length of each object array matches the number of arguments required by the method.
  * If so, yield the object array.
  * If not, invoke `GetData` on the next `DataAttribute`.
    * Throw away the first objects from the result in order to continue from where it ran out of objects from the first result set and concatenate the rest to the previous array.
[...]

PR #356, opened 2015-03-01
Composite data attribute - Refactor

alexfoxgill:
As discussed in #354 - submitted for feedback.

The main change is in the body of `CompositeDataAttribute`. I personally found the imperative code quite difficult to follow, and I think the functional alternative is easier to comprehend.

#### Extensions
I added two fairly straightforward helper functions in the `EnumerableExtensions` file:

##### [`Zip()`](/alexfoxgill/AutoFixture/blob/composite-data-attribute/Src/AutoFixture.xUnit.net/EnumerableExtensions.cs#L17)
should probably have been included in the original set of LINQ extensions (IMO), it's pretty handy. 

```CSharp
public static IEnumerable<TResult> Zip<T, TResult>(this IEnumerable<IEnumerable<T>> sequences,
    Func<IEnumerable<T>, TResult> resultSelector)
{
    var enumerators = sequences.Select(s => s.GetEnumerator()).ToList();
    while (enumerators.All(e => e.MoveNext()))
        yield return resultSelector(enumerators.Select(e => e.Current));
}
```

##### [`Collapse()`](/alexfoxgill/AutoFixture/blob/composite-data-attribute/Src/AutoFixture.xUnit.net/EnumerableExtensions.cs#L33)
is perhaps a little more difficult to grasp - this is the function that collapses a series of sequences into a single sequence

```CSharp
public static IEnumerable<T> Collapse<T>(this IEnumerable<IEnumerable<T>> sequences)
{
    var position = 0;
    foreach (var sequence in sequences)
    {
        var skipped = sequence.Skip(position);
        foreach (var item in skipped)
        {
            position++;
            yield return item;
        }
    }
}
```

For example, `[[1], [2,3], [4,5,6]].Collapse()` would yield `[1,3,6]`.

These two utility functions reduce the essence of `CompositeDataAttribute.GetData()` from over 60 lines to just three:

```CSharp
var dataSets = this.attributes
    .Select(attr => attr.GetData(methodUnderTest, parameterTypes))
    .Zip(dataSet => dataSet.Collapse().ToArray());
```

I think this a case where functional is clearer than imperative!

#### Tests

I removed two testcases from `CompositeDataAttributeInsufficientDataTest`. These both verified the rule that the first attribute would set the number of rows for the rest, but this rule now states that the shortest set of data is taken as the limit.

Similarly I removed a testcase from `CompositeDataAttributeSufficientDataTest`. This test provided four object parameters for an arity 3 function and verified that only the first 3 would be returned. I thought that this was slightly bad behaviour, and that if a sub-attribute returns too much data then the `CompositeDataAttribute` should complain, because that's what xUnit does. So I wrote a test for that in the Insufficient test set.

#### InfiniteDataAttribute

This is perhaps something to be looked at separately, but the idea is to combine this with `CompositeDataAttribute` to provide a recurring set of data. So your attribute might look like this:

```CSharp
public class PropertyAutoDataAttribute : CompositeDataAttribute
{
	public PropertyAutoDataAttribute(string propertyName)
		: base(new PropertyDataAttribute(propertyName),
		       new InfiniteDataAttribute(new AutoDataAttribute(new Fixture())))
   {
   }
}
```

**NB this will use the same fixture for every testcase so frozen objects will be resupplied - see  below**

#### Concerns

My only real concerns are the ones I mentioned [here](https://github.com/AutoFixture/AutoFixture/issues/354#issuecomment-76614495):

> Nikos, I'd be wary of supplying the same objects for different testcases. In an immutable world that's fine but there could be unintended consequences of reusing a mock, for example.

> This actually extends to the IFixture instance itself - **if you freeze an object in the fixture, the same instance will be resupplied to every test case, again exposing the user to unintended side-effects**. This is why I suggested a separate DataAttribute that takes a factory function, so it can instantiate the fixture for each test case. I really do believe that it is important to follow the principle of least astonishment, and I would certainly expect new instances to be provided per testcase.

> Mark, I did think that a zip is the most sensible way to do this. My only concern is that the user might accidentally restrict the number of test cases to one, by using a regular AutoDataAttribute. **As we've seen, in the current implementation this throws an InvalidOperationException, thereby warning the user that something is wrong. With a zip there is no error, but only the first testcase is executed.** Do you think that is acceptable? Perhaps the docs could warn the user?

I still haven't written the tests for `InfiniteDataAttribute` so I don't expect this PR to be accepted immediately but I wanted to create it as a starting point for discussion and to explain my suggested changes.


moodmosaic:
>This is why I suggested a separate DataAttribute that takes a factory function, so it can instantiate the fixture for each test case.

This can actually be an idea worth considering. – May I suggest to discuss this separately (e.g. as a new issue)?

Src/AutoFixture.xUnit.net/InfiniteDataAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit.Extensions;
+
+namespace Ploeh.AutoFixture.Xunit
+{
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class InfiniteDataAttribute : DataAttribute
```
moodmosaic: This can actually be an idea worth considering. – May I suggest to discuss this separately (e.g. as a new issue)?

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeInsufficientDataTest.cs
```diff
@@ -74,17 +74,7 @@ public IEnumerator<object[]> GetEnumerator()
             yield return CreateTestCase(
                 data: new[]
                 {
-                    new FakeDataAttribute(this.method, this.parameterTypes, new[] {  new object[] { 1, 2    },  new object[] {  3,  4     }, new object[] { 5, 6 } }),
```
moodmosaic: It's not clear to me why this test case should be removed. :confused: Isn't it equivalent to the following test?

```csharp
public class Tests
{
    public static IEnumerable<object[]> Property1
    {
        get
        {
            yield return new[] { "1", "2" };
            yield return new[] { "3", "4" };
        }
    }

    public static IEnumerable<object[]> Property2
    {
        get
        {
            yield return new[] { "5", "6" };
        }
    }

    [Theory]
    [ComposeTwoProperties("Property1", "Property2")]
    public void DoSomething(object x, object y, object z)
    {
        Assert.NotNull(x);
        Assert.NotNull(y);
        Assert.NotNull(z);
    }
}

internal class ComposeTwoPropertiesAttribute : CompositeDataAttribute
{
    internal ComposeTwoPropertiesAttribute(
        string fst,
        string snd)
        : base(
            new PropertyDataAttribute(fst),
            new PropertyDataAttribute(snd))
    {
    }
}
```

moodmosaic:
:+1: It looks like this Pull Request (PR) is making `CompositeDataAttribute` much better than it currently is.

-----

Here are a few successful tests I've made:

```csharp
public class Tests
{
    public static IEnumerable<object[]> Property1
    {
        get
        {
            yield return new[] { "1", "2" };
            yield return new[] { "3", "4" };
            yield return new[] { "5", "6" };
        }
    }

    public static IEnumerable<object[]> Property2
    {
        get
        {
            yield return new[] { "7", "8", "9" };
        }
    }

    [Theory]
    [ComposeTwoProperties("Property1", "Property2")]
    public void DoSomething(object x, object y, object z)
    {
        Assert.NotNull(x);
        Assert.NotNull(y);
        Assert.NotNull(z);
    }
}
```

Output:

```
x = 1,  y = 2, z = 9
```

Here's also another test:

```csharp
public class Tests
{
    public static IEnumerable<object[]> Property1
    {
        get
        {
            yield return new[] { "1", "2" };
            yield return new[] { "3", "4" };
            yield return new[] { "5", "6" };
        }
    }

    public static IEnumerable<object[]> Property2
    {
        get
        {
            yield return new[] {  "7",  "8",  "9" };
            yield return new[] { "10", "11", "12" };
        }
    }

    [Theory]
    [ComposeTwoProperties("Property1", "Property2")]
    public void DoSomething(object x, object y, object z)
    {
        Assert.NotNull(x);
        Assert.NotNull(y);
        Assert.NotNull(z);
    }
}
```

Output:

```
x = 1, y = 2, z = 9
x = 3, y = 4, z = 12
```

The `ComposeTwoPropertiesAttribute` was defined as:

```csharp
internal class ComposeTwoPropertiesAttribute : CompositeDataAttribute
{
    internal ComposeTwoPropertiesAttribute(
        string fst,
        string snd)
        : base(
            new PropertyDataAttribute(fst),
            new PropertyDataAttribute(snd))
    {
    }
}
```

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeInsufficientDataTest.cs
alexfoxgill: These both test whether the `CompositeDataAttribute` throws an `InvalidOperationException` if the first attribute provides more data sets than the rest. This used to be expected behaviour, but now the exception shouldn't be thrown - instead, the attribute providing the fewest number of data sets determines the number of data sets that are combined + provided to the theory. So I have moved these tests to the `CompositeDataAttributeSufficientDataTest` file, as they now contain sufficient data.

As a side note the `ComposeTwoPropertiesAttribute` described above *should* throw an exception: Neither `Property1` nor `Property2` provide a full set of data, so the third parameter `z` can never be populated

alexfoxgill:
As you see above I have opened two new issues for the other parts of this work. If you like I will remove `InfiniteDataAttribute` from this PR and open a new one?

moodmosaic:
:+1:

Yes, I think it makes sense to remove it, since #357 now contains the same information.

Whether or not you should create a Pull Request (PR) it's up to you. :smiley: – Though, IMO, it'd be beneficial if there's some discussion first around #357, before you put additional effort on a hypothetical PR.

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeInsufficientDataTest.cs
moodmosaic: Correct. – And also the test in the example above fails (I should have mentioned that, I apologize for this).

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeInsufficientDataTest.cs
```diff
@@ -74,17 +74,7 @@ public IEnumerator<object[]> GetEnumerator()
...
-                }
-            );
-
-            yield return CreateTestCase(
-                data: new[]
-                {
-                    new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] {  1,  2     }, new object[] {  3,  4     }, new object[] { 5, 6 } }),
-                    new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] {  7,  8,  9 }, new object[] { 10, 11, 12 }                        }),
-                    new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] { 13, 14, 15 }                                                     })
+                    new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] { 1, 2, 3, 4 } }),
```
moodmosaic: Agreed, xUnit.net data theories throw, if the number of data rows is bigger than the number of test parameters.

Though, in the context of `CompositeDataAttribute`, if the number of data rows is bigger, why not just return a subset?

-----

In other words, given a test with 3 parameters and a data theory with 4 rows:

```
new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] { 1, 2, 3, 4 } })
```

why not just feed the test with `1`, `2`, `3` and throw `4` out? – Since we *do* have some data, why not use it...

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeSufficientDataTest.cs
```diff
@@ -63,17 +63,6 @@ public IEnumerator<object[]> GetEnumerator()
             yield return CreateTestCase(
                 data: new[]
                     {
-                        new FakeDataAttribute(this.method, this.parameterTypes, new[] { new object[] { 1, 2, 3, 4 } })
```
moodmosaic: See [this](https://github.com/AutoFixture/AutoFixture/pull/356/files#r25605375) comment. – Would it make sense to run the test once, and feed it with `1`, `2`, `3`, while throwing `4` out?

moodmosaic:
I left a couple of verbose comments, as I'm trying to figure out if we could go without removing existing tests, for the benefits of having [append-only tests](http://blog.ploeh.dk/2013/04/02/why-trust-tests/) – if that's possible at all in this case.

Src/AutoFixture.xUnit.net.UnitTest/CompositeDataAttributeInsufficientDataTest.cs
alexfoxgill: It's a tricky one. You could ask the same of xUnit - why does it throw an exception, rather than just discard the excess? I think the answer is correctness: if a `DataAttribute` is providing data then it *knows* the type signature of the method and the number of parameters to provide. If you're receiving too many objects then there must be a bug in one of your attributes. If there's a bug, it's better to alert the user than silently ignore it.

Perhaps a compromise is to just return what you are given - don't trim the results, but don't throw an exception either. Let xUnit worry about what to do with an oversized data set.

alexfoxgill:
I agree that it would be nice to leave as many tests alone as possible. See my comment [here](https://github.com/AutoFixture/AutoFixture/pull/356#discussion_r25611719) - what do you think about simply forwarding on whatever the sub-attributes provide, and letting xUnit deal with under- or over-provision of data? It seems like a nice compromise rather than trying to make that decision as the middle-man. To some extent this is an application of the SRP - the Composite attribute should just bring other attributes together. If we want to trim excess perhaps that should be the job of a different attribute (although I can't imagine why you would ever purposefully return too many parameters from a DataAttribute)

moodmosaic:
I think, that's a pretty decent argument. If we *do* that, then we don't need to add [this](https://github.com/AutoFixture/AutoFixture/pull/356/files#diff-51ca684c33c48bce0e0f33b8fc4f7dd4R77) test case, isn't it?

alexfoxgill:
Yes - instead what we should do is remove the tests that check for an exception given too few or too many parameters, and instead replace them with two tests verifying that the input amount of data is returned.

So if we use the standard 3-parameter method, we would have one test which returns two parameters and one that returns four. Both of these should expect not to throw an exception, but to just return whatever they found.

alexfoxgill:
I have updated the two areas that we discussed:

- Removed `InfiniteDataAttribute`, to be discussed in #357 
- Removed bounds checking for `CompositeDataAttribute` (and thus removed tests)

Src/AutoFixture.xUnit.net/EnumerableExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        /// <summary>
+        /// Applies a specified function to the corresponding elements of any number of sequences.
+        /// </summary>
+        /// <typeparam name="T">The type of the elements of the input sequences.</typeparam>
+        /// <typeparam name="TResult">The type of the elements of the result sequence.</typeparam>
+        /// <param name="sequences">The input sequences.</param>
+        /// <param name="resultSelector">A function that specifies how to combine the corresponding elements of the sequences.</param>
+        /// <returns>An <see cref="IEnumerable{T}"/> that contains elements of the input sequences, combined by resultSelector.</returns>
+        public static IEnumerable<TResult> Zip<T, TResult>(this IEnumerable<IEnumerable<T>> sequences,
```
ploeh: Does this have to be public?

Src/AutoFixture.xUnit.net/EnumerableExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Collapses a series of sequences down by using items from the first sequence until it finishes,
+        /// then continuing from the same index through the second sequence, and so on until all sequences
+        /// have been exhausted.
+        /// </summary>
+        /// <typeparam name="T">The type of the elements of the input sequences.</typeparam>
+        /// <param name="sequences">The input sequences.</param>
+        /// <returns>Items from each sequence in turn, yielding those from the first sequence first.</returns>
+        public static IEnumerable<T> Collapse<T>(this IEnumerable<IEnumerable<T>> sequences)
```
ploeh: Does this have to be public?

alexfoxgill:
I think that's everything addressed now

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.24.3](http://www.nuget.org/packages/AutoFixture.Xunit/3.24.3).

moodmosaic:
This is indeed *''(...) a case where functional is clearer than imperative!*'', as @alexfoxgill wrote in the PR description. Thank you! :+1:

alexfoxgill:
Great, thanks guys!

PR #363, opened 2015-03-28
Enable tests in visual studio test runner

adamchester:
by referencing the `xunit.runner.visualstudio` 2.0.0 NuGet package from each unit test project.

Even though the `xunit.runner.visualstudio` is a 2.0 release, it supports discovering and running tests from v1.x.


adamchester:
Using the NuGet VS 2013 UI, the `xunit.runner.visualstudio` package made modifications to the project files which don't seem necessary (therefore I didn't commit them).

The only verification I did was to undo the automatic changes to all the project files, the load each solution and check the VS Test Explorer. In all cases, the tests from all projects were discovered and visible.

adamchester:
@TeaDeivenDev do you have any comments on this? Will it interfere with anything you plan to do in #294?

ploeh:
This seems to not cause any problems on my machine.

However, I'm wondering why only `xunit.runner.visualstudio.2.0.0.nupkg` is checked in. Isn't there a .nuspec file associated with this NuGet package? Or some binaries, for that matter?

adamchester:
> However, I'm wondering why only  xunit.runner.visualstudio.2.0.0.nupkg  is checked in. Isn't there a .nuspec file associated with this NuGet package? Or some binaries, for that matter?

That is an issue I had not noticed yet. Inside the unzipped package folder is a `build` folder, which is explicitly ignored by our `.gitignore`. I will get them checked in.

adamchester:
The missing files have been added to this PR now in d6c1e86.

ploeh:
Thank you for your contribution! No release is associated with this Pull Request, as it contains no changes in any of the published packages.

PR #367, opened 2015-03-31
Compatible version range for xUnit.net in AutoFixture.Xunit.nuspec

### LINKED ISSUE #366: AutoDataAttribute not compatible with xunit 2.0.0.
Hi,

xunit 2.0.0 has been released, and some of the classes in xunit extensions have been moved.
The latest nuget package for AutoFixture.Xunit is not compatible.

moodmosaic:
As it looks like, NuGet supports [specifying version ranges in *.nuspec* files](https://docs.nuget.org/create/versioning).  – This resolves #366. 

--

### Manual verification ###

The version now falls into a specified range:

image

NuGet will prevent updating project(s) using *AutoFixture.Xunit1* to v2 of xUnit.net:

image

### DIFF
--- a/Nuget/Ploeh.AutoFixture.Xunit.nuspec
+++ b/Nuget/Ploeh.AutoFixture.Xunit.nuspec
@@ -8,7 +8,7 @@
         <owners>Mark Seemann</owners>
         <dependencies>
             <dependency id="autofixture" version="2.2" />
-            <dependency id="xunit.extensions" version="1.8.0.1549" />
+            <dependency id="xunit.extensions" version="[1.8.0.1549,2.0.0)" />
         </dependencies>
         <summary>Extension to AutoFixture that integrates it with xUnit.net.</summary>
         <description>By leveraging the data theory feature of xUnit.net, this extension turns AutoFixture into a declarative framework for writing unit tests. In many ways it becomes a unit testing DSL (Domain Specific Language).</description>


ploeh:
Shouldn't it be any version `v`: `1.8.0.1549 <= v && v < 2.0.0`?

moodmosaic:
That's how this can be defined, according to the docs... (Is there a better way?)

moodmosaic:
The proposed change (`[1.8.0.1549,1.9.2]`) means `1.8.0.1549 ≤ x ≤ 1.9.2` according to [how NuGet specifies version ranges for *.nuspec* files](https://docs.nuget.org/create/versioning).

At least, that's the best combination I could spot from the docs, given that [1.9.2 is the latest *stable* version on NuGet prior to 2.0.0](https://www.nuget.org/packages/xunit/).

ploeh:
How about `[1.8.0.1549, 2.0.0)`?

moodmosaic:
Maybe it's just me but, I think, I can't spot that combination below (taken from [here](https://docs.nuget.org/create/versioning)):

```
1.0       = 1.0 ≤ x
(,1.0]    = x ≤ 1.0
(,1.0)    = x < 1.0
[1.0]     = x == 1.0
(1.0)     = invalid
(1.0,)    = 1.0 < x
(1.0,2.0) = 1.0 < x < 2.0
[1.0,2.0] = 1.0 ≤ x ≤ 2.0
empty     = latest version.
```

dcastro:
The constraint proposed by @ploeh is valid though (I've used it before), even though that exact combination isn't on the docs.

moodmosaic:
I'd *prefer* to use the constraint proposed by @ploeh, actually... Though, since it's not documented, could it be possible that NuGet suddenly stops supporting it?

ploeh:
It *is* documented, although not in the example table, but in the *Examples* section just below the table.

moodmosaic:
Oh, why I didn't see it from the beginning? :confused: – [Done](https://github.com/moodmosaic/AutoFixture/commit/6e926d5f85b5f869e9b15e186133f5f8492be60c#diff-7835a83ef78de16148d0ad5a2abbefaaR11) now.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.24.4](http://www.nuget.org/packages/AutoFixture.Xunit/3.24.4).

PR #368, opened 2015-04-03
Allow 0 as a count value

### LINKED ISSUE #365: CreateMany should accept zero
Maybe there's a better way to do this, but here's a scenario:
You want to create a collection of orders with random numbers of order lines.
This seems to be very difficult in AutoFixture. But still, here's an attempt.

It fails because it does not want to make empty collections. I need to explicitly write a case to fix this behavior. I'd almoste be better off populating the members with random data myself:

        private IEnumerable<Order> CreateOrdersWithLines(int collectionSize)
        {
            var fix = new Fixture();
            var lineCount = 0;

            for (var i = 0; i < collectionSize; i++)
            {
                var order= fix
                    .Build<Order>()
                    .With(o => o.Lines, fix.CreateMany<OrderLine>(lineCount).ToArray())
                    .Create();
                lineCount= ritCount++ % 5;
                yield return order;
            }
        }

moodmosaic:
Resolves #365.

### DIFF
--- a/Src/AutoFixture/Kernel/MultipleRelay.cs
+++ b/Src/AutoFixture/Kernel/MultipleRelay.cs
@@ -26,9 +26,9 @@ namespace Ploeh.AutoFixture.Kernel
             get { return this.count; }
             set
             {
-                if (value <= 0)
+                if (value < 0)
                 {
-                    throw new ArgumentOutOfRangeException("value", "Count cannot be zero or negative.");
+                    throw new ArgumentOutOfRangeException("value", "Count cannot be negative.");
                 }
                 this.count = value;
             }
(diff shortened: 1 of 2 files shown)


Src/AutoFixture/Kernel/MultipleRelay.cs
```diff
@@ -26,7 +26,7 @@ public int Count
             get { return this.count; }
             set
             {
-                if (value <= 0)
+                if (value < 0)
                 {
                     throw new ArgumentOutOfRangeException("value", "Count cannot be zero or negative.");
```
ploeh: Is that error message still valid?

moodmosaic: *Holy crap!* It's now [updated](https://github.com/AutoFixture/AutoFixture/pull/368/files#diff-cb78a5c3ba434ba944c25a25d7f6422bR31). – Thank you!

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.24.4](http://www.nuget.org/packages/AutoFixture/3.24.4).

Issue #369, opened 2015-04-05
MockRelay and abstract circular dependencies


dcastro:
Given the following abstract class with a circular dependency:

    public abstract class RecursiveAbstract
    {
        public abstract RecursiveAbstract Prop { get; set; }
    }

The following test passes as expected:

    [Fact]
    public void ResolveResursiveAbstract()
    {
        var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
        fixture.Behaviors.OfType<ThrowingRecursionBehavior>().ToList()
            .ForEach(b => fixture.Behaviors.Remove(b));
        fixture.Behaviors.Add(new OmitOnRecursionBehavior(1));

        fixture.Create<RecursiveAbstract>();
    }

However, the following test fails:

    [Fact]
    public void ResolveMockOfResursiveAbstract()
    {
        var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
        fixture.Behaviors.OfType<ThrowingRecursionBehavior>().ToList()
            .ForEach(b => fixture.Behaviors.Remove(b));
        fixture.Behaviors.Add(new OmitOnRecursionBehavior(1));

        fixture.Create<Mock<RecursiveAbstract>>();
    }

```
Ploeh.AutoFixture.ObjectCreationExceptionAutoFixture was unable to create an instance from Ploeh.AutoFixture.AutoMoq.UnitTest.QuickTest+RecursiveAbstract, most likely because it has no public constructor, is an abstract or non-public type.
```

(Note: at the moment, the user can work around the issue by using `Mock.Get(fixture.Create<RecursiveAbstract>())` instead)

To find out where the problem was, I followed the request path and found this:
```
SeededRequest(typeof(Mock<RecursiveAbstract>), null)
 -> typeof(Mock<RecursiveAbstract>)
 -> PropertyInfo "Prop"
 -> SeededRequest(PropertyInfo "Prop", "Prop")
 -> typeof(RecursiveAbstract)
 -> typeof(Mock<RecursiveAbstract>)
```

That last request, is made by [MockRelay (line 83)](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoMoq/MockRelay.cs#L83)

```csharp
public object Create(object request, ISpecimenContext context)
{
    //...
    var t = request as Type;
    //...

    var m = MockRelay.ResolveMock(t, context);
    if (m == null)
        return new NoSpecimen(request);

    return m.Object;
}

[...]

PR #370, opened 2015-04-05
Addresses Issue 369

dcastro:
This PR addresses issue #369.

I realized that the test cases in issue #369 also failed when the `OmitOnRecursionBehavior` is replaced by `NullRecursionBehavior`.

Therefore, not only does `MockRelay` need to "propagate" `OmitSpecimen`s returned by the underlying `ISpecimenContext`, it also needs to propagate `null` specimens. This PR covers both use cases.

---

Regarding regression tests: I was thinking I should copy the test cases presented in the issue ticket into the test suite.
These tests aren't unit tests, but more like integration/regression tests.

Is there a history of this being done in the past?
I found two cases, in the `Ploeh.AutoFixtureUnitTest.AbstractRecursionIssue` and `Ploeh.AutoFixtureUnitTest.NavigationPropertyRecursionIssue` namespaces, but I'm not sure this is a standard or whether this issue warrants regression tests.

### DIFF
--- a/Src/AutoMoqUnitTest/MockRelayTest.cs
+++ b/Src/AutoMoqUnitTest/MockRelayTest.cs
@@ -125,6 +125,42 @@ namespace Ploeh.AutoFixture.AutoMoq.UnitTest
             // Teardown
         }
 
+        [Fact]
+        public void CreateReturnsNoSpecimenWhenContextReturnsNoSpecimen()
+        {
+            // Fixture setup
+            var request = typeof (IInterface);
+            var mockType = typeof (Mock<>).MakeGenericType(request);
+
+            var contextStub = new Mock<ISpecimenContext>();
+            contextStub.Setup(ctx => ctx.Resolve(mockType)).Returns(new NoSpecimen());
+
+            var sut = new MockRelay();
+            // Exercise system
+            var result = sut.Create(request, contextStub.Object);
+            // Verify outcome
+            Assert.IsType<NoSpecimen>(result);
+            // Teardown
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


Src/AutoMoqUnitTest/MockRelayTest.cs
```diff
@@ -125,6 +126,35 @@ public void CreateReturnsCorrectResultWhenContextReturnsNonMock()
...
+            get
+            {
+                yield return new object[] {new NoSpecimen()};
+                yield return new object[] {new OmitSpecimen()};
+                yield return new object[] {null};
+            }
+        }
+
+        [Theory]
+        [PropertyData("ValidNonMockSpecimens")]
```
ploeh: Could you please turn this into a `[ClassData]`, so that it's not stringly typed?

ploeh:
IIRC, the reason I added those repros was that in both cases, I received a bug report, and I wasn't entirely sure that I understood the problem, so I added a repro to make sure that I actually ended up fixing the reported defect.

In this case, it seems you understand the problem very well, so for my sake an integration test repro isn't *required* - but not *prohibited* either. It's up to you :smile:

Src/AutoMoqUnitTest/MockRelayTest.cs
dcastro: Done, 0331889 :) I confess I wasn't even aware of the `ClassData` attribute!

dcastro:
I agree, the problem is not as complex and well understood, the unit tests provide good enough insurance ;)

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoMoq 3.24.5](http://www.nuget.org/packages/AutoFixture.AutoMoq/3.24.5).

PR #371, opened 2015-04-06
Missing Fact attributes

TeaDrivenDev:
I came across a few tests that were missing their [Fact] attributes. I have added those, and the tests pass.

The first commit also cleans up some stray whitespace in the affected file; this is incidental but I guess doesn't do any harm.


moodmosaic:
Thank you! Looks good to me :+1:

ploeh:
Thank you for catching that :+1: 

Since this is a test-only contribution, no published version is associated with it.

PR #373, opened 2015-04-07
Create solution for xUnit.net v2 support

TeaDrivenDev:
This adds a solution and empty project files for the xUnit v2 glue library and its test project. The solution has the same structure as that for xUnit v1 support. No existing files are affected.

### DIFF
--- a/BuildRelease.msbuild
+++ b/BuildRelease.msbuild
@@ -104,7 +104,7 @@
         <ItemGroup>
           <Xunit1TestAssemblies
               Include="Src\*Test\bin\$(Configuration)\*Test.dll"
-              Exclude="Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
+              Exclude="Src\AutoFixture.xUnit.net2.UnitTest\bin\$(Configuration)\Ploeh.AutoFixture.Xunit2.UnitTest.dll;Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
         </ItemGroup>
 
         <xunit
--- a/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
@@ -1,0 +1,64 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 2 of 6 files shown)


moodmosaic:
Thank you! The solution and project structure are looking good. – Though, running `build-release` fails:

```
Build FAILED.

"..\AutoFixture\BuildRelease.msbuild" (default target) (1) ->
(Test target) -> 
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    System.ArgumentException: Unknown test framework:
      Could not find xunit.dll or xunit.execution.desktop.dll.
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    Parameter name: ..\AutoFixture\Src\AutoFixture.xUnit.net2.UnitTest\
                      bin\Release\Ploeh.AutoFixture.Xunit2.UnitTest.dll
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    at Xunit.XunitFrontController.CreateInnerController()
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    at Xunit.XunitFrontController.get_InnerController()
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    at Xunit.XunitFrontController.Find(Boolean includeSourceInformation, ..
  ..\AutoFixture\BuildRelease.msbuild(110,9): error :
    at Xunit.Runner.MSBuild.xunit.ExecuteAssembly(String assemblyFileName, ..

    0 Warning(s)
    5 Error(s)
```

TeaDrivenDev:
Sorry, the new test project of course needs to be excluded from running the xUnit v1 tests. Done now.

moodmosaic:
It's looking good to me, and the `build-release` script runs fine now :+1:

BuildRelease.msbuild
```diff
@@ -104,7 +104,7 @@
         <ItemGroup>
           <Xunit1TestAssemblies
               Include="Src\*Test\bin\$(Configuration)\*Test.dll"
-              Exclude="Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
+              Exclude="Src\AutoFixture.xUnit.net2.UnitTest\bin\$(Configuration)\Ploeh.AutoFixture.Xunit2.UnitTest.dll;Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
```
ploeh: Can you elaborate on why this exclusion is necessary and how you intend to move forward?

While I did see the interchange regarding the build issue, it only informs me of the superficial reason for excluding the new unit test library.

However, I thought that the reason to add the new xUnit.net 2 runners (in a previous pull request) was to add support for also running xUnit.net 2 unit test libraries. Will this be possible with the build-release script, and if so, why not do whatever it takes first?

TeaDrivenDev:
The problem that the runner reported here is that no version of xUnit is referenced in the test project. This is because I tried to stay close to @moodmosaic's original plan from #294 and keep adding the projects and the xUnit v2 NuGet packages in two separate steps.

To answer your question, running the v2 tests would have been added with a separate `xunit` task. I have now dug a bit deeper and found that the separation in running v1 and v2 tests (that I likely picked up from xUnit's own build script) is not necessary - if the v2 test project does not cause errors.

I could adjust the approach like this:
- Add the xUnit v2 NuGet packages in a separate commit in this pull request
- Run all xUnit v1 and v2 tests in the same task right away, which works perfectly fine
- Skip the proposed "Install xUnit v2 packages" and "Integrate into build process" pull requests
- Add xUnit v2 glue library NuGet package creation as the very last pull request, as per my question in #375

ploeh:
@TeaDrivenDev, thank you for explaining. The easiest way forward now is, I assume, just to move forward with this PR, assuming that the exclusion is later removed when it's no longer necessary (if I understand your explanation correctly).

ploeh:
Thank you for your contribution! Since this is a test-only contribution, there's no published version associated with it.

PR #377, opened 2015-04-07
Add xUnit.net v2 packages

TeaDrivenDev:
This pull request adds the necessary xUnit v2 packages to the glue library and test projects. The test project uses the full `xunit` package, while the glue library only needs the reduced `xunit.core` package.

This means the xUnit MSBuild runner won't fail on the test project anymore; therefore I have also changed the MSBuild file to not exclude the project and run all xUnit tests in the same task, as outlined in #373. The new test project now shows up in the test runner output, but has not tests to be run yet.


ploeh:
Shouldn't the Glue Library itself only reference `xunit.extensibility.core`, instead of `xunit.core`, as [described in the xUnit.net documentation](http://xunit.github.io/docs/nuget-packages.html)?

TeaDrivenDev:
You are certainly correct; I had not seen that. I have updated the pull request accordingly.

ploeh:
Thank you for your contribution! Since this doesn't change any published code, there's no published version associated with it.

ploeh:
**Note:** after this PR was accepted, the build server has started failing the test `Ploeh.AutoFixture.AutoMoq.UnitTest.StubPropertiesCommandTest.DoesNotHangIfMockedTypeHasPropertiesWithCircularDependencies`:

```
Assert.True() Failure
   at Ploeh.AutoFixture.AutoMoq.UnitTest.StubPropertiesCommandTest.DoesNotHangIfMockedTypeHasPropertiesWithCircularDependencies() in z:\Builds\work\cba225c22a6f60ef\Src\AutoMoqUnitTest\StubPropertiesCommandTest.cs:line 66
```

(That's all the build server tells me.)

So far, I've been unable to reproduce this behaviour on my local machine.

TeaDrivenDev:
There was a note about potentially failing parallel tests on the example I lifted the `xunit` build task invocation from. Can you try setting the `MaxParallelThreads` value in the `Test` target to 1 (or `ParallelizeTests` to `false`)?

ploeh:
Thanks for the tip. Trying with `ParallelizeTests` set to `false` now...

ploeh:
That seems to have helped. Thank you, @TeaDrivenDev

PR #378, opened 2015-04-07
Added CustomizeAttribute and tests

TeaDrivenDev:
This pull request adds the `CustomizeAttribute` class (and its tests), which is used by almost all other classes in the glue library. The code is identical with the xUnit v1 version, with namespace adjustments.

### DIFF
--- a/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
@@ -51,6 +51,9 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="CustomizeAttributeTest.cs" />
+    <Compile Include="DelegatingCustomization.cs" />
+    <Compile Include="DelegatingCustomizeAttribute.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/Src/AutoFixture.xUnit.net2.UnitTest/CustomizeAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/CustomizeAttributeTest.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using Xunit;
+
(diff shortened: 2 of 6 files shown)


ploeh:
On my local machine, I'm getting one compiler warning when running `build-release.sh`:

```
"c:\Users\Mark\Documents\ploeh\Common\AutoFixture\BuildRelease.msbuild" (default target) (1) ->
"c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixture.xUnit.net2.sln" (default target) (20) ->
"c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixture.xUnit.net2\AutoFixture.xUnit.net2.csproj" (default target) (21) ->
(CoreCompile target) ->
  CustomizeAttribute.cs(8,20): warning CS1574: XML comment on 'Ploeh.AutoFixture.Xunit2.CustomizeAttribute' has cref attribute 'AutoDataAttribute' that could not be reso
lved [c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixture.xUnit.net2\AutoFixture.xUnit.net2.csproj]

    1 Warning(s)
    0 Error(s)
```

BTW, the new library should get a *Verify* build mode, and treat warnings as errors as soon as possible.

TeaDrivenDev:
Oh, sorry; I actually wanted to mention that, but forgot it. The class references `AutoDataAttribute` in an XML comment, which will only be added in the next pull request. As that is guaranteed to happen, I thought it wasn't necessary to remove the comment here and then restore it.

Is it okay if I add the `Verify` configuration and the flag in the pull request that adds NuGet package creation?

ploeh:
It's OK to add the `Verify` configuration and flag in a later pull request, but I'd like to request that we have a clean build without errors or warnings.

The reason is that while you and I know that you intend to address the warning later, someone else may come by and decide to clone the repository in order to make a contribution, and then that person will have a version of the repository where *build-release* isn't clean.

That seems a bit bureaucratic, I admit, but is a tax we pay for working in an environment where several people can work concurrently.

TeaDrivenDev:
I understand. I have updated the pull request; build runs without warnings now.

ploeh:
Thank you for your contribution! Since there's no published artefact associated with this pull request, no version is associated with it.

Keep up the great work :+1:

Issue #374, opened 2015-04-07
Make EnumeratorRelay a default residue collector


dcastro:
Should the `EnumeratorRelay` be added to the [default collection of residue collectors](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Fixture.cs#L116)?

I was planning on addressing the same issue as @mrinaldi did on PR #351, but for AutoMoq. I was going to simply add this line to both `AutoMoqCustomization` and `AutoConfiguredMoqCustomization`:

    fixture.ResidueCollectors.Add(new EnumeratorRelay());

But then it hit me: why not make it part of the default collection? It's just as important as the existing  `EnumerableRelay`. It's a cross-cutting concern, it's not tied to any specific mocking framework, or even mocking at all, and it would solve this issue for all the mocking glue libraries, including AutoFoq and AutoRhinoMock.

ploeh:
Yes, that sounds like a good idea. Great observation :+1:

ploeh:
Resolved in AutoFixture 3.26.0.

PR #379, opened 2015-04-08
Added EnumeratorRelay to the default collection of residue collectors

dcastro:
As discussed in #374, this PR adds the `EnumeratorRelay` created as part of #351 to the `ResidueCollectorNode`.

This way, all mocking glue libraries will return finite sequences for abstract types implementing `IEnumerable<T>`.

### DIFF
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -118,7 +118,8 @@ namespace Ploeh.AutoFixture
                                 new DictionaryRelay(),
                                 new CollectionRelay(),
                                 new ListRelay(),
-                                new EnumerableRelay())),
+                                new EnumerableRelay(),
+                                new EnumeratorRelay())),
                         new FilteringSpecimenBuilder(
                             new MutableValueTypeWarningThrower(),
                             new AndRequestSpecification(
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -5145,6 +5145,7 @@ namespace Ploeh.AutoFixtureUnitTest
         }
 
         [Theory]
+        [InlineData(typeof(EnumeratorRelay))]
         [InlineData(typeof(EnumerableRelay))]
         [InlineData(typeof(ListRelay))]
         [InlineData(typeof(CollectionRelay))]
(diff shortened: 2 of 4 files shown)


Src/AutoNSubstituteUnitTest/AutoConfiguredFixtureIntegrationTest.cs
```diff
@@ -427,23 +427,5 @@ public void InterfaceShadowedMethodsReturnValueFromFixture()
             Assert.Same(frozenString, (result as IInterfaceWithShadowedMethod).Method(0));
             // Teardown
         }
-
-        [Fact]
-        public void InterfacesImplementingIEnumerableReturnFiniteSequence()
```
ploeh: Why delete this test? Doesn't it represent a feature that we want to ensure keeps working?

dcastro: My reasoning to delete it was that, now that `Fixture` is able to create well-behaved enumerators on its own, this test only checks that `IMyList<T>.GetEnumerator` is correctly setup by the `AutoConfiguredNSubstituteCustomization`. Since there's nothing special about the method itself - `IMyList<T>.GetEnumerator` is just another method - this behavior is already covered by other tests.

But I can also see why keeping the test in would help as a piece of documentation and as a functional/regression test as well.

Whichever way you wanna go, I'll be happy to do it.

ploeh: While I understand that, under the hood, this test ends up invoking `GetEnumerator`, that's doesn't look like it's the purpose of this test. What I'm trying to say is that this unit test seems to operate on another abstraction level than the tests that explicitly go for `IEnumerator<T>`.

It looks to me like this is a slightly more behavioural test that's not particularly concerned with the implementation details of `GetEnumerator`. Perhaps this test was added to reproduce a defect. Isn't it worth to keep such a test around as a regression test?

dcastro: I agree, and that's a very good reason to keep it in. I'll revert this change.

dcastro: Done - force pushed to delete the last commit.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.26.0](http://www.nuget.org/packages/AutoFixture/3.26.0).

PR #380, opened 2015-04-08
Add AutoDataAttribute and tests for xUnit.net v2

TeaDrivenDev:
This pull request adds the `AutoDataAttribute` (and its tests) for xUnit v2. The code is identical to the v1 version, with the following exceptions:

- Namespaces adjustments
- xUnit's `DataAttribute.GetData()` does not take a second parameter of `Type[]` anymore
- Corrected typo in an XML comment

Also, the XML comment in `CustomizeAttribute` that caused the build warning in the previous pull request has been reinstated, as `AutoDataAttribute` now exists.


ploeh:

At this point, we actually have some functionality, so if we were to add a *Verify* build configuration and getting all of these other things set up, do we have a package we could publish?

TeaDrivenDev:
That is the next pull request I have lined up. I'm not sure whether it makes sense to already publish a package just with this, as it will hardly be enough to port *any* existing test suite out there to xUnit v2. But yes, this is the first "useful" piece of code, so that's your call.

ploeh:
I don't know if it's my call :flushed: You're doing the work, after all, and I think that counts for quite a lot!

You are probably correct that most test suites using *AutoFixture.Xunit* will need more than just `[AutoData]`, but that's just an assumption on our parts. We don't know if it it's true. What if someone out there is sitting and *waiting* for *AutoFixture.Xunit2*, and that's the *only* thing he or she needs?

As Dan North says, if software's not being used, it's worth nothing. If we don't publish it, we know for certain that it's worth nothing (so far). If we publish it, it's potentially worth *something*.

That was my motivation for getting something out early, but I also don't want to throw a wrench into your plans, so proceed in whichever way you find is the best - I trust you to find a good way :sunglasses:

PR #381, opened 2015-04-08
Integrate xUnit.net v2 glue library into build process

TeaDrivenDev:
This pull request does three things:

- Add the dependency constraints tests to make sure the glue library and test project don't reference third party assemblies used only in other AutoFixture projects; as per @moodmosaic's comment in the original pull request #294, I have updated these to check for all the different libraries for which glue projects exist
- Enable building a NuGet package for the glue library
- Add a `Verify` configuration to the glue library project and set the "Treat warnings as errors" flag

### DIFF
--- a/Nuget/Ploeh.AutoFixture.Xunit2.nuspec
+++ b/Nuget/Ploeh.AutoFixture.Xunit2.nuspec
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+    <metadata>
+        <id>AutoFixture.Xunit2</id>
+        <version>2.2</version>
+        <title>AutoFixture with xUnit.net v2 data theories</title>
+        <authors>Mark Seemann</authors>
+        <owners>Mark Seemann</owners>
+        <dependencies>
+            <dependency id="autofixture" version="2.2" />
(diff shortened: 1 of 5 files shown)


Nuget/Ploeh.AutoFixture.Xunit2.nuspec
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+    <metadata>
+        <id>AutoFixture.Xunit2</id>
+        <version>2.2</version>
+        <title>AutoFixture with xUnit.net v2 data theories</title>
+        <authors>Mark Seemann</authors>
```
ploeh: @TeaDrivenDev, you are welcome to replace my name with yours here, if you'd like!

TeaDrivenDev: Thank you; as the code is entirely taken from the v1 project and I'm not adding anything new, that doesn't seem appropriate, though. It's still your creation. :-)

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit2 3.25.0](http://www.nuget.org/packages/AutoFixture.Xunit2/3.25.0).

PR #385, opened 2015-04-08
Add CompositeDataAttribute to xUnit.net v2 glue library

TeaDrivenDev:
This pull request adds the `CompositeDataAttribute` class to the xUnit v2 glue library.

The differences from the xUnit v1 project are as follows:
- Namespace adjustments
- xUnit's `DataAttribute.GetData()` does not take a second parameter of `Type[]` anymore
- xUnit 2 removed `ClassDataAttribute`, which the tests for `CompositeDataAttribute` use. According to @bradwilson (https://twitter.com/bradwilson/status/585529188515479552), there is no straight-up replacement/alternative for that, so I opted for lifting the class from xUnit v1 and including it in the test project for the time being.

### DIFF
--- a/Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
+++ b/Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using Xunit.Sdk;
+
+namespace Ploeh.AutoFixture.Xunit2
+{
+    /// <summary>
+    /// An implementation of DataAttribute that composes other DataAttribute instances.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    [CLSCompliant(false)]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", Justification = "This attribute is the root of a potential attribute hierarchy.")]
+    public class CompositeDataAttribute : DataAttribute
+    {
+        private readonly IReadOnlyCollection<DataAttribute> attributes;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CompositeDataAttribute"/> class.
+        /// </summary>
+        /// <param name="attributes">The attributes representing a data source for a data theory.
+        /// </param>
+        public CompositeDataAttribute(IReadOnlyCollection<DataAttribute> attributes)
+            : this(attributes.ToArray())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CompositeDataAttribute"/> class.
+        /// </summary>
+        /// <param name="attributes">The attributes representing a data source for a data theory.
+        /// </param>
+        public CompositeDataAttribute(params DataAttribute[] attributes)
+        {
+            if (attributes == null)
+            {
+                throw new ArgumentNullException("attributes");
+            }
+
+            this.attributes = attributes;
+        }
+
+        /// <summary>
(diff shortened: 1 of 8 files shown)


Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
```diff
@@ -0,0 +1,77 @@
...
+    public class CompositeDataAttribute : DataAttribute
+    {
+        private readonly IEnumerable<DataAttribute> attributes;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CompositeDataAttribute"/> class.
+        /// </summary>
+        /// <param name="attributes">The attributes representing a data source for a data theory.
+        /// </param>
+        public CompositeDataAttribute(IEnumerable<DataAttribute> attributes)
```
ploeh: Since this library is already on .NET 4.5, I think `attributes` should be `IReadOnlyCollection<DataAttribute>` instead, for [this reason](http://blog.ploeh.dk/2013/07/20/linq-versus-the-lsp).

ploeh:
This looks good to me, but we've reached an area of the code base where I'm not the expert. If @moodmosaic or @alexfoxgill would care to glance at this, I would be grateful.

*build-release.sh* reports that all is good on my machine.

Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
TeaDrivenDev: That applies only to the instance field, not the constructor parameter, right?

moodmosaic:
It's looking good to me. – The difference with v1 (as @TeaDrivenDev noticed) is the xUnit.net's `GetData` method overload. The (adjusted) unit tests are also looking good to me.

*build-release.sh* reports that all is good on my machine, too.

Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
moodmosaic: IMO, the whole point of this is to apply it *at least* at the constructor parameter (because that's where the (potential bad) input may come from). :smiley:

moodmosaic:
I've also created locally an `InlineAutoDataAttribute` (which doesn't exist, yet) based on this branch and it worked fine. :+1:

Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
TeaDrivenDev: Updated to require `IReadOnlyCollection<DataAttribute>` in the constructor.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit2 3.29.0](http://www.nuget.org/packages/AutoFixture.Xunit2/3.29.0).

Src/AutoFixture.xUnit.net2/CompositeDataAttribute.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// by DataAttributes in ascending order. Data already returned is ignored on next
+        /// DataAttribute returned data.
+        /// </summary>
+        /// <param name="methodUnderTest">The method that is being tested.</param>
+        /// <returns>
+        /// Returns the composition of the theory data.
+        /// </returns>
+        /// <remarks>
+        /// The number of test cases is set from the first DataAttribute theory length.
+        /// </remarks>
```
alexfoxgill: This used to be the case but isn't any more - see https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/CompositeDataAttribute.cs#L63

alexfoxgill:
Sorry I came in late for this, my only comment is about a comment :) otherwise looks good!

TeaDrivenDev:
@alexfoxgill That one slipped through; thanks for catching it. I have sent another pull request to correct it.

Issue #348, opened 2015-02-19
MailAddressGenerator should use context to resolve name


ploeh:
As described in [this comment](https://github.com/AutoFixture/AutoFixture/pull/340/files#r23569387), `MailAddressGenerator` should delegate the creation of the name part of an email address to the `ISpecimenContext`.

There are two options for this:

- Ask the context for a `string`. This is the simplest solution, but could theoretically be problematic if the user has changed the string generation algorithm to create strings incompatible with the STMP address format.
- Ask the context for an instance of a hypothetical signal type like `EmailAddressNamePart`. We should look up the correct terminology for talking about the part of an email address that goes before the @ sign. If we take this route, we'd need to add this `EmailAddressNamePart` type, and make sure that AutoFixture can resolve it. This is similar to the implementation of how AutoFixture creates URLs.

ploeh:
The implementation of this feature would be an enhancement of #340.

dlongest:
Per RFC 3696, which I believe is controlling on the issue, the part before the @ is the `local-part`.  The RFC defines an email address as `local-part@domain`.  http://tools.ietf.org/pdf/rfc3696.pdf   RFC 5322 also mentions it.  So I submit `EmailAddressLocalPart` as the proper term in place of `EmailAddressNamePart`

ploeh:
`EmailAddressLocalPart` sounds good :+1: 

Thank you for investigating :smile:

dlongest:
Unless someone else is already doing it, I'd love to work on this one and the related #349 although I'll do them separately. I'm inclined to use the 2nd approach suggested by @ploeh with a typed request for the email local part instead of string.

ploeh:
As far as I know, no-one else is working on this, so please go ahead :+1: 

That 2nd approach is also my preferred solution.

moodmosaic:
In case it might helps, [`UriScheme`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/UriScheme.cs) works in a similar way when creating a `Uri`:

```csharp
var fixture = new Fixture();
fixture.Inject(new UriScheme("ftp"));
var uri = fixture.CreateAnonymous<Uri>(); 
// Prints -> ftp://bsdf6546-1df4-5d56-c7e9-1456dd226bd4/
```

Probably the implementation for `EmailAddressLocalPart` could be similar to `UriScheme`'s.

PR #386, opened 2015-04-08
Added EmailAddressLocalPart and EmailAddressLocalPartGenerator.

dlongest:
Here's my first pass at #348.  Quick synopsis:

- Added `EmailAddressLocalPart` and `EmailAddressLocalPartGenerator` using `UriScheme` as a model
- `EmailAddressLocalGenerator` fetches a string from the `ISpecimenContext` to use in creating the local part; that's how `UriGenerator` seems to do it, wasn't sure if that was preferable to using standalone local part creation logic to avoid a case when the context's string behavior has been altered
- Added unit tests for each component
- Updated `MailAddressGenerator` to request `EmailAddressLocalPart` from the context and modified/added tests
- Updated `DefaultPrimitiveBuilders` to include `EmailAddressLocalPartGenerator` and its related test

I did add a bit of validation to `EmailAddressLocalPart` based on the RFC specification, basically length and a regular expression I cited in the comments that evaluates it for character positioning.  I wasn't sure if all of that is desired or not.  

I ran build-release.sh and it completed without errors.  My only other question is whether I should add a test to `FixtureTest` or not.  There is one there for `MailAddressGenerator`, not sure if that's sufficient or we want a separate one.

### DIFF
--- a/Src/AutoFixture/EmailAddressLocalPart.cs
+++ b/Src/AutoFixture/EmailAddressLocalPart.cs
@@ -7,33 +7,29 @@ using System.Text.RegularExpressions;
 namespace Ploeh.AutoFixture
 {
     /// <summary>
-    /// Represents the local part of the email address, defined as everything up to, but not including the @ sign.  
-    /// The requirements enforced on the local part are defined in RFC 3696.  The regular expression for validating 
-    /// the local part borrowed from Phil Haack at 
-    /// http://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/
+    /// Represents the local part of the email address, defined as everything up to, but not including, the @ sign.  
+    /// Since EmailAddressLocalPart is used in constructing MailAddress, enforcement of rules on a valid email address
+    /// is performed by <see cref="System.Net.Mail.MailAddress"/> and not EmailAddressLocalPart other than as noted. 
     /// </summary>
-    public class EmailAddressLocalPart : IEquatable<EmailAddressLocalPart>
+    public class EmailAddressLocalPart
     {
-        private static string pattern = @"^(?!\.)(""([^""\r\\]|\\[""\r\\])*""|"
-                                               + @"([-A-Za-z0-9!#$%&'*+/=?^_`{|}~]|(?<!\.)\.)*)(?<!\.)$";
         private readonly string localPart;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EmailAddressLocalPart"/> class. Throws ArgumentNullException
-        /// if localPart is null or empty.  Throws ArgumentException if localPart is longer than 64 characters or
-        /// does not conform to the local pat requirements per RFC 3696. 
+        /// if localPart is null.  Throws ArgumentException if localPart is empty.
         /// </summary>
-        /// <param name="localPart">The local part.</param>        
+        /// <param name="localPart">The local part.</param>                
         public EmailAddressLocalPart(string localPart)
         {
-            if (string.IsNullOrEmpty(localPart))
+            if (localPart == null)
             {
-                throw new ArgumentNullException(localPart);
+                throw new ArgumentNullException("localPart");
             }
 
-            if (!EmailAddressLocalPart.IsValid(localPart))
+            if (localPart.Length == 0)
             {
-                throw new ArgumentException("Email address local part can only contain letters, digits, and special characters !#$%&'*+/=?^_`{|}~");
+                throw new ArgumentException("localPart cannot be empty");
             }
 
             this.localPart = localPart;
@@ -45,21 +41,6 @@ namespace Ploeh.AutoFixture
         public string LocalPart
         {
             get { return this.localPart; }
-        }
-
-        /// <summary>
-        /// Indicates whether the current object is equal to another object of the same type.
-        /// </summary>
-        /// <param name="other">An object to compare with this object.</param>
-        /// <returns>
-        /// true if the current object is equal to the other parameter; otherwise, false.
-        /// </returns>
-        public bool Equals(EmailAddressLocalPart other)
-        {
-            if (other == null)
-                return false; 
-
-            return this.localPart.Equals(other.localPart);
         }
 
         /// <summary>
@@ -79,7 +60,7 @@ namespace Ploeh.AutoFixture
             var other = obj as EmailAddressLocalPart;
             if (other != null)
             {
-                return this.Equals(other);
+                return this.localPart.Equals(other.localPart);
             }
 
             return base.Equals(obj);
@@ -106,37 +87,6 @@ namespace Ploeh.AutoFixture
         public override string ToString()
         {
             return this.localPart;
-        }       
-
-        /// <summary>
-        /// Returns true if the localPart is valid in terms of both length and conformity with allowable
-        /// characters and character ordering, false otherwise. 
-        /// </summary>
-        /// <param name="localPart"></param>
-        /// <returns></returns>
-        public static bool IsValid(string localPart)
-        {
-            if (string.IsNullOrEmpty(localPart))
-                return false;
-
-            return localPart.Length <= MaximumAllowableLength && Regex.IsMatch(localPart, ValidEmailAddressPattern);
-        }
-
-        /// <summary>
-        /// Gets the maximum length, inclusive, supported as the local part. 
-        /// </summary>
-        public static int MaximumAllowableLength
-        {
-            get { return 64; }
-        }
-
-        /// <summary>
-        /// Gets the pattern used by EmailAddressLocalPart to validate local parts.  This pattern is used
-        /// to validate for character content, but not for length.
-        /// </summary>
-        public static string ValidEmailAddressPattern
-        {
-            get { return pattern; }
         }
     }
 }
--- a/Src/AutoFixture/EmailAddressLocalPart.cs
+++ b/Src/AutoFixture/EmailAddressLocalPart.cs
@@ -1,0 +1,142 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Represents the local part of the email address, defined as everything up to, but not including the @ sign.  
+    /// The requirements enforced on the local part are defined in RFC 3696.  The regular expression for validating 
+    /// the local part borrowed from Phil Haack at 
+    /// http://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/
+    /// </summary>
+    public class EmailAddressLocalPart : IEquatable<EmailAddressLocalPart>
+    {
+        private static string pattern = @"^(?!\.)(""([^""\r\\]|\\[""\r\\])*""|"
+                                               + @"([-A-Za-z0-9!#$%&'*+/=?^_`{|}~]|(?<!\.)\.)*)(?<!\.)$";
+        private readonly string localPart;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailAddressLocalPart"/> class. Throws ArgumentNullException
+        /// if localPart is null or empty.  Throws ArgumentException if localPart is longer than 64 characters or
+        /// does not conform to the local pat requirements per RFC 3696. 
+        /// </summary>
+        /// <param name="localPart">The local part.</param>        
+        public EmailAddressLocalPart(string localPart)
+        {
+            if (string.IsNullOrEmpty(localPart))
+            {
+                throw new ArgumentNullException(localPart);
+            }
+
+            if (!EmailAddressLocalPart.IsValid(localPart))
+            {
+                throw new ArgumentException("Email address local part can only contain letters, digits, and special characters !#$%&'*+/=?^_`{|}~");
+            }
+
+            this.localPart = localPart;
+        }
+
+        /// <summary>
+        /// Get the local part.
+        /// </summary>
+        public string LocalPart
+        {
+            get { return this.localPart; }
+        }
+
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// true if the current object is equal to the other parameter; otherwise, false.
+        /// </returns>
+        public bool Equals(EmailAddressLocalPart other)
+        {
+            if (other == null)
+                return false; 
+
+            return this.localPart.Equals(other.localPart);
+        }
+
+        /// <summary>
+        /// Determines whether the specified <see cref="System.Object"/> is equal to this instance.
+        /// </summary>
+        /// <param name="obj">The <see cref="System.Object"/> to compare with this instance.
+        /// </param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="System.Object"/> is equal to this instance; 
+        /// otherwise, <c>false</c>.
+        /// </returns>
+        /// <exception cref="T:System.NullReferenceException">
+        /// The <paramref name="obj"/> parameter is null.
+        ///   </exception>
+        public override bool Equals(object obj)
+        {
+            var other = obj as EmailAddressLocalPart;
+            if (other != null)
+            {
+                return this.Equals(other);
+            }
+
+            return base.Equals(obj);
+        }
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data 
+        /// structures like a hash table. 
+        /// </returns>
+        public override int GetHashCode()
+        {
+            return this.localPart.GetHashCode();
+        }
+
+        /// <summary>
+        /// Returns a <see cref="System.String"/> that represents the local part for this instance.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="System.String"/> that represents the local part for this instance.
+        /// </returns>
+        public override string ToString()
+        {
+            return this.localPart;
+        }       
+
+        /// <summary>
+        /// Returns true if the localPart is valid in terms of both length and conformity with allowable
+        /// characters and character ordering, false otherwise. 
+        /// </summary>
+        /// <param name="localPart"></param>
+        /// <returns></returns>
+        public static bool IsValid(string localPart)
+        {
+            if (string.IsNullOrEmpty(localPart))
+                return false;
+
+            return localPart.Length <= MaximumAllowableLength && Regex.IsMatch(localPart, ValidEmailAddressPattern);
+        }
+
+        /// <summary>
+        /// Gets the maximum length, inclusive, supported as the local part. 
+        /// </summary>
+        public static int MaximumAllowableLength
+        {
+            get { return 64; }
+        }
+
+        /// <summary>
+        /// Gets the pattern used by EmailAddressLocalPart to validate local parts.  This pattern is used
+        /// to validate for character content, but not for length.
+        /// </summary>
+        public static string ValidEmailAddressPattern
+        {
+            get { return pattern; }
+        }
+    }
+}
--- a/Src/AutoFixture/MailAddressGenerator.cs
+++ b/Src/AutoFixture/MailAddressGenerator.cs
@@ -44,11 +44,7 @@ namespace Ploeh.AutoFixture
             try
             {
                 return TryCreateMailAddress(request, context);
-            }
-            catch (ArgumentNullException)
-            {
-                return new NoSpecimen(request);
-            }
+            }           
             catch (ArgumentException)
             {
                 return new NoSpecimen(request);
--- a/Src/AutoFixture/MailAddressGenerator.cs
+++ b/Src/AutoFixture/MailAddressGenerator.cs
@@ -36,10 +36,21 @@ namespace Ploeh.AutoFixture
                 return new NoSpecimen(request);
             }
 
-            var name = Guid.NewGuid().ToString();
-            var host = fictitiousDomains[(uint)name.GetHashCode() % 3];
+            if (context == null)
+            {
+                return new NoSpecimen(request);
+            }
+                    
+            var localPart = context.Resolve(typeof(EmailAddressLocalPart)) as EmailAddressLocalPart;
 
-            var email = string.Format(CultureInfo.InvariantCulture, "{0} <{0}@{1}>", name, host);
+            if (localPart == null)
+            {
+                return new NoSpecimen(request);
+            }
+
+            var host = fictitiousDomains[(uint)localPart.GetHashCode() % 3];
+
+            var email = string.Format(CultureInfo.InvariantCulture, "{0} <{0}@{1}>", localPart, host);
             return new MailAddress(email);
         }
     }
--- a/Src/AutoFixture/MailAddressGenerator.cs
+++ b/Src/AutoFixture/MailAddressGenerator.cs
@@ -44,11 +44,7 @@ namespace Ploeh.AutoFixture
             try
             {
                 return TryCreateMailAddress(request, context);
-            }           
-            catch (ArgumentException)
-            {
-                return new NoSpecimen(request);
-            }
+            }                    
             catch (FormatException)
             {
                 return new NoSpecimen(request);
--- a/Src/AutoFixture/MailAddressGenerator.cs
+++ b/Src/AutoFixture/MailAddressGenerator.cs
@@ -31,16 +31,36 @@ namespace Ploeh.AutoFixture
         /// </remarks>
         public object Create(object request, ISpecimenContext context)
         {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
             if (!typeof(MailAddress).Equals(request))
             {
                 return new NoSpecimen(request);
             }
 
-            if (context == null)
+            try
+            {
+                return TryCreateMailAddress(request, context);
+            }
+            catch (ArgumentNullException)
             {
                 return new NoSpecimen(request);
             }
-                    
+            catch (ArgumentException)
+            {
+                return new NoSpecimen(request);
+            }
+            catch (FormatException)
+            {
+                return new NoSpecimen(request);
+            }
+        }
+
+        private object TryCreateMailAddress(object request, ISpecimenContext context)
+        {
             var localPart = context.Resolve(typeof(EmailAddressLocalPart)) as EmailAddressLocalPart;
 
             if (localPart == null)
@@ -53,5 +73,5 @@ namespace Ploeh.AutoFixture
             var email = string.Format(CultureInfo.InvariantCulture, "{0} <{0}@{1}>", localPart, host);
             return new MailAddress(email);
         }
-    }
+    }       
 }
--- a/Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
+++ b/Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
@@ -1,0 +1,165 @@
+﻿using Ploeh.AutoFixture;
+using Ploeh.AutoFixture.Kernel;
+using Ploeh.AutoFixtureUnitTest.Kernel;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Ploeh.AutoFixtureUnitTest
+{
+    public class EmailAddressLocalPartGeneratorTest
+    {
+        [Fact]
+        public void SutIsSpecimenBuilder()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new EmailAddressLocalPartGenerator();
+            // Verify outcome
+            Assert.IsAssignableFrom<ISpecimenBuilder>(sut);
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithNullRequestReturnsCorrectResult()
+        {
+            // Fixture setup
+            var sut = new EmailAddressLocalPartGenerator();
+            // Exercise system
+            var dummyContext = new DelegatingSpecimenContext();
+            var result = sut.Create(null, dummyContext);
+            // Verify outcome
+            Assert.Equal(new NoSpecimen(), result);
+            // Teardown
+        }
+
+        [Fact]
+        public void CreateWithNullContextThrows()
+        {
+            // Fixture setup
+            var sut = new EmailAddressLocalPartGenerator();
+            var dummyRequest = new object();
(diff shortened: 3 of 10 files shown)


Src/AutoFixture/EmailAddressLocalPart.cs
```diff
@@ -0,0 +1,142 @@
...
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Represents the local part of the email address, defined as everything up to, but not including the @ sign.  
+    /// The requirements enforced on the local part are defined in RFC 3696.  The regular expression for validating 
+    /// the local part borrowed from Phil Haack at 
+    /// http://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/
+    /// </summary>
+    public class EmailAddressLocalPart : IEquatable<EmailAddressLocalPart>
```
ploeh: You're probably patterning the use of `IEquatable<T>` on older parts of AutoFixture where, I must admit, I hadn't fully researched the purpose of this interface, or the contract surrounding it.

As far as I've been able to determine so far, this interface is only used in certain generic collections in the BCL, and the main purpose is to prevent boxing when comparing value types. Thus, as [the documentation](https://msdn.microsoft.com/en-us/library/ms131187.aspx) states:

*"For a value type, you should always implement IEquatable<T>"*

However, for reference types like this one, it's a lot less useful, although there may be a microscopic performance improvement to be gained. However, the question is whether it warrants the extra complexity.

Do note that the documentation furthermore states:

- *"you should also override the base class implementations of Object.Equals(Object) and GetHashCode"*
- *"you should overload the op_Equality and op_Inequality operators"*
- *"you should also implement IComparable<T> if instances of your type can be ordered or sorted"*

Given all that, I'd be more inclined to *not* implement this interface at all. `EmailAddressLocalPart` can be a great Value Object without it.

AutoFixture's legacy use of `IEnumerable<T>` is all my fault, but once one realises one's errors, one should attempt to rectify them :flushed:

Src/AutoFixture/EmailAddressLocalPart.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// Initializes a new instance of the <see cref="EmailAddressLocalPart"/> class. Throws ArgumentNullException
+        /// if localPart is null or empty.  Throws ArgumentException if localPart is longer than 64 characters or
+        /// does not conform to the local pat requirements per RFC 3696. 
+        /// </summary>
+        /// <param name="localPart">The local part.</param>        
+        public EmailAddressLocalPart(string localPart)
+        {
+            if (string.IsNullOrEmpty(localPart))
+            {
+                throw new ArgumentNullException(localPart);
```
ploeh: If the argument is empty, an `ArgumentNullException` is thrown..?

Is that appropriate behaviour?

Src/AutoFixture/EmailAddressLocalPart.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Represents the local part of the email address, defined as everything up to, but not including the @ sign.  
+    /// The requirements enforced on the local part are defined in RFC 3696.  The regular expression for validating 
+    /// the local part borrowed from Phil Haack at 
+    /// http://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/
```
ploeh: Funny article, but I'm not even sure that this regular expression is liberal enough... Here are some examples that all evaluate to false - I wonder if at least one shouldn't be allowed:

```
> regex.IsMatch "ndøh@example.com";;
val it : bool = false
> regex.IsMatch "ndöh@example.com";;
val it : bool = false
> regex.IsMatch "åhnej@example.com";;
val it : bool = false
> regex.IsMatch "ñoñó1234@server.com";;
val it : bool = false
> regex.IsMatch "f#ers.com";;
val it : bool = false
```

(Unless I did something wrong when I copy-pasted the `pattern` string - which can't *entirely* be ruled out...)

Src/AutoFixture/EmailAddressLocalPart.cs
```diff
@@ -0,0 +1,142 @@
...
+        public static int MaximumAllowableLength
+        {
+            get { return 64; }
+        }
+
+        /// <summary>
+        /// Gets the pattern used by EmailAddressLocalPart to validate local parts.  This pattern is used
+        /// to validate for character content, but not for length.
+        /// </summary>
+        public static string ValidEmailAddressPattern
```
ploeh: Isn't this an implementation detail?

What if one day we'd like to refactor this implementation to use parser combinators instead of regular expressions?

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: Yes, you're correct, I saw that `UriScheme` implements `IEquatable` and presumed that was the desired pattern.  I wasn't aware of the larger points around it so I'll just remove that relationship.  Is it still a sound practice to define a typed `Equals` method i.e.. to leave `Equals(EmailAddressLocalPart other` in place?

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: Considering you covered this in your blog http://blog.ploeh.dk/2014/11/18/the-isnullorwhitespace-trap/ definitely not, a failure on my part.  I will modify it accordingly to only throw `ArgumentNullException` if the argument is truly null, otherwise `ArgumentException`.

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: To be honest, I didn't consider the possibility of an alternate implementation.  I had to look up parser combinators, I can see your point.  I will remove the public property that exposes the pattern.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -36,10 +36,21 @@ public object Create(object request, ISpecimenContext context)
                 return new NoSpecimen(request);
             }
 
-            var name = Guid.NewGuid().ToString();
-            var host = fictitiousDomains[(uint)name.GetHashCode() % 3];
+            if (context == null)
+            {
+                return new NoSpecimen(request);
```
ploeh: This is perhaps a little confusing, but while `request` is a bit special when it comes to null handling, `context` follows the [standard *design-by-contract* principles laid out by Bertrand Meyer](http://amzn.to/1claOin):

A method's *preconditions* describe what the method *needs* in order to do its work. Its *postconditions* provide a guarantee about the state of the system once the method returns - if the precondition was satisfied.

In this case, the postcondition should be that a `MailAddress` should be returned if the preconditions are met. In order to create a `MailAddress`, the new implementation *needs* the `context`. Thus, if the `context` is null, the precondition isn't satisfied.

That's an error.

Returning `NoSpecimen` means that the builder doesn't *want* to handle that particular `request`; it doesn't mean that it *can't*.

As I'm trying to explain this, I realize that this seems rather subtle, but that's the way the rest of AutoFixture works, so if the `context` is null, it should throw an `ArgumentNullException`.

dlongest: That makes total sense and if I had looked more closely, `UriGenerator` adheres to that policy as well of throwing `ArgumentNullException` when the context is null since it requires the context, but other generators don't evaluate it at all if they don't make use of it.   The build script caught this when I modified `MailAddressGenerator` and I missed that connection between the precondition and the exception.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -36,10 +36,21 @@ public object Create(object request, ISpecimenContext context)
...
+            if (context == null)
+            {
+                return new NoSpecimen(request);
+            }
+                    
+            var localPart = context.Resolve(typeof(EmailAddressLocalPart)) as EmailAddressLocalPart;
+
+            if (localPart == null)
+            {
+                return new NoSpecimen(request);
```
ploeh: (This implementation is correct - don't change it.)

Here it becomes really strange, but as far as AutoFixture is concerned, `null` is actually a valid return value... :sweat: That's not by any of my standard design principles, but because AutoFixture is attempting to be useful for more people than only me, and sometimes people want to be able to configure AutoFixture to return null in certain cases.

(This is the underlying reason that the `NoSpecimen` signal type exists in the first place. A Builder can't use `null` to signal that it doesn't want to handle the request, because it would then be impossible to distinguish between a `null` returned as an opt-out, and `null` returned as a *value*.)

If the returned `localPart` is null, it's not an error of the *caller* of the Builder, so at this point the preconditions are still satisfied: it would be wrong to throw an exception. Instead, it's correct to return `NoSpecimen` so that a potential fallback builder can get a chance to handle the request.

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        public void CreateReturnsEmailAddressLocalPartUsingLocalPartReceivedFromContext()
+        {
+            // Fixture setup
+            var request = typeof(EmailAddressLocalPart);
+            string expectedLocalPart = Guid.NewGuid().ToString();
+            var context = new DelegatingSpecimenContext
+            {
+                OnResolve = r =>
+                {
+                    if (typeof(string).Equals(r))
```
ploeh: How do we feel about [Conditional Test Logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html)?

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
```diff
@@ -0,0 +1,165 @@
...
+            // Fixture setup
+            var request = typeof(EmailAddressLocalPart);
+            string contextLocalPart = Guid.NewGuid().ToString() + Guid.NewGuid().ToString();
+            string expectedLocalPart = contextLocalPart.Substring(0, 64);
+
+            var context = new DelegatingSpecimenContext
+            {
+                OnResolve = r =>
+                {
+                    if (typeof(string).Equals(r))
```
ploeh: How do we feel about [Conditional Test Logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html)?

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            // Fixture setup
+            var request = typeof(EmailAddressLocalPart);
+            var invalidLocalPart = "@@Invalid";
+
+            var context = new DelegatingSpecimenContext
+            {
+                OnResolve = r =>
+                {
+                    if (typeof(string).Equals(r))
```
ploeh: How do we feel about [Conditional Test Logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html)?

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: So in terms of the local part only and the `EmailAddressLocalPart` constructor, I expect the first 4 to fail.  The local part cannot contain an @ sign and those do. While otherwise presumably legal email addresses, they don't conform to the requirement for the local part specifically and that's what `EmailAddressLocalPart` validates for.   When I tested it, "f#ers.com" is accepted as a valid local part as # is allowed and the .com has no particular significance to the local part.  

It occurs to me that perhaps it would be useful to have a way to create `EmailAddressLocalPart` from a complete email address as right now it would have to be done by hand. 

All the above having been said, you're right that the regular expression overall is not liberal enough in that it fails for the characters you included that are outside of the [A-Za-z0-9] set.  However, I'm unsure if the RFC actually allows for the addresses you posted and if it doesn't, what we should do about it.   I came across http://isemail.info/about which seems to have spent an intense amount of time studying the issue and when I pasted in the first 4 addresses you provided, that site rules all of them as invalid based on the non-US ASCII characters.  

Assuming the rules as enforced by that site are truly exact to the RFC, the larger question is: what should we do about it?  Presumably you've seen many email addresses that are seemingly invalid based on the RFC alone yet work just fine in the wild, and that seems to be because email vendors can and do support addresses that are not conformant.  If we don't use the RFC as-is or feel it's overly restrictive, we could consider relaxing our rules in some way (allowing non-US ASCII characters, for example).

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
```diff
@@ -0,0 +1,169 @@
...
+using Xunit.Extensions;
+
+namespace Ploeh.AutoFixtureUnitTest
+{
+    public class EmailAddressLocalPartTest
+    {
+        [Fact]
+        public void InitializeWithNullLocalPartThrows()
+        {
+            var sut = new EmailAddressLocalPart("good.localPart");
```
ploeh: Where is `sut` used?

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        [Fact]
+        public void ToStringReturnsCorrectResult()
+        {
+            // Fixture setup
+            string expected = "good.localPart";
+            var sut = new EmailAddressLocalPart("good.localPart");
+            // Exercise system
+            var result = sut.ToString();
```
ploeh: Could you pass this test by implementing `ToString` by always returning the hard-coded string "good.localPart"?

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
```diff
@@ -0,0 +1,169 @@
...
+            Assert.False(result);
+            // Teardown
+        }
+
+        [Fact]
+        public void SutDoesNotEqualNullSut()
+        {
+            // Fixture setup
+            var sut = new EmailAddressLocalPart("good.localPart");
+            UriScheme other = null;
```
ploeh: `UriScheme` isn't the same type as the SUT, so doesn't this test exercise the exact same path as `SutDoesNotEqualNullObject`?

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            // Fixture setup
+            var sut = new EmailAddressLocalPart("good.localPart");
+            var other = new EmailAddressLocalPart("good2.localPart");
+            // Exercise system
+            bool result = sut.Equals(other);
+            // Verify outcome
+            Assert.False(result);
+            // Teardown
+        }
```
ploeh: Could you pass this entire test suite by always returning `false` from `Equals`?

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -50,11 +50,36 @@ public void CreateWithNonMailAddressRequestReturnsCorrectResult()
...
         [Fact]
-        public void CreateWithMailAddressRequestReturnsCorrectResult()
+        public void CreatesNoSpecimenWhenContextIsNull()
         {
             // Fixture setup
             var request = typeof(MailAddress);
-            var context = new DelegatingSpecimenContext();
+            var sut = new MailAddressGenerator();
+            //Exercise system
+            var result = sut.Create(request, null);
```
ploeh: Should throw... (see other comment).

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -50,11 +50,36 @@ public void CreateWithNonMailAddressRequestReturnsCorrectResult()
...
+        public void CreateWithMailAddressRequestReturnsCorrectResultUsingLocalPartFromContext()
+        {
+            // Fixture setup
+            var request = typeof(MailAddress);
+            var expectedLocalPart = new EmailAddressLocalPart(Guid.NewGuid().ToString());
+            var context = new DelegatingSpecimenContext()
+            {
+                OnResolve = r =>
+               {
+                   if (typeof(EmailAddressLocalPart).Equals(r))
```
ploeh: How do we feel about [Conditional Test Logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html)?

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -62,5 +87,32 @@ public void CreateWithMailAddressRequestReturnsCorrectResult()
...
+        public void CreateWithMailAddressRequestReturnsNoSpecimenWhenContextReturnsNullLocalPart()
+        {
+            // Fixture setup
+            var request = typeof(MailAddress);
+            var expectedLocalPart = new EmailAddressLocalPart(Guid.NewGuid().ToString());
+            var context = new DelegatingSpecimenContext()
+            {
+                OnResolve = r =>
+                {
+                    if (typeof(EmailAddressLocalPart).Equals(r))
```
ploeh: How do we feel about [Conditional Test Logic](http://xunitpatterns.com/Conditional%20Test%20Logic.html)?

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
dlongest: In general should be avoided.  I realize I pulled this from I believe `UriGeneratorTest` and there it seems that conditional logic might be acceptable as the `UriGenerator` actually calls the context twice for two different types, but only one context can be provided to drive it.  In this case, there's only a single call to the context being made so there's no need for the conditional there.  That's true for all those in this module, I'll change them to simply return the expected value with no conditionals.

ploeh:
This is absolutely on the right track :+1: 

While I did leave quite a few comments, they are all details that can be quite easily addressed, so this PR just needs a little more love, and then it's ready for prime time :smile:

Src/AutoFixture/EmailAddressLocalPart.cs
ploeh: > Is it still a sound practice to define a typed Equals method

I usually don't do it, because I can't see that it serves any purpose (but I wouldn't mind to be proven wrong if you have good arguments).

Src/AutoFixture/EmailAddressLocalPart.cs
ploeh: Oh, dear, I messed it up :flushed: I should only have tested against the local part, but in my eagerness, I typed in whole email address.

If you test "f#ers.com" as valid, it probably means that I did an error when copy-pasting the regular expression...

The overall question, however, is whether we actually care if the local part is correct?

Why do we even need to validate that?

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
ploeh: Not so fast :stuck_out_tongue_winking_eye: 

We still want to [verify the data flow](http://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries). The test case here is that the `Resolve` method should *only* return `expectedLocalPart` if `r` is `typeof(string)`. Otherwise, the SUT has made an error calling the context. That data flow must still be protected.

Can you think of other ways to do that that doesn't involve conditional test logic?

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: It depends on how we view the utility of `EmailAddressLocalPart` within AutoFixture.  If we were writing a library for validating email addresses, it may make sense to do this level of validation.  However, as we've seen, even deciding to do that isn't exactly simple since there can definitely exist email addresses in use that aren't valid per the RFC so seems there's no actual standard that could be applied that wouldn't be overly restrictive.  

Within AutoFixture, I would think we'd use the local part more loosely to essentially mean "some token on the left side of the @ sign" or else we could make it very difficult to actually use `EmailAddressLocalPart`.  So I propose we have some basic rules, but otherwise allow a great deal of latitude in its content.  I'm thinking something along the lines of length <= 64, cannot contain @ sign, cannot contain whitespace characters, but otherwise any other character is fine.  I don't think I would even enforce the rules around periods.  These proposed rules are arbitrary, somewhat based on what I perceive as the core of the local part spec, but very liberal in terms of the character content.  We could also have no rules whatsoever other than not allowing it to be null or blank.  I'm open to however you'd like this to go.

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
dlongest: I read the blog post you linked, but to be honest I'm not sure how to do it without a conditional at all in this case.  I see that in the `DelegatingSpecimenContext` as I've set it up, it returns `NoSpecimen` when the type isn't string when I now think it should return null.  We can use the ? operator to express the condition as `r => typeof(string).Equals(r) ? expectedLocalPart : null` but that still uses a conditional.  In your blog post, the Mock appears to have the conditional logic implicit within it, which I think is the larger point being made, but without coding that into the `DelegatingSpecimenContext`, I'm a bit stuck.

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
dlongest: Yes, so I think your point is to not use a fixed expected value in this test.  I replaced that with a Guid.

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
dlongest: Was actually an oversight, `UriScheme` should have been `EmailAddressLocalPart` there.  Even with that change though, I don't think both tests are needed as they both exercise the same code path since the conditional within `Equals` is false regardless of the type if the object is set to null.  The other case (where a non-null non-`EmailAddressLocalPart` instance is sent to `Equals`) is also already covered, by `SutDoesNotEqualAnonymousObject` so I'm just going to remove this particular test.

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
dlongest: Adding a test named `SutEqualsOtherSutWhenLocalPartsAreEqual` that will test the other side of this case:  that two separate `EmailAddressLocalPart` instances with the same localPart will evaluate to equal.

Src/AutoFixture/EmailAddressLocalPart.cs
ploeh: If we take a step back and try to remember *why* we're adding this `EmailAddressLocalPart` class, *where* are we going to use it?

Src/AutoFixtureUnitTest/EmailAddressLocalPartGeneratorTest.cs
ploeh: You could add a [Guard Assertion](http://xunitpatterns.com/Guard%20Assertion.html) like this:

```C#
var context = new DelegatingSpecimenContext
{
    OnResolve = r =>
    {
        Assert.Equal(typeof(string), r);
        return expectedLocalPart;
    }
};
```

Very theoretically: this actually turns a [Stub](http://xunitpatterns.com/Test%20Stub.html) into a [Mock](http://xunitpatterns.com/Mock%20Object.html), which is [isn't really appropriate](http://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries), but is, IMO, the lesser of two evils, because in this case, the code is easier to reason about, and the Mock is very localised.

Src/AutoFixtureUnitTest/EmailAddressLocalPartTest.cs
ploeh: The question is whether `sut.Equals((EmailAddressLocalPart)null)` calls *the same method* as `sut.Equals((object)null)`?

Src/AutoFixture/EmailAddressLocalPart.cs
dlongest: We're adding it as we want to request `EmailAddressLocalPart` in the `MailAddressGenerator` instead of requesting a string because we want the mail address specimen to be in conformity with the SMTP format and we can't guarantee the logic for string generation hasn't been modified to be incompatible with that format.  That seems like an argument for having `EmailAddressLocalPart` perform validation on the local part although I could be completely missing what you're driving at.

ploeh: Good start :smile: 

And inside of `MailAddressGenerator`, how do we intend to use the `EmailAddressLocalPart`?

dlongest: We're going to combine it with a domain name and ultimately call `new MailAddress(...)` where `MailAddress` is the type `System.Net.Mail.MailAddress`.  And actually looking into that class, I see it will throw a `FormatException` if the provided email address is not in the proper format, which seems to include non-US ASCII characters.  Digging into that a bit, `MailAddressParser` actually handles the validation logic for `MailAddress`, unfortunately it's internal so can't be used as-is.  But seeing that we're planning to use `EmailAddressLocalPart` to construct `MailAddress` and given that the latter will validate the email address, I think we are obliged to have `EmailAddressLocalPart` enforce the same rules if it's to be useful.

ploeh: Bingo! :tada: 

Yes, it's unfortunate that the validation used by `MailAddress` is internal, but it also means that for our specific case, it's irrelevant to attempt to match 'the standard' - what we need is to match something that will work with `MailAddress`.

Personally, I don't think it's worthwhile to attempt to reverse-engineer the `MailAddress` validation rules, as they could change (e.g. because of bug fixes). AFAICT, the most pragmatic approach is to attempt to use the string passed into `EmailAddressLocalPart` with a valid domain name ("example.com") in a `try/catch` block. If no exception is thrown, it must be valid.

This will effectively couple `EmailAddressLocalPart` to `MailAddressGenerator`, so this may be one of those fairly rare cases where a nested public class may be appropriate...

moodmosaic: Using something that will work with `MailAddress` class for validation [seems to be a reasonable approach](http://stackoverflow.com/a/1374644). :+1:

dlongest: @ploeh When I move `EmailAddressLocalPart` to be a public inner class of `MailAddressGenerator`, the code analysis in the BuildRelease script throws an error as it only allows non-public inner classes.  Do you feel strongly enough that we should nest `EmailAddressLocalPart` that we should suppress the rule in this case or do you take that as evidence we should keep it in the outer scope despite the coupling?  For whatever it's worth, I'll note that the MSDN article on that warning advises to not suppress it - https://msdn.microsoft.com/en-us/library/ms182162.aspx

ploeh: Perhaps I was a bit too fast on the trigger there :flushed: 

I think I've misremembered the [Framework Design Guidelines on this topic](https://msdn.microsoft.com/en-us/library/ms229027.aspx), but now that I read them again, they seem to be aligned with the CA rule.

So, despite the coupling, it may be most appropriate to keep them unnested. Sorry about the inconvenience.

dlongest:
I think I've addressed all the feedback that was provided.  As discussed in one of the comments, `EmailAddressLocalPart` remains a non-inner class and we rely on `MailAddress` to validate the email specimen produced in `MailAddressGenerator`.  The BuildRelease script executes with no errors for me.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -31,16 +31,47 @@ public class MailAddressGenerator : ISpecimenBuilder
...
                 return new NoSpecimen(request);
             }
 
-            var name = Guid.NewGuid().ToString();
-            var host = fictitiousDomains[(uint)name.GetHashCode() % 3];
+            try
+            {
+                return TryCreateMailAddress(request, context);
+            }
+            catch (ArgumentNullException)
```
ploeh: What happens if you delete this `catch`?

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -31,16 +31,47 @@ public class MailAddressGenerator : ISpecimenBuilder
...
+            {
+                return new NoSpecimen(request);
+            }
+            catch (ArgumentException)
```
ploeh: What happens if you delete this `catch`?

dlongest: It still works correctly, `ArgumentNullException` derives from `ArgumentException` so here we should only catch the latter.

Src/AutoFixture/MailAddressGenerator.cs
```diff
@@ -31,16 +31,43 @@ public class MailAddressGenerator : ISpecimenBuilder
...
                 return new NoSpecimen(request);
             }
 
-            var name = Guid.NewGuid().ToString();
-            var host = fictitiousDomains[(uint)name.GetHashCode() % 3];
+            try
+            {
+                return TryCreateMailAddress(request, context);
+            }           
+            catch (ArgumentException)
```
ploeh: Which error case does this `catch` block handle? When would `TryCreateMailAddress` throw an `ArgumentException`?

ploeh:
This is embarrassing, but it seems like this perfectly fine pull request has been sitting here, neglected, for half a year :flushed: 

It's all my fault, and I apologise sincerely. Having tried being on the other end of a project that ignored my pull request for months, I know how disappointing that is.

Not that the following is any excuse, but GitHub doesn't send me notification emails when a pull request fork is updated, or new commits appear on it. I still should have checked back, but it might be good to know.

ploeh:
That you for your contribution! It's now live as [AutoFixture 3.33.0](https://www.nuget.org/packages/AutoFixture/3.33.0).

Once again, I'm sorry about the neglect!

Another time (if you'd ever stoop so low as to work with me again), please kick me with a notification if I've been silent for too long!

dlongest:
No reason to apologize at all as I made a few missteps in this one as well.  First, I usually add a comment when I update a PR, but in this one case I didn't and I had no idea GitHub doesn't send notifications in such a case so definitely learned something there.  Second, the early versions of the PR weren't particularly sharp and so that wasted quite a bit of your time in reviews than I had wanted/intended so I felt bad bugging you about it.  In the future I'll know to check in just in case (notifications can be missed, etc).  

Contributing to this project is a real pleasure for me because I've found AutoFixture to be such a great tool and I've learned so much from you over the years that it's the least I can do to try to contribute something back.  Thanks for being so patient and I would be happy to work with you again when I get some time :-)

ploeh:
For the record, I don't expect pull requests to be *perfect*; I'm quite happy to work with contributors to take a pull request the last stretch :smile:

PR #391, opened 2015-04-09
Correct typo in an XML comment

TeaDrivenDev:
This pull request corrects the (one-character) typing error in an XML comment that I already corrected in the xUnit 2 glue library.


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.30.1](http://www.nuget.org/packages/AutoFixture.Xunit/3.30.1).

PR #393, opened 2015-04-11
Disable xUnit2 'pre-discovery'

### LINKED ISSUE #392: incorrect test names when using autodata and xunit.testrunner.visualstudio
Hi,

I have updated to xunit 2 - thanks for all the work on this, it's great.
I am still having trouble with `autodata`, and `xunit.testrunner.visualstudio`.

Given a test like this:
```
        [Theory, AutoData]
        public void TestLastCommittedDsMessageDictionary(int x, int y)
        {
        {
```

In the visual studio test explorer the test is listed with its **name** containing an x and y value.
LastCommittedMessageTests.TestLastCommittedDsMessageDictionary(x: 30, y: 194)

Because `autodata` changes the test values on each test run, the test name doesn't match up with the name in test explorer.

I get this error message in the test output window. The value for x and y have changed, and therefore the test is not recognized as the same one that just executed.
```
Test adapter sent back a result for an unknown test case. Ignoring result for 
'LastCommittedMessageTests.TestLastCommittedDsMessageDictionary(x: 109, y: 97)'.
```

As you can see the **30 and 194** changed in the next test execution to **109 and 97**.
The test explorer doesn't match up the test.

adamchester:
This fixes #392, by disabling 'pre-discovery'.

I'm not sure if there's a good way to support this feature with tests.


Src/AutoFixture.xUnit.net2/InlineAutoDataDiscoverer.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Ploeh.AutoFixture.Xunit2
+{
+    internal class InlineAutoDataDiscoverer : DataDiscoverer
```
ploeh: Why do we need a second class that does the exact same thing as `AutoDataDiscoverer`?

Src/AutoFixture.xUnit.net2/AutoDataDiscoverer.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Ploeh.AutoFixture.Xunit2
+{
+    internal class AutoDataDiscoverer : DataDiscoverer
```
ploeh: Will the feature work when the class is `internal`?

adamchester: Note this class is internal. I couldn't find any documentation if this is acceptable, so perhaps @bradwilson could indicate if this is likely to break things?

Src/AutoFixture.xUnit.net2/InlineAutoDataDiscoverer.cs
adamchester: Good question, perhaps we don't.

Src/AutoFixture.xUnit.net2/AutoDataDiscoverer.cs
adamchester: @ploeh it does work; before this change, visual studio gave errors in the output window about mismatched test names for those scenario tests using the auto attributes. After this change, tests did not have data in their names, and no output errors occurred.

My concern is that we might be relying on undocumented or accidental behaviour.

ploeh: > My concern is that we might be relying on undocumented or accidental behaviour.

A reasonable concern. Why not make it public?

adamchester: Wouldn't internal be better, given that it works, and it doesn't add any value to the public API?

adamchester: You know what, I've thought about this further, and it seems unlikely to cause confusion or issues if we  expose it publicly, so I'll just make it public and avoid any potential issues.

ploeh: :+1:

adamchester:
Addressed feedback in 95190c8

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit2 3.30.2](http://www.nuget.org/packages/AutoFixture.Xunit2/3.30.2).

PR #396, opened 2015-04-17
CreateMany should accept zero - updated to include FiniteSequenceRequest

ztsmith:
An additional pull request to allow CreateMany to accept the value 0. An original fix was made, but that didn't allow zero as was intended.
This is for issue https://github.com/AutoFixture/AutoFixture/issues/365

### DIFF
--- a/Src/AutoFixtureUnitTest/Kernel/FiniteSequenceRequestTest.cs
+++ b/Src/AutoFixtureUnitTest/Kernel/FiniteSequenceRequestTest.cs
@@ -19,14 +19,16 @@ namespace Ploeh.AutoFixtureUnitTest.Kernel
             // Teardown
         }
 
-        [Fact]
-        public void InitializeWithInvalidCountThrows()
+        [Theory]
+        [InlineData(-1)]
+        [InlineData(-3)]
+        public void InitializeWithInvalidCountThrows(int count)
         {
             // Fixture setup
             var dummyRequest = new object();
             // Exercise system and verify outcome
             Assert.Throws<ArgumentOutOfRangeException>(() =>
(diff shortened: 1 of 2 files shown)


Src/AutoFixtureUnitTest/Kernel/FiniteSequenceRequestTest.cs
```diff
@@ -19,20 +19,19 @@ public void InitializeWithNullRequestAndValidNumberThrows()
             // Teardown
         }
 
-        [Theory]
-        [InlineData(0)]
-        [InlineData(-1)]
```
moodmosaic: Instead of removing these test cases entirely, would it be possible to just replace the one with the 0 with another negative number? For example -27, similar to what @ploeh did in [this](https://github.com/AutoFixture/AutoFixture/commit/72a9cb75fa512a213f2a87f7a7455659081fb2bc) commit.

moodmosaic:
Thank you for your Pull Request! I just left a comment but, other than that, it's looking good to me.

Src/AutoFixtureUnitTest/Kernel/FiniteSequenceRequestTest.cs
ploeh: I agree that it would be good with more than one test value, because it would protect against a Devil's Advocate implementation that specifically checks for `-1` only.

ztsmith:
Updated the tests to protect against a "devils advocate" implementation. Thanks for the quick feedback!

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.30.3](http://www.nuget.org/packages/AutoFixture/3.30.3).

Issue #394, opened 2015-04-14
AutoConfiguredNSubstituteCustomization throws ObjectCreationException when it can't populate a mock's property/method return value


alexfoxgill:
One of my interfaces exposes an `Expression<Func<...>>` as a method return value. When using `AutoConfiguredNSubstituteCustomization` to generate a mock of this interface, AutoFixture is unable to create an instance. That's fine, I don't need it to create an instance as I can do this myself, but what I would like it to do is **ignore** the method, rather than receive an ObjectCreationException. Perhaps [this method](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs#L185-L198) could swallow the exception?

dcastro:
I don't think the fixture should silently ignore the fact that it failed to setup a method... 
Would injecting a hand crafted `Expression<Func<...>>` solve your problem?

```csharp
fixture.Inject<Expression<Func<int>>>(() => 5);
fixture.Create<MyInterface>():
```

alexfoxgill:
I'm using the customization in an xUnit theory, and then immediately setting the expression in the test setup. I suppose I could include a customization that sets all `Expression<Func<...>>` to a standard value, but I'd rather be able to specify that it doesn't get set at all. I understand that you'd not want to do this in all cases, but I'd like the option to do it when it makes sense for me.

I tried adding a custom `SpecimenBuilder` to omit expressions:

```CSharp
public class OmitExpressionsSpecimenBuilder : ISpecimenBuilder
{
    public object Create(object request, ISpecimenContext context)
    {
        var t = request as Type;
        if (t == null || !t.IsGenericType)
            return new NoSpecimen();

        var genericType = t.GetGenericTypeDefinition();
        if (genericType != typeof (Expression<>))
            return new NoSpecimen();

        return new OmitSpecimen();
    }
}
```

However, this resulted in an `InvalidCastException`:

> Unable to cast object of type 'Ploeh.AutoFixture.Kernel.OmitSpecimen' to type 'System.Linq.Expressions.Expression`1...

Do you think the SpecimenBuilder should work?

dcastro:
I think it should, and I think the exception you're getting may be related to issue #369

alexfoxgill:
Yes I think you're right - how funny that you found an equivalent bug just last week :) in that case I'll close this issue and use an OmitSpecimen in future

PR #397, opened 2015-04-21
Handles OmitSpecimen in AutoConfiguredNSubstituteCustomization

### LINKED ISSUE #395: AutoConfiguredNSubstituteCustomization throws InvalidCastException when OmitSpecimen is returned from context
When enquiring about #394 I discovered that returning an `OmitSpecimen` throws an exception when trying to cast to a concrete return type. Specifically [line 190 of NSubstituteVirtualMethodsCommand](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs#L190) performs the cast.

Similar behaviour was also uncovered in #369.

I'm not sure what the implications are, but it looks like a similar fix could be made (resolve the object and compare to OmitSpecimen)

mrinaldi:
This PR fixes #395

### DIFF
--- a/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
+++ b/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
@@ -179,7 +179,7 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
                 ReturnsMethodInfo.Invoke(new object[] { value, returnThis });
             }
 
-            public void ReturnsUsingContext<T>(MethodInfo methodInfo)
+            public void ReturnsUsingContext(MethodInfo methodInfo)
             {
                 Substitute
                     .WhenForAnyArgs(_ => InvokeMethod(methodInfo))
@@ -194,7 +194,7 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
                             return;
                         methodCalls.Add(call);
 
-                        var value = Resolve<T>();
+                        var value = Resolve(methodInfo.ReturnType);
                         if (value is OmitSpecimen)
                             return;
 
@@ -203,17 +203,17 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
                     });
             }
 
-            private object Resolve<T>()
+            private object Resolve(Type type)
             {
                 return Task.Factory
-                    .StartNew(() => Context.Resolve(typeof (T)))
+                    .StartNew(() => Context.Resolve(type))
                     .Result;
             }
 
-            private void ReturnsFixedValue<T>(MethodInfo methodInfo, T value)
+            private void ReturnsFixedValue(MethodInfo methodInfo, object value)
             {
                 var refValues = GetFixedRefValues(methodInfo);
-                Returns<T>(default(T), x =>
+                Returns(null, x =>
                 {
                     SetRefValues(x, refValues);
                     return value;
@@ -233,7 +233,7 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
                 if (methodInfo.IsVoid())
                     return;
 
-                ReturnsUsingContextMethodInfo.MakeGenericMethod(methodInfo.ReturnType)
+                ReturnsUsingContextMethodInfo
                     .Invoke(this, new object[] { methodInfo });
             }
 
--- a/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
+++ b/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 using NSubstitute;
 using NSubstitute.Core;
 using NSubstitute.Exceptions;
@@ -150,6 +151,9 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
             private readonly object substitute;
             private readonly ISpecimenContext context;
 
+            private readonly List<Tuple<MethodInfo, object[]>> methodCalls =
+                new List<Tuple<MethodInfo, object[]>>();
+
             public SubstituteValueFactory(object substitute, ISpecimenContext context)
             {
                 if (substitute == null)
@@ -184,28 +188,43 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
 
             public void ReturnsUsingContext<T>(MethodInfo methodInfo)
             {
-                InvokeMethod(methodInfo);
-                ReturnsForAnyArgs<T>(default(T), callInfo =>
-                {
-                    var value = new Lazy<T>(() => (T)Context.Resolve(typeof(T)));
-                    object[] arguments = callInfo.Args();
-                    ReturnsFixedValue(methodInfo, callInfo, value);
-                    var returnValue = value.Value;
-                    InvokeMethod(methodInfo, arguments);
+                Substitute
+                    .WhenForAnyArgs(_ => InvokeMethod(methodInfo))
+                    .Do(callInfo =>
+                    {
+                        var arguments = callInfo.Args();
+                        var call = Tuple.Create(methodInfo, arguments.ToArray());
+                        if (methodCalls.Any(
+                            x =>
+                                call.Item1 == x.Item1 &&
+                                call.Item2.SequenceEqual(x.Item2)))
+                            return;
+                        methodCalls.Add(call);
 
-                    return returnValue;
-                });
+                        var value = Resolve<T>();
+                        if (value is OmitSpecimen)
+                            return;
+
+                        ReturnsFixedValue(methodInfo, value);
+                        InvokeMethod(methodInfo, arguments);
+                    });
             }
 
-            private void ReturnsFixedValue<T>(MethodInfo methodInfo, CallInfo callInfo, Lazy<T> value)
+            private object Resolve<T>()
+            {
+                return Task.Factory
+                    .StartNew(() => Context.Resolve(typeof (T)))
+                    .Result;
+            }
+
+            private void ReturnsFixedValue<T>(MethodInfo methodInfo, T value)
             {
                 var refValues = GetFixedRefValues(methodInfo);
                 Returns<T>(default(T), x =>
                 {
                     SetRefValues(x, refValues);
-                    return value.Value;
+                    return value;
                 });
-                SetRefValues(callInfo, refValues);
             }
 
             private IEnumerable<Tuple<int, Lazy<object>>> GetFixedRefValues(MethodInfo methodInfo)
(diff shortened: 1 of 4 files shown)


Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
```diff
@@ -197,6 +197,15 @@ public void ReturnsUsingContext<T>(MethodInfo methodInfo)
                 });
             }
 
+            private T Resolve<T>()
+            {
+                var value = Context.Resolve(typeof (T));
+                if (value is OmitSpecimen)
+                    return default(T);
```
ploeh: Are we sure that this is the desired behaviour?

For all reference types, `default(T)` means `null`. However, `OmitSpecimen` isn't the same as `null`. If the `Context` had found that `null` was the appropriate value to return, it would have returned `null`.

`OmitSpecimen` means that the caller should forget about the value - most of the cases by *not* assigning a value to a property or field. That's not the same as assigning `null` to a property or field.

Since this proposed change directly influences `ReturnsUsingContext`, which is a public function, this concerns me.

I also wonder why requesting a `Type` would return `OmitSpecimen`, which is a return value I'd mostly expect to see if a `PropertyInfo` or `FieldInfo` was requested...

mrinaldi:
I see your point.

Do you think ``OmitSpecimen`` should only be handled with ``PropertyInfo`` and ``FieldInfo``? If that's the case, I'll have to refactor the ``NSubstituteVirtualMethodsCommand`` to request a ``PropertyInfo`` instead of ``Type`` in such cases.

#394 describes the ``OmitSpecimen`` returned when a ``Type`` is requested. Do you think it makes sense? Should it be supported?

ploeh:
`OmitSpecimen` indicates a [NOOP](http://en.wikipedia.org/wiki/NOP). Its *purpose* is to indicate to the caller that whatever it's currently attempting to do (such as mutating a field or property), it should abort.

This makes sense for a [Command](http://en.wikipedia.org/wiki/Command%E2%80%93query_separation), but I fail to see how this could ever be appropriate for a Query. How can a Query ever be a NOOP?

In AutoFixture, Commands are usually used to mutate the state of an object by assigning a value to a property or field. This is the primary use case for `OmitSpecimen`, but there may be other I've forgotten about...

dcastro:
I think you raised an important question. `OmitSpecimen` usually means "don't assign anything to this field/property".

In the context of *setting up a method*, I guess it *could* mean:

> Don't set it up, at all. Instead, let the mocking framework decide what to do. The framework might then decide to either return null, 0, a mock (e.g., in Moq if `DefaultValue` is set to `DefaultValue.Mock`), or maybe even throw an exception.

This definition seems to fit the overall goal of `OmitSpecimen`.  

The problem with that is: the specimen is resolved (and `OmitSpecimen` is returned) lazily inside a delegate executed by the framework. When the delegate is executed, the framework already expects us to tell it what value to return - it's no longer up to the framework, it's up to us. It's simply not possible to halt and do a NOOP. 

Given this limitation, I believe the most sensible thing to do would be to return `default(T)`, even if that's not ideal.

ploeh:
> Don't set it up, at all. Instead, let the mocking framework decide what to do. The framework might then decide to either return null, 0, a mock (e.g., in Moq if DefaultValue is set to DefaultValue.Mock), or maybe even throw an exception.

I like that specification :+1: 

Converting `OmitSpecimen` to `default(T)` is wrong. The entire reason `OmitSpecimen` exists is exactly to enable downstream clients to make an informed decision about what it means. Converting to `default(T)` removes this information.

Originally, AutoFixture handled recursion issues by either throwing exceptions on detected recursions (to this day, this is the default, built-in behaviour), or alternatively, as an opt-in customization, by *assigning* `null`.

The problem with *assigning* `null` is that a writeable property may have a Guard Clause that protects against `null` values.

Therefore, the *motivation* for introducing `OmitSpecimen` was to signal that the request in question should be aborted.

This is clearly explained in the commit message for 4ffec5ea53f409406819aa97870891c8aa501fc0, which explains:

> Added the OmitSpecimen class, which is a pure marker object that can be used to indicate that a particular request should be omitted. The primary purpose is to notify the AutoPropertiesCommand that a property or field should be omitted.

mrinaldi:
I've managed to handle ``OmitSpecimen`` as it was supposed to.
Done in: 685d0e3 and 5567178.

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
```diff
@@ -172,40 +174,50 @@ public ISpecimenContext Context
                 get { return context; }
             }
 
-            public static void ReturnsForAnyArgs<T>(T value, Func<CallInfo, T> returnThis)
```
ploeh: Isn't it a breaking change to remove a public method?

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
```diff
@@ -172,40 +174,50 @@ public ISpecimenContext Context
...
-            {
-                ReturnsForAnyArgsMethodInfo.Invoke(new object[] { value, returnThis });
-            }
-
             public static void Returns<T>(T value, Func<CallInfo, T> returnThis)
             {
                 ReturnsMethodInfo.Invoke(new object[] { value, returnThis });
             }
 
-            public void ReturnsUsingContext<T>(MethodInfo methodInfo)
```
ploeh: Isn't it a breaking change to remove the type argument on a public method?

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
mrinaldi: The class is a private nested class, therefore I don't think it's a breaking change.

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
mrinaldi: The class is a private nested class, therefore I don't think it's a breaking change.

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
ploeh: Oh, foiled by GitHub's diff view :flushed:

My mistake - sorry about that!

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
ploeh: Oh, foiled by GitHub's diff view :flushed:

My mistake - sorry about that!

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoNSubstitute 3.30.4](http://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.30.4).

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
```diff
@@ -172,40 +174,50 @@ public ISpecimenContext Context
...
+                    });
+            }
 
-                    return returnValue;
-                });
+            private object Resolve(Type type)
+            {
+                return Task.Factory
+                    .StartNew(() => Context.Resolve(type))
+                    .Result;
```
ploeh: Isn't this going to block until the task is completed?

moodmosaic: Yes, it will... Essentially, it's the same with doing

```
Context.Resolve(type)
```

What was the motivation of wrapping this into a `Task`?

mrinaldi: This is needed when you have a method returning a circular reference like ``IInterfaceWithCircularDependency`` or ``IInterfaceWithCircularMethod``.

NSubstitute uses a static property ``SubstitutionContext.Current`` to get and set the state of its substitutes. However, it uses ThreadLocal fields so one thread does not interfere in another. For this reason, I have to force the ``Context.Resolve`` to run in another thread, so in cases where you have a circular reference, NSubstitute sets up the methods correctly.

This change has become needed because now I have to resolve the value, to check if it's an instance of ``OmitSpecimen``, before I set up the method. Before, the value was lazy and it was only evaluated after I had set up the method.

ploeh: @mrinaldi Thank you for explaining that. This sounds like a great candidate for a code comment, so that future developers understand the reasoning behind this. :wink:

moodmosaic: Wow... Thank you for the explanation! Lot's of stuff going on under the covers there. Indeed, as @ploeh wrote, a code comment would be great at this point. :smiley:

dcastro: @mrinaldi Does that mean you had to resolve the return value eagerly in order to meet this new `OmitSpecimen` requirement?

mrinaldi: @dcastro Not really.

When the mock is created, I set up a callback to be called when the method is called. This callback does not return the value. Instead, it sets up the method to return the value, or does not set up at all if it sees a ``OmitSpecimen``. 

To make a long story short, it will resolve the return value when the method is first called, but it may or may not set up the method to return the value.

dcastro: Ah, I see!

dcastro: Ah, I see!
Since I'm having the same issue with Moq, I tried something similar, something along these lines:

```
mock.Setup(m => m.Method()).
        .Callback(() => {
                 var specimen = context.Resolve(...);
                 if (! (specimen is OmitSpecimen))
                 {
                       mock.Setup(m => m.Method()).Returns(specimen);
                 }
         });
```

But Moq didn't play ball, it doesn't work :/ I guess Moq will have no choice but to return `null` instead...

mrinaldi: @dcastro It doesn't work the first time you call it, but it does if you call it twice.

I tried something out here and it seems to work:

```c#
mock.Setup(m => m.Method())
        .Callback(() => {
                 var specimen = context.Resolve(...);
                 if (!(specimen is OmitSpecimen))
                 {
                     setup.Returns(specimen);
                     return;
                 }
                 setup.Callback(() => { });
         })
         .Returns(() => mock.Object.Method());
```

Maybe you can go on from that.

dcastro: I found another solution, I have to reuse the initial instance created by `.Setup`, so something like:

```
var setup = mock.Setup(..);
setup.Callback( () => {
    //etc
    setup.Returns(...)
});
```

Back on track!

PR #398, opened 2015-04-22
Turns the member name comparison into a strategy.

ecampidoglio:
The identifiers used in `FieldSpecification`, `PropertySpecification` and `ParameterSpecification` can now be compared according to a strategy. The strategy is specified during initialization through an `IEqualityComparer<string>`.

For example, in order to make a _case-insensitive_ comparison of a property's name, one could say:

```csharp
new PropertySpecification(
    typeof(int),
    "Foo",
    StringComparer.OrdinalIgnoreCase);
```

If a strategy isn't specified at construction, the names will be compared in a _case-sensitive_ fashion, which is consistent with the current behavior.

This change enables to use the `FrozenAttribute` from #318 in scenarios where a field, property or parameter should be frozen by matching the name of the test parameter regardless of case.

### DIFF
--- a/Src/AutoFixture/Kernel/FieldSpecification.cs
+++ b/Src/AutoFixture/Kernel/FieldSpecification.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Reflection;
 
 namespace Ploeh.AutoFixture.Kernel
@@ -11,6 +12,7 @@ namespace Ploeh.AutoFixture.Kernel
     {
         private readonly Type targetType;
         private readonly string targetName;
+        private readonly IEqualityComparer<string> nameComparison;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="FieldSpecification"/> class.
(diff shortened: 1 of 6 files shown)


Src/AutoFixture/Kernel/FieldSpecification.cs
```diff
@@ -11,6 +12,7 @@ public class FieldSpecification : IRequestSpecification
     {
         private readonly Type targetType;
         private readonly string targetName;
+        private readonly IEqualityComparer<string> nameComparison;
```
ploeh: Given that the containing class is a `FieldSpecification`, what is it going to compare?

Src/AutoFixture/Kernel/ParameterSpecification.cs
```diff
@@ -11,6 +12,7 @@ public class ParameterSpecification : IRequestSpecification
     {
         private readonly Type targetType;
         private readonly string targetName;
+        private readonly IEqualityComparer<string> nameComparison;
```
ploeh: Given that the containing class is a `ParameterSpecification`, what is it going to compare?

Src/AutoFixture/Kernel/PropertySpecification.cs
```diff
@@ -11,6 +12,7 @@ public class PropertySpecification : IRequestSpecification
     {
         private readonly Type targetType;
         private readonly string targetName;
+        private readonly IEqualityComparer<string> nameComparison;
```
ploeh: Given that the containing class is a `PropertySpecification`, what is it going to compare?

ploeh:
Since this pull request had been sitting around for so long, I took it and finished it. It's now live as [AutoFixture 3.34.0](https://www.nuget.org/packages/AutoFixture/3.34.0). Thank you for your contribution.

PR #402, opened 2015-05-09
AutoMoq - Handle other valid specimens

dcastro:
As discussed in #369, this PR aims to refactor AutoMoq so that `null`, `NoSpecimen` and `OmitSpecimen` are properly validated.

Most cases are pretty straightforward to fix, but two stand out:

## Out parameters

This is more or less what AutoMoq currently does when setting up a method with an `out` parameter (reflection and abstractions aside):

```csharp
// AutoMoq
var outResult = fixture.Create<string>();
mock.Setup(m => m.Method(out outResult).Returns( ... );
```
```csharp
var frozenString = fixture.Freeze<string>();
var sut = fixture.Create<ISomething>();

string outResult = null;
sut.Method(out outResult);
Assert.Equal(frozenString, outResult);
```

In the context of resolving the value for an `out` parameter, what might the meaning of `OmitSpecimen` be? I can think of two possible definitions:

* Don't "assign" anything to the `out` parameter
    * (*Weeell*, `out` parameters must be assigned something, so "Assign `default(T)` to the `out` parameter" is a better description.)
* Don't setup the method

I ended up implementing the second definition - although, in hindsight, the first sounds more appropriate.
Opinions? Maybe there's a third definition?

## ReturnsUsingFixture / ReturnsUsingContext

This is a tricky one. How should `OmitSpecimen` be treated here?

In #397, I suggested the [following spec](https://github.com/AutoFixture/AutoFixture/pull/397#issuecomment-95675027):

> Don't set it up, at all. Instead, let the mocking framework decide what to do. The framework might then decide to either return null, 0, a mock (e.g., in Moq if `DefaultValue` is set to `DefaultValue.Mock`), or maybe even throw an exception.

However, I don't think this spec can be implemented in Moq.

This is the closest I got, by using a callback that would later decide whether or not to call `Returns`.

```csharp
internal static IReturnsThrows<TMock, TResult> ReturnsUsingContext<TMock, TResult>(
        this ISetup<TMock, TResult> setup, ISpecimenContext context)
    where TMock : class
{
    bool hasBeenSetup = false;
    return setup
        .Callback(() =>
        {
            if (hasBeenSetup)
                return;
            hasBeenSetup = true;

            var specimen = context.Resolve(typeof (TResult));

            // "specimen is TResult" returns false if specimen is null.
            // For that reason, we have to additionally check if null is a valid specimen for TResult,
            // i.e., if TResult is a reference type or Nullable<T>.
            if (specimen is TResult || (specimen == null && default(TResult) == null))
            {
                var result = (TResult)specimen;
                setup.Returns(result);
            }
        });
}
```

Before why I explain the drawbacks of this solution, let us break the spec in parts:

1. "let the mocking framework decide what to do. The framework might then decide to either return null, 0, ..." 
1. "... a mock (e.g., in Moq if `DefaultValue` is set to `DefaultValue.Mock`), ..."
1. "... or maybe even throw an exception."

##### (1) "let the mocking framework decide what to do. The framework might then decide to either return null, 0, ..." 

This point is possible, and can be achieved using the implementation above.

```csharp
public interface ITest
{
    string Property { get; set; }
}

// Fixture setup
var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());

var mock = fixture.Create<Mock<ITest>>();
// Exercise system
mock.Setup(m => m.Property).ReturnsUsingFixture(new Omitter());
// Verify outcome
Assert.Null(mock.Object.Property);
```

##### (2) "... a mock (e.g., in Moq if `DefaultValue` is set to `DefaultValue.Mock`), ..."

Not possible, this test fails:

```csharp
public interface ITest2
{
    ITest Property { get; set; }
}

// Fixture setup
var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());

var mock = fixture.Create<Mock<ITest2>>();
mock.DefaultValue = DefaultValue.Mock;
// Exercise system
mock.Setup(m => m.Property).ReturnsUsingFixture(new Omitter());
// Verify outcome
var propertyValue = mock.Object.Property;
Assert.NotNull(propertyValue);
```

##### (3) "... or maybe even throw an exception."

Technically possible - you could chain two calls, like `mock.Setup(...).ReturnsUsingContext(...).Throws<Exception>`.

The problem is, if we do that, our callback with the conditional setup won't ever be called. The call to `Throws` completely overrides it, and every member invocation will result in an exception being thrown. For example, this test fails:

```csharp
var mock = new Mock<ITest>();

mock.Setup(m => m.Property)
    .ReturnsUsingFixture(new Fixture())
    .Throws<Exception>();

Assert.DoesNotThrow(() => { var _ = mock.Object.Property; });
```

So, 2 out of the 3 requirements cannot be implemented.  A couple more cons:

* In order to internally be able to call `Callback`, we'd have to change the return type and the type of the first parameter of `ReturnsUsingFixture` (low impact breaking change).
* This change in the signature would allow for a non-intuitive API (e.g., `ReturnsUsingFixture().Returns()`, or `ReturnsUsingFixture().Throws()`)

To sum things up, I believe we're better off simply returning `default(T)` if we get an `OmitSpecimen`.

(This turned out to be longer than I thought :sweat: ).

### DIFF
--- a/Src/AutoMoq/MockType.cs
+++ b/Src/AutoMoq/MockType.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -64,13 +65,31 @@ namespace Ploeh.AutoFixture.AutoMoq
         }
 
         internal static IReturnsResult<TMock> ReturnsUsingContext<TMock, TResult>(this IReturns<TMock, TResult> setup,
-                                                                                  ISpecimenContext context)
+            ISpecimenContext context)
             where TMock : class
         {
             return setup.Returns(() =>
             {
                 var specimen = context.Resolve(typeof (TResult));
-                TResult result = specimen is TResult ? (TResult) specimen : default(TResult);
+
+                // check if specimen is null but member is non-nullable value type
+                if (specimen == null && (default(TResult) != null))
+                    throw new InvalidOperationException(
+                        string.Format(
+                            CultureInfo.CurrentCulture,
+                            "Tried to setup a member with a return type of {0}, but null was found instead.",
+                            typeof (TResult)));
+
+                // check if specimen can be safely converted to TResult
+                if (specimen != null && !(specimen is TResult))
+                    throw new InvalidOperationException(
+                        string.Format(
+                            CultureInfo.CurrentCulture,
+                            "Tried to setup a member with a return type of {0}, but an instance of {1} was found instead.",
+                            typeof (TResult),
+                            specimen.GetType()));
+
+                TResult result = (TResult) specimen;
 
                 //"cache" value for future invocations
                 setup.Returns(result);
--- a/Src/AutoMoq/MockType.cs
+++ b/Src/AutoMoq/MockType.cs
@@ -68,11 +68,14 @@ namespace Ploeh.AutoFixture.AutoMoq
             where TMock : class
         {
             return setup.Returns(() =>
-                {
-                    var result = (TResult) context.Resolve(typeof (TResult));
-                    setup.Returns(result);
-                    return result;
-                });
+            {
+                var specimen = context.Resolve(typeof (TResult));
+                TResult result = specimen is TResult ? (TResult) specimen : default(TResult);
+
+                //"cache" value for future invocations
+                setup.Returns(result);
+                return result;
+            });
         }
     }
 }
--- a/Src/AutoMoq/AutoMockPropertiesCommand.cs
+++ b/Src/AutoMoq/AutoMockPropertiesCommand.cs
@@ -23,7 +23,6 @@ namespace Ploeh.AutoFixture.AutoMoq
         /// <param name="context">The context that is used to create anonymous values.</param>
         public void Execute(object specimen, ISpecimenContext context)
         {
-            if (specimen == null) throw new ArgumentNullException("specimen");
             if (context == null) throw new ArgumentNullException("context");
 
             var mock = specimen as Mock;
--- a/Src/AutoMoq/MockPostprocessor.cs
+++ b/Src/AutoMoq/MockPostprocessor.cs
@@ -54,7 +54,11 @@ namespace Ploeh.AutoFixture.AutoMoq
                 return new NoSpecimen(request);
             }            
 
-            var m = this.builder.Create(request, context) as Mock;
+            var specimen = this.builder.Create(request, context);
+            if (specimen is NoSpecimen || specimen is OmitSpecimen || specimen == null)
+                return specimen;
+
+            var m = specimen as Mock;
             if (m == null)
             {
                 return new NoSpecimen(request);
--- a/Src/AutoMoq/MockVirtualMethodsCommand.cs
+++ b/Src/AutoMoq/MockVirtualMethodsCommand.cs
@@ -51,19 +51,22 @@ namespace Ploeh.AutoFixture.AutoMoq
                 var returnType = method.ReturnType;
                 var methodInvocationLambda = MakeMethodInvocationLambda(mockedType, method, context);
 
-                if (method.IsVoid())
+                if (methodInvocationLambda != null)
                 {
-                    this.GetType()
-                        .GetMethod("SetupVoidMethod", BindingFlags.NonPublic | BindingFlags.Static)
-                        .MakeGenericMethod(mockedType)
-                        .Invoke(this, new object[] {mock, methodInvocationLambda});
-                }
-                else
-                {
-                    this.GetType()
-                        .GetMethod("SetupMethod", BindingFlags.NonPublic | BindingFlags.Static)
-                        .MakeGenericMethod(mockedType, returnType)
-                        .Invoke(this, new object[] {mock, methodInvocationLambda, context});
+                    if (method.IsVoid())
+                    {
+                        this.GetType()
+                            .GetMethod("SetupVoidMethod", BindingFlags.NonPublic | BindingFlags.Static)
+                            .MakeGenericMethod(mockedType)
+                            .Invoke(this, new object[] {mock, methodInvocationLambda});
+                    }
+                    else
+                    {
+                        this.GetType()
+                            .GetMethod("SetupMethod", BindingFlags.NonPublic | BindingFlags.Static)
+                            .MakeGenericMethod(mockedType, returnType)
+                            .Invoke(this, new object[] {mock, methodInvocationLambda, context});
+                    }
                 }
             }
         }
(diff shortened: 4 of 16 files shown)


Src/AutoMoq/MockType.cs
```diff
@@ -68,11 +68,14 @@ internal static Type GetMockedType(this Type type)
...
         {
             return setup.Returns(() =>
-                {
-                    var result = (TResult) context.Resolve(typeof (TResult));
-                    setup.Returns(result);
-                    return result;
-                });
+            {
+                var specimen = context.Resolve(typeof (TResult));
+                TResult result = specimen is TResult ? (TResult) specimen : default(TResult);
```
ploeh: What happens if `specimen` is `OmitSpecimen` and `TResult` is a reference type? Wouldn't that cause `result` to be `null`?

dcastro: Yes, it would.
As explained in the PR description, there are a few reasons why returning `null` seems to be the most sensible thing to do.

> To sum things up, I believe we're better off simply returning default(T) if we get an OmitSpecimen.

It seems that, using NSubstitute, it is possible for the callback to conditionally setup the member. In Moq, this is not viable.

ploeh: Converting `OmitSpecimen` to `default(T)` shouldn't be the responsibility of an Auto-Mocking Container.

`OmitSpecimen` exists for the purpose of enabling recursion handlers and auto-property commands to skip the value. Prematurely converting the value takes away that ability.

Thus, I'm inclined to think that this conversion is undesirable.

What is the high-level scenario where this conversion is necessary?

dcastro: Sorry for the late reply!



> Converting OmitSpecimen to default(T) shouldn't be the responsibility of an Auto-Mocking Container.

>OmitSpecimen exists for the purpose of enabling recursion handlers and auto-property commands to skip the value. Prematurely converting the value takes away that ability.

>Thus, I'm inclined to think that this conversion is undesirable.

I agree completely.

> What is the high-level scenario where this conversion is necessary?

* The user asks the fixture to create an auto-configured mock for interface `A`.
* This interface has a method `M` that returns an instance of type `B`.
* The return value of type `B` needs to be resolved lazily, IFF the method is called.
* Ideally, we'd want AutoMoq to, somehow, tell Moq "Hey, when this method is called ask me whether I can create an instance of `B`. If i can create it, return it, otherwise fallback to whatever your default action is."

Turns out we can't ask Moq to do this. Not in a viable, fruitful way.

The best we can do, IMO, is tell Moq "Hey, when this method is called ask me whether I can create an instance of `B`. If i can create it, return it, otherwise return this other value instead (e.g., null)."

ploeh: Can you provide a repro?

dcastro:
```csharp
public interface ITest
{
    string M();
}

var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
fixture.Customize<string>(c => new Omitter());

var mock = fixture.Create<ITest>(); //will internally call ReturnsUsingContext
```

ploeh: Why would you want to do something like that?

dcastro: You mean, omit strings? Or have the method return nulls when an `OmitSpecimen` is found?

ploeh: Why would you omit all strings?

dcastro: Well, this was a rather contrived example, I don't think I would ever want to do that... 

In fact, I've never had the need to use `OmitSpecimen` directly - I've only used it through DSL customizations (`Customize<Test>(c => c.Without(...))`) which only affect concrete classes, not interfaces..

It makes sense to want to omit a property of a concrete class, because the setter may have unwanted side-effects, but I can't think of a scenario where you'd want to omit the setup of a method. I don't think this scenario will come up often, if at all.

ploeh: I agree, which is also the reason why I think it's wrong to convert `OmitSpecimen` to `null`.

dcastro: I suppose we could alternatively throw an exception (.e.g, `ObjectCreationException`) to signal that the return value could not be instantiated.
This would happen when the context returns an `OmitSpecimen`, a `NoSpecimen`, or anything that is not of type `TResult`.

ploeh: What would be the benefit of doing that?

dcastro: Well, as far as I can tell, we have 3 options:

* Leave it as it is, and throw an `InvalidCastException` when an `OmitSpecimen` is found
* Return `null`
* Throw an `ObjectCreationException`.

As you already pointed out, returning `null` is not desirable. Option 1 is clearly not desirable either.

The third option would be, IMO, appropriate. In the context of setting up a method, an `OmitSpecimen` does not make sense. There's nothing to omit. So we need to tell the user "hey, we tried to setup this method to return an instance of `string`, but we couldn't create one".

ploeh: What's the value of throwing `ObjectCreationException` over the `InvalidCastException` already thrown?

How will that help the caller?

dcastro: In my opinion, an `InvalidCastException` usually indicates a bug. On the other hand, an `ObjectCreationException` with a meaningful message (e.g., "Tried to create a string, found an OmitSpecimen" instead") will make it clear that there is no bug, and that the operation failed because `OmitSpecimen` is not supported.

ploeh: Would an `InvalidOperationException` with a proper message communicate the same information?

dcastro: Yes, I think it would. Should I go ahead and make this change then?

ploeh: Yes, that sounds good :+1:

mrinaldi: Should that be the expected behavior for AutoNSubstitute as well?
Although I've managed to handle ``OmitSpecimen`` the way it was supposed to be handled in #397, it has caused the issue #405 that I don't think I'll be able to handle.

ploeh: @mrinaldi The better we can follow the contract and actually do nothing when `OmitSpecimen` is encountered, the better.

However, if not possible, I prefer an exception over a conversion to `default(T)`.

dcastro: I've done this change: 9881496

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoMoq 3.30.6](http://www.nuget.org/packages/AutoFixture.AutoMoq/3.30.6).

Issue #337, opened 2015-01-16
OmitOnRecursionBehavior does not guard against recursionDepth < 1


dcastro:
The following code throws an `InvalidCastException` for `recursionDepth` < 1.

```csharp
var fixture = new Fixture();
fixture.Behaviors.OfType<ThrowingRecursionBehavior>().ToList()
    .ForEach(b => fixture.Behaviors.Remove(b));

fixture.Behaviors.Add(new OmitOnRecursionBehavior(recursionDepth));

fixture.Create<string>();
```

```
System.InvalidCastExceptionUnable to cast object of type 'Ploeh.AutoFixture.Kernel.OmitSpecimen' to type 'System.String'.
   at Ploeh.AutoFixture.SpecimenFactory.Create(ISpecimenContext context, T seed) in SpecimenFactory.cs: line 125
   at Ploeh.AutoFixture.SpecimenFactory.Create(ISpecimenContext context) in SpecimenFactory.cs: line 29
   at Ploeh.AutoFixture.SpecimenFactory.Create(ISpecimenBuilder builder) in SpecimenFactory.cs: line 73
   at Ploeh.AutoFixtureUnitTest.OmitOnRecursionBehaviorTest.FactMethodName() in OmitOnRecursionBehaviorTest.cs: line 90
```

It seems like `OmitOnRecursionBehavior` is not supposed to work for depth < 1, so a guard clause is needed somewhere. 

At first glance, it seems like the guard should go into the `OmitOnRecursionBehavior`'s constructor, but maybe this could be generalized for every `RecursionGuard` and be put into its constructor instead.

I can volunteer to make this change, but since I'm not very familiar with this part of the code, I'd like to know the owners/contributors' thoughts on this first.

ploeh:
I wonder why creation of `string` even triggers any recursion behaviour... `string` has a dedicated `ISpecimenBuilder`.

Pvlerick:
From what I understood, when `OmitOnRecursionBehavior` (or the default `ThrowingRecursionBehavior`) is added to the list of behaviors, their `Transform` function puts them first in the graph, which means that their respective Handlers returned get called first.

So in the example code above, execution doesn't reach the specimen builders.

Adding a guard in `RecursionGuard` would do the trick but then the exception would be thrown when `Transform` is called, contrarily to adding the guard clause in the constructor.

I'd be happy to pick this up either way, just let me know what is the preferred way and I'll send a PR.

PR #403, opened 2015-05-13
Guard clause in RecursionGuard and OmitOnRecursionBehavior constructors

Pvlerick:
Guard clause for the recursionDepth parameter that should be strictly higher than zero.
Unit tests added in both cases.

Addressing #337

### DIFF
--- a/Src/AutoFixture/Kernel/RecursionGuard.cs
+++ b/Src/AutoFixture/Kernel/RecursionGuard.cs
@@ -175,7 +175,7 @@ namespace Ploeh.AutoFixture.Kernel
                 throw new ArgumentNullException("comparer");
 
             if (recursionDepth < 1)
-                throw new ArgumentOutOfRangeException("recursionDepth");
+                throw new ArgumentOutOfRangeException("recursionDepth", "Recursion depth must be greater than 0.");
 
             this.builder = builder;
             this.recursionHandler = recursionHandler;
--- a/Src/AutoFixture/Kernel/RecursionGuard.cs
+++ b/Src/AutoFixture/Kernel/RecursionGuard.cs
@@ -160,7 +160,7 @@ namespace Ploeh.AutoFixture.Kernel
         /// or
         /// comparer
         /// </exception>
-        /// <exception cref="System.ArgumentOutOfRangeException">recursionDepth is less than zero.</exception>
+        /// <exception cref="System.ArgumentOutOfRangeException">recursionDepth is less than one.</exception>
         public RecursionGuard(
             ISpecimenBuilder builder,
             IRecursionHandler recursionHandler,
--- a/Src/AutoFixture/OmitOnRecursionBehavior.cs
+++ b/Src/AutoFixture/OmitOnRecursionBehavior.cs
@@ -30,7 +30,7 @@ namespace Ploeh.AutoFixture
         public OmitOnRecursionBehavior(int recursionDepth)
         {
             if (recursionDepth < 1)
-                throw new ArgumentOutOfRangeException("recursionDepth");
+                throw new ArgumentOutOfRangeException("recursionDepth", "Recursion depth must be greater than 0.");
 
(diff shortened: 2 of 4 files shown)


moodmosaic:
I've just verified the code in this Pull Request, with the code sample provided by @dcastro [here](https://github.com/AutoFixture/AutoFixture/issues/337#issue-54626803), and seems to be working as expected now:

```csharp
var fixture = new Fixture();
fixture.Behaviors.OfType<ThrowingRecursionBehavior>().ToList()
    .ForEach(b => fixture.Behaviors.Remove(b));
fixture.Behaviors.Add(new OmitOnRecursionBehavior(0));

fixture.Create<string>();
```
Output:

```
Test 'Ploeh.AutoFixtureUnitTest.CollectionFillerTest.Test' failed
    : System.ArgumentOutOfRangeException 
    : Specified argument was out of the range of valid values.
Parameter name: recursionDepth
    OmitOnRecursionBehavior.cs(33,0)
        : at Ploeh.AutoFixture.OmitOnRecursionBehavior..ctor(Int32 recursionDepth)
    CollectionFillerTest.cs(24,0)
        : at Ploeh.AutoFixtureUnitTest.CollectionFillerTest.Test()
```

Src/AutoFixtureUnitTest/OmitOnRecursionBehaviorTest.cs
```diff
@@ -77,5 +77,15 @@ public void TransformResultCorrectlyDecoratesInput()
...
             // Teardown
         }
+
+        [Fact]
+        public void ConstructorWithRecusionDepthLowerThanOneThrows()
+        {
+            // Fixture setup
+            // Exercise system and verify outcome
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+                new OmitOnRecursionBehavior(0));
```
moodmosaic: Wouldn't it be more clear, that recursion depth must be lower than `1`, if we'd parameterize the test? (e.g. `0`, `-7`, `-21`)

Pvlerick: Didn't think about it, indeed - I'll turn it into a Theory.

Src/AutoFixture/Kernel/RecursionGuard.cs
```diff
@@ -173,6 +174,9 @@ public RecursionGuard(ISpecimenBuilder builder, IEqualityComparer comparer)
             if (comparer == null)
                 throw new ArgumentNullException("comparer");
 
+            if (recursionDepth < 1)
+                throw new ArgumentOutOfRangeException("recursionDepth");
```
ploeh: Would it be more helpful to the user if the exception message explains what is the acceptable range?

Src/AutoFixture/OmitOnRecursionBehavior.cs
```diff
@@ -26,8 +26,12 @@ public OmitOnRecursionBehavior() : this(DefaultRecursionDepth)
         /// Initializes new instance of the <see cref="OmitOnRecursionBehavior" /> class with specific recursion depth.
         /// </summary>
         /// <param name="recursionDepth">The recursion depth at which the request will be omitted.</param>
+        /// <exception cref="System.ArgumentOutOfRangeException">recursionDepth is less than zero.</exception>
         public OmitOnRecursionBehavior(int recursionDepth)
         {
+            if (recursionDepth < 1)
+                throw new ArgumentOutOfRangeException("recursionDepth");
```
ploeh: Would it be more helpful to the user if the exception message explains what is the acceptable range?

Pvlerick:
Thanks for your reviews; should be good now.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.30.5](http://www.nuget.org/packages/AutoFixture/3.30.5).

Issue #408, opened 2015-06-07
RandomNumericSequenceGenerator.Create throws ArgumentOutOfRangeException when limits are not ascending


Pvlerick:
Simple repro:

```C#
[Fact]
public void ConstructorShouldThrow()
{
    var fixture = new Fixture();
    fixture.Customizations.Add(new RandomNumericSequenceGenerator(8, 3));

    fixture.Create<int>();
}
```

This is probably a bug as line 162 of [RandomNumericSequenceGenerator](https://github.com/AutoFixture/AutoFixture/blob/20b91b313daba05fd51714323ad0575e4f25d8c9/Src/AutoFixture/RandomNumericSequenceGenerator.cs) `this.random.Next((int)this.lower, (int)this.upper)` will throw an `ArgumentOutOfRangeException` when lower > upper.

The constructor documentation is a bit ambiguous; it states "A sequence of integers beginning with two positive or negative number optionally followed by a series of greater numbers." which doesn't say enough about the relation between the two first (second has to be equal or greater then the first). Actually the only constraint is that each number is >= than the previous.

I found this while investigating on #407.

I will send a PR.

moodmosaic:
Good catch :+1: – Thank you for reporting this!

PR #409, opened 2015-06-08
Constructor of RandomNumericSequenceGenerator now checks that given limits are ascending

Pvlerick:
Addressing #408 

I changed documentation on constructors. I also found out that if the two first numbers in limits are equal, there was an infinite loop, that's why I enforced that the sequence in strictly ascending (this wasn't an issue for numbers after the two first, but oh well...)

I also had to remove a test that wasn't relevant any more.

### DIFF
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -54,10 +54,9 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentException("Limits must be at least two ascending numbers.", "limits");
             }
 
-            this.limits = limits;
+            ValidateThatLimitsAreStrictlyAscending(limits);
 
-            this.ValidateThatLimitsAreStrictlyAscending();
-            
+            this.limits = limits;
             this.syncRoot = new object();
             this.random = new Random();
             this.numbers = new HashSet<long>();
@@ -95,9 +94,9 @@ namespace Ploeh.AutoFixture
             return this.CreateRandom(type);
         }
 
-        private void ValidateThatLimitsAreStrictlyAscending()
+        private static void ValidateThatLimitsAreStrictlyAscending(long[] limits)
         {
-            if (this.limits.Zip(this.limits.Skip(1), (a, b) => a >= b).Any(b => b))
+            if (limits.Zip(limits.Skip(1), (a, b) => a >= b).Any(b => b))
             {
                 throw new ArgumentOutOfRangeException("limits", "Limits must be ascending numbers.");
             }
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -51,17 +51,13 @@ namespace Ploeh.AutoFixture
 
             if (limits.Length < 2)
             {
-                throw new ArgumentException("limits must be two or more Int64.", "limits");
+                throw new ArgumentException("Limits must be at least two ascending numbers.", "limits");
             }
 
             //Check that limits are ascending
-            var previous = limits[0];
-            foreach (var item in limits.Skip(1))
+            if (limits.Zip(limits.Skip(1), (a, b) => a >= b).Any(b => b))
             {
-                if (previous >= item)
-                    throw new ArgumentOutOfRangeException("limits", "limits must be ascending Int64.");
-
-                previous = item;
+                throw new ArgumentOutOfRangeException("limits", "Limits must be ascending numbers.");
             }
 
             this.limits = limits;
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -30,8 +30,7 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// Initializes a new instance of the <see cref="RandomNumericSequenceGenerator" /> class.
         /// </summary>
-        /// <param name="limits">A sequence of integers beginning with two positive or negative 
-        /// number optionally followed by a series of greater numbers.</param>
+        /// <param name="limits">A sequence of at least two ascending numbers.</param>
         public RandomNumericSequenceGenerator(IEnumerable<long> limits)
             : this(limits.ToArray())
         {
@@ -40,8 +39,7 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// Initializes a new instance of the <see cref="RandomNumericSequenceGenerator" /> class.
         /// </summary>
-        /// <param name="limits">An array of integers beginning with two positive or negative 
-        /// number optionally followed by a series of greater numbers.</param>
+        /// <param name="limits">A array of at least two ascending numbers.</param>
         /// <exception cref="System.ArgumentNullException"></exception>
         /// <exception cref="System.ArgumentException"></exception>
         public RandomNumericSequenceGenerator(params long[] limits)
@@ -54,6 +52,16 @@ namespace Ploeh.AutoFixture
             if (limits.Length < 2)
             {
                 throw new ArgumentException("The limit must be a sequence of two or more integers.");
+            }
+
+            //Check that limits are ascending
+            var previous = limits[0];
+            foreach (var item in limits.Skip(1))
+            {
+                if (previous >= item)
+                    throw new ArgumentOutOfRangeException("limits", "limits must be ascending numbers");
+
+                previous = item;
             }
 
             this.limits = limits;
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -39,7 +39,7 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// Initializes a new instance of the <see cref="RandomNumericSequenceGenerator" /> class.
         /// </summary>
-        /// <param name="limits">A array of at least two ascending numbers.</param>
+        /// <param name="limits">An array of at least two ascending numbers.</param>
         /// <exception cref="System.ArgumentNullException"></exception>
         /// <exception cref="System.ArgumentException"></exception>
         public RandomNumericSequenceGenerator(params long[] limits)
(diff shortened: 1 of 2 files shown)


Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -30,8 +30,7 @@ public RandomNumericSequenceGenerator()
         /// <summary>
         /// Initializes a new instance of the <see cref="RandomNumericSequenceGenerator" /> class.
         /// </summary>
-        /// <param name="limits">A sequence of integers beginning with two positive or negative 
-        /// number optionally followed by a series of greater numbers.</param>
+        /// <param name="limits">A sequence of at least two ascending numbers.</param>
```
moodmosaic: Good – I think, this description is easier to understand than the original one, and it's also more precise.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -53,7 +51,17 @@ public RandomNumericSequenceGenerator(params long[] limits)
 
             if (limits.Length < 2)
             {
-                throw new ArgumentException("The limit must be a sequence of two or more integers.");
+                throw new ArgumentException("limits must be two or more Int64.", "limits");
```
moodmosaic: This is a bit confusing to me. –  Why is `Int64` written in the exception message?

Wouldn't it be easier by re-using the sentence provided in the XML documentation for the parameter?

>A sequence of at least two ascending numbers.

(Or similar.)

---

Also, I'd be nice if we'd use correct [grammar, spelling, and punctuation](http://blog.ploeh.dk/2014/12/23/exception-messages-are-for-programmers/), by starting with a capital letter.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -53,7 +51,17 @@ public RandomNumericSequenceGenerator(params long[] limits)
...
+            }
+
+            //Check that limits are ascending
+            var previous = limits[0];
+            foreach (var item in limits.Skip(1))
+            {
+                if (previous >= item)
+                    throw new ArgumentOutOfRangeException("limits", "limits must be ascending Int64.");
+
+                previous = item;
```
moodmosaic: FWIW, I just came across [a possible way of doing that using LINQ](http://stackoverflow.com/a/14815511/467754). – What do you think? :smiley:

Src/AutoFixture/RandomNumericSequenceGenerator.cs
Pvlerick: I have to admit I don't like the term "integer" because it implies 'int', which is not the same as a 'long', so I generally use the BCL type names outside code; but I can change it back to integer or maybe simply "number"?

Will do for the capital L.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
Pvlerick: Ah, I didn't came across that when I looked for it. Using Zip, that's smart :-)

I'll make the change.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
moodmosaic: Using *numbers* sounds good – Though, I think *numbers in ascending order* can be more precise :wink:

Pvlerick: Definitely; I'll switch the message to that.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -40,8 +39,7 @@ public RandomNumericSequenceGenerator(IEnumerable<long> limits)
         /// <summary>
         /// Initializes a new instance of the <see cref="RandomNumericSequenceGenerator" /> class.
         /// </summary>
-        /// <param name="limits">An array of integers beginning with two positive or negative 
-        /// number optionally followed by a series of greater numbers.</param>
+        /// <param name="limits">A array of at least two ascending numbers.</param>
```
moodmosaic: *A array* should probably be *An array*, isn't it? :confused:

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -53,7 +51,13 @@ public RandomNumericSequenceGenerator(params long[] limits)
 
             if (limits.Length < 2)
             {
-                throw new ArgumentException("The limit must be a sequence of two or more integers.");
+                throw new ArgumentException("Limits must be at least two ascending numbers.", "limits");
+            }
+
+            //Check that limits are ascending
+            if (limits.Zip(limits.Skip(1), (a, b) => a >= b).Any(b => b))
```
moodmosaic: This looks good! – Would it make sense to extract this into a function? Then, [we wouldn't need the comment](http://butunclebob.com/ArticleS.TimOttinger.ApologizeIncode), right? :smiley:

moodmosaic:
@Pvlerick do you have more commits to add? :smiley:

ploeh:
Oh, it looks like this PR has been sitting here for a while, with no love from me :flushed: 

It looks good to me. Is anything more needed? If not, I'll take it and merge it!

Pvlerick:
@moodmosaic Nope, I think I addressed all the comments you made.

ploeh:
Running `build-release`, I get build errors...

Pvlerick:
@ploeh I will have a look

Pvlerick:
@ploeh Can you tell me what error you have when running the buildrelease script? It builds fine on my end, I'm not sure where to check for possible differences.

ploeh:
That's strange...

Here's the error:

```
c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixture\RandomNumericSequenceGenerator.cs(102): error CA2208: Microsoft.Usage : Method 'RandomNumericSequenceGen
erator.ValidateThatLimitsAreStrictlyAscending()' passes 'limits' as the 'paramName' argument to a 'ArgumentOutOfRangeException' constructor. Replace this argument with o
ne of the method's parameter names. Note that the provided parameter name should have the exact casing as declared on the method. [c:\Users\Mark\Documents\ploeh\Common\A
utoFixture\Src\AutoFixture\AutoFixture.csproj]
```

moodmosaic:
@ploeh @Pvlerick Indeed, that's strange.

I'm wondering if the error persists if `ValidateThatLimitsAreStrictlyAscending` gets turned into a static method taking `limits` as parameters...

ploeh:
What's strange is that @Pvlerick doesn't see the error. The error itself makes sense to me.

@moodmosaic, do you see the error if you run `build-release`?

moodmosaic:
@ploeh Yes, I see it:

```
h:\Projects\AutoFixture\Src\AutoFixture\RandomNumericSequenceGenerator.cs(102):
 error CA2208: Microsoft.Usage : Method 'RandomNumericSequenceGenerator.Validat
eThatLimitsAreStrictlyAscending()' passes 'limits' as the 'paramName' argument
to a 'ArgumentOutOfRangeException' constructor. Replace this argument with one
of the method's parameter names. Note that the provided parameter name should h
ave the exact casing as declared on the method. [h:\Projects\AutoFixture\Src\Au
toFixture\AutoFixture.csproj]

Code Analysis Complete -- 1 error(s), 0 warning(s)
MSBUILD : error : Code Analysis detected errors.  See Code Analysis results win
dow or log file for details. [h:\Projects\AutoFixture\Src\AutoFixture\AutoFixtu
re.csproj]
```

Pvlerick:
It seems that I don't get the code analysis when I build. I used the .ps1 file but judging from the script I don't see it doing something different than the .sh script.

I'll run it again and analyse the output to check if there are any clues there.

moodmosaic:
@Pvlerick What happens if you open Git Bash, navigate to your repo, and run:

```bash
build-release.sh
```
?

Pvlerick:
@moodmosaic Indeed, using Git Bash I get the error, while when using Posh Git and the .ps1 I didn't...

I'll fix it and investigate why it isn't showing up in the later case.

ploeh:
FWIW, I also get these errors when running `BuildRelease.ps1` in a plain vanilla PowerShell window.

moodmosaic:
Same here, getting errors when running `BuildRelease.ps` from a PowerShell window.

ploeh:
Which version of Visual Studio are you using?

I'm still on Visual Studio Premium 2013 Update 4.

Pvlerick:
@ploeh Well spotted, that's the issue :+1: 

I switched to Community 2015 over the weekend. Running the script as it is now - with the `/property:VisualStudioVersion=12.0` switch) - builds without error while running it with the `/property:VisualStudioVersion=14.0` switch gets me the error you have above.

As a side note, I uninstalled VS 2013 prior to installing 2015, but it seems that are leftovers in $env:
`VS110COMNTOOLS                 C:\Program Files (x86)\Microsoft Visual Studio 11.0\Common7\Tools\`
`VS120COMNTOOLS                 C:\Program Files (x86)\Microsoft Visual Studio 12.0\Common7\Tools\`
`VS140COMNTOOLS                 C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools\`

The condition in the script will have to be updated, I'll open a separate issue for that, and I'll fix the code analysis error.

Pvlerick:
I hope it'll work on your side now.

I do have something new on my side, building with the `/property:VisualStudioVersion=14.0` switch: 

    MSBUILD : error CA1020: Microsoft.Design : Consider merging the types defined in 'Ploeh.SemanticComparison.Fluent' with another namespace. [C:\Users\Phil\Documents\GitHub\AutoFixture\Src\SemanticComparison\SemanticComparison.csproj]
    c:\Users\Phil\Documents\GitHub\AutoFixture\Src\SemanticComparison\Likeness.cs(186): error CA1011: Microsoft.Design : Consider changing the type of parameter 'propertyPicker' in 'Likeness<TSource, TDestination>.With<TProperty>(Expression<Func<TDestination, TProperty>>)' from 'Expression<Func<TDestination, TProperty>>' to its base type 'LambdaExpression'. This method appears to only require base class members in its implementation. Suppress this violation if there is a compelling reason to require the more derived type in the method signature. [C:\Users\Phil\Documents\GitHub\AutoFixture\Src\SemanticComparison\SemanticComparison.csproj]`

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.31.2](https://www.nuget.org/packages/AutoFixture/3.31.2).

Sorry this pull request turned out to be so painful, and take so long! It's usually not that bad. A bit of it was only bad luck, but most of the troubles were my fault, because I was too slow reacting to this originally :flushed:

Pvlerick:
No harm done, we all do this on our spare time :-)

PR #411, opened 2015-06-09
Handles substitutes that have return values setup using argument matchers

### LINKED ISSUE #405: AutoConfiguredNSubstituteCustomization cannot set up override using Arg matching
Substitutes created with the fixture using `AutoConfiguredNSubstituteCustomization` cannot subsequently be overridden using NSubstitute's `Arg` matching facility.

The following five test cases are all variations of the theme

### Test Case 1: Pass
This test already exists in the project [here](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstituteUnitTest/AutoConfiguredFixtureIntegrationTest.cs#L282). It sets up a substitute, calls a method, then mocks the method and calls it again to check that the returned value is the same as the mock. So far, so good!

```CSharp
[Fact]
public void SetupCallsOverrideValueFromFixture()
{
    var substitute = fixture.Create<IInterfaceWithMethod>();
    var anonymousString = fixture.Create<string>();
    var result1 = substitute.Method(anonymousString);

    var expected = fixture.Create<string>();
    substitute.Method(anonymousString).Returns(expected);

    var result2 = substitute.Method(anonymousString);

    Assert.NotEqual(result1, result2);
    Assert.Equal(expected, result2);
}
```

### Test Case 2: Pass
This is the same as the original test case, but *without* the initial call to the substitute's method. It still passes.

```CSharp
[Fact]
public void SetupCallsOverrideValueFromFixture2()
{
    var substitute = fixture.Create<IInterfaceWithMethod>();
    var anonymousString = fixture.Create<string>();
    //var result1 = substitute.Method(anonymousString);

    var expected = fixture.Create<string>();
    substitute.Method(anonymousString).Returns(expected);

    var result2 = substitute.Method(anonymousString);

    //Assert.NotEqual(result1, result2);
    Assert.Equal(expected, result2);
}
```

### Test Case 3: Fail
The above test case passes, but when we use NSubstitute's `Arg.Any<>` matching parameter to mock the method return, it *doesn't* return the expected result

```CSharp
[Fact]
public void SetupCallsOverrideValueFromFixture3()
{
    var substitute = fixture.Create<IInterfaceWithMethod>();
    var anonymousString = fixture.Create<string>();
    //var result1 = substitute.Method(anonymousString);

    var expected = fixture.Create<string>();
    substitute.Method(Arg.Any<string>()).Returns(expected);

    var result2 = substitute.Method(anonymousString);

    //Assert.NotEqual(result1, result2);
    Assert.Equal(expected, result2);
}
```

### Test Case 4: Fail
It's a similar story for the `Arg.Is<>` parameter matcher.

```CSharp
[Fact]
public void SetupCallsOverrideValueFromFixture4()
{
    var substitute = fixture.Create<IInterfaceWithMethod>();
    var anonymousString = fixture.Create<string>();
    //var result1 = substitute.Method(anonymousString);

    var expected = fixture.Create<string>();
    substitute.Method(Arg.Is<string>(_ => true)).Returns(expected);

    var result2 = substitute.Method(anonymousString);

    //Assert.NotEqual(result1, result2);
    Assert.Equal(expected, result2);
}
```

### Test Case 5: Pass
This is where it gets interesting. By uncommenting the first call to `substitute.Method`, we can use the `Arg.Any<>` matcher to return the expected value. It's not clear what's going on behind the scenes in NSubstitute here, but my guess is that calling the method somehow clears it for future stubbing.

```CSharp
[Fact]
public void SetupCallsOverrideValueFromFixture5()
{
    var substitute = fixture.Create<IInterfaceWithMethod>();
    var anonymousString = fixture.Create<string>();
    var result1 = substitute.Method(anonymousString);

    var expected = fixture.Create<string>();
    substitute.Method(Arg.Any<string>()).Returns(expected);

    var result2 = substitute.Method(anonymousString);

    Assert.NotEqual(result1, result2);
    Assert.Equal(expected, result2);
}
```

I hope that I explained the problem clearly! I took a look at the `AutoConfiguredNSubstituteCustomization` but I don't have any knowledge of the NSubstitue codebase so it was quite difficult to understand what was going on at a glance. Hopefully if I get some time I'll do some more in-depth reading.

mrinaldi:
This PR fixes #405.

This fix checks if the call is already setup in NSubstitute and only sets a return value if it is not.
To be able to do that, I must have access to the ``ISubstituteState`` and I couldn't do that before I updated NSubstitute to 1.5.0.


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoNSubstitute 3.30.7](http://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.30.7).

alexfoxgill:
:smiley:

Issue #410, opened 2015-06-09
NumericSequenceGenerator internally uses Int32 as counter


Pvlerick:
[NumericSequenceGenerator.cs](https://github.com/AutoFixture/AutoFixture/blob/20b91b313daba05fd51714323ad0575e4f25d8c9/Src/AutoFixture/NumericSequenceGenerator.cs) internally uses an Int32 for the value it keeps increasing and returning.

Wouldn't it be more appropriate to use an Int64/long so that when a long is requested there is a chance that it actually returns something that is in that space? For the moment it will never be a long with a value higher than Int32.MaxValue.

This might be more of a question, as they might be a good reason for it - and admittedly this generator would have to be hammered quit a bit before reaching Int32.MaxValue since it starts from 0 :-)

I found this while investigating on #407

moodmosaic:
Thank you for reporting this. AFAICT, `NumericSequenceGenerator` shouldn't relate to #407, isn't it?

Pvlerick:
I stumbled upon it while investigating and trying [this](https://github.com/AutoFixture/AutoFixture/issues/407#issuecomment-108200933).

But it is not strictly related.

moodmosaic:
@Pvlerick should we close this now that #412 is merged? :smiley:

Pvlerick:
Right on!

PR #412, opened 2015-06-10
Changed internal counter to Int64

Pvlerick:
Addressing #410

I simply changed the internal counter to long/Int64 and adapted the code accordingly. I tested it by setting the internal value to something like Int64.MaxValue - 5 and it worked fine.

### DIFF
--- a/Src/AutoFixture/NumericSequenceGenerator.cs
+++ b/Src/AutoFixture/NumericSequenceGenerator.cs
@@ -9,7 +9,7 @@ namespace Ploeh.AutoFixture
     /// </summary>
     public class NumericSequenceGenerator : ISpecimenBuilder
     {
-        private int value;
+        private long value = 0;
 
         /// <summary>
         /// Creates an anonymous number.
@@ -44,9 +44,9 @@ namespace Ploeh.AutoFixture
                 case TypeCode.Int16:
                     return (short)this.GetNextNumber();
                 case TypeCode.Int32:
-                    return this.GetNextNumber();
+                    return (int)this.GetNextNumber();
                 case TypeCode.Int64:
-                    return (long)this.GetNextNumber();
+                    return this.GetNextNumber();
                 case TypeCode.SByte:
                     return (sbyte)this.GetNextNumber();
                 case TypeCode.Single:
@@ -62,7 +62,7 @@ namespace Ploeh.AutoFixture
             }
         }
 
-        private int GetNextNumber()
+        private long GetNextNumber()
         {
             return Interlocked.Increment(ref this.value);
         }


adamchester:
Shouldn't this have at least one test that guarantees the intended behaviour remains working and documented?

Pvlerick:
Well, all previous unit tests are still working.

The only way to test this would be to request longs until the internal counter reaches that space, which takes a considerable amount of time.
Alternatively we could add a constructor that takes the initial value of the internal counter and test from there.

adamchester:
So, it takes such a long time to execute before this behaviour is externally visible, that it's not practical to test it?

The obvious question then, is it ever going to be used by anyone?

Pvlerick:
That, we don't know, but it is not impossible. It is more of a theoretical issue.

The specification of the Create method is "The next number in a consecutive sequence, if request is a request for a numeric value; otherwise, a NoSpecimen instance."; in case of a long, it is not true :-)

adamchester:
Just to see what sort of time range we're dealing with here, I ran a simple test.

On my machine, it takes the following amounts of time:

Test         | Times        | Elapsed
------------ | ------------- | -------------
loop: Interlocked.Increment(long) | int.MaxValue + 1 | 00:00:11.6898759
loop: NumericSequenceGenerator TypeCode.Int64 | int.MaxValue + 1 | 00:00:24.3372879
loop: Fixture.Create long | int.MaxValue + 1 | 13+ minutes (I gave up)

To give an idea of how long it _would_ take (if I could wait), it takes around 15 seconds to create 100,000 Int64 values from Fixture.Create.

If I needed to generate so many numbers (`> Int32.MaxValue`), I'm not sure it's wise to use `AutoFixture` at all (takes far too long).

Pvlerick:
Indeed, the overhead is considerable to generate a large amount of numbers.

Oh well, that was just for consistency. The case could be made to use an Int16 internally :smile: , (or for a constructor that takes the initial value that is to be used instead of 1).

Anyway, if this is deemed useless, let's just close it.

adamchester:
I like the consistency argument, and I like the idea of adding an initial value to the constructor.

Let's see what @ploeh and/or @moodmosaic have to say :smile:

ploeh:
I ran `build-release.sh` on this PR in order to get an idea if it even looked like there would be a measurable performance difference, but I get this error:

```
Build FAILED.

"c:\Users\Mark\Documents\ploeh\Common\AutoFixture\BuildRelease.msbuild" (default target) (1) ->
(Test target) ->
  c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixtureUnitTest\FixtureTest.cs(1021): error : Ploeh.AutoFixtureUnitTest.FixtureTest.CreateAnonymousWithDoubleD
ateTimePropertiesWillAssignDifferentDates: Assert.NotEqual() Failure [c:\Users\Mark\Documents\ploeh\Common\AutoFixture\BuildRelease.msbuild]
  c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixtureUnitTest\FixtureTest.cs(1021): error :    at Ploeh.AutoFixtureUnitTest.FixtureTest.CreateAnonymousWithD
oubleDateTimePropertiesWillAssignDifferentDates() in c:\Users\Mark\Documents\ploeh\Common\AutoFixture\Src\AutoFixtureUnitTest\FixtureTest.cs:line 1021 [c:\Users\Mark\Doc
uments\ploeh\Common\AutoFixture\BuildRelease.msbuild]

    0 Warning(s)
    2 Error(s)
```

moodmosaic:
I think it's fine the way this PR is. I'm not currently sure how we could reflect the proposed change in tests.

Pvlerick:
@ploeh I'm puzzled by this build error. I don't have it when I run BuildRelease.ps1, and that tests doesn't make use NumericSequenceGenerator :confounded:

ploeh:
@Pvlerick Running it again, I don't get the same failure. I've tried quite a few times, so it looks like we have ourselves an [Erratic Test](http://xunitpatterns.com/Erratic%20Test.html) in `CreateAnonymousWithDoubleDateTimePropertiesWillAssignDifferentDates`. Looking it over, it might hint at a bug in in `RandomNumericSequenceGenerator`, which `RandomDateTimeSequenceGenerator` depends on. However, that's not your fault, and doesn't have anything to do with this PR, as far as I can tell.

I should have been more diligent, and checked multiple times, before reporting this error back. Please accept my apologies for pulling the trigger too quickly.

FWIW, I did some non-rigorous comparison between before and after this PR, but I couldn't see any difference in either execution time or memory use, so I agree that it seems reasonable from a consistency perspective to make this change.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.30.7](http://www.nuget.org/packages/autofixture/3.30.7).

Pvlerick:
@ploeh No worries, I was afraid I missed something on my end :smiley: 

I will open an issue concerning `CreateAnonymousWithDoubleDateTimePropertiesWillAssignDifferentDates`, just in case. I had a look earlier and I think I found a reason why it could fail due to randomness, but could not assess what are the chances for it to happen.

PR #416, opened 2015-06-16
Handles interfaces implementing another in AutoConfiguredNSubstitute

### LINKED ISSUE #415: AutoConfiguredNSubstituteCustomization does not set up properties when interface extends another interface from a different assembly
Apologies for the contrived title..! I have stumbled upon some strange behaviour in the AutoConfiguredNSubstituteCustomization.

I tried to make the simplest case possible to demonstrate what's happening:

**In OtherAssembly.dll**
```CSharp
public interface IExternalBar { }
```

***

**In Tests.dll**
```CSharp
public interface IFooBar : OtherAssembly.IExternalBar
{
    int MakeInteger();
    string MakeString();
}

public interface IInternalBar { }

public interface IFoo : IInternalBar
{
    int MakeInteger();
    string MakeString();
}

public class Test
{
    [Fact]
    public void SomethingFishy()
    {

        var fooBar = fixture.Create<IFooBar>();
        fooBar.MakeInteger(); // 0
        fooBar.MakeString();  // ""

        var foo = fixture.Create<IFoo>();
        foo.MakeInteger(); // 228
        foo.MakeString(); // "8725ecce-d382-4acd-8645-9a88df31e2ce"
    }
}
```

As the comments in the above demonstrate, when the fixture creates an instance of `IFooBar`, the properties are not set up as per the customization. However, the instance of `IFoo` is correctly set up. The other assembly is a simple class library with a single interface, `IExternalBar`. It does not matter whether the extending interface is in this assembly or another one: if it extends an interface in a separate assembly, it will not be configured.

mrinaldi:
This PR fixes #415.

### DIFF
--- a/Src/AutoNSubstitute/NSubstituteType.cs
+++ b/Src/AutoNSubstitute/NSubstituteType.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using NSubstitute;
-using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture.AutoNSubstitute
 {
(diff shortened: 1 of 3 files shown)


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoNSubstitute 3.30.8](https://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.30.8).

Issue #414, opened 2015-06-16
CreateAnonymousWithDoubleDateTimePropertiesWillAssignDifferentDates in FixtureTest behaves erraticaly


Pvlerick:
While inspecting/merging #412, it became apparent that the [CreateAnonymousWithDoubleDateTimePropertiesWillAssignDifferentDates](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/AutoFixtureUnitTest/FixtureTest.cs#L1014) test is [erratic](http://xunitpatterns.com/Erratic%20Test.html)

The test creates a [`DoublePropertyHolder`](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/TestTypeFoundation/DoublePropertyHolder.cs) that hold two date and then checks that the dates of the two DateTime are different.

By default, Fixture uses the default constructor of [RandomDateTimeSequenceGenerator](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/AutoFixture/RandomDateTimeSequenceGenerator.cs#L35) that uses [Now -/+ 2 years](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/AutoFixture/RandomDateTimeSequenceGenerator.cs#L22) as limits, which in turn creates a [RandomNumericSequenceGenerator](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/AutoFixture/RandomNumericSequenceGenerator.cs#L47) with the ticks of [the upper and lower as limits](https://github.com/AutoFixture/AutoFixture/blob/2aea953f6c4e532b3021579fb1f87605aa15e390/Src/AutoFixture/RandomDateTimeSequenceGenerator.cs#L42).

My calculations are the following:
- One day is 864 000 000 000 ticks
- Four years is 1 262 304 000 000 000 = (365 * 4 + 1) * 864 000 000 000 -- the +1 being for the leap year

So, when drawing a random long from RandomNumericSequenceGenerator, we have __~1 262 304 000 000 000__ values to draw from, but since every __864 000 000 000__ will give us the same date, we have __1/1461__ chance to draw two ticks that are in the same range. It makes sense since there are 1461 days in four years.

[...]

PR #417, opened 2015-06-17
Fixed erratic tests that were relying on previous behaviour

Pvlerick:
Addressing #414 

I fixed test that I could find (there were more than one actually) and deleted a test that was freezing int was deleted (I guess previous behaviour was using ints to increase DateTime and it was meant to make sure the Freeze call didn't affect that).

I ran each test in a loop 5000 times to check and they all passed.


ploeh:

Issue #404, opened 2015-05-13
Support for CoreCLR


tommck:
It seems that AutoFixture currently doesn't support the new DNX platform.

Are there plans to add support this?

ploeh:
Not yet, at least. What is DNX?

tommck:
Haha.. it's the new cross-platform asp.net framework. https://github.com/aspnet/DNX

the runtime is "dnx451" instead of "net451", etc.

ploeh:
What does "support for" mean in this context? Are you saying that it's impossible to test ASP.NET 5 from a test library with a dependency on a .NET 4 library?

tommck:
ASPNET 5 runs on multiple platforms.  We're only writing ASPNET 5 on the dnx platform right now, so it is a different set of runtime libraries.  Our code currently is not compiled for "net451", so the libraries are effectively targeting incompatible platforms

hoetz:
My tests run just fine with this project.json:

{
	"dependencies": {

		 "xunit.runner.dnx": "2.1.0-*",
		 "xunit":"2.1.0-*",
		"AutoFixture.Xunit2":"3.30-*",
		"NSubstitute": "1.8.0",
		"ManagementWeb": "",
		"Microsoft.AspNet.Mvc": "6.0.0-beta4",
		"AutoFixture": "3.30.4-*",
		"AutoFixture.AutoNSubstitute": "3.30.4-*",
		"Microsoft.Azure.Documents.Client": "0.9.2-preview",
		"WindowsAzure.Storage": "4.4.1-*",
		"DeepEqual": "1.1-*"
	},
	 "frameworks": {
        "dnx451": { }
    },
	"commands": {
		"test": "xunit.runner.dnx"
	}
}

tommck:
interesting.  I'll have to bug the guy that's working on this to see what the difference is.
Did you do anything special to make it work?

hoetz:
Not really, no. IIRC Autofixture always worked with aspnet 5 but in the past I had trouble with the xUnit extension. Now that xUnit 2 support for AF is out and DNX & xUnit play together nicely it just works.

aggieben:
Seems to the the idea here is really "support for CoreCLR", not DNX.  Anything that works for the .NET Framework 4.5 will work for 4.5 on DNX.  CoreCLR will likely require some effort to support, but I have no idea how much.  The best way to find out is to try building it for CoreCLR and see what breaks.

tommck:
[...]

Issue #132, opened 2013-06-27
Support for creating mocks from concrete types when using NSubstitute and AutoData


giulianob:
Currently, it's easy to create a mock from a concrete type with Moq since you can just use Mock<SomeType> in your test method parameters. When using NSubstitute, it's difficult to do this since it doesn't wrap the type like Moq. One idea I had was to give an attribute like [Mock] to tell AutoFixture how to create the object. I was going to do this myself but I needed to pair both [Mock] and [Freeze] and wasn't able to really follow how the whole specimen builder process works in the FreezeCustomization to make them play nicely together.

olegsych:
While understand the point @dhilgarth made regarding consistency with other AutoFixture mocking extensions, I think that it's even more important for these extensions to be consistent with their respective frameworks. In this particular case, implementing a class like `SubstituteFor<T>` doesn't fit the current design of NSubstitute and possibly the philosophy behind it. Although an implicit conversion operator, as @dhilgarth suggested, would make the experience less jarring, it can have side effects in weakly-typed code and still require use of `mock.Object` syntax in tests. I don't know if that's the case, but I wouldn't be surprised if the idea was rejected by the NSubstitute team. 
 
I like the original suggestion @dhilgarth made regarding a special parameter attribute, which could look like this.
```C#
public void SomeTest([Substitute] ConcreteClass dependency)
{
}
```

As @ploeh mentioned, CLR explicitly leaves order of attributes undefined, so the `[Frozen]` attribute could customize the fixture before the `[Substitute]` does and you can't predict if a concrete type or a substitute will get created. However, I agree with @giulianob in that this shouldn't limit AutoFixture and the fact that it does smells like a design limitation in AutoFixture or an implementation problem with the `FrozenAttribute`.

[...]

Issue #118, opened 2013-06-12
AutoFixture.Extensions


ecampidoglio:
While looking at the attributes defined in the [AutoFixture.xUnit](http://nuget.org/packages/AutoFixture.Xunit) glue library, I noticed that only few of them are actually dependent on xUnit:
- [AutoDataAttribute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/AutoDataAttribute.cs)
- [CompositeDataAttribute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/CompositeDataAttribute.cs)
- [InlineAutoDataAttribute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs)

These attributes are tied to the xUnit.Extensions API in which they derive, directly or indirectly, from [DataAttribute](http://xunit.codeplex.com/SourceControl/latest#src/xunit.extensions/DataTheories/DataAttribute.cs). The other attributes in the package, on the other hand, stand on their own and could therefore easily be used together with other test frameworks.

On top of that, by looking at the `AutoDataAttribute`, I noticed that the implementation itself isn't coupled to xUnit any more than by the signature of the `GetData` method it overrides:

``` csharp
public override IEnumerable<object[]> GetData(MethodInfo methodUnderTest, Type[] parameterTypes)
{
    if (methodUnderTest == null)
    {
      throw new ArgumentNullException("methodUnderTest");
    }

    var specimens = new List<object>();
    foreach (var p in methodUnderTest.GetParameters())
    {
      this.CustomizeFixture(p);

      var specimen = this.Resolve(p);
      specimens.Add(specimen);
    }

    return new[] { specimens.ToArray() };
}
```

Things become even more interesting when we look at the [ParameterDataAttribute](http://bazaar.launchpad.net/~nunit-core/nunitv2/trunk/view/head:/src/NUnitFramework/framework/Attributes/ValuesAttribute.cs) class, used to declare [parameterized tests](http://nunit.org/index.php?p=parameterizedTests&r=2.5) in NUnit:

``` csharp
/// <summary>
/// Abstract base class for attributes that apply to parameters 
/// and supply data for the parameter.
/// </summary>
[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
public abstract class ParameterDataAttribute : Attribute
{
[...]

PR #418, opened 2015-06-19
Extract common GetData logic from framework-specific AutoDataAttribute classes

olegsych:
Defining test data customization attributes that are specific to a _mocking_ framework without taking a dependency on a specific _test_ framework is not possible today because the `CustomizeAttribute` classes are defined in the _test framework assemblies_, like `Ploeh.AutoFixture.Nunit2`, `Ploeh.AutoFixture.Xunit` and `Ploeh.AutoFixture.Xunit2`, while the mocking attributes, such as the `SubstituteAttribute` added in #430, have to be defined in the _mocking framework assemblies_, like the `Ploeh.AutoFixture.NSubstitute`. Because of this limitation, the usage of the `SubstituteAttribute` can produce obscure test failures, described in #430, due to missing fixture customizations.

Here is the desired usage scenario:

```C#
using Ploeh.AutoFixture.NSubstitute;
using Ploeh.AutoFixture.xUnit;
using Xunit;

public class MyTest
{
    [Fact, AutoData]
    public void Test1([Substitute]MyType sample)
    {
        // Fixture setup
        // Exercise system
        sample.HelloWorld();
        // Verify results
        sample.Received().HelloWorld();
        // Teardown
    }
}
```

Today, this test fails with an obscure error saying that the object instance for which the `Received` method is called is not a substitute. Developer writing this test has to know that an `AutoNSubstituteCustomization` or an `AutoConfiguredNSubstituteCustomization` has to be used for this test to work correctly.

To solve this problem, this pull request 
- introduces a new Kernel interface, `IParameterCustomizationProvider` that can be implemented by attributes defined by any of the _mocking_ assemblies, like `AutoFixture.NSubstitute` and `AutoFixture.AutoMoq`, without additional dependencies. 
- implements a new `TestDataProvider` utility class that uses `IParameterCustomizationProvider` and consolidates the fixture customization logic from the three separate `AutoDataAttribute` implementations in `AutoFixture.Nunit2`, `AutoFixture.xUnit` and `AutoFixture.xUnit2`. 
- modifies the existing `AutoDataAttribute` implementationsto use the common logic in the `TestDataProvider`.

Once this pull request is merged, the existing `SubstituteAttribute` can be changed to implement the new `IParameterCustomizationProvider` interface, which will make the test in the scenario above work as expected.

### DIFF
--- a/Src/AutoFixture.NUnit2.UnitTest/AutoDataAttributeTest.cs
+++ b/Src/AutoFixture.NUnit2.UnitTest/AutoDataAttributeTest.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
-using Ploeh.AutoFixture.NUnit2.Addins;
 using NUnit.Framework;
+using Ploeh.AutoFixture.Kernel;
+using Ploeh.AutoFixture.NUnit2.Addins;
 using Ploeh.TestTypeFoundation;
 
 namespace Ploeh.AutoFixture.NUnit2.UnitTest
@@ -147,6 +149,60 @@ namespace Ploeh.AutoFixture.NUnit2.UnitTest
             // Verify outcome
             Assert.True(new[] { expectedResult }.SequenceEqual(result.Single()));
             // Teardown
+        }
+
+        [Test]
+        public void GetDataReturnsValuesSuppliedByTestDataProvider()
+        {
+            // Fixture setup
+            var arguments = new[] { new object(), new object() };
+            var provider = new DelegatingTestDataProvider { OnGetData = m => arguments };
+            var sut = new TestableAutoDataAttribute(new DelegatingFixture()) { OnCreateDataProvider = () => provider };
+            // Excercise system
+            IEnumerable<object[]> result = sut.GetData(new Action<object>(ParameterizedMethod).Method);
+            // Verify outcome
+            Assert.AreEqual(arguments, result.Single());
+            // Teardown
+        }
+
+        [Test]
+        public void CreateDataProviderReturnsTestDataProviderConstructedWithGivenFixture()
+        {
+            // Fixture setup
+            bool fixtureInvoked = false;
+            var fixture = new DelegatingFixture { OnCreate = (request, context) => fixtureInvoked = true };
+            var sut = new TestableAutoDataAttribute(fixture);
+            // Excercise system
+            ITestDataProvider provider = sut.TestableCreateDataProvider();
+            // Verify outcome
+            provider.GetData(new Action<object>(ParameterizedMethod).Method);
+            Assert.IsTrue(fixtureInvoked);
+            // Teardown
+        }
+
+        private static void ParameterizedMethod(object parameter)
+        {
+        }
+
+        private class TestableAutoDataAttribute : AutoDataAttribute
+        {
+            public Func<ITestDataProvider> OnCreateDataProvider;
+
+            public TestableAutoDataAttribute(IFixture fixture)
+                : base(fixture)
+            {
+                this.OnCreateDataProvider = base.CreateDataProvider;
+            }
+
+            public ITestDataProvider TestableCreateDataProvider()
+            {
+                return base.CreateDataProvider();
+            }
+
+            protected override ITestDataProvider CreateDataProvider()
+            {
+                return this.OnCreateDataProvider();
+            }
         }
     }
 }
--- a/Src/AutoFixture.NUnit2.UnitTest/AutoFixture.NUnit2.UnitTest.csproj
+++ b/Src/AutoFixture.NUnit2.UnitTest/AutoFixture.NUnit2.UnitTest.csproj
@@ -12,7 +12,7 @@
     <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <TargetFrameworkProfile />
-    <CodeAnalysisCulture>en-US</CodeAnalysisCulture> 
+    <CodeAnalysisCulture>en-US</CodeAnalysisCulture>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
@@ -57,6 +57,7 @@
     <Compile Include="DelegatingCustomizeAttribute.cs" />
     <Compile Include="DelegatingFixture.cs" />
     <Compile Include="DelegatingSpecimenBuilder.cs" />
+    <Compile Include="DelegatingTestDataProvider.cs" />
     <Compile Include="DependencyConstraints.cs" />
     <Compile Include="FakeDataAttribute.cs" />
     <Compile Include="FavorArraysAttributeTest.cs" />
@@ -91,6 +92,9 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="packages.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
--- a/Src/AutoFixture.NUnit2.UnitTest/CustomizeAttributeTest.cs
+++ b/Src/AutoFixture.NUnit2.UnitTest/CustomizeAttributeTest.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using NUnit.Framework;
+using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture.NUnit2.UnitTest
 {
@@ -25,6 +26,16 @@ namespace Ploeh.AutoFixture.NUnit2.UnitTest
             var sut = new DelegatingCustomizeAttribute();
             // Verify outcome
             Assert.IsInstanceOf<Attribute>(sut);
+            // Teardown
+        }
+
+        [Test]
+        public void SutImplementsIParameterCustomizationProviderToBeRecognizedByTestDataProvider()
+        {
+            // Fixture setup
+            // Exercise system 
+            // Verify outcome
+            Assert.True(typeof(IParameterCustomizationProvider).IsAssignableFrom(typeof(CustomizeAttribute)));
             // Teardown
         }
     }
--- a/Src/AutoFixture.NUnit2.UnitTest/DelegatingTestDataProvider.cs
+++ b/Src/AutoFixture.NUnit2.UnitTest/DelegatingTestDataProvider.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.NUnit2.UnitTest
+{
+    internal class DelegatingTestDataProvider : ITestDataProvider
+    {
+        public Func<MethodInfo, IEnumerable<object>> OnGetData = method => Enumerable.Empty<object>();
+
+        public IEnumerable<object> GetData(MethodInfo testMethod)
+        {
+            return this.OnGetData(testMethod);
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit2/AutoDataAttribute.cs
+++ b/Src/AutoFixture.NUnit2/AutoDataAttribute.cs
@@ -3,8 +3,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using Ploeh.AutoFixture.NUnit2.Addins;
 using Ploeh.AutoFixture.Kernel;
+using Ploeh.AutoFixture.NUnit2.Addins;
 
 namespace Ploeh.AutoFixture.NUnit2
 {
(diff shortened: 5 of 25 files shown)


ploeh:
Thank you for your interest in AutoFixture.

AFAICT, this contribution creates new public types that we will have to maintain for a long time. There ought to be some benefits that outweigh that disadvantage; what are they?

This is reminiscent of [a discussion we've already had](https://github.com/AutoFixture/AutoFixture/issues/118). Additionally, I'd like to point out Udi Dahan's excellent article [Beware the Share](http://programmer.97things.oreilly.com/wiki/index.php/Beware_the_Share).

olegsych:
Thanks for responding so quickly, @ploeh. I know you're busy, so let me try to answer your concerns as completely and succinctly as I can. I will also update the pull request description to include a better description of the scenarios this pull request enables.

> AFAICT, this contribution creates new public types that we will have to maintain for a long time. There ought to be some benefits that outweigh that disadvantage; what are they?

- The `IParameterCustomizationProvider` interface enables development of mocking framework-specific attributes _independently_ of the test framework-specific attributes. 
- Implementing an attribute like `[Substitute]` discussed in #132 without the `IParameterCustomizationProvider` interface, would require breaking the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) to support all three test frameworks currently supported by AutoFixture.
- The `TestDataProvider` consolidates 3 separate, identical implementation of GetData methods from `AutoFixture.Nunit2`, `AutoFixture.xUnit.net` and `AutoFixture.xUnit.net2`, reducing the future maintenance cost.
- The `TestDataProviderTest` provides comprehensive test coverage for the previously _untested_ logic of the `AutoDataAttribute` implementations that apply `ICustomization` instances returned by `CustomizeAttribute` classes to the test `IFixture`. 
- Implementing support for the `IParameterCustomizationProvider` interface in the `AutoDataAttribute` classes in the test framework assemblies, without introduction of the `TestDataProvider` would break the _rule of three_.

> This is reminiscent of a discussion we've already had. Additionally, I'd like to point out Udi Dahan's excellent article Beware the Share.

-  Since #118 was closed 2 years ago, the introduction of support for XUnit v2 has already broke the _rule of three_ :smile:. AutoFixture now has 3 separate copies of the `AutoDataAttribute` and `CustomizeAttribute` classes. Taking into consideration that the `CustomizeAttribute` seems to have withstood the test of time and reuse, I think it's worthwhile to consider reopening this discussion.
- While consolidating separate `CustomizeAttribute` definitions into a single class in a separate NuGet package referenced by the test framework packages would be ideal, the proposed solution in this pull request makes a step toward that solution without breaking backward compatibility for existing consumers of v3.x. 
- A separate assembly to hold the consolidated `CustomizeAttribute` could also be introduced without breaking backward compatibility with the help of [TypeForwardedToAttribute](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.typeforwardedtoattribute(v=vs.110).aspx) in the test framework assemblies. The advantage of doing it this way is that it allows eliminating the `IParameterCustomizationProvider` interface at the cost of adding a new project/assembly/package to AutoFixture. Depending on how many other types can be immediately moved there, this could be a better alternative.
- While I completely agree with Udi's points regarding the cost of sharing code, in this particular case, _not sharing_ the `GetData` implementation creates a bigger problem because excessive duplication can affects the code similarly to the over-aggressive sharing does. Because of the reasons listed above, evolution of the mocking framework assemblies, such as `AutoFixture.NSubstitute`, independently from test framework assemblies, such as `AutoFixture.NUnit2` is effectively discouraged today. The current implementation of `AutoDataAttribute` classes encourages additional duplication and dependencies because requires creation of assemblies, like `AutoFixture.NSubstitute.Nunit2`, `AutoFixture.NSubstitute.xUnit.net` and `AutoFixture.NSubstitute.xUnit.net2` and so on, while discouraging thorough testing due to the geometric rate of code growth.

ploeh:
I'm torn.

There are many good points here. OTOH, I'm beginning to think that the AutoFixture core library is becoming quite bloated, and I'm loath to put even more stuff into it.

From my perspective, the Rule of Three describes a *minimum threshold:* before one has the same code in triplicate, there's too little information to separate the differences from the similarities. IME, this doesn't meant that one *must* refactor to remove the duplication as soon as the duplication count reaches three, but rather that this refactoring shouldn't be done *before* that threshold.

With the imminent release of *.NET Core*, it looks like there's much work ahead getting AutoFixture working in other configurations (re: #404). I don't think it'd be possible to add .NET Core support to AutoFixture without breaking changes, so we're looking at a potential AutoFixture 4.

The reason I bring this up is that this will most likely involve removing a lot of code (and features) from the core AutoFixture library, so I'm not sure now is the best time to *add* code.

As initially stated, I'm in doubt, so I'd like to solicit other opinions; particularly those of the other core contributors @moodmosaic, @ecampidoglio, and @adamchester.

moodmosaic:
This looks quite impressive piece of work :smiley:

I think it goes off on a tangent, as the core AutoFixture library isn't aware of test methods and parameters – that's where AutoFixture with xUnit.net/NUnit Glue Libraries come into play, optionally followed by one of the auto-mocking Glue Libraries.

After all, it becomes quite common scenario to compose multiple Customizations, so taking a dependency on a specific test framework shouldn't be a problem per se.

That being said, I can imagine users of the core AutoFixture library being quite confused by the proposed `IParameterCustomizationProvider`.

(On a side note: I haven't gone through all the changes, but I'm wondering whether some implementations of `IParameterCustomizationProvider` can be subject to the [Constrained Construction anti-pattern](http://blog.ploeh.dk/2011/04/27/Providerisnotapattern/).)

IMHO, taking into account @ploeh's [comment](https://github.com/AutoFixture/AutoFixture/pull/418#issuecomment-120075929) and given all the above, I wouldn't move forward with this...

olegsych:
Now that the `SubstituteAttribute` is implemented, I'd like to continue working on this pull request because it provides the foundation required to prevent the obscure test failures due to missing NSubstitute customization described in #430.

ploeh:
I must admit that I forgot all about this pull request :flushed: 

Does it add anything that AutoFixture *needs?* In other words, are there features we can't add to AutoFixture, that we could add with the code in this pull request?

olegsych:
I have updated description of the pull request to answer these questions and accurately reflect the current implementation of `AutoFixture`.

ploeh:
The design philosophy behind AutoFixture is that it tries to follow the 80-20 rule. If it can understand 80 % of your API, it's already done a lot of work for you, but given that a user could do *anything* with their API, it'd be impossible to address all quirks.

AutoFixture is quite opinionated in how it expects an API to be designed. Those opinions are based on fundamental principles of OOD, such as *design by contract*.

Even though I'm the main designer of AutoFixture, and the opinions embedded in it reflect mine, I rarely use AutoFixture without tweaking it.

First, you may find yourself in a situation where you want to write a test with this signature:

```C#
[Fact, AutoData]
public void Test1(MyType sample, IMyInterface iface)
```

That's not going to be possible, because vanilla AutoFixture doesn't know how to create an instance of an interface. You'll have to create a custom attribute, and pull in one or more customizations, such as `AutoNSubstituteCustomization`.

Yes, you'll have to know how to do that, but in the motivating example here, you'll also have to know that a `[Substitute]` attribute exists.

Even so, it rarely stops there. Often (in practice: *always*) you'll have to encapsulate some domain rule that you'll have to tell AutoFixture about. Imagine, for example, that you'll have to deal with [ISBNs](https://en.wikipedia.org/wiki/International_Standard_Book_Number) in your code. You may introduce a Value Object that encapsulates the rules about ISBN (valid digits, check sums), but you still need to populate the Value Object with values. You may, for example, pass a `string` to the constructor, but AutoFixture's default string algorithm isn't going to produce valid ISBNs. This means that you have to customise how AutoFixture generates ISBNs.

The point here is that no matter how much you try, you can't eliminate the need for customising AutoFixture. Thus, I'm not particularly concerned about the need for explicitly pulling in `AutoNSubstituteCustomization`.

Another design principle behind AutoFixture is from the [Zen of Python](https://www.python.org/dev/peps/pep-0020) - particularly the second rule: *Explicit is better than implicit*.

For that reason, I don't wish to pull AutoFixture in a direction where it's becoming even more 'auto-magical' than it already is.

In closing, I don't think this proposed change solves a real problem, and while I think you for your interest, I don't want to inflate the kernel's public API without a sound reason.

olegsych:
Thanks for sharing your thoughts on this. I do appreciate it. I hope I can trouble you with one last question. :fearful:   

Could you help me understand why you are rejecting a new implementation of what appears to be an already established AutoFixture design pattern? Using examples from you previous comment, you don't want to allow the following usage of AutoFixture.

```C#
[Theory, AutoData]
public void Test1([Substitute]IMyInterface iface)
```

While the following usage is allowed and already works _as is_, no extra code needed.

```C#
[Theory, AutoData]
public void Test1([RegularExpression(@"\d{9}[\d|X]")]string isbn)
```

As a user of AutoFixture looking at this hypothetical code written by someone else on my team, I'd be surprised that the `System.ComponentModel.DataAnnotations` attributes work out of the box even though they have no relation to AutoFixture, while a built-in attribute that comes with `Ploeh.AutoFixture.NSubstitute` does not. Wouldn't both of these scenarios fit in the 80%?

I completely agree that explicit test fixture configuration is better than implicit. That is why I'd prefer the usage scenarios listed _above_ instead of the following.

```C#
[Theory, MyTestConventions]
public void Test1(string isbn, IMyInterface iface);
```

The test fixture here is implicit. I think of it as a [standard fixture](http://xunitpatterns.com/Standard%20Fixture.html) implemented as a test conventions attribute. When used in teams with mixed expertise levels, _standard fixtures_ tend to devolve into [shared fixtures](http://xunitpatterns.com/Shared%20Fixture.html), which I prefer to avoid. By defining a global test convention attribute like the one above, I am discouraging the use of [minimal fixtures](http://xunitpatterns.com/Minimal%20Fixture.html) for individual tests and parameters (based on the `CustomizeAttribute` and attributes in the `DataAnnotations` namespace) and encouraging the use of a _standard fixture_. 

I hope we agree that _minimal_ and _standard_ fixture designs are both valid and have benefits making them uniquely suitable in different circumstances. Based on my interpretation, the AutoFixture design today supports both, by allowing users to implement _standard_ fixtures by subclassing the `AutoDataAttribute` and to implement _minimal_ fixtures by reusing a number of attributes from the `DataAnnotations` namespace and by subclassing the `CustomizateAttribute` when necessary.

The changes proposed in this pull request are an _incremental_ improvement of the _existing_ design. They enable implementing first-class support for attributes for minimal fixtures in other _AutoFixture_ assemblies and not just the core or the customer assemblies supported today.

I hope I was able to convince you that this pull request does not try to pull AutoFixture in a new direction. With that hope, could you address the _specific_ changes in this pull request? 
- Introduction of the `IParameterCustomizationProvider` interface in the `AutoFixture.Kernel`? (this is a key design change, but naming, definition, visibility and location of the new abstraction are flexible)
- Implementation of the `TestDataProvider` utility class in Kernel? (this is optional and can be easily eliminated)
- Changing of the `AutoDataAttribute` implementations to recognize the new `IParameterCustomizationProvider` interface? (this is key implementation change, but the implementation is flexible)
- Changing of the `AutoDataAttribute` implementations to use the `TestDataProvider` utility class? (this is also optional and can be removed)

ploeh:
What's already in AutoFixture is a result of a 6½ year journey (so far). Lots of things would be different if I'd known then what I know now.

I try to manage AutoFixture a bit like one would captain an oil tanker: if you want to stay clear of an obstacle, you have to start turning in good time before reaching it. While I'm sympathetic to the argument about established precedents, some common idioms have turned out to be less beneficial than others. If I think that a certain established way of doing things pull AutoFixture in the wrong direction, I'll stop it from spreading.

Adorning methods and arguments with attributes fall in a grey zone.

The main reason the attributes from `System.ComponentModel.DataAnnotations` just work out of the box is that these types are part of the Base Class Library, on which AutoFixture already has a dependency. In order to implement them, all that was needed was some concrete classes.

I have a much higher tolerance for adding concrete classes than I have for adding interfaces. Well-designed interfaces should represent *abstractions*, and these should be discovered instead of invented. The AutoFixture kernel already has too many poorly designed interfaces (some of which are fortunately already deprecated). I'm actively looking for ways to get rid of some of them. Adding new interfaces to the AutoFixture core library runs contrary to that effort.

If there's a way to do what you want that doesn't introduce new interfaces (or abstract base classes) it might be worth looking into.

Ultimately, I still don't find that this is much of an issue anyway. Some time back, someone did an analysis of what it would take to turn AutoFixture into a Portable Class Library, and one of the observations was that the data annotations are missing from some configurations. This might be the case for CoreCLR as well. In that case, my plan was to pull support for data annotations out of the core library, and add it to another library with extensions that you can pull in if you're running on the full platform. If this happens (it would be a breaking change), support for the `[RegularExpression]` attribute would be in exactly the same position as the `[Substitute]` attribute, and I would have no problem with that.

If you pull in types from a downstream dependency, I think it's only natural that you have to configure AutoFixture to tell it how to deal with that type. That's the current situation of `[Substitute]`.

If you pull in types from an upstream dependency, I wouldn't be surprised if AutoFixture just works with that type out of the box. That's the case of the data annotations.

olegsych:
Thank you for the response. I am disappointed that the prescribed solution for problems like #419 and #430 is for the users to know why they occur and how to configure AutoFixture and their test environment to fix them. While I realize that the answers may be obvious to someone who developed AutoFixture or spent enough time to understand how it works, the combination of the obscure errors with complexity of the core take too much value out of using it in the first place. With no productive way of addressing these issues in AutoFixture, it's hard to justify using it in projects I'm working on.

ploeh:
If AutoFixture errors are obscure, we can work on making them better.

PR #421, opened 2015-06-21
[WIP] Substitute attribute enabling creation of substitutes for concrete types with NSubstitute

olegsych:
The proposed `SubstituteAttribute` enables changing the default fixture behavior and forcing creation of substitutes of concrete types for individual test parameters. The implementation follows the pattern established by the `AutoFixture` for data annotation attributes, such as `[Range]`, `[RegularExpression]` and `[StringLength]`.

Why is this useful? The `SubstituteAttribute` enables declarative use of AutoFixture to isolate SUT from dependencies defined as concrete types with virtual members. The idea was previously discussed in #132.

Here are samples that illustrate this attribute's use:

```C#
[Theory, AutoDataWithAutoSubstituteCustomization]
public void ConcreteTypesAreSubstitutedWhenParameterIsMarkedWithSubstituteAttribute(
    [Substitute]ConcreteType t, object obj)
{
    t.VirtualMethod(obj);
    t.Received().VirtualMethod(obj);
}

[Theory, AutoDataWithAutoSubstituteCustomization]
public void SubstituteAttributeAppliesToSingleParameterAndNotItsEntireType(
    [Substitute]ConcreteType t1, ConcreteType t2, object obj)
{
    t2.VirtualMethod(obj);
    Assert.Throws<NotASubstituteException>(() => t2.Received().VirtualMethod(obj));
}
```

Applying the `SubstituteAttribute` to a parameter of a concrete type fails the test, such as the one in example below, _before_ running it.

```C#
[Theory, AutoDataWithAutoSubstituteCustomization]
public void AttemptingToSubstituteSealedTypeFailsTestBeforeItStarts(
    [Substitute]SealedType t, object obj)
{
}
```

Applying the `SubstituteAttribute` in a method with a regular `AutoDataAttribute` results in an obscure failure within the test itself because the attribute has no behavior and requires the `AutoSubstituteCustomization` to be applied to the test's `Fixture`. Below is a sample that illustrates the problem. A separate pull request #418 was started earlier to enable addressing it.

```C#
[Theory, AutoData]
public void SubstituteAttributeShouldWorkWithPlainAutoDataAttribute(
    [Substitute]ConcreteType t, object obj)
{
    t.VirtualMethod(obj);
    // The following line throws NotASubstituteExeption, which is confusing because the
    // Substitute attribute is clearly used.
    t.Received().VirtualMethod(obj);
}
```

#### AutoFixture.NSubstitute
- Implement `SubstituteAttribute`.
- Implement `SubstituteRequest` used to request a substitute of a given type.
- Implement `SubstituteBuilder` to handle substitute requests.
- Implement `SubstituteAttributeRelay` to convert requests for parameters, properties and fields marked with the `SubstituteAttribute` to substitute requests of their respective types.
- Implement `SubstituteRelay` to convert requests for abstract types and interfaces to requests for substitutes.
- Implement `AutoSubstituteCustomization` to encapsulate addition of the new components to an `IFixture`. 

#### AutoFixture
- Extract common logic from the existing `RangeAttributeRelay`, `RegularExpressionAttributeRelay` and `StringLengthAttributeRelay` classes to a new abstract base class, called `AttributeRelay<T>`, which also serves as a base class for the new `SubstituteAttributeRelay`.

### Open Questions
- [ ] Should the new `SubstituteBuilder` logic be merged into the existing `NSubstituteBuilder`? Note that the new class is meant to be a customization, while the existing class serves as a residue collector.
- [ ] Should the new `AutoSubstituteCustomization` logic be merged into the existing `AutoNSubstituteCustomization` and the `AutoConfiguredNSubstituteCustomization`? I don't foresee any observable behavior changes in pure usage scenarios. However, the existing classes expose internal implementation components through properties and there is a potential that someone might have taken a dependency on them.


ploeh:
Thank you for your interest in AutoFixture, and please accept my apologies that I've been unresponsive for so long. (Not that it's any excuse, but I've been busy the last month or so, but am now coming off some work, and may have more time in the near future. Some of that time is about to be used for a vacation, though.)

At the moment, I haven't done a full review of this PR, because it contains at least one blocker in its current form.

While I realise that it's possible to find examples of inheritance in AutoFixture, the guiding principle throughout is to *favour composition over inheritance*. For years, I've been trying to remove what inheritance remains in AutoFixture, and I don't want to add more today.

There can be special cases, but inheritance introduced for the sake of reuse isn't one of them. It's always possible to find other ways to reduce duplication without relying on inheritance.

For this reason, I'm closing this pull request. This doesn't mean that I'm rejecting the overall idea, but please try to [make each pull request do only one thing](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests), so that we can discuss each small change individually.

PR #424, opened 2015-07-10
Introduce AttributeRelay<T> class

olegsych:
This pull request extracts common logic from the existing `RangeAttributeRelay`,  `RegularExpressionAttributeRelay` and `StringLengthAttributeRelay` classes to a new concrete class, called  `AttributeRelay<T>`. The existing classes are modified to reuse the new class in their implementations of the `ISpecimenBuilder.Create` method.

### Motivation
Implementations of the `ISpecimenBuilder.Create` method in the three existing classes contain complex logic for finding custom attributes applied to a code element (a parameter, a property, etc), creating a relayed request based on a well-known attribute and resolving it from the `ISpecimenContext`. Most of this logic does not depend on the specific attribute type or the code element to which it is applied and thus could be reused by the existing classes. 

The new `AttributeRelay<T>` class can also be reused by the `SubstituteAttribute`, which was previously submitted in pull request #421 and will be re-submitted in a separate, smaller pull request. The `SubstituteAttribute` is a major motivation for this refactoring, because without the `AttributeRelay<T>`, one would need to create yet another copy of the attribute relay boilerplate code, brining the total number of nearly identical copies to 4. The effort required to extract existing duplicate logic into a reusable component is comparable to re-implementing another copy, but unlike the latter, it reduces the overall amount of code in AutoFixture and improves test coverage.

### Usage
In addition to using `AttributeRelay<T>` when implementing the `ISpecimenBuilder.Create` method, the new class could also be used directly, without the need to introduce a separate `ISpecimenBuilder` class. Here is an example of how it could be used to support the upcoming `SubstituteAttribute`.

```C#
public class AutoSubstituteCustomization : ICustomization
{
  {
    // ...
    fixture.Customizations.Insert(0, new AttributeRelay<SubstituteAttribute>(SubstituteRequest.Create));
  }
}

public class SubstituteRequest
{
  public static object Create(ICustomAttributeProvider codeElement, SubstituteAttribute attribute)
  {
    return new SubstituteRequest(/* ... */);
  }
}
```

### Out of scope
The following changes may be worth considering in the future. They were not included in this pull request to keep it small and speed up the code review. 
* Moving the private `CreateRelayedRequest` method from the existing `AttributeRelay` classes to their respective `Request` classes and making them public would allow to significantly simplify tests. Because these methods are private today, testing them requires stubbing and mocking the `ISpecimenBuilder` and `ISpecimenContext` interfaces. Simply making them public in the `Relay` classes would feel like exposing internal implementation details. However, if they were defined on the respective `Request` classes, they would be fairly typical factory methods and making them public would be natural.
* Eliminating the existing `AttributeRelay` classes and using the new `AttributeRelay<T>` directly would help to reduce the amount of product and test code. With `ISpecimenBuilder.Create` implemented by the `AttributeRelay<T>` and the `CreateRelayedRequest` methods moved to their respective `Request` classes, the existing `AttributeRelay` classes would become empty shells with 2 lines of code(excluding declarations) per class, but still require a significant amount of test infrastructure. This would constitute a breaking change and have to wait until v4.

### DIFF
--- a/Src/AutoFixture/DataAnnotations/AttributeRelay.cs
+++ b/Src/AutoFixture/DataAnnotations/AttributeRelay.cs
@@ -1,0 +1,109 @@
+﻿using System;
+using System.ComponentModel.DataAnnotations;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture.DataAnnotations
+{
+    /// <summary>
+    /// Relays requests made for properties, fields or parameters to different requests, based on custom attributes 
+    /// applied to the code elements. 
+    /// </summary>
+    /// <remarks>
+    /// A typical example is relaying a request for a parameter that has <see cref="RangeAttribute"/> applied to it 
+    /// to a <see cref="RangedNumberRequest"/>. Another way to describe it, a request for a <see cref="ParameterInfo"/>
+    /// instance that contains a <see cref="RangeAttribute"/>, is transformed to a <see cref="RangedNumberRequest"/>
+    /// with parameters extracted from the attribute.
+    /// </remarks>
+    /// <typeparam name="TAttribute">
+    /// An <see cref="Attribute"/> type that can be applied to a property, a field or a method. The attribute is 
+    /// expected to allow applying only once. When the same attribute is applied multiple times to the same code 
+    /// element, an <see cref="InvalidOperationException"/> will be thrown.
+    /// </typeparam>
+    public class AttributeRelay<TAttribute> : ISpecimenBuilder
+        where TAttribute : Attribute
+    {
+        private readonly Func<ICustomAttributeProvider, TAttribute, object> createRelayedRequest;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AttributeRelay{TAttribute}"/> class.
+        /// </summary>
+        /// <param name="createRelayedRequest">
+        /// A delegate responsible for creating a relayed request based on an <see cref="Attribute"/> applied to to the 
+        /// code element represented by an <see cref="ICustomAttributeProvider"/>.
+        /// </param>
+        public AttributeRelay(Func<ICustomAttributeProvider, TAttribute, object> createRelayedRequest)
+        {
+            if (createRelayedRequest == null)
+            {
+                throw new ArgumentNullException("createRelayedRequest");
+            }
+
+            this.createRelayedRequest = createRelayedRequest;
+        }
+
+        /// <summary>
+        /// Gets a <see cref="MethodInfo"/> representing the delegate specified in the constructor.
+        /// </summary>
+        /// <remarks>
+        /// This property is meant to be used in structural inspection tests for classes that use the
+        /// <see cref="AttributeRelay{T}"/>. It returns a <see cref="MethodInfo"/> instead of the actual
+        /// delegate to clearly state that the delegate is not meant to be invoked directly by this class'
+        /// consumers.
+        /// </remarks>
+        public MethodInfo CreateRelayedRequestMethod 
+        {
+            get { return this.createRelayedRequest.Method; }
+        }
+
+        /// <summary>
+        /// Creates a relayed request based on a custom attribute applied to a code element and resolves it from the 
+        /// given <paramref name="context"/>.
+        /// </summary>
+        /// <returns>
+        /// A specimen resolved from the <paramref name="context"/> based on a relayed request.
+        /// If the <paramref name="request"/> code element does not have attribute of expected type applied to it, 
+        /// returns <see cref="NoSpecimen"/>.
+        /// </returns>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            var customAttributeProvider = request as ICustomAttributeProvider;
+            if (customAttributeProvider == null)
+            {
+                return new NoSpecimen(request);
+            }
+
+            TAttribute attribute;
+
+            try
+            {
+                attribute = customAttributeProvider.GetCustomAttributes(typeof(TAttribute), true)
+                    .OfType<TAttribute>().SingleOrDefault();
+            }
+            catch (InvalidOperationException e)
+            {
+                throw new InvalidOperationException(
+                    string.Format(
+                        CultureInfo.CurrentCulture, 
+                        "Multiple instances of {0} were found for {1}.", 
+                        typeof(TAttribute), request), 
+                    e);
+            }
+
+            if (attribute == null)
+            {
+                return new NoSpecimen(request);
+            }
+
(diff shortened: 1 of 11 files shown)


Src/AutoFixture/DataAnnotations/AttributeRelay.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// to a <see cref="RangedNumberRequest"/>. Another way to describe it, a request for a <see cref="ParameterInfo"/>
+    /// instance that contains a <see cref="RangeAttribute"/>, is transformed to a <see cref="RangedNumberRequest"/>
+    /// with parameters extracted from the attribute.
+    /// </remarks>
+    /// <typeparam name="TAttribute">
+    /// An <see cref="Attribute"/> type that can be applied to a property, a field or a method. The attribute is 
+    /// expected to allow applying only once. When the same attribute is applied multiple times to the same code 
+    /// element, an <see cref="InvalidOperationException"/> will be thrown.
+    /// </typeparam>
+    public class AttributeRelay<TAttribute> : ISpecimenBuilder
```
ploeh: Could this class be made `internal`?

olegsych: For this particular set of changes, the `AttributeRelay<TAttribute>` could be `internal`. However, to help with implementation of attribute relays outside of the `Ploeh.AutoFixture` assembly it would need to be either public or `[InternalsVisibleTo]`. In particular, I'd like to reuse it to implement the `SubstituteAttribute` in the `Ploeh.AutoFixture.NSubstitute` assembly.
from the `Ploeh.AutoFixture.NSubstitute` assembly. (This was initially discussed in #132 and previously submitted in #421)

ploeh: My problem with having this class being a public type is that it introduces a new class (i.e. a new abstraction) that we'll have to maintain for at least two major versions of AutoFixture; that's a couple of years, at minimum. This abstraction is abstruse; while I understand what it does, it seems that its only purpose is to eliminate duplication.

Duplication is worth eliminating [if there's a cost associated with it](http://blog.ploeh.dk/2014/08/07/why-dry). In the case of `RangeAttributeRelay`, `RegularExpressionAttributeRelay`, and `StringLengthAttributeRelay`, these classes are extremely stable, so the cost of the duplication is low; the last time any of these classes were modified were in late 2012 (in a5b89aca2c9bb3ed0082a09ad26c227da9fc8175), and that was even in a part of its code not related to what's being extracted here.

These three classes are abstract enough as it is, but at least it seems fairly clear that they relay a data annotation to a 'proper' request object.

The proposed `AttributeRelay<TAttribute>` may benefit the developers of AutoFixture, but does it benefit the *users*?

olegsych:
I agree that creating a new type can be considered as added cost from the standpoint of maintaining existing code. However, it's absence also adds cost from the standpoint of creating new "auto" attributes. 

I may not be a typical user of AutoFixture, but here is my specific scenario. 
- I want to use AutoFixture and NSubstitute in [one of my projects](https://github.com/olegsych/T4Toolbox/commit/0986731f21f5244096049aa354ac2192bf690d19).
- I feel that using the `Fixture` class directly in my tests makes them more obscure, so I prefer using the `[Theory]` and `[AutoData]` attributes
- This approach doesn't allow me get partial substitutes for non-abstract classes I want to use as SUT and test doubles in my tests. 
- I want to have a `[Substitute]` attribute that I can apply to parameters in my tests to force creation of a partial substitute instead of a regular instance of a non-abstract class.
- I look at the AutoFixture code and find three separate existing examples of attribute-based relays. 
- I could create another copy of the relay code, but it's about the same size as the actual feature code I'd need to write to support partial substitutes. I'd rather not write, test and maintain a separate copy of the boilerplate code.

I included a code example of how I'd want to use the `AttributeRelay<T>` is in the Usage section of this pull request. Without it, I'd have to write about 50 lines of product and 100 lines test code instead of about 1 and 10 lines respectively. These are total lines, including declarations and comments, but I'd still rather write fewer of them. 

Code reuse is important to me in this case because I would prefer using specific attributes instead of global test conventions to keep my test fixtures [minimal](http://xunitpatterns.com/Minimal%20Fixture.html), so I expect I'll need more attributes like `[Substitute]`. If that assumption turns out to be wrong and `AttributeRelay<T>` is used only with these 4 attributes, the cost of maintaining it will be no different than maintaining 4 separate copies of attribute relay logic, because this code hasn't changed since 2012 so there is no reason to expect that it will have to in the future.

moodmosaic:
IIRC, `RangeAttributeRelay`, `RegularExpressionAttributeRelay`, and `StringLengthAttributeRelay`, didn't introduce an `AttributeRelay<T>` for the exact reason that @ploeh [described](https://github.com/AutoFixture/AutoFixture/pull/424#discussion_r35497371):

>The proposed `AttributeRelay<TAttribute>` may benefit the developers of AutoFixture, but does it benefit the *users*?

@olegsych, wouldn't it be possible to define an `AttributeRelay<T>` outside the AutoFixture codebase?

olegsych:
Technically? Yes. However, what I really need is a `SubstituteAttribute` in `Ploeh.AutoFixture.NSubstitute` (#132, #421). It could use the `AttributeRelay<T>` as described in the Usage section of this PR or introduce a new `SubstituteAttributeRelay` class, essentially a copy & paste of the `RegularExpressionAttributeRelay` product and test code. Do you see the latter as the better option?

moodmosaic:
Yes, it can be a lot easier with a hypothetical `SubstituteAttributeRelay` class. @ploeh do you also agree?

ploeh:
An open source *library* like AutoFixture is different than the software most people write. Most people I've met write some sort of software where they control the environment in which it's running: web sites, intranet apps, web services, microservices, etc. In such environments, the cost of making a mistake in the public (C#) API isn't too big, because you can just fix the mistake by changing the API again and redeploy the software. 

This becomes even more pronounced the more towards Continuous Deployment you get.

Thus, in such environments, the cost of duplication tends to outweigh the cost of making an API design mistake.

On the other hand, when it comes to a *library* with an unknown body of users, the cost of making a mistake in the public API of the library is much bigger: you'll have to deal with it for years.

If we were considering removing duplication from an area of the code base where we'd seen a lot of bugs, I'd be more ready to accept that the duplication should be addressed. In this case, though, there's no evidence that the duplication is costly, so I don't think the added complexity introduced in this pull request is warranted.

For this reason, I'm sorry to decline the pull request.

olegsych:
I understand. Are you still open to the introduction of the `SubstituteAttribute`, which is the other half of the original pull request #421?

ploeh:
Where would you put such an attribute?

olegsych:
I would implement the `SubstituteAttribute` in the `Ploeh.AutoFixture.NSubstitute` assembly. Please see #421 for details.

ploeh:
Ah, OK, for a moment I though you had in mind deriving from `CustomizeAttribute` from *AutoFixture.Xunit*, but I now understand that that's not the case.

Go for it :+1:

Issue #413, opened 2015-06-12
The 255th number returned by RandomNumericSequenceGenerator is always 255.


Pvlerick:
I found something else while investigating #407

```C#
[Fact]
public void The255thDrawIsAlways255()
{
    // Fixture setup
    var dummyContext = new DelegatingSpecimenContext();
    var sut = new RandomNumericSequenceGenerator(1, 255);
    // Exercise system
    var result = Enumerable
        .Range(0, 256)
        .Select(_ => sut.Create(typeof(long), dummyContext))
        .Cast<long>()
        .ToList();
    // Verify outcome
    Assert.Equal(255, result[254]);
    // Teardown
}
```

First, notice that `GetNextRandom` method in `RandomNumericSequenceGenerator` uses `this.random.Next((int)this.lower, (int)this.upper)`, the upper bound in Random.Next being exclusive - the first 254 will be drawn between 1 and 254.

Next, in the `EvaluateRange` method, notice that the count check is done on `this.upper - this.lower`.

```C#
private void EvaluateRange()
{
    if (this.count == (this.upper - this.lower))
    {
        this.count = 0;
        this.CreateRange();
    }

    this.count++;
}
```

When comes the 255th draw, the `count` is 254 and upper = 255 and lower = 1, so the `if` is entered, and `count` is reset and the `CreateRange` method is called.

```C#
private void CreateRange()
{
    var remaining = this.limits.Where(x => x > this.upper - 1).ToArray();
    if (remaining.Any() && this.numbers.Any())
    {
        this.lower = this.upper;
        this.upper = remaining.Min() + 1;
    }
    else
    {
        this.lower = limits[0];
        this.upper = limits[1];
    }

    this.numbers.Clear();
}
```

The `remaining` variable will hold only one element: 255 (since it's bigger than `upper` (255) - 1) and `numbers` has 254 elements, so the `if` is entered.
In there, `lower` is set to `upper` (255) and upper is set to the minimum in `remaining` + 1 (256).

Next call to Random.Next will be with 255 and 256, so it'll always be 255.

I think this only happens with the initial lower and upper limits, but I have to check: I will investiage more when time permits. It seems it was discussed in #148 but I didn't have time to read the full discussion yet.

Pvlerick:
After reading #148, I can that this issue was raised in the past but that no decision was taken on the upper limit being inclusive or exclusive.

[...]

Issue #407, opened 2015-06-02
Creating dictionaries with relatively small key type fails ocasionally


dzendras:
There is a problem with generating dictionaries that have a relatively small key type (like System.Byte).
Sometimes unit tests fail because of duplicate keys:

*System.ArgumentException : An item with the same key has already been added.*

So even though AutoFixture creates dictionaries that have only 3 elements (well at least that's what I can observe), it doesn't check whether the generated keys are unique.

In order to reproduce it, please run the following test:

        [Fact]
        public void Duplicate_entries()
        {
            var fixture = new Fixture();
            for (int i = 0; i < 1000; ++i)
            {
                fixture.Create<IDictionary<byte, string>>();
            }
        }

The question is: is it a bug that can be fixed, or is it a corner case where AutoFixture is not supposed to be used?

moodmosaic:
Thank you for reporting this. –  By default, the `Fixture` is configured to create numbers using a [constrained non-deterministic sequence](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/RandomNumericSequenceGenerator.cs).

One way to solve your problem is by making the `Fixture` instance *more* deterministic:

```c#
fixture.Customizations.Add(new NumericSequenceGenerator());
``` 

This changes the `Fixture` so that it generates sequences of consecutive numbers, starting at `1`.

---

It looks weird to me the fact that this behaves erratically *as-is*, since [a `HashSet<T>` is used internally](https://github.com/AutoFixture/AutoFixture/blob/33d0c7850c16d6f3f6c52efa77af908d9fece4ca/Src/AutoFixture/RandomNumericSequenceGenerator.cs#L16)... This *might* be a bug.

dzendras:
Thanks. Using NumericSequenceGenerator worked.

ploeh:
@moodmosaic, according to the documentation of `RandomNumericSequenceGenerator`, it

> Creates a sequence of random, unique, numbers starting at 1.

Note the word *uniqe* here...

This looks strange to me, then, because `fixture.Create<IDictionary<byte, string>>()` should create a dictionary with only three elements, and I would expect the keys to be unique.

Here's a theory, though:

Since a default `Fixture` is used in the above repro, `RepeatCount` is `3`, which means that it's going to create in all 3000 `byte` values. This means that it's going to exhaust the `byte` range of 256 numbers multiple (approximately 12) times.

[...]

PR #426, opened 2015-07-25
attempt at addressing issue #413.  Upper bound is always last generat…

jryanmoseley:
I took a stab at fixing this issue.  

I made the upper bound inclusive for the random generator only for Int32 values.

I also modified one test to include the upper bound in the range check (I hope that is OK).

All of the existing tests pass and there does not look to be any performance hit either.

Thanks.

### DIFF
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -56,13 +56,7 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentException("The limit must be a sequence of two or more integers.");
             }
 
-            for (var i = 0; i < limits.Count() - 1; i++)
-            {
-                if (limits[i + 1] < limits[i])
-                    throw new ArgumentException("The limit must be an increasing sequence.");
-            }
-
-                this.limits = limits;
+            this.limits = limits;
             this.syncRoot = new object();
             this.random = new Random();
             this.numbers = new HashSet<long>();
@@ -201,7 +195,7 @@ namespace Ploeh.AutoFixture
             else
             {
                 this.lower = limits[0];
-                this.upper = limits[1] > Int32.MaxValue
+                this.upper = limits[1] >= Int32.MaxValue
                     ? limits[1]
                     : limits[1] + 1;
             }
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -156,7 +156,7 @@ namespace Ploeh.AutoFixture
                 long result;
                 do
                 {
-                    if (this.lower >= int.MinValue && 
+                    if (this.lower >= int.MinValue &&
                         this.upper <= int.MaxValue)
                     {
                         result = this.random.Next((int)this.lower, (int)this.upper);
@@ -195,7 +195,9 @@ namespace Ploeh.AutoFixture
             else
             {
                 this.lower = limits[0];
-                this.upper = limits[1];
+                this.upper = limits[1] > Int32.MaxValue
+                    ? limits[1]
+                    : limits[1] + 1;
             }
 
             this.numbers.Clear();
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -56,7 +56,13 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentException("The limit must be a sequence of two or more integers.");
             }
 
-            this.limits = limits;
+            for (var i = 0; i < limits.Count() - 1; i++)
+            {
+                if (limits[i + 1] < limits[i])
+                    throw new ArgumentException("The limit must be an increasing sequence.");
+            }
+
+                this.limits = limits;
             this.syncRoot = new object();
             this.random = new Random();
             this.numbers = new HashSet<long>();
--- a/Src/AutoFixture/RandomNumericSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomNumericSequenceGenerator.cs
@@ -195,12 +195,24 @@ namespace Ploeh.AutoFixture
             else
             {
                 this.lower = limits[0];
-                this.upper = limits[1] >= Int32.MaxValue
-                    ? limits[1]
-                    : limits[1] + 1;
+                this.upper = this.GetUpperRangeFromLimits();
             }
 
             this.numbers.Clear();
+        }
+
+        /// <summary>
+        /// Returns upper limit + 1 when expecting to use upper as max value in Random.Next(Int32,Int32).
+        /// This ensures that the upper limit is included in the possible values returned by Random.Next(Int32,Int32)
+        /// 
+        /// When not expecting to use Random.Next(Int32,Int32).  It returns the original upper limit.
+        /// </summary>
+        /// <returns></returns>
+        private long GetUpperRangeFromLimits()
+        {
+            return limits[1] >= Int32.MaxValue
+                    ? limits[1]
+                    : limits[1] + 1;
         }
 
         private long GetNextInt64InRange()
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -948,8 +948,8 @@ namespace Ploeh.AutoFixtureUnitTest
             var result = sut.Create<DoublePropertyHolder<int, long>>();
             // Verify outcome
             Assert.True(
-                (result.Property1 >= lower && result.Property1 < upper) &&
-                (result.Property2 >= lower && result.Property2 < upper)
+                (result.Property1 >= lower && result.Property1 <= upper) &&
+                (result.Property2 >= lower && result.Property2 <= upper)
                 );
         }
 
@@ -5630,8 +5630,8 @@ namespace Ploeh.AutoFixtureUnitTest
         [InlineData(10)]
         public void CreateComplexArrayTypeReturnsArrayReflectingCorrectRepeatCount(int repeatCount)
         {
-            var sut = new Fixture{ RepeatCount = repeatCount };
-            
+            var sut = new Fixture { RepeatCount = repeatCount };
+
             var actual = sut.Create<int[,][]>();
 
             Assert.Equal(repeatCount, actual.GetLength(0));
@@ -5721,7 +5721,7 @@ namespace Ploeh.AutoFixtureUnitTest
             int count = 0;
             while (result.MoveNext())
             {
-                count ++;
+                count++;
                 Assert.True(count <= expectedCount);
             }
             Assert.Equal(expectedCount, count);
--- a/Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
+++ b/Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
@@ -108,9 +108,9 @@ namespace Ploeh.AutoFixtureUnitTest
             // Fixture setup
             var unexpectedLimits = new long[]
             { 
-                (byte)10,
-                (Int16)50,
-                (Int32)20,
+                (byte)byte.MaxValue,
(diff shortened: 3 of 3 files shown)


Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -56,7 +56,13 @@ public RandomNumericSequenceGenerator(params long[] limits)
                 throw new ArgumentException("The limit must be a sequence of two or more integers.");
             }
 
-            this.limits = limits;
+            for (var i = 0; i < limits.Count() - 1; i++)
+            {
+                if (limits[i + 1] < limits[i])
+                    throw new ArgumentException("The limit must be an increasing sequence.");
```
moodmosaic: Isn't this already addressed by #409? /cc @Pvlerick

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -195,7 +201,9 @@ private void CreateRange()
             else
             {
                 this.lower = limits[0];
-                this.upper = limits[1];
+                this.upper = limits[1] > Int32.MaxValue
+                    ? limits[1]
+                    : limits[1] + 1;
```
moodmosaic: I'm probably missing something obvious, but I can't seem to understand what's the purpose of this change. :confused:

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
```diff
@@ -108,9 +121,9 @@ public void LimitsDoesNotMatchParamsArray()
             // Fixture setup
             var unexpectedLimits = new long[]
             { 
-                Byte.MaxValue, 
-                Int16.MaxValue, 
-                Int32.MaxValue
+                (byte)10,
+                (Int16)50,
+                (Int32)20,
```
moodmosaic: Do the `unexpectedLimits` had to change?

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
```diff
@@ -65,6 +65,19 @@ public void InitializeWithSingleLimitThrows(long[] limits)
...
+        [InlineData(new long[] { 3, 4, 5, 1 })]
+        [InlineData(new long[] { 0, -1, 2 })]
+        [InlineData(new long[] { -8, -3, -4, 0 })]
+        [InlineData(new long[] { 1, 1, 2, 1 })]
+        public void InitializeWithNonIncreasingLimitThrows(long[] limits)
+        {
+            Assert.Throws<ArgumentException>(() =>
+                new RandomNumericSequenceGenerator(limits));
+        }
+
```
moodmosaic: Isn't this already being addressed by #409? /cc @Pvlerick

moodmosaic:
Thank you for your contribution! :smiley: I left a few comments but, other than that, I think this is looking good.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
jryanmoseley: It looks like it is LOL!  I didn't notice because I forked the main branch and missed the pull request addressing it.  Apologies.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
jryanmoseley: This change is what will allow the upper limit to be included in the call to Random.Next for Int32 ranges.  This is my fix for Issue #413.

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
jryanmoseley: If issue #409 is fixed then this change will be required.  In the test the Int32 is added to the constructor before the Int16.  This would violate the increasing sequence requirement.  Unless my fix to issue #409 was incorrect.

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
jryanmoseley: It looks like it is LOL! I didn't notice because I forked the main branch and missed the pull request addressing it. Apologies.

jryanmoseley:
Thanks for the feedback.  Should I remove my changes that were fixed in #409?

Src/AutoFixture/RandomNumericSequenceGenerator.cs
ploeh: I'm not sure I follow this either. How does including `Int32.MaxValue` prevent the 255th number from being 255?

jryanmoseley: The limits[1] + 1 will prevent the 255th number from being 255.  

Random.Next(1,255) will never return 255 because it excludes the upper limit.  My change will make the call Random.Next(1,256) so that 255 could be returned.  However, Random.Next takes (Int32,Int32) which means it will break if the upper bound is larger than Int32.MaxValue.  So I only do this if the limit[1] is less than Int32.MaxValue.  Now that I think about it, the check should be >= Int32.MaxValue. 

Hope this explains it better.

ploeh: @jryanmoseley, thank you, it does :+1: 

This is one of the rare cases where I'd think that a code comment might be in place.

One of the facts that mislead me was that it wasn't clear to me why `Int32.MaxValue` was particularly interesting, since a request could also be for an `Int64` value. However, the fact that `Random.Next(int, int)` operates on `Int32` instead of `Int64` explains that part.

ploeh:
It seems that I inadvertently allowed #409 to be dropped on the floor for a while. Let's see what happens with it first, now that I've attempted to reactivate it.

Based on what I've seen, though, I expect to accept it, in which case you should consider updating this PR.

Sorry about the lapse on my part!

jryanmoseley:
Hey Guys,

Thanks for all of the feedback.  I have removed the change I made duplicating #409 in anticipation of that being accepted.

Thanks.

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
moodmosaic: OK, thank you for explaining this :+1:

 #409 takes care of that unit test (so we don't have to, in this PR).

I think, if we can rollback this change, we might avoid unnecessary merge conflicts :smiley:

Pvlerick:
Hello,

The whole issue should have been gone with the fix for #407 that I am working on, but since I'm being a bit slow, it cannot hurt that someone else fixes it :+1:

Src/AutoFixtureUnitTest/RandomNumericSequenceGeneratorTest.cs
jryanmoseley: My last commit rolls this change back.

Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -948,8 +948,8 @@ public void InjectCustomUpperLimitWillCauseSutToReturnNumbersInLimit()
             var result = sut.Create<DoublePropertyHolder<int, long>>();
             // Verify outcome
             Assert.True(
-                (result.Property1 >= lower && result.Property1 < upper) &&
-                (result.Property2 >= lower && result.Property2 < upper)
+                (result.Property1 >= lower && result.Property1 <= upper) &&
+                (result.Property2 >= lower && result.Property2 <= upper)
```
ploeh: How does this change ensure that #413 is resolved?

Where did we decide to change the specification of the algorithm to let the upper limit be inclusive?

jryanmoseley: Maybe I have the wrong understanding of the bug.  The bug states that if I set a lower and upper limit of 1 and 10 respectively.  The final returned value will always be 10.  I believe the test above would never assert result.Property1 or 2 as 10 and so the test would not fail.  When I implemented my fix this was the only test that failed because 10 may possibly be the first value returned.  It is my understanding that the test was not complete.  If the upper limit is not inclusive then my fix is not needed.

Please let me know if that makes sense.

moodmosaic: I agree with @ploeh here. The algorithm's specification must remain the same, otherwise we might end up introducing a breaking change.

What needs to addressed by this Pull Request is that the upper limit's value shouldn't be the same all the time.

jryanmoseley: Based on my understanding of the issue.  If the issue is addressed, this test case will begin to fail.  The test case never expects the generator to return the upper limit.  If this is by design then I am not sure how to address the issue without changing the test case.  Again, maybe I am have misunderstood the issue.

Pvlerick: The previous implementation had the upper limit inclusive, as far as I understand. If you gave `[1; 255; 300]` as the limits, you would draw the first 254 from [1;254], __the 255th would be 255__ and then the last 256-300 woud be taken from [256;300] - this is due to [line 193 in RandomNumericSequenceGenerator.cs](https://github.com/jryanmoseley/AutoFixture/blob/master/Src/AutoFixture/RandomNumericSequenceGenerator.cs#L193). The issue here is only about the 255th draw in the previous example.

I believe the test wasn't correct to start with (at least from what I understand from the code); otherwise the code didn't follow the specs (but there is no mention in that in the class' comments, maybe this should be added too).

Pvlerick: Ok, let me try to step back a little and recap my understanding.

The limits given to `RandomNumericSequenceGenerator` are "steps" in which define sets from which random number should be drawn. The last limit is the maximum number you can ever draw before the generator is "reset" (and in the following examples, since we start at 1 it'll be the number of draws that can be made).

Examples:
* limits = [1;42] -> first 42 numbers should be unique and between 1 and 42, after that the generator is reset;
* limits = [1;42;80] -> first 42 numbers should be unique and between 1 and 42, then the next 38 numbers will be unique and between 43 and 80; after that the generator is reset;
* limits = [1;42;80;1500] -> first 42 numbers should be unique and between 1 and 42, then the next 38 numbers will be unique and between 43 and 80; then the next 1420 numbers will be unique and between 81 and 1500, after that the generator is reset;
* etc

In all cases, I can draw as many uniques as the given upper limit before the generator is reset (42, 80, 1500).

The current implementation differs in that in all the examples above, the first 41 draws will be between 1 and 41, and the 42th will be 42, then as above.

The test here is correct against the implementation (it checks that it is between 1 and 41) but not against what I believe to be the correct implementation (between 1 and 42). So if #413 if fixed, this test will fail if it stays as is, hence the required modification.

jryanmoseley: My understanding is in line with @Pvlerick.  Thanks.

ploeh: OK, thank you, @Pvlerick and @jryanmoseley, at this point I think you understand this better than I do :flushed:

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.31.3](https://www.nuget.org/packages/AutoFixture/3.31.3).

jryanmoseley:
Thanks a lot!  I really appreciate you all working through this with me and not just rejecting my efforts.

Src/AutoFixtureUnitTest/FixtureTest.cs
moodmosaic: Impressive analysis @Pvlerick and @jryanmoseley :+1: 

I must admit that even I, who created this class in the first place, couldn't answer all the questions without doing analysis :open_mouth:

---

As a side note, the fact that such an analysis is required makes me think of [this](http://fsharpforfunandprofit.com/posts/is-your-language-unreasonable/) article, and [this](http://blog.ploeh.dk/2015/04/13/less-is-more-language-features/), and [this](http://blog.ploeh.dk/2015/04/15/c-will-eventually-get-all-f-features-right/), and others as well :relieved:

Issue #420, opened 2015-06-20
Undefined behavior of FrozenAttribute combined with construction attributes like [Greedy], [Modest], [Favor?], etc.


olegsych:
This topic was discussed in the previously closed thread #132. I'd like to reopen it in hope of discovering a solution. Here is the gist of it.

The CLR does not guarantee the order of attributes in assembly metadata. Because of that, combining the `FrozenAttribute` with construction attributes, like `GreedyAttribute` and `FavorListsAttribute`, produces undefined results. With the current implementation, one of the following two tests will always fail.

``` C#
using Ploeh.AutoFixture.Xunit2;
using Xunit;

public class Subject
{
    public Subject() { }
    public Subject(int greedy) { this.IsGreedy = true; }
    public bool IsGreedy { get; private set; }
    public override string ToString() { return "Subject " + this.GetHashCode().ToString(); }
}

public class UnitTest1
{
    [Theory, AutoData]
    public void FrozenAndGreedy([Frozen, Greedy] Subject p1, Subject p2)
    {
        Assert.Same(p1, p2);
        Assert.True(p1.IsGreedy);
    }

    [Theory, AutoData]
    public void GreedyAndFrozen([Greedy, Frozen] Subject p1, Subject p2)
    {
        Assert.Same(p1, p2);
        Assert.True(p1.IsGreedy);
    }
}
```

This behavior is a trap for those without deep understanding of both CLR and AutoFixture because "freezing" and "construction" are conceptually independent operations and combining two attributes like `[Frozen]` and `[Greedy]` feels natural. However, with the current implementation, one should _never_ apply these attributes to the same parameter at the risk of previously working tests starting to produce unpredictable results. I suspect that people will only learn about this problem after using AutoFixture for a while and accumulating a fair number of tests. By that time fixing the problem in the user code may be costly and they may have pulled out a few hairs trying to understand why the same test fails in different environments or starts failing in the same environment without obvious reasons.

olegsych:
The undefined ordering of attribute metadata by the CLR probably cannot be fixed by AutoFixture :smile:. But I'm hoping that the current implementation of the `FreezingCustomization` used by the `[Frozen]` attribute implementations could be changed to solve this problem. 

[...]

PR #430, opened 2015-07-31
Substitute attribute, 

olegsych:
The proposed `SubstituteAttribute` enables changing the default fixture behavior and forcing creation of concrete type substitutes for individual test parameters. The implementation follows the pattern established by the `AutoFixture` for data annotation attributes, such as `[Range]`, `[RegularExpression]` and `[StringLength]`.

The idea was previously discussed in #132 and submitted in #421. The proposed implementation does _not_ include the previously rejected `AttributeRelay<T>` refactoring #424.

Here are samples that illustrate this attribute's use:

```C#
[Theory, AutoDataWithAutoSubstituteCustomization]
public void ConcreteTypesAreSubstitutedWhenParameterIsMarkedWithSubstituteAttribute(
    [Substitute]ConcreteType t, object obj)
{
    t.VirtualMethod(obj);
    t.Received().VirtualMethod(obj);
}

[Theory, AutoDataWithAutoSubstituteCustomization]
public void SubstituteAttributeAppliesToSingleParameterAndNotItsEntireType(
    [Substitute]ConcreteType t1, ConcreteType t2, object obj)
{
    t2.VirtualMethod(obj);
    Assert.Throws<NotASubstituteException>(() => t2.Received().VirtualMethod(obj));
}
```

Applying the `SubstituteAttribute` to a parameter of a concrete type fails the test, such as the one in example below, _before_ running it.

```C#
[Theory, AutoDataWithAutoSubstituteCustomization]
public void AttemptingToSubstituteSealedTypeFailsTestBeforeItStarts(
    [Substitute]SealedType t, object obj)
{
}
```

Applying the `SubstituteAttribute` in a method with a regular `AutoDataAttribute` results in an obscure failure within the test itself because the attribute has no behavior and requires the `AutoSubstituteCustomization` to be applied to the test's `Fixture`. Below is a sample that illustrates the problem. A separate pull request #418 was started earlier to enable addressing it.

```C#
[Theory, AutoData]
public void SubstituteAttributeShouldWorkWithPlainAutoDataAttribute(
    [Substitute]ConcreteType t, object obj)
{
    t.VirtualMethod(obj);
    // The following line throws NotASubstituteExeption, which is confusing because the
    // Substitute attribute is clearly used.
    t.Received().VirtualMethod(obj);
}
```

### Changes in AutoFixture.NSubstitute
- Implement `SubstituteAttribute`.
- Implement `SubstituteRequest` used to request a substitute of a given type.
- Implement `SubstituteRequestHandler` to handle substitute requests.
- Implement `SubstituteAttributeRelay` to convert requests for parameters, properties and fields marked with the `SubstituteAttribute` to substitute requests of their respective types.
- Implement `SubstituteRelay` to convert requests for abstract types and interfaces to requests for substitutes.
- Change existing `AutoNSubstituteCustomization` and `AutoConfiguredNSubstituteCustomization` to use the new components instead of the existing `NSubstituteBuilder`.


Src/AutoNSubstitute.sln
```diff
@@ -11,6 +11,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoNSubstituteUnitTest", "
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoNSubstitute", "AutoNSubstitute\AutoNSubstitute.csproj", "{189C9474-B406-4356-A367-FBC882E9BA4B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.xUnit.net", "AutoFixture.xUnit.net\AutoFixture.xUnit.net.csproj", "{266B9F0E-28E5-47CA-9816-ACB6737E7E4D}"
```
ploeh: Why is *AutoFixture.Xunit* being added to this solution?

olegsych: To implement [scenario tests](https://github.com/olegsych/AutoFixture/commit/f925a115dd711811f9891ad79655fe3c3d554633).

ploeh: I see.

These integration tests do serve a purpose (at least because they validate that what you want do to actually works).

What if someone comes by and wants to add a similar set of integration tests for *AutoFixture.NUnit2?* Or what if we get a new unit testing framework glue library? Should we add references to NUnit etc. to this solution as well?

It reasonable to have these integration tests, although we've traditionally haven't had them; IIRC, I don't think we have any integration tests that demonstrate the combination of xUnit.net and Moq.

Still, if we want to keep these integration tests, we'll need to create a new solution for them, as I don't want the coupling between xUnit.net and NSubstitute in the *AutoNSubstitute* solution.

Src/AutoNSubstituteUnitTest/Scenario.cs
```diff
@@ -0,0 +1,97 @@
...
+        [Theory, AutoDataWithAutoSubstituteCustomization]
+        public void ConcreteTypesAreNotSubstitutedByDefault(ConcreteType t, object obj)
+        {
+            t.VirtualMethod(obj);
+            Assert.Throws<NotASubstituteException>(() => t.Received().VirtualMethod(obj));
+        }
+
+        [Theory, AutoDataWithAutoSubstituteCustomization]
+        public void ConcreteTypesAreSubstitutedWhenParameterIsMarkedWithSubstituteAttribute(
+            [Substitute]ConcreteType t, object obj)
```
ploeh: Now that I see what this looks like, I wonder why, instead of an attribute, we don't introduce a `Substitute<T>` class?

As we've previously discussed, the order of attributes isn't guaranteed by the .NET framework, so when `[Substitute]` is an attribute, you can't combine it with e.g. `[Frozen]`.

IMO, it wouldn't hurt readability, and it's the same number of characters you'd have to type:

```C#
[Theory, AutoDataWithAutoSubstituteCustomization]
public void ConcreteTypesAreSubstitutedWhenParameterIsMarkedWithSubstituteAttribute(
    Substitute<ConcreteType> t, object obj)
```

That would solve the combination problem.

Src/AutoNSubstitute.sln
olegsych: I've added scenario tests to illustrate the usage of the new API because this approach was [already common in AutoFixture](https://github.com/AutoFixture/AutoFixture/search?utf8=%E2%9C%93&q=scenario). I don't have a strong preference on this matter. It's your call. 

To answer your questions, I don't find scenario tests to be particularly useful after the design has settled down and would be happy to remove them before the PR is merged. I like the idea of moving scenario tests to a separate solution less. It would reduce their value by separating them from the APIs they are meant to help you learn. Moving scenario tests to a separate project within the NSubstitute solution may be a better option.

ploeh: I don't want the dependency on *AutoFixture.Xunit* in this solution.

If you don't think the scenario tests provide much value anyway, I'd be inclined to simply delete them.

Src/AutoNSubstituteUnitTest/Scenario.cs
olegsych: As someone who chose NSubstitute for its terse, low-ceremony API, I'd find having to use `Substitute<T>` awkward. The [. Operator](https://msdn.microsoft.com/en-us/library/6zhxzbds(v=vs.71).aspx) cannot be overloaded to hide the wrapper type, so the `Substitute<T>` would bleed into my test code. It would start looking Moq-ish, only more confusing, because `Received()` has to be called for the actual substitute instance of type `T` instead of the instance of `Substitute<T>`.

As we discussed in #420, the `[Frozen]` attribute cannot be combined with `[Substitute]` or other creation attributes today because `FreezingCustomization` uses a fixture customization instead of fixture behavior. 

While the type-based approach would be sufficient to support the `[Frozen]`/Substitute scenario today, the attribute-based approach is more open to future extension. The `SubstituteAttribute` attribute can be parameterized or combined other attributes to support creation of [partial substitutes](http://nsubstitute.github.io/help/partial-subs/), [unconfigured substitutes](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/AutoConfiguredNSubstituteCustomization.cs) and, once #420 is addressed,  _frozen unconfigured partial substitutes_ for those who need them.

Src/AutoNSubstitute.sln
olegsych: Sounds good. I can force-push my branch without commit [f925a11](https://github.com/olegsych/AutoFixture/commit/f925a115dd711811f9891ad79655fe3c3d554633) so that it won't show up in the history of the main repository. Would you like me to do it now?

ploeh: Sounds good :+1:

Src/AutoNSubstituteUnitTest/Scenario.cs
ploeh: It's a fair point about the reason for choosing NSubstitute in the first place; I accept that as a motivating factor for designing this particular feature as an attribute, although I consider the overall design of the NSubstitute API fundamentally flawed.

Anyway, with this overall design question resolved, I will now do a more detailed review of this pull request.

Src/AutoNSubstitute.sln
ploeh: Perhaps I wrote too soon when I wrote that [I'd proceed with a more detailed review](https://github.com/AutoFixture/AutoFixture/pull/430#discussion_r36284251). I'll first wait for the change that removes the scenario tests.

Will you please add a notification here when it's done? GitHub doesn't send me a notification when pull request branches are updated, but I *do* get a notification for every new comment.

olegsych: @ploeh, I have force-pushed the `substitute-attribute` branch without the [scenario tests](https://github.com/olegsych/AutoFixture/commit/f925a115dd711811f9891ad79655fe3c3d554633).

Src/AutoNSubstitute/AutoNSubstitute.csproj
```diff
@@ -26,7 +26,7 @@
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet>AutoNSubstitute.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>bin\Release\Ploeh.AutoFixture.AutoNSubstitute.XML</DocumentationFile>
-    <RunCodeAnalysis>true</RunCodeAnalysis>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
```
ploeh: Does this turn off static code analysis? If so, why?

ploeh: Oh, I see now that this is in Release mode. This may actually be an original error... (I'll have to investigate)

ploeh: It's not even in Release mode, but in Debug mode, I see now.

AFAICT, this was added in 627c6d85b9482bbd9424561b06e491985de52980, and went unnoticed for 2½ years :flushed:

Good catch :+1:

Src/AutoNSubstitute/AutoNSubstitute.csproj
```diff
@@ -64,6 +64,7 @@
   <ItemGroup>
     <Compile Include="AutoConfiguredNSubstituteCustomization.cs" />
     <Compile Include="AutoNSubstituteCustomization.cs" />
+    <Compile Include="AutoSubstituteCustomization.cs" />
```
ploeh: Could we think of another name for this class?

The name makes sense, but it only differs by a single character from `AutoNSubstituteCustomization`, so I'm a bit concerned that newcomers would mistype the name, or not understand the difference.

That would give a bad first impression of the overall library, and lead to many questions here or on Stack Overflow.

Src/AutoNSubstitute/SubstituteBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            this.substituteConstructor = substituteConstructor;
+        }
+
+        /// <summary>
+        /// Returns an <see cref="ISpecimenBuilder"/> responsible for creating a substitute instance based on a target
+        /// <see cref="Type"/> type.
+        /// </summary>
+        public ISpecimenBuilder SubstituteConstructor
```
ploeh: Would `SubstituteFactory` or `SubstituteBuilder` be better names?

Given that the type in question is an `ISpecimenBuilder`, there's no guarantee that it'll invoke a constructor...

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        [Fact]
+        public void CreateReturnsSpecimenResolvedByContextFromSubstituteRequest()
+        {
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<ParameterInfo>();
+            request.ParameterType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(Arg.Any<Type>(), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
```
ploeh: What if you changed this line to the following?

```C#
request
    .GetCustomAttributes(typeof(SubstituteAttribute), Arg.Any<bool>())
    .Returns(new[] { new SubstituteAttribute() });
```

Wouldn't the `CreateGetsInheritedAttributesFromProviderBecauseAttributesAreInheritedByDefault` test be redundant, then?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+        {
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = new object();
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            object specimen = sut.Create(request, context);
+            // Verify outcome
+            var noSpecimen = Assert.IsType<NoSpecimen>(specimen);
+            Assert.Same(request, noSpecimen.Request);
```
ploeh: `NoSpecimen` has structural equality, so it'd be easier to read the assertion if it was expressed like this:

```C#
var expected = new NoSpecimen(request);
Assert.Equal(expected, specimen);
```

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+        public void CreateReturnsSpecimenResolvedByContextFromSubstituteRequest()
+        {
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<ParameterInfo>();
+            request.ParameterType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(Arg.Any<Type>(), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
+            var expectedSpecimen = new object();
+            var context = Substitute.For<ISpecimenContext>();
+            context.Resolve(Arg.Any<SubstituteRequest>()).Returns(expectedSpecimen);
```
ploeh: With `Any<SubstituteRequest>` couldn't we pass the test by always attempting to resolve `SustituteRequest(typeof(object))`?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<ParameterInfo>();
+            request.ParameterType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(Arg.Any<Type>(), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            sut.Create(request, context);
+            // Verify outcome
+            context.Received().Resolve(Arg.Is<SubstituteRequest>(r => r.TargetType == request.ParameterType));
```
ploeh: What does this prove? Do we expect the call to `Resolve` to have a side effect?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<PropertyInfo>();
+            request.PropertyType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(Arg.Any<Type>(), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            sut.Create(request, context);
+            // Verify outcome
+            context.Received().Resolve(Arg.Is<SubstituteRequest>(r => r.TargetType == request.PropertyType));
```
ploeh: What does this prove? Do we expect the call to `Resolve` to have a side effect?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<FieldInfo>();
+            request.FieldType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(Arg.Any<Type>(), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            sut.Create(request, context);
+            // Verify outcome
+            context.Received().Resolve(Arg.Is<SubstituteRequest>(r => r.TargetType == request.FieldType));
```
ploeh: What does this prove? Do we expect the call to `Resolve` to have a side effect?

Src/AutoNSubstituteUnitTest/SubstituteAttributeTest.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Linq;
+using Xunit;
+using Xunit.Extensions;
+
+namespace Ploeh.AutoFixture.AutoNSubstitute.UnitTest
+{
+    public class SubstituteAttributeTest
```
ploeh: I'd be OK with deleting this test class. It tests no behaviour, and the type and its annotations are more concisely expressed by the class itself.

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
```diff
@@ -0,0 +1,78 @@
...
+            var constructedSubstute = Substitute.For(new[] { targetType}, new object[0]);
+            var constructor = Substitute.For<ISpecimenBuilder>();
+            constructor.Create(Arg.Any<object>(), Arg.Any<ISpecimenContext>()).Returns(constructedSubstute);
+            var sut = new SubstituteBuilder(constructor);
+            var request = new SubstituteRequest(targetType);
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            object result = sut.Create(request, context);
+            // Verify outcome
+            constructor.Received().Create(targetType, context);
```
ploeh: What does this prove? Do we expect the call to `Create` to have a side effect?

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
```diff
@@ -0,0 +1,102 @@
...
+        [InlineData(typeof(AbstractType))]
+        [InlineData(typeof(IInterface))]
+        public void CreateReturnsObjectResolvedFromContextWhenRequestedTypeIsAbstractOrInterface(Type requestedType)
+        {
+            // Fixture setup
+            var sut = new SubstituteRelay();
+            object request = requestedType;
+            object substitute = Substitute.For(new Type[] { requestedType }, new object[0]);
+            var context = Substitute.For<ISpecimenContext>();
+            context.Resolve(Arg.Any<object>()).Returns(substitute);
```
ploeh: Wouldn't that allow us to change the implementation to resolve `new object()` instead?

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        [Fact]
+        public void CreateThrowsInvalidOperationExceptionWhenResolvedObjectIsNotSubstituteAssumingInvalidConfiguration()
+        {
+            // Fixture setup
+            var sut = new SubstituteRelay();
+            var request = typeof(IComparable);
+            var notASubstitute = new object();
+            var context = Substitute.For<ISpecimenContext>();
+            context.Resolve(Arg.Any<object>()).Returns(notASubstitute);
```
ploeh: Wouldn't that allow us to change the implementation to resolve `new object()` instead?

Src/AutoNSubstituteUnitTest/SubstituteRequestTest.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        [Fact]
+        public void TargetTypeReturnsValueSpecifiedInConstructor()
+        {
+            // Fixture setup
+            var expectedType = typeof(int);
+            // Excercise system
+            var sut = new SubstituteRequest(expectedType);
+            // Verify outcome
+            Assert.Same(expectedType, sut.TargetType);
```
ploeh: Couldn't a Devil's Advocate pass the test by hard-coding `typeof(int)` in the implementation?

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
```diff
@@ -0,0 +1,78 @@
...
+            // Fixture setup
+            var constructor = Substitute.For<ISpecimenBuilder>();
+            var sut = new SubstituteBuilder(constructor);
+            var request = typeof(IComparable);
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            object result = sut.Create(request, context);
+            // Verify outcome
+            var noSpecimen = Assert.IsType<NoSpecimen>(result);
+            Assert.Same(request, noSpecimen.Request);
```
ploeh: `NoSpecimen` has structural equality, so it'd be easier to read the assertion if it was expressed like this:

```C#
var expected = new NoSpecimen(request);
Assert.Equal(expected, specimen);
```

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            // Fixture setup
+            var sut = new SubstituteRelay();
+            object request = "beer";
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            object result = sut.Create(request, context);
+            // Verify outcome
+            var noSpecimen = Assert.IsType<NoSpecimen>(result);
+            Assert.Same(request, noSpecimen.Request);
```
ploeh: `NoSpecimen` has structural equality, so it'd be easier to read the assertion if it was expressed like this:

```C#
var expected = new NoSpecimen(request);
Assert.Equal(expected, specimen);
```

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            // Fixture setup
+            var sut = new SubstituteRelay();
+            object request = typeof(string);
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            object result = sut.Create(request, context);
+            // Verify outcome
+            var noSpecimen = Assert.IsType<NoSpecimen>(result);
+            Assert.Same(request, noSpecimen.Request);
```
ploeh: `NoSpecimen` has structural equality, so it'd be easier to read the assertion if it was expressed like this:

```C#
var expected = new NoSpecimen(request);
Assert.Equal(expected, specimen);
```

ploeh:
This pull request is looking good :+1: 

I had a few comments, but nothing major.

Src/AutoNSubstitute/AutoNSubstitute.csproj
olegsych: What if the new logic is implemented in the existing `AutoNSubstituteCustomization` class?

Src/AutoNSubstitute/SubstituteBuilder.cs
olegsych: `SubstituteBuilder` property in a `SubstituteBuilder` class would be confusing. If the class name stays the same, I'd say `SubstituteFactory` would be a better property name. On the other hand, the class could also have a more descriptive name, like `SubstituteRequestHandler`, which would make `SubstituteBuilder` an adequate property name.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: Ok, I'll update this and other tests in this class to take advantage of structural equality in `NoSpecimen`.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: This test proves that `SubstituteAttributeRelay` creates a `SubstituteRequest` with correct `TargetType` given a _parameter_ with `SubstituteAttribute`.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: This test proves that `SubstituteAttributeRelay` creates a `SubstituteRequest` with correct `TargetType` given a _property_ with `SubstituteAttribute`.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: This test proves that `SubstituteAttributeRelay` creates a `SubstituteRequest` with correct `TargetType` given a _field_ with `SubstituteAttribute`.

Src/AutoNSubstituteUnitTest/SubstituteAttributeTest.cs
olegsych: The key logic tested here is the applicability of the attribute to the same code elements (properties, parameters and fields) that are also supported by the `SubstituteAttributeRelay`. Unfortunately the test name `AttributeCanBeAppliedToParametersOfTestMethods` is not very accurate after collapsing several facts into a single theory. Assuming it is corrected, if you still feel that its not valuable enough, I can remove the class. Would you like me to do that?

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
olegsych: The line above proves that `SubstituteBuilder.Create` extracts `TargetType` from the given `SubstituteRequest`, passes it along with the given `ISpecimenContext` to the `SubstituteConstructor.Create`. The next line below proves that `SubstituteBuilder.Create` returns the specimen constructed by `SubstituteConstructor.Create`. Together they prove that `SubstituteBuilder.Create` successfully returns a substitute specimen given a valid `SubstituteRequest`.

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
olegsych: Good point. Do you prefer expressing this test condition as `Arg.Is<Type>(t => t == requestedType)` in the fixture setup code or as the `context.Received().Resolve(requestedType)` assertion in the outcome verification?

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
olegsych: With this concern addressed by the test above, do you think that `Arg.Any<Type>()` would be sufficient here?

Src/AutoNSubstituteUnitTest/SubstituteRequestTest.cs
olegsych: Replaced with `IInterface` - a type that cannot be hardcoded in `SubstituteRequest`

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
olegsych: Resolved.

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
olegsych: Resolved.

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
olegsych: Resolved.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: If the definition of _redundant_ does not take into account documentation, then  `CreateGetsInheritedAttributesFromProviderBecauseAttributesAreInheritedByDefault` to become redundant if this line was changed to also expect a specific `bool` value:
```C#
request
    .GetCustomAttributes(typeof(SubstituteAttribute), true)
    .Returns(new[] { new SubstituteAttribute() });
```

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: That would make the three tests below fail.

Src/AutoNSubstitute/SubstituteBuilder.cs
ploeh: > `SubstituteBuilder` property in a `SubstituteBuilder` class would be confusing.

Good point. I completely agree. I hadn't really had my eye on the slightly bigger picture, there :flushed: 

Although I see your point about giving the class itself a more descriptive name, I don't mind `SubstituteBuilder` as a class name, because this is, after all, the 'top' class responsible for building substitutes. In that case, I prefer `SubstituteFactory` as the property name.

Src/AutoNSubstitute/AutoNSubstitute.csproj
ploeh: > What if the new logic is implemented in the existing `AutoNSubstituteCustomization` and `NSubstituteBuilder` classes instead of defining new classes with similar names?

That might work. What's most important is that we don't introduce breaking changes, or paint ourselves into a corner that we can't get out of.

When it comes to `AutoNSubstituteCustomization`, would it change the overall behaviour, or would it just add a new feature?

Usually, the easiest way to answer that question is to try it and see if anything breaks. If the tests are still green, it's a strong indication that it isn't a breaking change :smile: 

(There may be a few tests that break because they are too tightly coupled to a particular implementation detail, and it *may* be okay to modify such tests, but it's always something we'd need to carefully review. The point with this side remark is that if you try to make the change, and there's one or two failing tests, it may still be OK.)

What if we include the code from  `AutoSubstituteCustomization` into `AutoNSubstituteCustomization` and no tests fail? Could we think of situations where someone else would want that new feature to go away? I can't think of any such situations, but wanted to raise the question for others to ponder...

My gut feeling about this is that it'd probably be fine to merge `AutoSubstituteCustomization` into `AutoNSubstituteCustomization`.

> Do we want all new classes to include `N` in their name?

I think that if a class explicitly uses the NSubstitute API, it should be named accordingly (with the `N`); otherwise, not.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: Using the *Devil's Advocate* technique, I can rewrite the `Create` method to to this, and all tests still pass:

```C#
public object Create(object request, ISpecimenContext context)
{
    if (context == null)
    {
        throw new ArgumentNullException("context");
    }

    var customAttributeProvider = request as ICustomAttributeProvider;
    if (customAttributeProvider == null)
    {
    }

    var attribute = customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), true)
            .OfType<SubstituteAttribute>().FirstOrDefault();
    if (attribute == null)
    {
    }

    object substituteRequest = CreateSubstituteRequest(customAttributeProvider, attribute);
    context.Resolve(substituteRequest);
    return context.Resolve(new SubstituteRequest(typeof(object)));
}
```

Notice the two last lines of code.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: See [my other comment](https://github.com/AutoFixture/AutoFixture/pull/430/files#r36579077).

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: See [my other comment](https://github.com/AutoFixture/AutoFixture/pull/430/files#r36579077).

Src/AutoNSubstituteUnitTest/SubstituteAttributeTest.cs
ploeh: You decide :smile: 

I don't think it adds any value, but I don't think it hurts either.

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
ploeh: Using the *Devli's Advocate* technique, I can change the implementation of `Create` to the following, and all tests still pass:

```C#
public object Create(object request, ISpecimenContext context)
{
    var substituteRequest = request as SubstituteRequest;
    if (substituteRequest == null)
    {
    }

    this.substituteConstructor.Create(substituteRequest.TargetType, context);
    return this.substituteConstructor.Create(null, null);
}
```

Notice the two last lines of code.

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
ploeh: Why would you need `Resolve(Arg.Is<Type>(t => t == requestedType), ...)`? Wouldn't it be enough to write `Resolve(requestedType, ...)`?

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
ploeh: Please disregard this comment of mine.

I now realise that this test exercises a failure scenario, in which case it's the right thing to do to state that, *regardless* of the input to `Resolve`, when it returns `notASubstitute`, the behaviours should be as follows...

I wasn't paying enough attention here. The fault is all mine; I'm sorry.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: As [demonstrated elsewhere, that's not the case](https://github.com/AutoFixture/AutoFixture/pull/430/files#r36579077).

olegsych: Changed fixture setup to close this gap.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: `CreateReturnsSpecimenResolvedByContextFromSubstituteRequest` now ensures that specimen resolved from correct `substituteRequest` is returned. With that gap closed, this test is still needed to verify that `ParameterType` is supported.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: `CreateReturnsSpecimenResolvedByContextFromSubstituteRequest` now ensures that specimen resolved from correct `substituteRequest` is returned. With that gap closed, this test is still needed to verify that `PropertyType` is supported.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: `CreateReturnsSpecimenResolvedByContextFromSubstituteRequest` now ensures that specimen resolved from correct `substituteRequest` is returned. With that gap closed, this test is still needed to verify that `FieldType` is supported.

Src/AutoNSubstituteUnitTest/SubstituteBuilderTest.cs
olegsych: Thanks for the explanation. I have made changes to close the gap in test logic.

Src/AutoNSubstituteUnitTest/SubstituteRelayTest.cs
olegsych: I'm sorry I got the first option wrong. I've made the changes to close the test gap.

Src/AutoNSubstituteUnitTest/SubstituteAttributeTest.cs
olegsych: In that case I'll leave it with corrected test name. :smile:

Src/AutoNSubstitute/SubstituteBuilder.cs
olegsych: Renamed the property to `SubstituteFactory`.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: I can still do this, and all tests pass:

```C#
public object Create(object request, ISpecimenContext context)
{
    if (context == null)
    {
        throw new ArgumentNullException("context");
    }

    var customAttributeProvider = request as ICustomAttributeProvider;
    if (customAttributeProvider == null)
    {
    }

    customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), true);
    var attribute = customAttributeProvider.GetCustomAttributes(null, false)
            .OfType<SubstituteAttribute>().FirstOrDefault();
    if (attribute == null)
    {
    }

    object substituteRequest = CreateSubstituteRequest(customAttributeProvider, attribute);
    return context.Resolve(substituteRequest);
}
```

Src/AutoNSubstitute/AutoNSubstitute.csproj
olegsych: `AutoNSubstituteCustomization` has a constructor that takes an `ISpecimenBuilder` instance which is stored and returned by the `Builder` property. What is their intended use in consumer code?

With the exception of the `SubstituteAttributeRelay`, none of the new classes call NSubstitute directly. Would you leave their names without the `N`?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: This is already covered by the `CreateGetsInheritedAttributesFromProviderBecauseAttributesAreInheritedByDefault` test, but I'll update this test to catch this as well.

Src/AutoNSubstitute/AutoNSubstitute.csproj
ploeh: > `AutoNSubstituteCustomization` has a constructor that takes an `ISpecimenBuilder` instance which is stored and returned by the `Builder` property. What is their intended use in consumer code?

The `Builder` property is an [Inspection Property](http://blog.ploeh.dk/2013/04/04/structural-inspection). It has no explicit intended use in consumer code, but (like all other Inspection Properties) exist as a courtesy to any caller that may want to inspect it.

When it comes to the constructor overload taking an `ISpecimenBuilder`, I can't remember the reason. Perhaps the Git log can answer that question...

> With the exception of the `SubstituteAttributeRelay`, none of the new classes call NSubstitute directly. Would you leave their names without the N?

That would make sense according to the above heuristic, but we don't need to make it a hard rule. The only thing that concerns me here is if we have two classes with *almost* identical names, because users are bound to mistype those from time to time.

ploeh:
I can see that you added some commits 10 days ago (Github doesn't send me a notification when that happens), but you added some comments recently. Should I review again, or should I wait?

olegsych:
I pushed several commits to take care of the minor issues pointed out earlier. You can review them if you'd like. However, the design issues listed in the pull request description still need to be resolved, so it may make sense to wait with another deep review until then.

Src/AutoNSubstitute/AutoNSubstitute.csproj
olegsych: Would you consider changing the `Builder` property to return an instance of different type, or changing where the `Builder` is added to the `Fixture` by `Customize` a breaking change?

ploeh: Formally, it would be a breaking change.

Sometimes we do accept that a few unit tests break (until we update them), if the breaks happen because the tests are a bit too overspecified. It's a judgement call every time, though.

I'd tend to consider changing the type of the `Builder` property a breaking change. Regarding the other change, it'd depend on whether or not it changes the behaviour of the system.

What do you have in mind?

ploeh:
In that case, I'll wait a bit, if you don't mind :smile:

Src/AutoNSubstitute/AutoNSubstitute.csproj
olegsych: To keep the number of classes small, I'd prefer to change behavior of the existing `AutoNSubstituteCustomization` and the `AutoConfiguredNSubstituteCustomization` to support the new `SubstituteAttribute`. Their `Builder` inspection properties mayhave to be modified to return an `ISpecimenBuilder` instance of a different type - `SubstituteBuilder` or `SubstituteAttributeRelay` instead of the `NSubstituteBuilder`. The type of property itself would remain an `ISpecimenBuilder`. 

If you consider that change to be _backward compatible_, I would recommend 
- merging the new `AutoSubstituteCustomization` logic into the two existing `NSubstituteCustomization` classes and eliminating the new customization class before merging this PR;
- replacing the existing `NSubstituteBuilder` implementation with that of the new `SubstituteBuilder` and eliminating the new builder class before merging this PR;
- renaming the new `SubstituteXxx` classes to follow the existing `NSubstituteXxx` naming convention.

Alternatively, if returning a different `ISpecimenBuilder` from the `Builder` properties of the existing customization classes _breaks_ backward compatibility, I would recommend 
- renaming the new `AutoSubstituteCustomization` to simply `SubstituteCustomization` to emphasize its connection to the other new classes
- leaving existing and new classes separate.

In other words, it's largely a tradeoff between the number of public classes and backward compatibility with implications on long term evolution of the API. How would you like to proceed?

ploeh: > Their `Builder` inspection properties mayhave to be modified to return an `ISpecimenBuilder` instance of a different type - `SubstituteBuilder` or `SubstituteAttributeRelay` instead of the `NSubstituteBuilder`. The type of property itself would remain an `ISpecimenBuilder`.

> If you consider that change to be *backward compatible*

I do. :smile: If clients use this property and attempt to downcast to the concrete type, they are on their own.

> merging the new `AutoSubstituteCustomization` logic into the two existing `NSubstituteCustomization` classes and eliminating the new customization class before merging this PR;

Sounds promising :+1: 

> replacing the existing `NSubstituteBuilder` implementation with that of the new `SubstituteBuilder` and eliminating the new builder class before merging this PR;

Do they do the same thing? They don't look identical to me, but I'm not so deep in the bowels of this part of the code base as you are. If it's possible to implement one using the other, it sounds like a beneficial refactoring.

Nevertheless, unless it's *necessary* to do as part of this PR, I would prefer that it's not made a part of this PR. [Keep pull requests as small and focused as possible](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests).

> renaming the new SubstituteXxx classes to follow the existing NSubstituteXxx naming convention.

*That* would be a breaking change, so unfortunately, we can't do that.

What we *can* do is to copy the old class to a new class with the desired name, and then deprecate the old class, but that's not going to make the number of classes smaller. Again, if we do that, we should do that in a separate pull request.

olegsych: I will update the existing customization classes and remove the `AutoSubstituteCustomization` class then.

You've raised a good point - the existing `NSubstituteBuilder` and the new `SubstituteBuilder` are indeed different; the first handles abstract `Type` requests while the second handles the new `SubstituteRequest`. I agree that replacing the existing type with the new would be a breaking change. The new `SubstituteRelay`, however, is a much closer match because it also handles abstract `Type` requests, so if you prefer to deprecate the existing `NSubstituteBuilder` this class could be a replacement.

I think we still need to agree on the naming convention for the new classes. I don't understand why renaming the new `SubstituteXxx` classes to include the `N` prefix would be a breaking change. Aside from the new `SubstituteBuilder` there are no naming collisions with the existing classes and making the new class use a different, more descriptive name, like `NSubstituteRequestHandler` would solve this particular problem.

ploeh: > I don't understand why renaming the new SubstituteXxx classes to include the N prefix would be a breaking change

Perhaps I miswrote somewhere. Renaming the *new* types in this pull request would *not* be a breaking change.

What I meant to state was that renaming *existing* types would be a breaking change. Sorry about the confusion.

olegsych: Thanks for the clarification. Here are the remaining changes I'd like to make.
- move code from the new `AutoSubstituteCustomization` class into the two existing  `AutoNSubstituteCustomization` and `AutoConfiguredNSubstituteCustomization` classes;
- eliminate the new `AutoSubistituteCustomization` class;
- rename the new `SubstituteAttribute` to `NSubstituteAttribute`;
- rename the new `SubstituteAttributeRelay` to `NSubsituteAttributeRelay`;
- rename the new `SubstituteBuilder` to `NSubstituteRequestHandler`;
- rename the new `SubstituteRelay` to `NSubstituteAbstractTypeRelay`;
- rename the new `SubstituteRequest` to `NSubstituteRequest`;
- `[Obsolete]` the existing `NSubstituteBuilder` which is being replaced by the new `NSubstituteAbstractTypeRelay`; 

Do you agree?

ploeh: > rename the new `SubstituteAttribute` to `NSubstituteAttribute`

I don't particularly mind, but this type is, IIRC, the original use case that this pull request is all about. Originally, you wanted to be able to write `[Substitute]foo`. With the proposed change, you'd have to write `[NSubstitute]foo`. Do you think that's better?

> rename the new `SubstituteAttributeRelay` to `NSubsituteAttributeRelay`

If you rename `SubstituteAttribute` to `NSubstituteAttribute`, then this makes sense as well. If you decide not to rename `SubstituteAttribute` to `NSubstituteAttribute`, then I also think the name should remain `SubstituteAttributeRelay`, because it relays that particular attribute.

> rename the new `SubstituteRequest` to `NSubstituteRequest`

I don't know... Are other objects requesting a *substitute*, or an *nsubstitute*?

> `[Obsolete]` the existing `NSubstituteBuilder` which is being replaced by the new `NSubstituteAbstractTypeRelay`

Is it necessary to do this as part of this pull request?

olegsych: I don't have a strong preference on either of these points, simply trying to get to an agreement. Since you don't seem to like the idea of consistently applying the `N` prefix consistently, how about this: 

- move code from the new `AutoSubstituteCustomization` class into the two existing  `AutoNSubstituteCustomization` and `AutoConfiguredNSubstituteCustomization` classes;
- eliminate the new `AutoSubistituteCustomization` class;
- rename the new `SubstituteBuilder` to `SubstituteRequestHandler` to avoid confusion with the existing `NSubstituteBuilder`

ploeh: Sounds good :+1:

olegsych:
I have implemented the changes we agreed on (remove `AutoSubstituteCustomization`, rename `SubstituteBuilder` to `SubstituteRequestHandler` and changed the existing `AutoNSubstituteCustomization` and `AutoConfiguredNSubstituteCustomization` to use the new components instead of the existing `NSubstituteBuilder`). Please take a look.

ploeh:
Thank you; I'll take a look as soon as possible, but did you do something to whitespace in certain files? If you look in the diff view here on GitHub, it looks like some classes have been replaced wholesale.

ploeh:
Since this pull request also touches on `AutoConfiguredNSubstituteCustomization`, I'm hereby inviting its original author, @mrinaldi to take a look as well.

A review is completely voluntary, but may be prudent to 'protect your investment' :wink:

olegsych:
Oops, my bad :fearful: Line ending changes have been removed.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,164 @@
...
+        public void CreateGetsInheritedAttributesFromProviderBecauseAttributesAreInheritedByDefault()
+        {
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<ICustomAttributeProvider>();
+            var context = Substitute.For<ISpecimenContext>();
+            // Exercise system
+            sut.Create(request, context);
+            // Verify outcome
+            request.Received().GetCustomAttributes(typeof(SubstituteAttribute), true);
```
ploeh: Does `GetCustomAttributes` have a side-effect that we care about?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        [Fact]
+        public void CreateReturnsSpecimenResolvedByContextFromSubstituteRequest()
+        {
+            // Fixture setup
+            var sut = new SubstituteAttributeRelay();
+            var request = Substitute.For<ParameterInfo>();
+            request.ParameterType.Returns(typeof(IInterface));
+            request.GetCustomAttributes(typeof(SubstituteAttribute), Arg.Any<bool>()).Returns(new[] { new SubstituteAttribute() });
```
ploeh: Don't we care about the value of the `inherit` argument?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: I don't understand the question. Can you elaborate?

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: Yes, we do. It is verified by the `CreateGetsInheritedAttributesFromProviderBecauseAttributesAreInheritedByDefault` test.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: The above test verifies `request` as though the return value doesn't matter. [If a test doesn't care about the return value of a function, then the reason for verifying it must be because calling the method has a side-effect that the test cares about.](http://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries)

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: Not really. Using the `Devil's Advocate` technique, I can sabotage the implementation like this:

```C#
public object Create(object request, ISpecimenContext context)
{
    if (context == null)
    {
        throw new ArgumentNullException("context");
    }

    var customAttributeProvider = request as ICustomAttributeProvider;
    if (customAttributeProvider == null)
    {
    }

    customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), true);
    var attribute = customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), false)
            .OfType<SubstituteAttribute>().FirstOrDefault();
    if (attribute == null)
    {
    }

    object substituteRequest = CreateSubstituteRequest(customAttributeProvider, attribute);
    return context.Resolve(substituteRequest);
}
```

Not that I would, but the `Devil's Advocate` is a review technique to demonstrate that the test may not be testing what one thinks it's testing.

With that above incorrect implementation, all tests still pass.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: Thanks for the explanation. I have updated the tests to address this comment.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
olegsych: Thanks for the explanation. I've updated the test to close this gap.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: Given this test, I can still sabotage the implementation like this:

```C#
public object Create(object request, ISpecimenContext context)
{
    if (context == null)
    {
        throw new ArgumentNullException("context");
    }

    var customAttributeProvider = request as ICustomAttributeProvider;
    if (customAttributeProvider == null)
    {
    }

    var attribute = customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), true)
            .OfType<SubstituteAttribute>().FirstOrDefault();
    if (attribute == null)
    {
    }

    object substituteRequest = CreateSubstituteRequest(customAttributeProvider, attribute);
    if(customAttributeProvider is ParameterInfo)
        return context.Resolve(substituteRequest);
    context.Resolve(substituteRequest);
    return context.Resolve(new SubstituteRequest(typeof(object)));
}
```

Notice the last four lines of code, which are obviously malicious and incorrect.

Src/AutoNSubstituteUnitTest/SubstituteAttributeRelayTest.cs
ploeh: Given this test, I can still sabotage the implementation like this:

```C#
public object Create(object request, ISpecimenContext context)
{
    if (context == null)
    {
        throw new ArgumentNullException("context");
    }

    var customAttributeProvider = request as ICustomAttributeProvider;
    if (customAttributeProvider == null)
    {
    }

    var attribute = customAttributeProvider.GetCustomAttributes(typeof(SubstituteAttribute), true)
            .OfType<SubstituteAttribute>().FirstOrDefault();
    if (attribute == null)
    {
    }

    object substituteRequest = CreateSubstituteRequest(customAttributeProvider, attribute);
    if(customAttributeProvider is ParameterInfo)
        return context.Resolve(substituteRequest);
    context.Resolve(substituteRequest);
    return context.Resolve(new SubstituteRequest(typeof(object)));
}
```

Notice the last four lines of code, which are obviously malicious and incorrect.

olegsych:
Additional comments in `SubstituteAttributeRelayTest` addressed.

Src/AutoNSubstituteUnitTest/SubstituteRequestHandlerTest.cs
```diff
@@ -0,0 +1,78 @@
...
+            Type targetType = typeof(IComparable);
+            var constructedSubstute = Substitute.For(new[] { targetType}, new object[0]);
+            var constructor = Substitute.For<ISpecimenBuilder>();
+            constructor.Create(targetType, context).Returns(constructedSubstute);
+            var sut = new SubstituteRequestHandler(constructor);
+            var request = new SubstituteRequest(targetType);
+            // Exercise system
+            object result = sut.Create(request, context);
+            // Verify outcome
+            constructor.Received().Create(targetType, context);
```
ploeh: This may be a stupid question, since I don't know NSubstitute well, but isn't this expression redundant, given how `constructor` is configured above?

mrinaldi: I agree with @ploeh on this one.

mrinaldi:
Despite that one comment, this PR looks good to me. :+1:

Src/AutoNSubstituteUnitTest/SubstituteRequestHandlerTest.cs
olegsych: Removed.

ploeh:
Thank you for your contribution, and your great patience! It's now live as [AutoFixture.AutoNSubstitute 3.32.0](https://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.32.0).

jrnail23:
@olegsych I've been trying to use [Substitute] with the NUnit2 AutoData stuff,
but I haven't been able to get it to freeze my substitute specimen (i.e., `[Substitute][Frozen] MyType specimen` ).

Any ideas there?

ploeh:
@jrnail23, you may want to look at #420

jrnail23:
yes, @ploeh, thanks for that.

sergeyshushlyapin:
@jrnail23 
> I haven't been able to get it to freeze my substitute specimen (i.e., [Substitute][Frozen] MyType specimen

This problem is described in #637.

PR #439, opened 2015-09-15
Fixed exception when adding duplicated entries to a dictionary

Pvlerick:
Adressing #407.

Guard clause added by calling the `ContainsKey` prior to calling `Add`.


moodmosaic:
This is looking good to me, and the build-release.sh scripts runs fine on my machine :+1:

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.34.1](https://www.nuget.org/packages/AutoFixture/3.34.1).

PR #444, opened 2015-09-28
FrozenAttribute with matching strategy for xUnit2

ecampidoglio:
Ports the `FrozenAttribute` with matching strategy to the **xUnit2** Glue Library.

For example you can freeze a parameter for any interfaces implemented by its type:

```csharp
[Theory, AutoData]
public void FreezeByImplementedInterface(
    [Frozen(Matching.ImplementedInterfaces)]Foo p1,
    IFoo p2)
{
    Assert.Equal(p1, p2);
}
```

or freeze a parameter for any properties with the same name (case-insensitive) and a compatible type: 

```csharp
[Theory, AutoData]
public void FreezeByMatchingPropertyTypeAndName(
    [Frozen(Matching.PropertyName)]string bar,
    Foo p)
{
    Assert.Equal(bar, p.Bar);
}
```

### DIFF
--- a/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
@@ -163,11 +163,27 @@ namespace Ploeh.AutoFixture.Xunit2.UnitTest
         }
 
         [Theory, AutoData]
+        public void FreezeFirstParameterByExactTypeShouldNotAssignSameInstanceToSecondParameterOfDifferentType(
+            [Frozen(Matching.ExactType)]ConcreteType p1,
+            object p2)
+        {
+            Assert.NotEqual(p1, p2);
+        }
+
+        [Theory, AutoData]
         public void FreezeFirstParameterByDirectBaseTypeShouldAssignSameInstanceToSecondParameter(
             [Frozen(Matching.DirectBaseType)]ConcreteType p1,
             AbstractType p2)
         {
             Assert.Equal(p1, p2);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByDirectBaseTypeShouldNotAssignSameInstanceToSecondParameterOfIndirectBaseType(
+            [Frozen(Matching.DirectBaseType)]ConcreteType p1,
+            object p2)
+        {
+            Assert.NotEqual(p1, p2);
         }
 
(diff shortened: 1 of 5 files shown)


Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
```diff
@@ -136,10 +138,92 @@ public void FavorArraysCausesArrayConstructorToBeInjectedWithFrozenItems([Frozen
...
+            Assert.Equal(property, p2.Property);
+        }
+
+        [Theory, AutoData]
+        public void FreezeFirstParameterByMemberWithSameNameShouldAssignSameInstanceToMatchingField(
+            [Frozen(Matching.MemberName)]string field,
+            FieldHolder<object> p2)
+        {
+            Assert.Equal(field, p2.Field);
+        }
```
ploeh: I like having these tests here :+1: 

These tests verify the *behaviour* of the `[Frozen]` attribute, which is what we actually care about.

All these tests verify that values are matched, but not all values should be matched. Shouldn't there be some tests that verify that values are not matched as well?

Src/AutoFixture.xUnit.net2/FrozenAttribute.cs
```diff
@@ -11,22 +13,63 @@ namespace Ploeh.AutoFixture.Xunit2
     [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
     public sealed class FrozenAttribute : CustomizeAttribute
     {
+        private readonly Matching by;
+        private IRequestSpecification matcher;
```
ploeh: Could we please get rid of this [Temporary Field](http://blog.ploeh.dk/2015/09/18/temporary-field-code-smell)?

Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
ecampidoglio: Added in 2de100e.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit2 3.35.1](https://www.nuget.org/packages/AutoFixture.Xunit2/3.35.1) (there was an error publishing to NuGet, so 3.35.0 was never published, and I had to bump up the patch version to get it out).

PR #445, opened 2015-09-28
FrozenAttribute with matching strategy for NUnit2

ecampidoglio:
Ports the `FrozenAttribute` with matching strategy to the **NUnit2** Glue Library.

For example you can freeze a parameter for any interfaces implemented by its type:

```csharp
[Test, AutoData]
public void FreezeByImplementedInterface(
    [Frozen(Matching.ImplementedInterfaces)]Foo p1,
    IFoo p2)
{
    Assert.AreEqual(p1, p2);
}
```

or freeze a parameter for any properties with the same name (case-insensitive) and a compatible type: 

```csharp
[Test, AutoData]
public void FreezeByMatchingPropertyTypeAndName(
    [Frozen(Matching.PropertyName)]string bar,
    Foo p)
{
    Assert.AreEqual(bar, p.Bar);
}
```


ecampidoglio:
Resolved the same issues raised during the code review of #444.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit2 3.36.0](https://www.nuget.org/packages/AutoFixture.NUnit2/3.36.0).

PR #448, opened 2015-10-06
Resolves issue #447

### LINKED ISSUE #447: Latest version of Frozen attribute no longer works with mocked instances
Somehow the latest version of the Frozen attribute (in 3.36.0 release), which features the new policy enumeration, no longer seems to work when freezing mocked interfaces.

For instance, for a `MyTheory([Frozen] Mock<IRepository> repository, SomeSut sut)`, with `SomeSut` constructor accepting `IRepository` in parameter, a new mocked instance will be created.

ecampidoglio:
Fixes #447, where the `FreezeOnMatchCustomization` conflicted with the `AutoMoqCustomization` – and all other auto-mocking customizations for that matter – causing the frozen mock not to be returned for requests of the implemented interface type.

### DIFF
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -99,7 +99,8 @@ namespace Ploeh.AutoFixture
 
         private void FreezeTypeForMatchingRequests(IFixture fixture)
         {
-            fixture.Customizations.Add(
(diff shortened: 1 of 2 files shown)


ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.36.1](https://www.nuget.org/packages/AutoFixture/3.36.1).

PR #449, opened 2015-10-07
Fixes the shotgun freeze issue in xUnit

ecampidoglio:
I just discovered an issue in the new policy-based `[Frozen]` attribute: freezing by using a matching strategy erroneously freezes the _exact type of the test parameter_ as well.

For example, this test passes:

```csharp
[Theory, AutoData]
public void FreezeByPropertyOfDifferentNameAndTypeShouldNotAssignSameInstanceToParameter(
    [Frozen(Matching.PropertyName)]string p1,
    PropertyHolder<object> p2)
    {
        Assert.NotEqual(p1, p2.Property);
    }
```

but this one doesn't:

```csharp
[Theory, AutoData]
public void FreezeByPropertyOfDifferentNameButSameTypeShouldNotAssignSameInstanceToParameter(
    [Frozen(Matching.PropertyName)]string p1,
    PropertyHolder<string> p2)
    {
        Assert.NotEqual(p1, p2.Property);
    }
```

This PR eliminates this unintended side effect by making sure that the specified matching strategy is treated as an exclusive condition, like it was originally intended.

### DIFF
--- a/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
@@ -212,6 +212,14 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         [Theory, AutoData]
+        public void FreezeFirstParameterByInterfaceShouldNotAssignSameInstanceToSecondParameterOfSameType(
+            [Frozen(Matching.ImplementedInterfaces)]NoopInterfaceImplementer p1,
+            NoopInterfaceImplementer p2)
+        {
+            Assert.NotEqual(p1, p2);
+        }
+
+        [Theory, AutoData]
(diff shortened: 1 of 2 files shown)


ploeh:
FYI, I added two more scenarios in d8130acc016b6d164a14125c3673b28fec892fe8

ecampidoglio:
Thanks. I'll issue two new PRs for xUnit2 and NUnit2.

ploeh:
Will you also add the two scenario combos I added in d8130acc016b6d164a14125c3673b28fec892fe8?

ecampidoglio:
Done :+1:

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.36.8](https://www.nuget.org/packages/AutoFixture.Xunit/3.36.8).

(I had quite some problems with the CodeBetter CI server, and I don't even think *all* the 3.36.8 packages were published, but this one was.)

PR #451, opened 2015-10-07
Fixes the shotgun freeze issue in NUnit2

ecampidoglio:
I just discovered an issue in the new policy-based `[Frozen]` attribute for **NUnit2**: freezing by using a matching strategy erroneously freezes the _exact type of the test parameter_ as well.

For example, this test passes:

```csharp
[Test, AutoData]
public void FreezeByPropertyOfDifferentNameAndTypeShouldNotAssignSameInstanceToParameter(
    [Frozen(Matching.PropertyName)]string p1,
    PropertyHolder<object> p2)
    {
        Assert.AreNotEqual(p1, p2.Property);
    }
```

but this one doesn't:

```csharp
[Test, AutoData]
public void FreezeByPropertyOfDifferentNameButSameTypeShouldNotAssignSameInstanceToParameter(
    [Frozen(Matching.PropertyName)]string p1,
    PropertyHolder<string> p2)
    {
        Assert.AreNotEqual(p1, p2.Property);
    }
```

This PR eliminates this unintended side effect by making sure that the specified matching strategy is treated as an exclusive condition, like it was originally intended.

### DIFF
--- a/Src/AutoFixture.NUnit2.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.NUnit2.UnitTest/Scenario.cs
@@ -190,6 +190,14 @@ namespace Ploeh.AutoFixture.NUnit2.UnitTest
         }
 
         [Test, AutoData]
+        public void FreezeFirstParameterByParameterWithDifferentNameShouldNotAssignSameInstanceToSecondParameterOfSameType(
+            [Frozen(Matching.ParameterName)]string p1,
+            SingleParameterType<string> p2)
+        {
+            Assert.AreNotEqual(p1, p2.Parameter);
+        }
+
+        [Test, AutoData]
         public void FreezeFirstParameterByPropertyWithSameNameShouldAssignSameInstanceToSecondParameter(
             [Frozen(Matching.PropertyName)]string property,
             PropertyHolder<object> p2)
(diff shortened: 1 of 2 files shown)


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit2 3.36.8](https://www.nuget.org/packages/AutoFixture.NUnit2/3.36.8).

(I had quite some problems with the CodeBetter CI server, and I don't even think *all* the 3.36.8 packages were published, but this one was.)

jrnail23:
I just discovered this myself (at least I think it's the same issue).
Here's my failing test case:

```csharp

        [Test, AutoData]
        public void FreezeCompositeService(
            [Frozen(Matching.ParameterName)] NoopInterfaceImplementer p1,
            [Frozen(Matching.ParameterName)] NoopInterfaceImplementer p2,
            CompositeService holder)
        {
            Assert.AreSame(p1, holder.P1);
            Assert.AreSame(p2, holder.P2);

            // this part fails
            Assert.AreNotSame(p1, p2);
        }

    public class CompositeService
    {
        public CompositeService(IInterface p1, IInterface p2)
        {
            P1 = p1;
            P2 = p2;
        }

        public IInterface P1 { get; private set; }

        public IInterface P2 { get; private set; }
    }
```

jrnail23:
yep, looks like my test case passes now on the latest version

PR #433, opened 2015-08-06
Adapting build scripts for Visual Studio 2015

Pvlerick:
A start for fixing #432.

I do get the error reporter in the issue thread when building, so these will have to be fixed.

PR #455, opened 2015-10-19
Added more detailed exception messages

ploeh:
when an interface or abstract class is requested.

It seems to be a FAQ that people don't understand what to do when running
into the previous error, so hopefully this will help such users.

Since this modification touches on some of @adamchester's work that I don't feel that I fully understand, I'd appreciate a review :smile:

### DIFF
--- a/Src/AutoFixture/Kernel/TerminatingWithPathSpecimenBuilder.cs
+++ b/Src/AutoFixture/Kernel/TerminatingWithPathSpecimenBuilder.cs
@@ -89,8 +89,7 @@ namespace Ploeh.AutoFixture.Kernel
                 {
                     throw new ObjectCreationException(string.Format(
                         CultureInfo.CurrentCulture,
-                        "AutoFixture was unable to create an instance from {0}, most likely because it has no " +
-                        "public constructor, is an abstract or non-public type.{1}{1}Request path:{1}{2}",
+                        BuildCoreMessageTemplate(request),
                         request,
                         Environment.NewLine,
                         BuildRequestPathText(this.SpecimenRequests)));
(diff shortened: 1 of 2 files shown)


ploeh:
AFAICT, the AppVeyor build belongs to @ecampidoglio; could you please tell me what I did wrong? :cry:

adamchester:
It looks like the AppVeyor build was enabled by accident.

The error just means @ecampidoglio hasn't specified which msbuild project to use. If you actually did intend to create an AppVeyor build, you could do it [like this](https://github.com/AutoFixture/AutoFixture/pull/433#issuecomment-128278066).

adamchester:
@ploeh I assume you removed the `requestPath` from the message when it's caused by being abstract or an interface on purpose, is that correct?

ploeh:
@adamchester, good that you asked! This is the reason reviews are valuable!

You assumption is incorrect. I *did* intend to include `requestPath` in those messages as well, but then I got caught up in formatting the messages, and plain forgot about it :flushed: 

Pull request is now updated.

ecampidoglio:
@adamchester @ploeh Yes, I created the AppVeyor build but never intended it to run, just yet. Sorry about that. :flushed:

Since we've been having [some problems](https://github.com/AutoFixture/AutoFixture/compare/381977fabdf5397458ba203f936b3b520eef0ed8...356ea3a691e487b94b85e9f5f16731cd850fcd19) with [our TC build on CodeBetter](http://teamcity.codebetter.com/project.html?projectId=AutoFixture&tab=projectOverview), I started to investigate alternative build servers, like [Travis CI](https://travis-ci.org) and [AppVeyor](https://ci.appveyor.com).

AppVeyor seems to be the way to go for .NET, so I created a project to kick the tires. Apparently, it started running even if I didn't specify a build process. :disappointed: 

I'll try to get the existing AutoFixture build process up and running _as-is_ in AppVeyor and report back.

ecampidoglio:
@adamchester Your build configuration looks good. :+1:  I'll create an organization account for AutoFixture on AppVeyor and get it up and running there.

ecampidoglio:
Unfortunately there's no way to remove the [status check](https://developer.github.com/v3/repos/statuses/) from the PR. Sorry for the inconvenience. :disappointed:

ploeh:
We'll have to ignore the status check for this pull request, and proceed as though it isn't there.

Are there any other review comments?

adamchester:
Looks good. :ship:

ploeh:
I get the impression that there are no further comments.

You clearly aren't used to receiving pull requests from me, but another time, feel free to merge if you think it looks good :smile:

adamchester:
You are correct, I didn't even think about that :smile:

Issue #457, opened 2015-10-30
InlineAutoData no way to specify custom AutoDataAttribute


stevenao:
Although the ctor has a AutoDataAttribute parameter, there is no way to use it in practice. It is confusing the real intention of it. If it is in fact not the intention to take in an AutoDataAttribute. It is better to make that ctor protected for the case that you can inherit from it.

ploeh:
There *is* a way to use it in practice, although I admit that it can be difficult to find information about that. That constructor overload exists to enable you to create a derived class that customizes the behaviour of the attribute. You can see an [example in AtomEventStore](https://github.com/GreanTech/AtomEventStore/blob/master/AtomEventStore.UnitTests/InlineAutoAtomDataAttribute.cs), including [usage examples](https://github.com/GreanTech/AtomEventStore/blob/master/AtomEventStore.UnitTests/LifoEventsTests.cs#L144-L192).

Since this information is too difficult to find, we ought to add an example to the documentation.

ploeh:
Added *enhancement* and *jump in* labels to indicate that we should enhance the documentation. When that's done, I think we can close this issue, unless @stevenao objects.

stevenao:
I was confused at the first place when I was trying to use it as an attribute, but later I figured that I should make a class to inherit from it. I agree that a little more documentation will be nice. That can be just a little comment in the source code <summary> section. Thanks @ploeh!

StefanoChiodino:
I've found the constructor used in the example given by @ploeh is obsolete and for whatever reason didn't work. I just extracted the `Func<IFixture>`  and used that instead of the and worked just fine.

PR #460, opened 2015-11-05
Document inline auto data attribute constructor overload

ploeh:
Addresses #457.

### DIFF
--- a/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
@@ -82,6 +82,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
         {
             Assert.Equal(1337, x);
+            // y can vary, so we can't express any meaningful assertion for it.
             Assert.Equal(42, z);
         }
 
--- a/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
@@ -73,6 +73,42 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
             Assert.NotEqual("bar", s3);
         }
 
+        // This test and its associated types is used to document one of the
+        // InlineAutoDataAttribute constructor overloads.
+        [Theory]
+        [MyCustomInlineAutoData(1337)]
+        [MyCustomInlineAutoData(1337, 7)]
+        [MyCustomInlineAutoData(1337, 7, 42)]
+        public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
+        {
+            Assert.Equal(1337, x);
+            Assert.Equal(42, z);
+        }
+
+        private class MyCustomInlineAutoDataAttribute : InlineAutoDataAttribute
+        {
+            public MyCustomInlineAutoDataAttribute(params object[] values) :
+                base(new MyCustomAutoDataAttribute(), values)
+            {
+            }
+        }
+
+        private class MyCustomAutoDataAttribute : AutoDataAttribute
+        {
+            public MyCustomAutoDataAttribute() :
+                base(new Fixture().Customize(new TheAnswer()))
+            {
+            }
+
+            private class TheAnswer : ICustomization
+            {
+                public void Customize(IFixture fixture)
+                {
+                    fixture.Inject(42);
+                }
+            }
+        }
+
         [Theory, AutoData]
         public void FreezeFirstParameter([Frozen]Guid g1, Guid g2)
         {
--- a/Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
+++ b/Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
@@ -51,6 +51,7 @@ namespace Ploeh.AutoFixture.Xunit
         /// public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
         /// {
         ///     Assert.Equal(1337, x);
+        ///     // y can vary, so we can't express any meaningful assertion for it.
         ///     Assert.Equal(42, z);
         /// }
         /// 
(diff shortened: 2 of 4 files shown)


Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
```diff
@@ -72,6 +72,42 @@ public void InlineAutoDataSuppliesDataSpecimensOnlyForNonProvidedValues(string s
             Assert.NotEqual("bar", s3);
         }
 
+        // This test and its associated types is used to document one of the
+        // InlineAutoDataAttribute constructor overloads.
+        [Theory]
+        [MyCustomInlineAutoData(1337)]
+        [MyCustomInlineAutoData(1337, 7)]
```
adamchester: It took me a while to understand why `y` was missing from the test assertions. I wonder if a named constant that indicates we don't care about the value is easier to read?

moodmosaic: :+1: I think it'd be more clear for the test reader if we can do what @adamchester writes.

Src/AutoFixture.xUnit.net2/InlineAutoDataAttribute.cs
```diff
@@ -34,6 +34,55 @@ public InlineAutoDataAttribute(params object[] values)
...
+        /// MyCustomAutoDataAttribute, and transitively in
+        /// MyCustomInlineAutoDataAttribute. A parameterized test demonstrates
+        /// how it can be used.
+        /// <code>
+        /// [Theory]
+        /// [MyCustomInlineAutoData(1337)]
+        /// [MyCustomInlineAutoData(1337, 7)]
+        /// [MyCustomInlineAutoData(1337, 7, 42)]
+        /// public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
+        /// {
```
moodmosaic: See @adamchester's [comment](https://github.com/AutoFixture/AutoFixture/pull/460/files#r43981442).

Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
```diff
@@ -30,6 +30,55 @@ public InlineAutoDataAttribute(params object[] values)
...
+        /// MyCustomAutoDataAttribute, and transitively in
+        /// MyCustomInlineAutoDataAttribute. A parameterized test demonstrates
+        /// how it can be used.
+        /// <code>
+        /// [Theory]
+        /// [MyCustomInlineAutoData(1337)]
+        /// [MyCustomInlineAutoData(1337, 7)]
+        /// [MyCustomInlineAutoData(1337, 7, 42)]
+        /// public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
+        /// {
```
moodmosaic: See @adamchester's [comment](https://github.com/AutoFixture/AutoFixture/pull/460/files#r43981442).

Src/AutoFixture.xUnit.net.UnitTest/Scenario.cs
```diff
@@ -73,6 +73,42 @@ public void InlineAutoDataSuppliesDataSpecimensOnlyForNonProvidedValues(string s
...
         }
 
+        // This test and its associated types is used to document one of the
+        // InlineAutoDataAttribute constructor overloads.
+        [Theory]
+        [MyCustomInlineAutoData(1337)]
+        [MyCustomInlineAutoData(1337, 7)]
+        [MyCustomInlineAutoData(1337, 7, 42)]
+        public void CustomInlineDataSuppliesExtraValues(int x, int y, int z)
+        {
```
moodmosaic: See @adamchester's [comment](https://github.com/AutoFixture/AutoFixture/pull/460/files#r43981442).

Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
ploeh: I don't understand. Which concrete change to the code are you suggesting?

adamchester: The second inline data value (I.e 7, which corresponds to the `y` argument when the test method runs) has no assertion and is unused inside the method.

Instead of using the literal value `7`, we could use a `const` to give it a name.

I'm not sure what name would suit here.

moodmosaic: Or perhaps we could rename `y` to `dummy`?

adamchester: `dummy` sounds good too

ecampidoglio: How about removing the second parameter altogether? I would probably also rename the custom auto data attribute to be more specific:

```csharp
[Theory]
[AllNumbersAre42InlineAutoData(1337)]
[AllNumbersAre42InlineAutoData(1337, 42)]
public void CustomInlineDataSuppliesExtraValues(int x, int y)
{
    Assert.Equal(1337, x);
    Assert.Equal(42, y);
}
```

I think this makes the intent of replacing customized types with specific values a little clearer.

ploeh: Assigning a `const` requires yet another line of code in the code example. It's also another level of indirection. Would that make the example easier to understand?

Renaming `y` to `dummy` is an option, but does it make the example easier to understand? If we do that, the argument list would be `(int x, int dummy, int z)`. Would the reader wonder what happened to `y` from the range `y..z`?

Would `(int x, int dummyY, int z)` be better?

Or would it be better to entirely remove the unused argument? The reason I included it was to show that you could have an argument that could be provided by a constant, or by AutoFixture.

Another option is to add a comment to the body of the test, like this:

```
Assert.Equal(1337, x);
// y can vary, so we can't express any meaningful assertion for it.
Assert.Equal(42, z);
```

moodmosaic: Good points, all. I think the comment would be just fine in this case :smiley:

ploeh:
Comment now added to test code.

Issue #214, opened 2013-12-01
AutoFixture.AutoFoq should use Foq's new Return Strategy


moodmosaic:
Originally discussed and proposed by Ruben Bartelink (@bartelink) in [Foq project site discussions](https://foq.codeplex.com/discussions/470568).

Foq 1.4 and above (currently available using the latest Foq source code) accepts a Strategy for properties and methods that have not been explicitly setup:

```
Mock<AClient>(returnStrategy: SpecimenContext(fixture).Resolve)
```
The above can be automated with AutoFixture.AutoFoq and could also be the default behavior.

This would be a breaking change (new behavior and Foq package update), so should be done in AutoFixture 4.

moodmosaic:
Foq's default behavior for mock objects that have not been explicitly setup, is to return `null` if the return type of a method is a reference type (e.g. a `string`):

```f#
type IInterface =
    abstract ReturnSomething : unit -> string

let fixture = Fixture().Customize(AutoFoqCustomization())

let sut = fixture.Create<IInterface>()
// No expectations have been setup - return null if
// the return type of a method is a reference type.

let actual = sut.ReturnSomething()
// -> actual: null
```
AutoFixture is an opinionated library, and one of the opinions it holds is that [nulls are invalid return values](http://stackoverflow.com/a/18170070/467754).

By the time AutoFixture.AutoFoq 3 was published, Foq was still in version 1.0 with no specific hooks to override this default behavior.

Since [we've updated Foq](https://github.com/AutoFixture/AutoFixture/pull/264), all the necessary hooks to override its default behavior are already [available](https://foq.codeplex.com/discussions/470568) to us. This means we'll be able to do the following, out of the box, once this issue is worked on:

```f#
type IInterface =
    abstract ReturnSomething : unit -> string

let fixture = Fixture().Customize(AutoFoqCustomization())

let sut = fixture.Create<IInterface>()
// No expectations have been setup - fallback to
// Foq's returnStrategy function.

let actual = sut.ReturnSomething()
// Foq uses a fresh new Fixture instance internally.
// -> actual: "f5cdf6b1-a473-410f-95f3-f427f7abb0c7"
```

---

Since this is happening in `v4`, I'd prefer to modify the existing `AutoFoqCustomization`. — Any thoughts?

ploeh:
@moodmosaic Agreed; I'v removed the *4.0* milestone from this issue.

moodmosaic:
[...]

PR #464, opened 2015-11-09
Use Foq's return strategy in AutoFoq

moodmosaic:
Related to the issue described in #214.

### DIFF
--- a/Src/AutoFoq/AutoFoqCustomization.fs
+++ b/Src/AutoFoq/AutoFoqCustomization.fs
@@ -7,28 +7,7 @@ open System
 /// <summary>
 /// Enables auto-mocking with Foq.
 /// </summary>
-type AutoFoqCustomization(relay: ISpecimenBuilder) =
-    do if relay = null then raise (ArgumentNullException("relay"))
-
-    /// <summary>
-    /// Initializes a new instance of the <see cref="AutoFoqCustomization"/> 
-    /// class.
-    /// </summary>
-    new() = 
-        AutoFoqCustomization(
-            FilteringSpecimenBuilder(
-                MethodInvoker(
-                    FoqMethodQuery(
-                        Fixture())),
-                AbstractTypeSpecification()))
-
-    /// <summary>
-    /// Gets the relay that will be added to 
-    /// <see cref="IFixture.ResidueCollectors"/> when <see cref="Customize"/>
-    /// is invoked.
-    /// </summary>
-    member this.Relay = relay
-
+type AutoFoqCustomization() =
     /// <summary>
     /// Customizes an <see cref="IFixture"/> to enable auto-mocking with Foq.
     /// </summary>
@@ -41,4 +20,9 @@ type AutoFoqCustomization(relay: ISpecimenBuilder) =
         member this.Customize fixture = 
             match fixture with
             | null -> raise (ArgumentNullException("fixture"))
-            | _    -> fixture.ResidueCollectors.Add(this.Relay)
+            | _    -> fixture.ResidueCollectors.Add(
+                          FilteringSpecimenBuilder(
+                              MethodInvoker(
+                                  FoqMethodQuery(
+                                      fixture)),
+                              AbstractTypeSpecification()))
--- a/Src/AutoFoq/AutoFoqCustomization.fs
+++ b/Src/AutoFoq/AutoFoqCustomization.fs
@@ -18,7 +18,8 @@ type AutoFoqCustomization(relay: ISpecimenBuilder) =
         AutoFoqCustomization(
             FilteringSpecimenBuilder(
                 MethodInvoker(
-                    FoqMethodQuery()),
+                    FoqMethodQuery(
+                        Fixture())),
                 AbstractTypeSpecification()))
 
     /// <summary>
--- a/Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
+++ b/Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
@@ -82,12 +82,7 @@ let ``Fixture customizations are propagated in Foq for non-explicitly setup test
     let fixture = Fixture()
     let expected = "Friday, 14 December 1984"
     fixture.Register<string>(fun () -> expected)
-    fixture.ResidueCollectors.Add(
-        FilteringSpecimenBuilder(
-            MethodInvoker(
-                FoqMethodQuery(
-                    fixture)),
-            AbstractTypeSpecification()))
+    fixture.Customize(AutoFoqCustomization()) |> ignore
     let sut = fixture.Create<IInterface>()
 
     let actual = sut.ReturnSomething()
--- a/Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
+++ b/Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
@@ -3,7 +3,9 @@
 open Ploeh.AutoFixture
 open Ploeh.AutoFixture.AutoFoq
 open Ploeh.AutoFixture.AutoFoq.UnitTest.TestDsl
+open Ploeh.AutoFixture.Kernel
 open Ploeh.TestTypeFoundation
+open Swensen.Unquote
 open System
 open System.Collections.Generic
 open Xunit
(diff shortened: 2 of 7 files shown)


Src/AutoFoqUnitTest/FoqMethodQueryTest.fs
```diff
@@ -87,6 +105,17 @@ let SelectMethodsDefineCorrectParameters (request: Type) =
...
+[<Fact>]
+let ``Builder is correct`` () =
+    let expected =
+        { new ISpecimenBuilder with
+              member this.Create(r, c) = obj() }
+    let sut = FoqMethodQuery expected
+
+    let actual = sut.Builder
+
+    verify <@ expected = actual @>
```
moodmosaic: I've just noticed, by looking at the Pull Request, that this can be changed to `expected =? actual`.

moodmosaic: Actually, I've just noticed that we're using an old version of Unquote, and these old operators are deprecated in newer versions, [as shown in the Unquote source code](http://git.io/v8KPF):

```
///These '?' suffixed operators conflict with F# 3.0's nullable operators and
/// have been replaced by equivalent '!' suffixed operators.

[<System.Obsolete(
    "These '?' suffixed operators conflict with F# 3.0's nullable operators and
     have been replaced by equivalent '!' suffixed operators.")>]
module Obsolete =
    let inline (=?) x y = test <@ x = y @>
    // ...

let inline (=!) x y = test <@ x = y @>
//...
```

So I'll leave this one as-is, but I'll change an other test added in this PR so it doesn't use that deprecated `=?` operator.

**Tl;dr**

We should consider upgrading Unquote, in a separate Pull Request.

moodmosaic: Added 3a5e842, in order to address [this](https://github.com/AutoFixture/AutoFixture/pull/464#discussion_r44374058) comment of mine. – Done adding commits.

Src/AutoFoq/AutoFoqCustomization.fs
```diff
@@ -18,7 +18,8 @@ type AutoFoqCustomization(relay: ISpecimenBuilder) =
         AutoFoqCustomization(
             FilteringSpecimenBuilder(
                 MethodInvoker(
-                    FoqMethodQuery()),
+                    FoqMethodQuery(
+                        Fixture())),
```
ploeh: An `ICustomization` is normally something you apply to a given `IFixture` instance. Wouldn't this mean that the default `AutoFoqCustomization` applied to an `IFixture` would contain *another* `Fixture` instance? Isn't that going to be confusing to users?

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
```diff
@@ -66,3 +67,29 @@ let FixtureSuppliesValuesToAbstractGenericTypeWithNonDefaultConstructorWithMulti
...
+    let sut = (Fixture().Customize(AutoFoqCustomization())).Create<IInterface>()
+    let actual = sut.ReturnSomething()
+    verify <@ not (isNull actual) @>
+
+[<Fact>]
+let ``Fixture customizations are propagated in Foq for non-explicitly setup test-doubles`` () =
+    let fixture = Fixture()
+    let expected = "Friday, 14 December 1984"
+    fixture.Register<string>(fun () -> expected)
+    fixture.ResidueCollectors.Add(
```
ploeh: Shouldn't this instead add `AutoFoqCustomization` in order to demonstrate that other Customizations take effect through Foq?

Src/AutoFoqUnitTest/FoqMethodQueryTest.fs
```diff
@@ -74,7 +89,10 @@ let SelectMethodsDefineCorrectParameters (request: Type) =
             ||| BindingFlags.Instance
             ||| BindingFlags.NonPublic)         
         |> Seq.map(fun ci -> ci.GetParameters())
-    let sut = FoqMethodQuery()
+    let dummyBuilder =
```
ploeh: Would it make sense to make `dummyBuilder` a module value?

Src/AutoFoq/AutoFoqCustomization.fs
moodmosaic: To be honest, I was going a bit back and forth on that one :neutral_face:

In the beginning I tried to move the whole object graph in the `Customize` method, in order to supply the given `IFixture` instance, but then I got failing tests which were not easy to make them passing.

---

Then I thought that supplying a *default* `Fixture` can be considered an implementation detail; just so that Foq has something reasonable to return. – After all, this affects *only* those Test Doubles that have not being explicitly setup.

So I went by supplying a default `Fixture` instance, but I've also added a façade test [here](https://github.com/moodmosaic/AutoFixture/blob/use-autofoq-return-strategy-to-avoid-nulls/Src/AutoFoqUnitTest/FixtureIntegrationTest.fs#L82-L95) to demonstrate that it's possible to use the types defined in AutoFoq with a customized `Fixture` instance.

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
moodmosaic: This happens already in that other test, above this one. This one exists mostly for the reason described [here](https://github.com/AutoFixture/AutoFixture/pull/464/files#r44875292).

Src/AutoFoqUnitTest/FoqMethodQueryTest.fs
moodmosaic: Yes, I think it would make sense :+1:

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
ploeh: > This happens already in that other test, above this one.

They aren't testing the same behaviour. The first one only checks to see if `actual` isn't null. This one is more explicit.

Src/AutoFoq/AutoFoqCustomization.fs
ploeh: I think we need to find a solution to this. Having *two* Fixture instances in play is bound to blow up in our faces sooner or later.

moodmosaic: Agreed, let's do it. After all, since this PR targets the `v4` branch things are a bit more flexible.

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
moodmosaic: [I'll get rid of that default `Fixture` instance](https://github.com/AutoFixture/AutoFixture/pull/464/files#r44896077) in order to make it possible to just add the `AutoFoqCustomization`.

Src/AutoFoqUnitTest/FoqMethodQueryTest.fs
moodmosaic: Done in [ef5280e](https://github.com/moodmosaic/AutoFixture/commit/ef5280ed4a30827caaf1ee33999937b7b9c7781a).

Src/AutoFoq/AutoFoqCustomization.fs
moodmosaic: Done in [3702089](https://github.com/AutoFixture/AutoFixture/commit/3702089b1bd10cfd5c2d7221949636a01e367cc0).

moodmosaic:
Review comments addressed. – Done adding commits.

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
```diff
@@ -66,3 +67,24 @@ let FixtureSuppliesValuesToAbstractGenericTypeWithNonDefaultConstructorWithMulti
...
+let ``Fixture customizations are propagated in Foq for non-explicitly setup test-doubles`` () =
+    let fixture = Fixture()
+    let expected = "Friday, 14 December 1984"
+    fixture.Register<string>(fun () -> expected)
+    fixture.Customize(AutoFoqCustomization()) |> ignore
+    let sut = fixture.Create<IInterface>()
+
+    let actual = sut.ReturnSomething()
+
+    verify <@ expected = actual @>
```
ploeh: Fewer unit tests (deletions from `AutoFoqCustomizationTest`) and more [Facade Tests](http://blog.ploeh.dk/2012/06/27/FacadeTest): yay :+1:

ploeh:
Thank you. It's merged into *v4* only, since it contains breaking changes. AFAIU, that was also your intent all along.

Src/AutoFoqUnitTest/FixtureIntegrationTest.fs
moodmosaic: :wink:

moodmosaic:
Thank you for a great review! Yes, that's correct, this one is for the v4 branch.

PR #465, opened 2015-11-09
Updated icon URL in all .nuspec files

ploeh:
to point to the logo file in GitHub.


moodmosaic:
I've just visited the new URLs and they point to the logo file on GitHub, so it's looking good. :+1:

PR #469, opened 2015-11-12
Resolves #467 by relaying non-seeded requests.

### LINKED ISSUE #467: Customizing AutoFixure using FromSeed Causes Exception
Given two classes:

    class Foo
    {
        ...
    }

    class Bar
    {
        public Foo FooBar { get; set; }
    }

I have set up the following test:

    void Test()
    {
        var fixture = new Fixture();

        fixture.Customize<Foo>(x => x.FromSeed(TestFooFactory));

        var fooWithoutSeed = fixture.Create<Foo>();
        var fooWithSeed = fixture.Create<Foo>(new Foo());

        var bar = fixture.Create<Bar>(); //error occurs here
    }
        
    Foo TestFooFactory(Foo seed)
    {
        //do something with seed...

        return new Foo();
    }

I can create `Foo` objects directly with and without seed values without any problem.  But once I try to create a `Bar` object that has a `Foo` property, I get an `ObjectCreationException`:

> The decorated ISpecimenBuilder could not create a specimen based on the request: Foo. This can happen if the request represents an interface or abstract class; if this is the case, register an ISpecimenBuilder that can create specimens based on the request. If this happens in a strongly typed Build<T> expression, try supplying a factory using one of the IFactoryComposer<T> methods.

I'd expect `TestFooFactory` to get passed a `null` seed value during the creation of `Bar`, just as when I created `Foo` without a seed value.

In my real-world scenario, I want to customize how AutoFixture would use seeded values for certain objects when I pass seeded values in, but I still want AutoFixture to default to normal behavior if no seed is provided.

ecampidoglio:
This PR modifies the `SeededFactory<T>` class to relay also [_non-seeded_ requests](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/SeededFactory.cs#L47-L51) for `T` to the custom factory function passing the _default value_ of `T` as a seed. This is to solve #467.

### Compatibility

I'm torn between considering this a _breaking change_ or not. On one side, none of the _existing_ tests were broken by this change, so one might consider it to be backwards compatible. On the other hand, this modifies the behavior of `SeededFactory<T>` by handling a new case or, put differently, by handling an existing case in a _different way_.

I proposed to merge it into the `v4` branch just to be on the safe side. However, I'd like to discuss whether it could be merged into `master`, since this does resolve a concrete issue.


moodmosaic:
Now that #470 is merged, would it help if we first find out [whether #470 solves #467](https://github.com/AutoFixture/AutoFixture/issues/467#issuecomment-156227595) or not? :smiley:

moodmosaic:
Oh, from what I read [here](https://github.com/AutoFixture/AutoFixture/issues/467#issuecomment-156429567), #470 won't also solve #467.

ecampidoglio:
Since #470 has been merged into `master` (b7b6e30), I'm closing this PR and opening a new one for the `master` branch.

PR #471, opened 2015-11-16
Resolves #467 by relaying non-seeded requests.

### LINKED ISSUE #467: Customizing AutoFixure using FromSeed Causes Exception

ecampidoglio:
This PR modifies the `SeededFactory<T>` class to relay also [_non-seeded_ requests](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/SeededFactory.cs#L47-L51) for `T` to the custom factory function passing the _default value_ of `T` as a seed. This is to solve #467.

### DIFF
--- a/Src/AutoFixture/Kernel/SeededFactory.cs
+++ b/Src/AutoFixture/Kernel/SeededFactory.cs
@@ -44,6 +44,11 @@ namespace Ploeh.AutoFixture.Kernel
         /// </returns>
         public object Create(object request, ISpecimenContext context)
         {
+            if (request != null && request.Equals(typeof(T)))
+            {
+                return this.create(default(T));
+            }
+
             var seededRequest = request as SeededRequest;
             if (seededRequest == null)
             {
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -3523,6 +3523,29 @@ namespace Ploeh.AutoFixtureUnitTest
         }
 
         [Fact]
+        public void CustomizeFromSeedWithUnmodifiedSeedValueWillPopulatePropertyOfSameType()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            // Exercise system
+            fixture.Customize<Version>(c => c.FromSeed(s => s));
+            // Verify outcome
+            Assert.Null(fixture.Create<PropertyHolder<Version>>().Property);
+        }
+
+        [Fact]
+        public void CustomizeFromSeedWithFixedSeedValueWillPopulatePropertyOfSameType()
+        {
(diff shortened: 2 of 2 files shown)


Src/AutoFixtureUnitTest/Kernel/SeededFactoryTest.cs
```diff
@@ -129,5 +129,23 @@ public void CreateWithCorrectlyTypedSeededRequestWithNullSeedReturnsCorrectResul
             Assert.Equal(expectedResult, result);
             // Teardown
         }
+
+        [Fact]
+        public void CreateWithCorrectlyTypedRequestReturnsCorrectResult()
```
ploeh: Doesn't this test slightly overspecify the implementation details?

If we didn't have this test, wouldn't it leaves us more flexibility with regards to future refactoring?

ecampidoglio: Yes, that test was superseded by this [scenario test](https://github.com/AutoFixture/AutoFixture/blob/467-create-value-from-seed/Src/AutoFixtureUnitTest/FixtureTest.cs#L3525;L3534) introduced in 28bb833. I'll remove it.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.36.12](https://www.nuget.org/packages/AutoFixture/3.36.12).

Issue #349, opened 2015-02-19
MailAddressGenerator should use context to resolve domain name


ploeh:
As described in [this comment](https://github.com/AutoFixture/AutoFixture/pull/340#issuecomment-71556125), `MailAddressGenerator` should delegate the creation of the domain name part of an email address to the `ISpecimenContext`.

The generated domain names should by default be one of the domain names "example.com", "example.net", or "example.org", so that accidental collisions with real resources on the internet can be avoided. These domain names are reserved for such purposes.

ploeh:
The implementation of this feature would be an enhancement of #340.

tiesmaster:
I'm interested in picking up this issue. Is this still wanted?

As for the naming of the class, should I follow the same pattern as #348, and name it `EmailAddressDomainPart` (that's how the RFC talks about it)? Or follow the naming of the referenced comment, use `DomainName`, so it might be used in other places where you need a domain name?

ploeh:
@tiesmaster, you are welcome to contribute this feature :smile: 

Good question regarding the name. I'd lean slightly towards `DomainName`, because it'd also be useful for other purposes, such as WWW domain names. By default we should only use `example.org` etc. but calling the class `DomainName` leaves it open to also serve up names such as `www.example.org`. That might seem like a slightly odd part of an email address, but as far as I know, `foo@www.example.org` is still a valid email address.

I'm open for counter-arguments, though :smile:

tiesmaster:
Ok, good to know ;)

`DomainName` sounds like the most appropriate then, I'll take that.

PR #472, opened 2015-11-18
Issue #349: use context to resolve domain name

tiesmaster:
@ploeh As discussed, a PR for #349. My first PR for AutoFixture ;)

A few heads-up: I cheated a bit, I looked heavily at the PR of #348, since it was very similar to this, and to look at naming of test scenario's, and general code style of AutoFixture. So it might look a bit familiar.

Next, I could reuse the `.GetHashCode() % 3`, so I used `Random` for this. I don't know if that the right approach is, but getting a random `int` from the context seemed wrong to me, since that can get frozen, and that will fix the `DomainName` as well, isn't it?

Finally, for testing the random domain names, I have a `foreach` loop in my test, which is usually bad, but I didn't know how to test it otherwise. Let me know what you thing of it.

### DIFF
--- a/Src/AutoFixture/DomainName.cs
+++ b/Src/AutoFixture/DomainName.cs
@@ -1,0 +1,87 @@
+﻿using System;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Represents a domain name.
+    /// </summary>
+    public class DomainName
+    {
+        private string domainName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DomainName"/> class. Throws ArgumentNullException
+        /// if domainName is null. Throws ArgumentException if domainName is empty.
+        /// </summary>
+        public DomainName(string domainName)
+        {
+            if (domainName == null)
+            {
+                throw new ArgumentNullException("domainName");
+            }
+
+            if (domainName == string.Empty)
+            {
+                throw new ArgumentException("domainName");
+            }
+
+            this.domainName = domainName;
+        }
+
+        /// <summary>
+        /// Get the name of the domain.
+        /// </summary>
+        public string Domain
+        {
+            get
+            {
+                return this.domainName;
+            }
+        }
+
+        /// <summary>
+        /// Returns a <see cref="System.String"/> that represents the domain name for this instance.
+        /// </summary>
+        /// <returns>
+        /// A <see cref="System.String"/> that represents the domain name for this instance.
+        public override string ToString()
+        {
+            return this.domainName;
+        }
+
+        /// <summary>
+        /// Determines whether the specified <see cref="System.Object"/> is equal to this instance.
+        /// </summary>
+        /// <param name="obj">The <see cref="System.Object"/> to compare with this instance.
+        /// </param>
+        /// <returns>
+        ///   <c>true</c> if the specified <see cref="System.Object"/> is equal to this instance; 
+        /// otherwise, <c>false</c>.
+        /// </returns>
+        /// <exception cref="T:System.NullReferenceException">
+        /// The <paramref name="obj"/> parameter is null.
+        ///   </exception>
+        public override bool Equals(object obj)
+        {
+            var other = obj as DomainName;
+
+            if (other != null)
+            {
+                return this.domainName.Equals(other.domainName);
+            }
+            return base.Equals(obj);
+        }
+
+        /// <summary>
+        /// Returns a hash code for this instance.
+        /// </summary>
+        /// <returns>
+        /// A hash code for this instance, suitable for use in hashing algorithms and data 
+        /// structures like a hash table. 
+        /// </returns>
+        public override int GetHashCode()
+        {
+            return this.domainName.GetHashCode();
+        }
+    }
+}
--- a/Src/AutoFixture/DomainName.cs
+++ b/Src/AutoFixture/DomainName.cs
@@ -20,9 +20,9 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentNullException("domainName");
             }
 
-            if (domainName == string.Empty)
+            if (domainName.Length == 0)
             {
-                throw new ArgumentException("domainName");
+                throw new ArgumentException("domainName cannot be empty");
             }
 
             this.domainName = domainName;
--- a/Src/AutoFixture/DomainName.cs
+++ b/Src/AutoFixture/DomainName.cs
@@ -44,6 +44,7 @@ namespace Ploeh.AutoFixture
         /// </summary>
         /// <returns>
         /// A <see cref="System.String"/> that represents the domain name for this instance.
+        /// </returns>
         public override string ToString()
         {
             return this.domainName;
--- a/Src/AutoFixture/DomainName.cs
+++ b/Src/AutoFixture/DomainName.cs
@@ -19,12 +19,6 @@ namespace Ploeh.AutoFixture
             {
                 throw new ArgumentNullException("domainName");
             }
-
-            if (domainName.Length == 0)
-            {
-                throw new ArgumentException("domainName cannot be empty");
-            }
-
             this.domainName = domainName;
         }
 
--- a/Src/AutoFixture/DomainNameGenerator.cs
+++ b/Src/AutoFixture/DomainNameGenerator.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Creates new <see cref="DomainName"/> instances.
+    /// </summary>
+    public class DomainNameGenerator : ISpecimenBuilder
+    {
+        private readonly string[] fictitiousDomains =
+        {
+            "example.com",
+            "example.net",
+            "example.org"
+        };
+
+        private readonly Random random = new Random();
+
+        /// <summary>
+        /// Creates a new specimen based on a request.
+        /// </summary>
+        /// <param name="request">The request that describes what to create</param>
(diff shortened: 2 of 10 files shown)


ploeh:
Thank you for submitting this. I'm currently travelling, and with a client, so I may only have time to review this in a couple of days from now. Meanwhile, I can see that the build is failing because of some malformed XML comments. These ought to be easy to fix, but must still be addressed.

When you've done that, please leave a message to that effect, because GitHub doesn't send out notifications when someone push new commits to a pull request. The ensures that I, or someone else here, are reminded to take a second look at your pull request as soon as possible :smile:

tiesmaster:
Yes, after I submitted the PR, I realized that I hadn't run the build release script (forgot all about that, and had read the contributing guideline earlier on, so didn't read it when committing :().

I've fixed the xml doc, and ran the build release, but I get a lot more errors. I think because I only have VS 2015 installed, and the buildserver is configured on an older appveyor image. Is that right?

Let's see if this goes through.

tiesmaster:
@ploeh Build is green ;)

Src/AutoFixtureUnitTest/DomainNameTest.cs
```diff
@@ -0,0 +1,136 @@
...
+            object other = null;
+            // Exercise system
+            bool result = sut.Equals(other);
+            // Verify outcome
+            Assert.False(result);
+            // Teardown
+        }
+
+        [Fact]
+        public void SutDoesNotEqualNullSut()
```
ploeh: Does this test exercise a different code path than `SutDoesNotEqualNullObject`?

Src/AutoFixtureUnitTest/DomainNameTest.cs
```diff
@@ -0,0 +1,136 @@
...
+            object other = new DomainName(Guid.NewGuid().ToString());
+            // Exercise system
+            bool result = sut.Equals(other);
+            // Verify outcome
+            Assert.False(result);
+            // Teardown
+        }
+
+        [Fact]
+        public void SutDoesNotEqualOtherSutWhenDomainNamesDiffer()
```
ploeh: Does this test exercise a different code path than `SutDoesNotEqualOtherObjectWhenDomainNamesDiffer`?

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
```diff
@@ -87,13 +94,39 @@ public void CreateWithMailAddressRequestReturnsCorrectResultUsingLocalPartFromCo
...
+            {
+                OnResolve = r =>
+                {
+                    Assert.True(typeof(EmailAddressLocalPart).Equals(r) || typeof(DomainName).Equals(r));
+                    return typeof(DomainName).Equals(r) ? anonymousDomainName : null;
+                }
+            };
+            var sut = new MailAddressGenerator();
+            // Exercise system
+            var expectedResult = new NoSpecimen(request);
```
ploeh: Why is `expectedResult` part of the *Exercise system* phase?

Src/AutoFixture/DomainName.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// if domainName is null. Throws ArgumentException if domainName is empty.
+        /// </summary>
+        public DomainName(string domainName)
+        {
+            if (domainName == null)
+            {
+                throw new ArgumentNullException("domainName");
+            }
+
+            if (domainName.Length == 0)
```
ploeh: Why can't `domainName` be empty?

ploeh:
Thank you; it's looking good :+1: 

The comments I added are almost superficially minor, and I could easily work with the code to take it the last stretch myself. Often, however, at least part of the reason people tend to submit pull requests to AutoFixture is in order to learn. If that's the case for you as well, you can easily address the comments; otherwise, let me know, and I'll pull it as it is (and make a few changes myself).

tiesmaster:
Indeed, a big reason to contribute is to learn. It's not at all a problem to process the feedback. I'm away for the weekend so I'll put in the changes on Monday or Tuesday.

Src/AutoFixtureUnitTest/DomainNameTest.cs
tiesmaster: No, indeed, same path. I've added a commit to remove this superfluous unit test, as well as the other one.

I've also taken the liberty to remove them from the `EmailAddressLocalPart` unit test in a separate PR (#474).

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
tiesmaster: Indeed, I also found it weird that it was part of that phase. It was already like that, but now that I've looked around elsewhere in the code, I see that's normally part of the assert phase.

Frankly, I usually make my expected outcome part of the arrange phase, since it's very often derived from the input values, and then it closer to the setup of the rest of the fixture, and you only need to do actual assert calls for the assert phase. However, after looking around, I'd say it's part of the assert phase, so I've moved it there.

Src/AutoFixture/DomainName.cs
tiesmaster: Mostly, because the `EmailAddressLocalPart` also didn't allow that. I saw your discussion in the other PR about the RFC implementation, and agree that you shouldn't reimplement the validation logic of the MailAddress class. With that in mind, it's perfectly valid for the `domainName` to be emtpy, and let the `MailAddressGenerator` handle invalid domain names.

I've removed the constraint. Should it also be removed from `EmailAddressLocalPart`?

tiesmaster:
@ploeh I've pushed new commits to resolve the issues. Please have a look again.

Src/AutoFixtureUnitTest/MailAddressGeneratorTests.cs
ploeh: Technically, there's no difference :smile: 

FWIW, I find the related heuristics described in [Code Complete](http://amzn.to/1xioCmw) reasonable: introduce variables as close to where they're used as possible. That makes the code easier to read, because the distance between introduction and use is as short as possible.

Src/AutoFixture/DomainName.cs
ploeh: I can't even recall having that discussion related to `EmailAddressLocalPart` :flushed: I wonder why we left it in...

Src/AutoFixture/DomainNameGenerator.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="request">The request that describes what to create</param>
+        /// <param name="context">A context that can be used to create other specimens.</param>
+        /// <returns>
+        /// The requested specimen if possible; otherwise a <see cref="NoSpecimen"/> instance.
+        /// </returns>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (request == null || !typeof(DomainName).Equals(request))
+            {
+                return new NoSpecimen();
```
ploeh: This looks *slightly* incorrect to me.

The convention in AutoFixture is to return `new NoSpecimen(request)` if `request` isn't null.

IIRC, it's actually not used anywhere, but I'd like to be consistent. Also, since AutoFixture is a reusable library, we don't know if someone else relies on this convention, even if AutoFixture itself doesn't. I doubt it, however.

(If I should really do something about this, since it's not used, I ought to remove the overload that takes the request, but that would be a breaking change...)

Src/AutoFixtureUnitTest/DomainNameGeneratorTest.cs
```diff
@@ -0,0 +1,77 @@
...
+        [Fact]
+        public void CreateReturnsOneOfTheFictiousDomains()
+        {
+            // Fixture setup
+            var sut = new DomainNameGenerator();
+            // Exercise system
+            var result = sut.Create(typeof(DomainName), null);
+            // Verify outcome
+            Assert.IsAssignableFrom<DomainName>(result);
+            var actualDomainName = (DomainName)result;
```
ploeh: Wouldn't the following be easier?

```C#
var actualDomainName = Assert.IsAssignableFrom<DomainName>(result);
```

ploeh:
Getting closer :+1:

Src/AutoFixture/DomainNameGenerator.cs
ploeh: BTW, this is one of those conventions that are quite difficult to discover. It represents a design failure on my part.

ploeh: Based on this, I've created #475.

tiesmaster: Well, now you mention it, I had already seen it, when looking around in the code elsewhere. However, as you mention, it's easy to do wrong, and APIs, of course, should make it easy to do the right thing, and difficult to do the wrong thing ;)

Looking at the references of `Request`, it's indeed not being used (at least, not in the `AutoFixture.sln` solution, I'll do a more thorough check, if I pick it up :smiley:).

image

Src/AutoFixtureUnitTest/DomainNameGeneratorTest.cs
tiesmaster: :+1: I didn't knew the `IsAssignableFrom` method wasn't void. (I'm still a bit unfamiliar with the `Xunit` assert methods; I use the FluentAssertions library at work, and also for my private projects.)

tiesmaster:
@ploeh Can you have another look?

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.37.0](https://www.nuget.org/packages/AutoFixture/3.37.0).

Issue #221, opened 2013-12-08
Consider performance impact of attribute-based default relays


adamchester:
During my performance testing (e.g. #218), I discovered that the `ISpecimenBuilder` relays, which query for attributes via an `ICustomAttributeProvider`, have a major impact on performance. The negative impact exists regardless if the caller/user requires them.

My testing revealed that by commenting out [these lines from `DefaultRelays`](https://github.com/AutoFixture/AutoFixture/blob/04c0d4c0/Src/AutoFixture/DefaultRelays.cs#L25-L28) (i.e. removing the `StringLengthAttributeRelay`, `RangeAttributeRelay`, and `RegularExpressionAttributeRelay`) resulted in a huge performance improvement.

For example, with some simple tests I used in #218, the following was observed:

```
WITH-ATTR-RELAYS>msbuild BuildRelease.msbuild /t:test /m:8 | find /i /n "quick"
[3787]             Ploeh.AutoFixtureUnitTest.FixtureTest.Create10000ImmutableSpecimensQuickly
[3789]               Create10000ImmutableSpecimensQuickly, Elapsed: 00:00:00.6987564
[3802]             Ploeh.AutoFixtureUnitTest.FixtureTest.Create1000SpecimensQuickly
[3804]               Create1000SpecimensQuickly, Elapsed: 00:00:01.5871797
[3840]             Ploeh.AutoFixtureUnitTest.FixtureTest.InstantiateFixtureQuickly
[3842]               InstantiateFixtureQuickly, Elapsed: 00:00:00.0291962

NO-ATTR-RELAYS>msbuild BuildRelease.msbuild /t:test /m:8 | find /i /n "quick"
[3553]             Ploeh.AutoFixtureUnitTest.FixtureTest.Create10000ImmutableSpecimensQuickly
[3555]               Create10000ImmutableSpecimensQuickly, Elapsed: 00:00:00.3815511
[3713]             Ploeh.AutoFixtureUnitTest.FixtureTest.Create1000SpecimensQuickly
[3715]               Create1000SpecimensQuickly, Elapsed: 00:00:00.6354635
[3780]             Ploeh.AutoFixtureUnitTest.FixtureTest.InstantiateFixtureQuickly
[3782]               InstantiateFixtureQuickly, Elapsed: 00:00:00.0260725
```

[...]

Issue #475, opened 2015-11-24
Investigate removing Request from NoSpecimen


ploeh:
IIRC, the `Request` property of `NoSpecimen` is never used. If so, it should be removed. Since this is a breaking change, it should be done as part of AutoFixture 4.

If it turns out that it's used after all, we can simply close this issue.

adamchester:
It was used by some tests a long time ago. #221 [discusses removal](https://github.com/AutoFixture/AutoFixture/issues/221#issuecomment-31623706) and I have an [old branch](https://github.com/adamchester/AutoFixture/tree/nospecimen-singleton) that may help.

ploeh:
To ease into this breaking change, we ought to deprecate the `Request` property and the corresponding constructor, using the `[Obsolete]` attribute.

JoyPeterson:
I am seeing the warning that this is now obsolete in code that I adapted from http://stackoverflow.com/a/27774580/1656318.  Is there another way to accomplish what this solution on StackOverflow does?

Thanks.

ploeh:
@JoyPeterson `new NoSpecimen()` isn't obsolote.

ploeh:
As far as I can tell, we're done removing use of `new NoSpecimen(request)` in the *v4* branch, apart from the definition of the constructor overload and the property itself, and the tests that cover these members.

Should we introduce a Singleton property for `NoSpecimen` instead of requiring people to use `new NoSpecimen()`?

As @adamchester has commented, it doesn't seem to have much performance impact, but on the other hand, would it hurt?

If we introduce a `NoSpecimen.Instance` static property, we can also deprecate the parameterless constructor. This would be possible to do already in AutoFixture 3.

moodmosaic:
I would leave it as-is, without introducing a static member in the `NoSpecimen` definition.

AFAICT, in the AutoFixure codebase static members aren't exposed; they might be just used internally here and there as an implementation detail. I could be wrong though...

---

At the end, it really is a matter of personal style, the fact that I can't see a place for static members in OOP.

adamchester:
I think it's a good idea, it seems needlessly wasteful to create so many instances of it, and I can't think of a good reason why it would be better as separate instances.

moodmosaic:
@adamchester Granted, it can be overhead for the GC. [Should we care?](http://c2.com/cgi/wiki?PrematureOptimization)

---

If not, then, there's also this:

| Text  | Chars  |
|---|---|---|
[...]

PR #480, opened 2015-12-02
Deprecated NoSpecimen.Request,

ploeh:
and the corresponding constructor. The reason for this is that Request is
never used, but the design makes it look like it is. For consistency's
sake, we ask contributors to adhere to the established standard of
returning `new NoSpecimen(request)` even when everything works fine
without it. The design would be slightly simpler if Request was simply
omitted.

This also paves the way towards making NoSpecimen a Singleton, although
measurements carried out by Adam Chester indicate that it would only give
a marginal performance boost
(https://github.com/AutoFixture/AutoFixture/issues/221#issuecomment-31623706).

Only two [Obsolete] attributes are added in this commit, both in
Src/AutoFixture/Kernel/NoSpecimen.cs. The rest of the files only contain
myriads of warning suppressions, because of the deprecation.

This commit changes no behaviour, because that might be a breaking change
(although I doubt it). Instead, all warnings related to deprecation of
NoSpecimen.Request are suppressed, in order to keep the behaviour as is.

### DIFF
--- a/Src/AutoFakeItEasy/FakeItEasyBuilder.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyBuilder.cs
@@ -68,19 +68,25 @@ namespace Ploeh.AutoFixture.AutoFakeItEasy
             var type = request as Type;
             if (!type.IsFake())
             {
+#pragma warning disable 618
                 return new NoSpecimen(request);
+#pragma warning restore 618
             }
 
             var fake = this.builder.Create(request, context) as FakeItEasy.Configuration.IHideObjectMembers;
             if (fake == null)
             {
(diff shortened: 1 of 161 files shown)


ploeh:
**Note to reviewers:**

`Src/AutoFixture/Kernel/NoSpecimen.cs` contains the interesting change. The rest is super-repetitive warning suppressions.

ploeh:
BTW, this commit is related to #475, but is only the first step. It doesn't complete #475, so don't close it even if accepting this PR.

moodmosaic:
Wow, what a massive piece of work :+1:

Build succeeds without warnings on my machine.

Issue #482, opened 2015-12-04
Can't customize byte property


karstenfk:
var fixture = new Fixture();
fixture.Customize<WorkStationTimeLog>(f => f.With(l => l.Position, 0));

If property Position is byte I get the following exception. When I change the type to int, it works.

An exception of type 'System.ArgumentException' occurred in Ploeh.AutoFixture.dll but was not handled in user code
Additional information: The expression's Body is not a MemberExpression. Most likely this is because it does not represent access to a property or field.

moodmosaic:
Thank you for reporting this! AFAICT, this looks like a bug.

---

Until this is issue is resolved, explicitly casting to `byte` works:

```c#
public class Foo
{
    public byte Position { get; set; }
}

[Fact]
public void CustomizeBytePropertyWorks()
{
    var fixture = new Fixture();
    fixture.Customize<Foo>(c => c.With(x => x.Position, (byte)0));

    Assert.DoesNotThrow(() => fixture.Create<Foo>());
}
```

moodmosaic:
The issue (*if* it's a bug) seems to be in the [`ExpressionReflector`](https://github.com/AutoFixture/AutoFixture/blob/046c18b532bfa56d4bf64d66979233018b9dfdaa/Src/AutoFixture/Kernel/ExpressionReflector.cs) file:

![image](https://cloud.githubusercontent.com/assets/287532/11610113/d5650d68-9b90-11e5-9849-e8e588937bab.png)

The `propertyPicker.Body` is an `UnaryExpression` on runtime, so an exception is thrown.

PR #483, opened 2015-12-07
With conversion issue

ploeh:
Attempts to address #482.


moodmosaic:
Awesome :+1: Yes, this seems to be solving the problem :smiley:

moodmosaic:
Builds without warnings and errors on my machine.

Issue #485, opened 2015-12-08
Invalid cast exception after installing 3.37.2


voroninp:
Here is the stack trace:

>Test method Wumte.Server.Distribution.Core.UnitTests.InfoSec.InfoSecDistributionDbAccessorTest.GetTerminalsInfo_CorrectQueryLogic threw exception: 
>System.InvalidCastException: Invalid cast from 'System.DateTime' to 'System.Nullable`1[[System.DateTime, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]'.
>
>    at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
>   at System.DateTime.System.IConvertible.ToType(Type type, IFormatProvider provider)
>   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
>   at Ploeh.AutoFixture.Kernel.BindingCommand`2.ConvertIfAppropriate(Object candidate, Type targetType)
>   at Ploeh.AutoFixture.Kernel.BindingCommand`2.Execute(Object specimen, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.FilteringSpecimenBuilder.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Dsl.NodeComposer`1.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.RecursionGuard.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Dsl.CompositeNodeComposer`1.Create(Object request, ISpecimenContext context)
>   at Ploeh.AutoFixture.Kernel.SpecimenContext.Resolve(Object request)
[...]

Issue #453, opened 2015-10-14
DataAnnotations RangeAttribute causes arithmetic OverflowException


kristofferahl:
We have a simple model that looks like this:

**Model**

```
public class AddRequest
{
    [Range(1, long.MaxValue, ErrorMessage = "Id is not in range")]
    public long CustomerId { get; set; }

    public int PartnerId { get; set; }

    public string Text { get; set; }
}
```

**Test**

```
var fixture = new Fixture();
var request = fixture.Freeze<AddRequest>();
```

**Exception**

When using autofixures Freeze method to create an instance we get the following exception:

```
  System.OverflowException : Arithmetic operation resulted in an overflow.
     at System.Double.System.IConvertible.ToInt64(IFormatProvider provider)
     at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
     at Ploeh.AutoFixture.DataAnnotations.RangeAttributeRelay.Create(RangeAttribute rangeAttribute, Object request)
     at Ploeh.AutoFixture.DataAnnotations.RangeAttributeRelay.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.CustomizationNode.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.TracingBuilder.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.TerminatingWithPathSpecimenBuilder.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.BehaviorRoot.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.RecursionGuard.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Fixture.Create(Object request, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.SpecimenContext.Resolve(Object request)
     at Ploeh.AutoFixture.Kernel.AutoPropertiesCommand`1.Execute(Object specimen, ISpecimenContext context)
     at Ploeh.AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
[...]

PR #486, opened 2015-12-08
Supported implicit conversions in With expressions

ploeh:
It turns out that the support for conversions in `With` expressions added
previously caused a regression that we've never tested for: when an
implicit conversion exists (implemented by op_Implicit), the `With`
expression used to work, but after explicitly attempting to convert using
Convert.ChangeType, this doesn't always work (e.g. for Nullable<T>).

There seems to be no easy way to use Reflection to figure out if an
implicit conversion already exists, so the pragmatic solution seems to be
to attemp making the conversion in a try block, and then retry with the
conversion if the first attempt fails.

If someone learns of a better way to test for this without resorting to
exception handling, a pull request will be welcome.

### DIFF
--- a/Src/AutoFixture/Kernel/BindingCommand.cs
+++ b/Src/AutoFixture/Kernel/BindingCommand.cs
@@ -210,29 +210,43 @@ namespace Ploeh.AutoFixture.Kernel
 
             var pi = this.member as PropertyInfo;
             if (pi != null)
-                pi.SetValue(
+                TrySetValue(
                     specimen,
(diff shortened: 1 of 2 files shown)


ploeh:
This is an attempt to address #485

moodmosaic:
Damn, those *IConvertibles*...

moodmosaic:
FWIW, I was struggling with similar issues [while investigating #453](https://github.com/AutoFixture/AutoFixture/issues/453#issuecomment-149463476)...

PR #487, opened 2015-12-10
Use no specimen without request

ploeh:
This pull request is a partial, incremental move towards further deprecating `NoSpecimen.Request`, as outlined in #475, although that work is far from complete with this.

Since all these changes are breaking, the commits should only be applied to the *v4* branch (in case someone actually decides to accept this).

PR #489, opened 2015-12-17
Remove some more usages of request overload for no specimen

ploeh:
Similar to PR #487 this PR takes us a little further towards addressing #475.

Since it contains breaking changes, it should only be applied to the *v4* branch, if accepted.


moodmosaic:
Looking good! The `build-release.sh` script succeeds without warnings on my machine.

moodmosaic:
Very nice! We're getting closer to #475, which is great :smiley:

PR #491, opened 2015-12-20
Remove some more NoSpecimen requests

tiesmaster:
I'd thought, I'll pitch in a little bit for this. I hope that's ok ;)


moodmosaic:

This is looking good to me. Since it contains breaking changes, it will only be applied to the v4 branch, similar to PR #489.

moodmosaic:

moodmosaic:
Very nice! We're getting closer to #475, which is great :smiley:

ploeh:
:+1: Thank you!

PR #492, opened 2015-12-20
Remove last of deprecated NoSpecimen from main AutoFixture project

tiesmaster:
Some more removals. These were the last occurrences in the main AutoFixture project ;)


moodmosaic:

This is looking good to me. Since it contains breaking changes, it will only be applied to the v4 branch, similar to PR #491.

moodmosaic:

PR #494, opened 2015-12-20
Remove deprecated nospecimen ctor calls from AutoMoq

tiesmaster:
This is the last one for me tonight ;)


moodmosaic:

This is looking good to me. Since it contains breaking changes, it will only be applied to the v4 branch, similar to PR #493.

moodmosaic:

PR #495, opened 2015-12-21
Fix #459: import xunit.core.props in xunit.net2 project

### LINKED ISSUE #459: Test failure in AutoFixture.xUnit.net2 in Visual Studio
When I open *AutoFixture.xUnit.net2.sln* in Visual Studio 2015 and attempt to run all tests with [TestDriven.Net](http://testdriven.net), I get this result:

```
------ Test started: Assembly: Ploeh.AutoFixture.Xunit2.dll ------

Test '' failed: File not found: C:\Users\mark\Documents\ploeh\AutoFixture\Src\AutoFixture.xUnit.net2\bin\Debug\xunit.execution.desktop.dll
Parameter name: assemblyFileName
	System.ArgumentException: File not found: C:\Users\mark\Documents\ploeh\AutoFixture\Src\AutoFixture.xUnit.net2\bin\Debug\xunit.execution.desktop.dll
	Parameter name: assemblyFileName
	at Guard.FileExists(String argName, String fileName)
	at Xunit.Xunit2Discoverer..ctor(ISourceInformationProvider sourceInformationProvider, IAssemblyInfo assemblyInfo, String assemblyFileName, String xunitExecutionAssemblyPath, String configFileName, Boolean shadowCopy, String shadowCopyFolder, IMessageSink diagnosticMessageSink)
	at Xunit.Runner.TdNet.TdNetRunnerHelper..ctor(Assembly assembly, ITestListener testListener)
	at Xunit.Runner.TdNet.TdNetRunner.CreateHelper(ITestListener testListener, Assembly assembly)
	at Xunit.Runner.TdNet.TdNetRunner.RunAssembly(ITestListener testListener, Assembly assembly)
	at TestDriven.TestRunner.AdaptorTestRunner.Run(ITestListener testListener, ITraceListener traceListener, String assemblyPath, String testPath)
	at TestDriven.TestRunner.ThreadTestRunner.Runner.Run()

0 passed, 1 failed, 0 skipped, took 0,44 seconds (xUnit.net 2.0.0 build 2929).

------ Test started: Assembly: Ploeh.AutoFixture.Xunit2.UnitTest.dll ------
134 passed, 0 failed, 0 skipped, took 3,46 seconds (xUnit.net 2.0.0 build 2929).

------ Test started: Assembly: Ploeh.AutoFixtureUnitTest.dll ------
3691 passed, 0 failed, 0 skipped, took 12,02 seconds (xUnit.net 1.9.2 build 1705).

========== Total Tests: 3825 passed, 1 failed, 0 skipped, took 16,17 seconds ==========
```

The `build-release-sh` script runs without error, so it seems to be a Visual Studio-specific issue. It may also be present in other solutions, but I've yet to try. This seems to have been present at least since 7ab27e6dcbf4ee49a2a67cb0e4c0adb7e7fa5720, but I haven't attempted to trace it further back than that.

tiesmaster:
This fixes #459.

### DIFF
--- a/Src/AutoFixture.xUnit.net2/AutoFixture.xUnit.net2.csproj
+++ b/Src/AutoFixture.xUnit.net2/AutoFixture.xUnit.net2.csproj
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\Packages\xunit.core.2.0.0\build\portable-net45+win+wpa81+wp80+monotouch+monoandroid+Xamarin.iOS\xunit.core.props" Condition="Exists('..\..\Packages\xunit.core.2.0.0\build\portable-net45+win+wpa81+wp80+monotouch+monoandroid+Xamarin.iOS\xunit.core.props')" />
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>


ploeh:
This looks like it solves the problem. Thank you.

As far as I can tell, this pull request is entirely beneficial. It also builds without errors on my computer.

I can't claim that I understand why that import is necessary, though, so I'd like to solicit a second opinion, particularly from, but not limited to, @moodmosaic, @ecampidoglio, or @adamchester.

moodmosaic:
I don't know why we need to import `xunit.core.props` but it seems like it's solving our problem.

/cc @bradwilson

tiesmaster:
I wasn't really sure if my explanation in the issue was clear enough, but the previous two comments indicate it wasn't ;) Below, I'll try to elaborate a little bit more on this. I hope that that will give enough background why this happens.

The reason this is needed, is because the xunit2 library doesn't work without it. They use a technique, that's called "Bait, and Switch Trick". There are [a couple blogposts](https://www.google.nl/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=pcl+bait+and+switch) on this topic, but the general gist is like this: the PCL implements a small portion of the library, and a full implementation comes from a separate assembly, one for each platform it supports. During runtime, the appropriate assembly is loaded using reflection, and the full platform is available to the component, instead of the restricted API which you get as a PCL.

One of the nuget package that I know of that uses this technique is AutoMapper, but I expect that there are plenty of more that do this. Xunit2 appears to be another one that uses this, although, it seems that they don't have a separate assembly for each platform, but they do have one for the full desktop framework (that is the `xunit.execution.desktop.dll` assembly, the [`Xunit2Discoverer`](https://github.com/xunit/xunit/blob/master/src/xunit.runner.utility/Frameworks/v2/Xunit2Discoverer.cs#L112) class shows how that file is located, and loaded).

The problem with this approach is that you reference the PCL, but during runtime, you also need the platform specific assembly that contains the rest of the implementation. AutoMapper solves that by just adding the platform specific assembly as reference when you install the nuget package, even though you only consume the API of the library solely through the PCL.

Xunit solves that a little bit differently: they don't force you to reference the desktop assembly, but add it as a item in the solution (through the [`xunit.core.props`](https://github.com/xunit/xunit/blob/master/src/xunit.core/build/xunit.core.props) project), hide it from the list, and enable "copy to output directory" on compilation, so it's there at runtime. Probably, they add this line when you install the xunit nuget package, though, I just tried to reproduce that, but wasn't able to accomplish that :(

Finally, just in case this was missed, the reason that this doesn't fail for the `AutoFixture.xUnit.net2.UnitTest`, is that that project has that project reference, so the file is available for that project.

bradwilson:
The situation here is a little different than bait-and-switch. It is a mandated bunch of nonsense to enable support in the Visual Studio runner.

The Visual Studio test runner allows you to run tests with a variety of project types, including desktop and device projects. When you are writing a non-desktop unit test library (including PCL), you link against our non-desktop execution implementation. However, the architecture of the Visual Studio runner demands that discovery be done from the desktop framework, even when the unit tests are not desktop unit tests. We use the `.props` file to copy the desktop execution library in there, so that the desktop discovery can continue to work. We load the execution library dynamically based on the runtime environment, so in this one scenario, we need both (even though you aren't writing desktop unit tests).

ploeh:
@tiesmaster @bradwilson Thank you for taking the time to explain this to me.

There are still a couple of items I don't understand:

**1.** Why are we seeing problems specifically with *TestDriven.Net*, while the batch build (`build-release.sh` or `BuildRelease.ps1`) executes without error?

Is it because the batch build uses the xUnit.net MSBuild runner, and that runner already does everything required, as described above?

**2.** Why are we even seeing the error described in #459 for the `Ploeh.AutoFixture.Xunit2.dll` library?

While this library references xUnit.net, it's a Glue Library, and it contains no tests.

Is it because the xUnit.net test discovery mechanism kicks in simply because of the reference to *xunit.core*, and then fails because it can't run (as explained above)?

bradwilson:
> Why are we seeing problems specifically with TestDriven.Net

It's hard to pinpoint down, but it's possible that in some combinations, TD.NET also requires the desktop execution library because it always runs in the desktop context. We have simply always expected that the `.props` file would be properly installed, because NuGet installs it for us, so I have zero experience testing what happens when that's missing.

> Is it because the xUnit.net test discovery mechanism kicks in simply because of the reference to xunit.core, and then fails because it can't run (as explained above)?

Actually, it's the presence of `xunit.execution.*.dll` that triggers it, and then the absence of the desktop library causes the failure, but otherwise this analysis is correct. We don't know there are no tests until we go looking for them, and we can't look for them in VS without the desktop DLL.

bradwilson:
BTW, what versions of the runners are you using? I wouldn't expect the TD.NET problem to happen if you were using the 2.1 runner, even with the props file missing.

tiesmaster:
@bradwilson I stand corrected ;) It looked quite similar to the Bait and Switch technique, but indeed it looks like it, but it's not the same.

> 1. Why are we seeing problems specifically with TestDriven.Net, while the batch build (build-release.sh or BuildRelease.ps1) executes without error?

Because it simply doesn't include that assembly: [the build script](https://github.com/AutoFixture/AutoFixture/blob/master/BuildRelease.msbuild#L198) makes a list of test assemblies based on the convention that unit tests libraries should end with `Test`, so it won't end up in that list. But if you do include it (like below), then it will fail in the exact same way.

````
-              Include="Src\*Test\bin\$(Configuration)\*Test.dll"
+              Include="Src\*Test\bin\$(Configuration)\*Test.dll;Src/AutoFixture.xUnit.net2/bin/Release/Ploeh.AutoFixture.Xunit2.dll"
````

As for 2, what Brad Wilson said.

ploeh:
@bradwilson Thank you for chiming in.

> BTW, what versions of the runners are you using?

I'm not even sure how to answer that question, at least for TestDriven.Net... The currently installed runner NuGet package is *xunit.runner.visualstudio.2.0.0*, but I don't know if TestDriven.Net even uses it.

FWIW, I tried replacing *xunit.runner.visualstudio.2.0.0* with *xunit.runner.visualstudio.2.1.0*, but it doesn't make any difference...

Not that I expect you to troubleshoot the issue further, but if that solution isn't viable, would you consider the changes in the pull request (one added line of code) to be an appropriate fix?

bradwilson:
> I'm not even sure how to answer that question, at least for TestDriven.Net

In that case, it's whatever version of the `xunit.extensibility.core` package you're using, since it comes in that way.

> Not that I expect you to troubleshoot the issue further, but if that solution isn't viable, would you consider the changes in the pull request (one added line of code) to be an appropriate fix?

I'm not sure you're talking about here. Do you have some fix for xUnit.net that you want? Obviously, if something's broken, we're always happy to take PRs and/or issue reports.

tiesmaster:
@bradwilson He's talking about the fix for this project: [this one](https://github.com/AutoFixture/AutoFixture/pull/495/files#diff-08b4b4c6c8045b28ecb1b5479087a49fR3).

bradwilson:
As I originally stated, I have no idea why that line wasn't already there, because NuGet should've added it for you. It is absolutely 100% required.

ploeh:
@bradwilson Thank you for confirming that the fix in this PR is the appropriate way to resolve the issue.

> I have no idea why that line wasn't already there, because NuGet should've added it for you.

Can it be because that project depends exclusively on the NuGet package *xunit.extensibility.core*, and not on the 'big' *xunit* package?

ploeh:
@tiesmaster Thank you for helping resolving this issue, and please accept my apologies that it took so long :flushed:

tiesmaster:
No problem. Glad I could help ;) Happy new year! :tada:

moodmosaic:
:+1:

Happy New Year :tada:

bradwilson:
Any project which depends exclusively on `xunit.extensibility.core` is expected to be an extensibility project, not a unit test project. We have **absolutely** no intention to support someone who takes just `xunit.extensibility.core` as a dependency to be able to run tests.

ploeh:
@bradwilson That completely describes that library: it *is* an extensibility library, and it contains no tests.

PR #496, opened 2015-12-23
Remove more uses of NoSpecimen with Request

ploeh:
Another incremental step towards resolving #475.

For the *v4* branch only.


moodmosaic:
:+1: Looking good to me. The `build-release.sh` script runs fine locally.

PR #497, opened 2015-12-23
Remove some more usages of NoSpecimen with Request

moodmosaic:
Here's another incremental step towards #475.

It should be merged only with the v4 branch, if accepted.


ploeh:
Thank you :+1:

PR #498, opened 2015-12-27
Make AutoFixture.Xunit2 a bit more discoverable

moodmosaic:
This Pull Request attempts to	make AutoFixture.Xunit2 a bit more discoverable, by adding a note in the NuGet Package description of AutoFixture.Xunit, as suggested by @adamralph in #294.

Apart from this change, I'm not sure we can add any formatting, as [it looks like it's not currently supported by the .nuspec](http://stackoverflow.com/q/23370532/467754) file format.


adamralph:
:+1:

ploeh:
@moodmosaic Thank you.

Issue #476, opened 2015-11-26
From constrained set and min/max for common types


chaitanyagurrapu:
It seems that if I want to limit the length of strings the recommended way is to use a custom specimen builder as discussed here http://stackoverflow.com/questions/10125199/autofixture-configure-fixture-to-limit-string-generation-length

It would be nice if Autofixture came with some of this constraints out of the box for common types such as strings, ints, datetimes, etc. So that users don't have to write them themselves.

It would also be very convenient if there was a way to specify a constrained list (of IEnumerable) for a particular property. For example, in Australia bank accounts have a BSB number, which is basically a branch number of a particular bank.

```c#
var allowedBSBs = new [] { "061237", "061238", "061239"};
var newAccount = fixture.Build<Account>()
    .With(x => x.BSB, allowedBSBs);
```

or alternate, more expressive syntax

```c#
var allowedBSBs = new [] { "061237", "061238", "061239"};
var newAccount = fixture.Build<Account>()
    .With(x => x.BSB)
    .FromConstrainedSet(allowedBSBs);
```

ploeh:
AutoFixture was originally conceived as a TDD tool, and that's still a guiding principle behind its design. I don't mind if people use it for other purposes as well, but TDD is the main focus.

TDD is a feedback mechanism; as [GOOS](http://bit.ly/growingoos) explains, you should *listen to your tests*. AutoFixture tends to amplify this feedback mechanism, and that's one reason I find it valuable.

Whenever I encounter the need to slice and dice primitive values in various ways, I consider it a sign that I should pause and reflect. Often, the solution isn't to introduce special ways to slice and dice primitives, but instead to [refactor from primitives to Domain Objects](http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling). When you do that, not only does your domain model become much clearer, but it also [becomes easier to tweak AutoFixture in your desired direction](http://blog.ploeh.dk/2009/05/01/DealingWithConstrainedInput).

Instead of treating e.g. *BSB number* as a string, make it a [Value Object](http://martinfowler.com/bliki/ValueObject.html) and then configure AutoFixture to automatically generate values drawn from `{ "061237", "061238", "061239"}` whenever a request is made for a `BsbNumber` value.

chaitanyagurrapu:
[...]

PR #499, opened 2015-12-27
Added a ElementsBuilder generic class that draws items from a given collection

Pvlerick:
Addressing #476 

The constructor expects a collection that contains at least two element; each time a request is received a random element will be drawn from that collection and returned. It internally uses the RandomNumericSequenceGenerator to select at least each element once before it is reset.

### DIFF
--- a/Src/AutoFixture/ElementsBuilder.cs
+++ b/Src/AutoFixture/ElementsBuilder.cs
@@ -16,7 +16,7 @@ namespace Ploeh.AutoFixture
         private readonly RandomNumericSequenceGenerator sequence;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ElementsBuilder<T>"/> class.
+        /// Initializes a new instance of the <see cref="ElementsBuilder{T}"/> class.
         /// </summary>
         /// <param name="elements">The collection from which elements should be drawn from.
         /// It must contain at least two element.</param>
--- a/Src/AutoFixture/ElementsBuilder.cs
+++ b/Src/AutoFixture/ElementsBuilder.cs
@@ -18,7 +18,7 @@ namespace Ploeh.AutoFixture
         /// Initializes a new instance of the <see cref="ElementsBuilder{T}"/> class.
         /// </summary>
         /// <param name="elements">The collection from which elements should be drawn from.
-        /// It must contain at least two element.</param>
+        /// It must contain at least one element.</param>
         public ElementsBuilder(IEnumerable<T> elements)
         {
             if (elements == null)
--- a/Src/AutoFixture/ElementsBuilder.cs
+++ b/Src/AutoFixture/ElementsBuilder.cs
@@ -23,10 +23,10 @@ namespace Ploeh.AutoFixture
         public ElementsBuilder(ICollection<T> elements)
         {
             if (elements == null)
-                throw new ArgumentNullException(nameof(elements));
+                throw new ArgumentNullException("elements");
 
             if (elements.Count < 2)
-                throw new ArgumentException("The collection must contain at least two elements.", nameof(elements));
+                throw new ArgumentException("The collection must contain at least two elements.", "elements");
 
             this.elements = elements;
             this.sequence = new RandomNumericSequenceGenerator(0, elements.Count - 1);
--- a/Src/AutoFixture/ElementsBuilder.cs
+++ b/Src/AutoFixture/ElementsBuilder.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Ploeh.AutoFixture
 {
@@ -12,7 +11,7 @@ namespace Ploeh.AutoFixture
     /// <typeparam name="T"></typeparam>
     public sealed class ElementsBuilder<T> : ISpecimenBuilder
     {
-        private readonly ICollection<T> elements;
+        private readonly T[] elements;
         private readonly RandomNumericSequenceGenerator sequence;
 
         /// <summary>
@@ -20,16 +19,19 @@ namespace Ploeh.AutoFixture
         /// </summary>
         /// <param name="elements">The collection from which elements should be drawn from.
         /// It must contain at least two element.</param>
-        public ElementsBuilder(ICollection<T> elements)
+        public ElementsBuilder(IEnumerable<T> elements)
         {
             if (elements == null)
                 throw new ArgumentNullException("elements");
 
-            if (elements.Count < 2)
-                throw new ArgumentException("The collection must contain at least two elements.", "elements");
+            this.elements = elements.ToArray();
 
-            this.elements = elements;
-            this.sequence = new RandomNumericSequenceGenerator(0, elements.Count - 1);
+            if (this.elements.Length < 1)
+                throw new ArgumentException("elements must contain at least one element.", "elements");
+
+            //The RandomNumericSequenceGenerator is only created for collections of minimum 2 elements
+            if (this.elements.Length > 1)
+                this.sequence = new RandomNumericSequenceGenerator(0, this.elements.Length - 1);
         }
 
         /// <summary>
@@ -44,14 +46,19 @@ namespace Ploeh.AutoFixture
         public object Create(object request, ISpecimenContext context)
         {
             if (!typeof(T).Equals(request))
-                return new NoSpecimen();
+#pragma warning disable 618
+                return new NoSpecimen(request);
+#pragma warning restore 618
 
-            return this.elements.ElementAt(this.GetNext());
+            return this.elements[this.GetNextIndex()];
         }
 
-        private int GetNext()
+        private int GetNextIndex()
         {
-            return (int)this.sequence.Create(typeof(int));
+            if (this.elements.Length == 1)
+                return 0;
+            else
+                return (int)this.sequence.Create(typeof(int));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Pvlerick:
Side note: after struggling with the XML comments that made the build server unhappy, I noticed that now it passes while the build script doesn't because I used C# 6.0 ``nameof``.

Pvlerick:
I removed the ``nameof`` usages, I will open a separate issue for that.

Src/AutoFixture/ElementsBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        private readonly ICollection<T> elements;
+        private readonly RandomNumericSequenceGenerator sequence;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ElementsBuilder{T}"/> class.
+        /// </summary>
+        /// <param name="elements">The collection from which elements should be drawn from.
+        /// It must contain at least two element.</param>
+        public ElementsBuilder(ICollection<T> elements)
```
ploeh: What is the motivation for requiring `ICollection<T>`? Does `ElementsBuilder<T>` need to add, remove, or modify the elements?

Src/AutoFixture/ElementsBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (elements == null)
+                throw new ArgumentNullException("elements");
+
+            if (elements.Count < 2)
```
ploeh: Why is a collection with a single element not allowed?

Granted, a singleton collection would be degenerate, and result in deterministic behaviour, but it'd still work.

From a user experience perspective, I think it's quite clear than when you have a function that select a random element from a list, and that list only has one element, then the return value will always be that single element.

FWIW, [FsCheck's](http://fscheck.github.io/FsCheck) `Gen.elements` allows the use of a singleton list, as demonstrated by this passing property:

```F#
[<Property>]
let test (i : int) = Prop.forAll (Gen.elements [i] |> Arb.fromGen) ((=) i)
```

ploeh:
Thank you for submitting this contribution. It's looking good :+1: 

The few comments I had are hardly a big deal.

I can easily address them myself, but usually contributors prefer to polish off their own work, so I'll leave this PR open to give you the opportunity to do that (or dispute my comments).

Src/AutoFixture/ElementsBuilder.cs
Pvlerick: I choose ``ICollection<T>`` because the implementation needs to know the number of elements in the "collection". I considered using ``IEnumerable<T>`` instead, but the Count() extension method because it might trigger side effects.

Src/AutoFixture/ElementsBuilder.cs
Pvlerick: I was struggling with that as well; my initial unit tests were using a single element collection.

My idea was to internally use ``RandomNumericSequenceGenerator`` in order to return elements from the collection in a random order but still returning each element once before resetting.

Because of the use of ``RandomNumericSequenceGenerator`` which constructor needs at least an upper and a lower limit that are strictly ascending (see #408 and #409), these were my options:
* Make special case for one element collections, which is a bit ugly
* Force the collection to have at least 2 elements to have a simple implementation
* Stop using ``RandomNumericSequenceGenerator``, for example by using ``NumericSequenceGenerator`` instead but then element would always be returned in the same order - I kind of liked the idea that element would be returned at random :smile: 
* Use ``Random`` and give up on the idea of returning each element once

Any advice welcome, I made the compromise based solely on personal preferences

Src/AutoFixture/ElementsBuilder.cs
ploeh: I grant that calling `Count()` could have side-effects, if traversing the sequence produces side-effects. It shouldn't, but who knows what a user might do...

I'm happy taking on that risk, and essentially declaring that side-effecty sequences constitute a user error.

Using any sort of mutable collection comes with the risk of side-effects as well, in the sense that a user could pass in a `List<T>` and then subsequently modify the collection. Essentially, the 'not empty' constructor Guard Clause is no guarantee, because the user could empty the collection after the constructor runs:

```C#
var l = new List<int> { 1, 2, 3 };
var b = new ElementsBuilder<int>(l);
l.Clear();
``` 

FWIW, the [most appropriate type to use is `IReadOnlyCollection<T>`](http://blog.ploeh.dk/2013/07/20/linq-versus-the-lsp), but unfortunately, that's only available on .NET 4.5+.

Another option is to use an array, but arrays are also 'slightly mutable'.

In AutoFixture, the most idiomatic type to use would be `IEnumerable<T>`, despite the flawed guarantees it gives.

Src/AutoFixture/ElementsBuilder.cs
ploeh: I think that singleton collections should be possible as well, even if that requires a special case in the implementation. That is, after all, an implementation detail.

Whether or not each element should be returned exactly once before restarting the 'pool' is not something I feel strongly about. It *would* be in the spirit of how AutoFixture currently tends to do things, so I'd slightly lean towards that behaviour for the sake of consistency.

Src/AutoFixture/ElementsBuilder.cs
Pvlerick: ``IReadOnlyCollection<T>`` would be nice ideal, the type requirements being that has count and seek operations executing in O(1).

I'll modify the code to use ``IEnumerable<T>``.

Src/AutoFixture/ElementsBuilder.cs
Pvlerick: I'll add a special case for single element collections.

Src/AutoFixture/ElementsBuilder.cs
ploeh: :+1: 

(Please be sure to add a notification when you've created the update, since GitHub doesn't send me any emails when commits are added to a branch.)

Src/AutoFixture/ElementsBuilder.cs
ploeh: :+1:

Pvlerick:
@ploeh All done. Maybe I shall add an example in the [Cheat Sheet](https://github.com/AutoFixture/AutoFixture/wiki/Cheat-Sheet)?

Src/AutoFixture/ElementsBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="request">The request that describes what to create.</param>
+        /// <param name="context">Not used.</param>
+        /// <returns>
+        /// One of the element present in the collection given to the constructor if <paramref name="request"/>
+        /// is a request for <typeparamref name="T"/>; otherwise, a <see cref="NoSpecimen"/> instance.
+        /// </returns>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (!typeof(T).Equals(request))
+#pragma warning disable 618
```
ploeh: Instead of suppressing this particular warning, please follow it instructions.

Not that I don't want to help you by telling you explicitly what to do here, but I also consider this a test of the wording of that warning message. If you're in doubt about what to do, it means that the warning is poorly worded, and I need to rephrase it.

Src/AutoFixture/ElementsBuilder.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// It must contain at least one element.</param>
+        public ElementsBuilder(IEnumerable<T> elements)
+        {
+            if (elements == null)
+                throw new ArgumentNullException("elements");
+
+            this.elements = elements.ToArray();
+
+            if (this.elements.Length < 1)
+                throw new ArgumentException("elements must contain at least one element.", "elements");
```
ploeh: I'm almost embarrassed to mention this, because is such an incredibly insignificant issue, but I prefer that exception messages look like complete sentences. This one, I realise, *is* a complete sentence, because *elements* refer to the name of the argument, but due to the lower-case starting character, at first glance it *looks* incorrect.

Could we rephrase to something like the following?

> The supplied collection of elements must contain at least one element. This collection is expected to contain the elements from which the randomized algorithm will draw; if the collection is empty, there are no elements to draw.

If you don't care to make this change, I'm completely happy still accepting the pull request, but then I'll probably make the edit myself.

Pvlerick: No questions, good exception messages are what makes a good library :smile:

Src/AutoFixture/ElementsBuilder.cs
Pvlerick: I added the warning suppression in a subsequent commit (ed5c062), I initially used ``new NoSpecimen()``` which is apparently what it should be. I think the warning is fine as it is.

What got me confuses is the message associated with commit 5b459db6:

> For consistency's
sake, we ask contributors to adhere to the established standard of
returning 'new NoSpecimen(request)' even when everything works fine
without it. The design would be slightly simpler if Request was simply
omitted.

I'm not sure about the meaning of the second sentence: does it mean that I should use the old way or the new way :question:

ploeh: Yes, that comment described the way it had been, which then prompted me to make the change in that commit. Now that I re-read it, I admit that that statement is ambiguous.

Clarity in writing is *hard*.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.38.0](https://www.nuget.org/packages/AutoFixture/3.38.0).

PR #501, opened 2015-12-30
Provide an msbuild target for just running tests

adamchester:
Sometimes, you want to build and run the tests repeatedly from the command line (especially if, like me, you don't trust the built-in visual studio test runners). These new msbuild targets makes that easier, while still preserving the existing behaviour of `/t:Test`.

The new msbuild targets are:
 * `/t:TestOnly` - only runs the test commands, without attempting clean/rebuild//verify.
 * `/t:BuildAndTestOnly` - only runs the build and test targets, without attempting clean/rebuild/verify.


adamchester:
I had the wrong target name in this PR diff. Will open a new one.

PR #505, opened 2016-01-03
Added ctor that takes a list element using params

Pvlerick:
Added a constructor that takes ``elements`` with ``params``. It was a bit silly of me not to add it in the original PR (#499) since it makes creation easier.


ploeh:
Thank you for your contribution. It's now live as [AutoFixture 3.38.1](https://www.nuget.org/packages/AutoFixture/3.38.1).

PR #506, opened 2016-01-03
Use method group where possible

ErikSchierboom:
This PR converts to lambda expressions to method groups. Makes the code slightly easier to read IMHO.


ploeh:
Thank you for your contribution. It's now live as [AutoFixture 3.38.1](https://www.nuget.org/packages/AutoFixture/3.38.1).

Issue #500, opened 2015-12-28
Modify build scripts to allow the use of C# 6.0 features


Pvlerick:
As mentioned in PR #499, the current build scripts do not allow for C# 6.0 features, while VS 2015 and the build server are perfectly happy with them.

This does not happen when building with a more recent version of msbuild (such as the one that ships with VS 2015 in "C:\Program Files (x86)\MSBuild\14.0\Bin\MSBuild.exe")

Side note: with the new msbuild, 4 warnings are produced
``warning MSB3884: Could not find rule set file "AutoFixture.NUnit2.ruleset``
``warning MSB3884: Could not find rule set file "AutoFixture.NUnit2.ruleset``
``warning MSB3884: Could not find rule set file "AutoFixture.xUnit.net.ruleset``
``warning MSB3884: Could not find rule set file "Migrated rules for AutoRhinoMock.ruleset``

ploeh:
Thank you for bringing this to our attention.

First of all, just to make sure that I understand the current capabilities correctly, please correct me if the following is incorrect:

- AutoFixture 3 targets .NET 4.
- It's possible to use C# 6 to compile .NET 4 binaries.
- To use C# 6 one must have Visual Studio 2015. Earlier versions of Visual Studio can't compile C# 6 code.

I must admit that I'm not completely sure about all of these statements, which is the reason I'm listing them.

Pvlerick:
1 and 2 are ok.

For 3 AFAIK only Visual Studio 2015 supports C# 6.0's syntax indeed - including C# 6.0 in the code base would force everyone to move to Visual Studio 2015 (or other C# 6.0 supporting IDEs).

On a side note again, my understanding is that MSBuild can now be retrieved as a package (https://www.nuget.org/packages/MSBuild/), maybe that could be explored.

ploeh:
For my part, I have no problem with a Visual Studio 2015 requirement, but we ought to consider the wider community.

Again, I'd like to solicit comments for or against **making Visual Studio 2015 a requirement** for working with AutoFixture, particularly, but not exclusively, from @moodmosaic, @ecampidoglio, @adamchester.

adamchester:
:+1: to VS 2015, no reason I can think of to stick with 2013.

Also :+1: to putting those build warnings in a separate issue; I get them too. We might also need to upgrade the sln files to be all 2015 format, for consistency. In the past, appveyor used this to help decide on msbuild version but that may not be necessary now.

[...]

PR #507, opened 2016-01-04
Use C# 6 features (see #500)

ErikSchierboom:
The first version of my PR is done. It includes the following things:

- Argument strings are converted to `nameof` operator calls.
- Getter-only properties (essentially readonly properties) are converted to auto-implemented getter-only properties.
- Getter-only properties that could not be converted to auto-implemented properties are converted to expression-bodied members.
- Converted the `buildrelease.ps1` script to use MSBuild as located in the "Program files (x86)\MSBuild" directory, instead of the version in the "Windows\.NET Framework" directory.

To make it easy to see what happened in each step, this PR has several, separate commits.

What I did not do is convert methods to expression-bodies, as most people I know don't really like them. As an example, the following code:

```csharp
public int GetHashCode(Type obj)
{
    return 0;
}
```

could be converted to:

```csharp
public int GetHashCode(Type obj) => 0;
```

Is that something you'd like to see?

Furthermore, as written in the comments in #500, what should I do with the `buildrelease.sh` file? It refers to $WINDIR/Microsoft.NET/Framework/v4.0.30319/MSBuild.exe, so I suspect that it won't work with C# 6. To what should I change the call?

I will probably not be able to do anything internet-related the next week, so I think that is a nice time for everyone to consider if they like what I have done here :)

### DIFF
--- a/BuildRelease.ps1
+++ b/BuildRelease.ps1
@@ -10,4 +10,4 @@ function Get-VsVersion ()
     }
 }
 
-Start-Process -NoNewWindow -Wait -FilePath $env:windir\Microsoft.NET\Framework\v4.0.30319\MSBuild -ArgumentList ('BuildRelease.msbuild', (Get-VsVersion))
+Start-Process -NoNewWindow -Wait -FilePath ${env:ProgramFiles(x86)}\MSBuild\14.0\Bin\MsBuild.exe -ArgumentList ('BuildRelease.msbuild', (Get-VsVersion))
--- a/Src/AutoFixture/AutoPropertiesTarget.cs
+++ b/Src/AutoFixture/AutoPropertiesTarget.cs
@@ -38,7 +38,7 @@ namespace Ploeh.AutoFixture
         public AutoPropertiesTarget(ISpecimenBuilder builder)
         {
             if (builder == null)
-                throw new ArgumentNullException("builder");
+                throw new ArgumentNullException(nameof(builder));
 
             this.builder = builder;
         }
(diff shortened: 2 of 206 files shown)


ploeh:
Thank you for your interest in contributing to AutoFixture. Could you please split your pull request into several smaller pull requests?

When I attempt to look at the diff in order to review, GitHub gives me this message:

> Sorry, we could not display the entire diff because it was too big.

Since the pull request comes with a bullet list, would it be possible to split it up into several pull requests, one for each bullet?

Even more pull requests than 4 would be preferable, since 206 files / 4 pull requests still suggests more than 50 changed files per pull request.

Consider reading these [10 tips for better Pull Requests](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests).

For the above reasons, I'm going to close this pull request. Not because I don't want the contribution, but because I can't even tell if the contribution is something I would find appropriate for AutoFixture.

ErikSchierboom:
Okay, understood. I didn't think of that. First, I'll try to implement FAKE (#500), then I'll move on to the C# 6 features.

By the way, I was wondering why the packages folder was included in the repository, I almost never see people do that. Is there a specific reason for this?

adamchester:
@ErikSchierboom While I can't speak for @ploeh, I can link to his blog :smile: 

http://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful/

ploeh:
That is exactly the reason :+1:

PR #508, opened 2016-01-06
Port the reflection code to TypeInfo.

lbargaoanu:
First step for #404.

### DIFF
--- a/Src/AutoFixture/Dsl/NodeComposer.cs
+++ b/Src/AutoFixture/Dsl/NodeComposer.cs
@@ -412,7 +412,7 @@ namespace Ploeh.AutoFixture.Dsl
             Expression<Func<T, TProperty>> propertyPicker)
         {
             var m = propertyPicker.GetWritableMember().Member;
-            if (m.ReflectedType != typeof(T))
+            if (m.DeclaringType != typeof(T))
                 m = typeof(T).GetProperty(m.Name);
 
             return (NodeComposer<T>)this.ReplaceNodes(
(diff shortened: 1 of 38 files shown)


lbargaoanu:
Code analysis complains about unused members (in the extension methods class). It would be easier to let them in and only remove them when the port is over (if we really don't use them).

lbargaoanu:
This touches quite a few files, but the changes are trivial.

ploeh:
Thank you for your interest in contributing to AutoFixture. FTR, I'm currently travelling, and have only limited to time to evaluate pull requests. In due time, I'll work through my backlog, but it may take weeks :flushed: :sweat:

Src/AutoFixture/Dsl/NodeComposer.cs
```diff
@@ -411,7 +412,7 @@ public IPostprocessComposer<T> WithAutoProperties()
             Expression<Func<T, TProperty>> propertyPicker)
         {
             var m = propertyPicker.GetWritableMember().Member;
-            if (m.ReflectedType != typeof(T))
+            if (m.DeclaringType != typeof(T))
```
ploeh: What is the motivation for changing from `ReflectedType` to `DeclaringType`?

lbargaoanu: https://github.com/dotnet/coreclr/issues/926

ploeh:
Thank you for this pull request. As you can see, I had a single question, but apart from that, it's looking good :+1: 

Not only does it build on the CI server, but it also builds on my machine :wink:

Src/AutoFixture/Dsl/NodeComposer.cs
ploeh: Thank you; I didn't know that. Apparently, this change isn't an issue, since all tests still pass.

lbargaoanu: I thought so too :) Anyway, I can fix any issues that might appear, but the tests seem to cover a lot.

Issue #509, opened 2016-01-07
.NET Core & Serializable Exceptions


lbargaoanu:
In .NET exceptions need to be serializable in order to be able to cross AppDomain boundaries. Not so in .NET Core (no AppDomains). Do AutoFixture exceptions need to be passed across AppDomains? The testing infrastructure in VS uses AppDomains, but I'm not sure whether this affects AutoFixture. The simplest option is not to serialize exceptions. Another would be to solve it at compile time. Or at runtime with a shim.

ploeh:
> Do AutoFixture exceptions need to be passed across AppDomains?

We don't know. AutoFixture is a library. What people do with it is beyond our control. They *may* be using it in such a way that exceptions need to be serialised across AppDomains.

IIRC, the only way that affects the implementation of Exceptions is that they are adorned with the `[Serializable]` attribute, and that they have an extra protected constructor overload. Is that an impediment when it comes to .NET Core?

lbargaoanu:
There is no default binary serialization in .NET Core. So if the code is to remain unchanged we need to provide some replacement types just to compile.

ploeh:
Does that mean that the `[Serializable]` attribute doesn't exist?

Just trying to understand...

lbargaoanu:
Yes. And the types passed in the serialization constructor.

ploeh:
I see. Thank you for explaining that.

Changing the existing AutoFixture types by removing the offending code would be a breaking change, so isn't something we can do in AutoFixture 3.

In AutoFixture 4, we *could* make that breaking change, but would that be appropriate?

I haven't seen if the .NET team has issued any guidance on this question, but it'd be interesting to read if they have.

Assuming that 'normal' .NET will be around for a while yet, wouldn't it still be appropriate to release a version of AutoFixture that targets the 'old' framework? If so, wouldn't it still be appropriate that Exceptions are serializable?

If so, I don't see any other option than conditional compilation, but that may only be my lack of imagination speaking...

lbargaoanu:
I'm sure most custom exception types out there are not serializable and they get away with it :) But strictly speaking, yes, they should be serializable. We can solve it with conditional compilation. We can also add a reference only in the .NET Core configuration to a library containing the placeholder types. Partial classes and conditional compilation could be simple enough I believe.

ploeh:
OK. What should we do with this issue?

Is it a work item? Or did we arrive at an answer, so that we ought to close the issue?

[...]

PR #511, opened 2016-01-12
Exceptions don't need to be serializable in .NET Core because there a…

lbargaoanu:
…re no AppDomains
@ploeh If you're still not [convinced](https://github.com/AutoFixture/AutoFixture/issues/509#issuecomment-170224097), I'll add the placeholder types for .NET Core.

### DIFF
--- a/Src/AutoFixture/Kernel/IllegalRequestException.cs
+++ b/Src/AutoFixture/Kernel/IllegalRequestException.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Runtime.Serialization;
 
 namespace Ploeh.AutoFixture.Kernel
 {
@@ -13,7 +12,6 @@ namespace Ploeh.AutoFixture.Kernel
     /// </para>
     /// </remarks>
     /// <seealso cref="IntPtrGuard"/>
-    [Serializable]
     public class IllegalRequestException : Exception
     {
         /// <summary>
@@ -43,28 +41,6 @@ namespace Ploeh.AutoFixture.Kernel
         /// <param name="innerException">The inner exception.</param>
         public IllegalRequestException(string message, Exception innerException)
             : base(message, innerException)
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="IllegalRequestException"/> class.
-        /// </summary>
-        /// <param name="info">
-        /// The <see cref="T:System.Runtime.Serialization.SerializationInfo"/> that holds the
-        /// serialized object data about the exception being thrown.
-        /// </param>
-        /// <param name="context">
-        /// The <see cref="T:System.Runtime.Serialization.StreamingContext"/> that contains
-        /// contextual information about the source or destination.
-        /// </param>
-        /// <exception cref="T:System.ArgumentNullException">
-        /// The <paramref name="info"/> parameter is null.
-        /// </exception>
-        /// <exception cref="T:System.Runtime.Serialization.SerializationException">
-        /// The class name is null or <see cref="P:System.Exception.HResult"/> is zero (0).
-        /// </exception>
-        protected IllegalRequestException(SerializationInfo info, StreamingContext context)
-            : base(info, context)
         {
         }
     }
--- a/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
+++ b/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
@@ -2,7 +2,6 @@ using System;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.Serialization;
 
 namespace Ploeh.AutoFixture.Kernel
 {
@@ -10,7 +9,6 @@ namespace Ploeh.AutoFixture.Kernel
     /// The exception that is thrown when AutoFixture is unable to infer the type 
     /// parameters of a generic method from its arguments.
     /// </summary>
-    [Serializable]
     public class TypeArgumentsCannotBeInferredException : Exception
     {
         /// <summary>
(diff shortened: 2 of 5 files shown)


ploeh:
Thank you for your interest in contributing to AutoFixture. FTR, I'm currently travelling, and have only limited to time to evaluate pull requests. In due time, I'll work through my backlog, but it may take weeks :flushed: :sweat:

ploeh:
Thank you for your interest in contributing to AutoFixture.

While it may be true that there are no AppDomains in .NET Core, they are still present in the 'full' .NET framework, and I don't see that going away in the near future.

For people who want to use AutoFixture with the existing .NET framework(s), it could be a problem if we remove the `[Serializiable]` attribute from Exceptions. How do we resolve that issue?

lukasz-pyrzyk:
> How do we resolve that issue?

Probably we can use if statement:

    #if DNX451
        [Serializable]
    #endif
    public class IllegalRequestException : Exception


Not tested.

lbargaoanu:
I understand you've read the MS position and you disagree. Then I guess you can close this PR and I'll add the placeholder types for .NET Core in another PR.

ploeh:
Which MS position is that?

(Sorry if I'm being dense here, but I'm currently in an airport, roaming on my phone, so it's a bit difficult for me to master the entire context.)

ploeh:
And now, after having posted the previous message, I see that you have provided a link... Please disregard my previous comment.

ploeh:
I need help in order to understand the current status of .NET Core. As [others have pointed out](http://weblog.west-wind.com/posts/2016/Jan/21/Microsoft-renames-ASPNET-5-to-ASPNET-Core-10), the messaging around this has been confusing. While I'm *fairly* sure that I understand that Roslyn is a compiler, I no longer understand if *Core CLR* and *.NET Core* is the same thing. I also have absolutely no clue as to the status of it, but I can see that ASP.NET Core is some months away yet.

What can we expect once .NET Core 1.0 see the light of day? Will it replace .NET, or will it exist side-by-side with the 'full' framework?

When not preparing presentations or courses on F#, most of my energy currently goes into trying to learn Haskell, so I haven't been able to follow the .NET Core situation closely.

It'd be nice with a status summary, and how it applies to AutoFixture.

I must admit that I don't (yet) find [the linked discussion](https://github.com/dotnet/roslyn/issues/3528#issuecomment-116827732) particularly illuminating. It's essentially a semi-random discussion where an issue gets closed with a curt remark that this is no longer relevant. How does a discussion about Roslyn (a compiler) even relate to .NET Core? I'm sure there's a connection, but I need help understanding what it is.

I do understand that the trend points towards AppDomains being increasingly irrelevant, but I also need to consider existing clients and users of AutoFixture. One of the difficult things about publishing and maintaining a public library like AutoFixture is that I have little insight into how people use it...

lbargaoanu:
The original [link] (https://github.com/AutoFixture/AutoFixture/issues/509#issuecomment-170224097).
Going forward exceptions are not going to be serializable (someone said in the issue that the guidance will be changed). It doesn't matter what framework, .NET Core or the full framework. I chose two examples, EF7 and Roslyn, of projects that will support both frameworks and will not use serializable exceptions. The reason is that they are trying to stay away from conditional compilation because it can get messy. They run one binary on all frameworks. .NET Core is RC, so probably a few months away as you said.

lbargaoanu:
The second [link](https://github.com/dotnet/roslyn/issues/3528#issuecomment-116827732) was to show how you can workaround the fact that exceptions are not serializable anymore. I guess they consider it's a corner case and the extra work is doable. Anyway, it's just as easy to add placeholder types for .NET Core :) The real problem is that we introduce conditional compilation...

lbargaoanu:
The strategy is to wait? :)

ploeh:
Perhaps I have things mixed up. What little knowledge I have of .NET Core is the noise regarding ASP.NET Core that filter my way. Perhaps my view of this work is tainted by that. Perhaps .NET Core is only superficially related to ASP.NET Core, in which case I look forward to become illuminated.

Based on the public information regarding progress of ASP.NET Core, however, I see no evidence that this project is stable. As long as it's unstable, I don't wish to start thrashing the AutoFixture source code, in order to try to keep it in sync.

If .NET Core, as opposed to ASP.NET Core, shows sign of stability, I will happily accept being taken to school on the topic. This is also the reason I previously [asked the questions I did](https://github.com/AutoFixture/AutoFixture/pull/511#issuecomment-173909621).

lbargaoanu:
AutoFixture depends only on the runtime and the BCL, that's what .NET Core is. ASP.NET is higher in the stack. The changes seem to be more about naming then anything. RC means the API is frozen, RTM is [Q1 2016](https://github.com/dotnet/core/blob/master/roadmap.md) Also versioning is not all or nothing. You depend on a number of packages and you choose what version you want for each package. Any question I missed? :)
About the issue at hand. This seems clear enough. I don't see how that could change in the near future.
About the [MailAddress issue](https://github.com/AutoFixture/AutoFixture/issues/404#issuecomment-171943540). That might be implemented eventually...
It doesn't look like anything serious to me. It's a little tricky to support both frameworks but that will always be the case. Many projects did it already (nUnit, xUnit, AutoMapper, Json.NET). I guess waiting is also a plan, but I doubt that in six months time the conversation will be much different.

ploeh:
> RC means the API is frozen

It's possible that I'm misinformed, but everything I hear about this is that all the beta releases were really alphas, and that the Release Candidate isn't a release candidate in any usual sense of the term.

As far as I understand, alone the decision to rename everything to .NET Core has triggered uniform renaming of namespaces.

Is there any official documentation that states that .NET Core really, unconditionally will have no breaking changes from now until it actually releases? And even so, can we trust it, in light of the project's history?

lbargaoanu:
I don't think the real world works that way :) I guess if you don't trust it, you just don't.

ploeh:
[I'm not alone](https://twitter.com/adamralph/status/697791925668200449).

lbargaoanu:
I would guess that's because people actually use it on .NET Core. And that makes sense.

PR #512, opened 2016-01-13
Add support for creating encodings

### LINKED ISSUE #490: Creating MailMessage doesn't work
Similar to #441, creating a `MailMessage` doesn't work without customizing an instance of AutoFixture, as shown by this test:

````csharp
    [Fact]
    public void FixtureCanCreateMailMessage()
    {
        var fixture = new Fixture();
        var actual = fixture.Create<MailMessage>();
        Assert.NotNull(actual);
    }
````

Which [throws `TargetInvocationException`](https://gist.github.com/tiesmaster/491c9cd706fc536d80ff).

This is because the encoding properties (`BodyEncoding`, `HeadersEncoding`, and `SubjectEncoding`) require an `System.Text.Encoding` instance, and it appears that AutoFixture is not able to create those as well.

Injecting something like `Encoding.UTF8` solves the problem. Maybe the same approach could be taken like #427, by adding a `Utf8EncodingGenerator`.

### PR DESCRIPTION
**Pull request by tiesmaster**
This adds a new `Utf8EncodingGenerator` class that works similar to the `InvariantCultureGenerator`: this will always return `Encoding.UTF8`. This makes it possible to create `MailMessage`s, out of the box, and therefore fixes #490 (though, I didn't add an unit test for creating them, since this PR is about making sure types of `Encoding` can be created, and that it also enables the creation `MailMessage`s to be fixed is nice to have).

I choose UTF8, since that is the most popular encoding these days, so made sense to me. I also considered the fact that this might break stuff for people, however, after some though, I realized that's not really the case. It will make it possible for people that have classes with `Encoding` dependencies to be created without customizations. Before, it either didn't work (the unit test would throw), or they would have customized the fixture instance, and those come first, so those customizations won't be overriden. Therefore, doesn't break anything.

@ploeh I saw you put in a couple issues and PRs that you're travelling, so I don't expect feedback real quickly. That's ok, I'll be on holiday to Iceland starting this Friday, so I might not be able to respond that quickly ;)

/cc @Pvlerick

### DIFF
--- a/Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
+++ b/Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
@@ -1,0 +1,68 @@
+﻿using Ploeh.AutoFixture;
+using Ploeh.AutoFixture.Kernel;
+using Ploeh.AutoFixtureUnitTest.Kernel;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Ploeh.AutoFixtureUnitTest
+{
+    public class Utf8EncodingGeneratorTest
+    {
+        [Fact]
+        public void SutIsSpecimenBuilder()
+        {
+            var sut = new Utf8EncodingGenerator();
+            Assert.IsAssignableFrom<ISpecimenBuilder>(sut);
+        }
+
+        [Fact]
+        public void CreateWithEncodingRequestWillReturnUtf8Encoding()
+        {
+            // Arrange
+            var sut = new Utf8EncodingGenerator();
+
+            // Act
+            var result = sut.Create(typeof(Encoding), new DelegatingSpecimenContext());
+
+            // Assert
+            Assert.Equal(Encoding.UTF8, result);
+        }
+
+        [Fact]
+        public void CreateWithNullRequestWillReturnNoSpecimen()
+        {
+            // Arrange
+            var sut = new Utf8EncodingGenerator();
+
+            // Act
+            var result = sut.Create(null, new DelegatingSpecimenContext());
+
+            // Assert
+            Assert.Equal(new NoSpecimen(), result);
+        }
+
+        [Fact]
+        public void CreateWithNullContextDoesNotThrow()
+        {
+            // Arrange
+            var sut = new Utf8EncodingGenerator();
+
+            // Act
+            Assert.DoesNotThrow(() => sut.Create(typeof(object), null));
+        }
+
+        [Fact]
+        public void CreateWithNonTypeRequestWillReturnNoSpecimen()
+        {
+            // Arrange
+            var sut = new Utf8EncodingGenerator();
+
+            // Act
+            var result = sut.Create(new object(), new DelegatingSpecimenContext());
+            // Assert
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by ploeh** [comment]

**Comment by Pvlerick** [comment]
I never noticed that there was such an exceptional case for ``CultureInfo`` via ``InvariantCultureGenerator``.

I guess that paves the way to add support in the same way for ``IPAddress`` and ``IPEndPoint`` :smile:

File: Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+
+        [Fact]
+        public void CreateWithNonTypeRequestWillReturnNoSpecimen()
+        {
+            // Arrange
+            var sut = new Utf8EncodingGenerator();
+
+            // Act
+            var result = sut.Create(new object(), new DelegatingSpecimenContext());
+            // Assert
```
ploeh: Should there be an assertion here?

Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            // Act
+            var result = sut.Create(null, new DelegatingSpecimenContext());
+
+            // Assert
+            Assert.Equal(new NoSpecimen(), result);
+        }
+
+        [Fact]
+        public void CreateWithNullContextDoesNotThrow()
```
ploeh: Is this an important specification of this class, or is only (incidentally) how it happens to behave at the moment?

What would happen to this test case if someone needs to make a change to the class in the future, and needs the context?

Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
tiesmaster: Indeed, that was missing. Added that.

Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
tiesmaster: I was already wondering how to handle something like that. I have seen tests to verify that, but indeed it's not really a requirement, but just a relaxed requirement that might change in the future.

Of course, when you start using that parameter, then you would add a test to also verify that it **does** throw when you give it `null`. So in that situation, you would just modify the test to meet its new requirement.

Would you say that that is a violation of the open/closed principle? I always wonder in what extend something like the OCP applies. Isn't changing the class that it uses the parameter already a violation of OCP, or is that stretching it?

Anyway, I've removed the test case, since I agree that it's not necessary for this class.

tiesmaster:
@ploeh I've updated the code. Let me know if you see anything else.

Src/AutoFixtureUnitTest/Utf8EncodingGeneratorTest.cs
ploeh: > Would you say that that is a violation of the open/closed principle?

You could argue that. In the case of AutoFixture (and, I suppose, many other open source libraries), the reason is more practical, though. The tests constitute the public contract that we (the maintainers) swear to uphold.

As a general rule, as long as one can fix a bug or introduce a new feature without breaking any of the existing tests, we regard that as a non-breaking change. If we need to change tests, we regard it as a breaking change.

This is important because AutoFixture explicitly follows semantic versioning.

This would all be perfect if it was true, but alas it isn't. There are some tests (particularly `Ploeh.AutoFixtureUnitTest.DefaultPrimitiveBuildersTest.SutHasCorrectContents` - we really ought to delete that test) that tend to change on a regular basis, but at least these are well-known exceptions to the rule.

For that reason, if I can predict in advance that a test might be one that we might want to change in the future, that's a sign to me that perhaps it only tests an implementation detail. If that's the case, I'd prefer to not have it in the first place.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.39.0](https://www.nuget.org/packages/AutoFixture/3.39.0).

PR #513, opened 2016-01-15
Replace Type.GetTypeCode

lbargaoanu:
The Type extension method GetTypeCode trades an allocation for simpler code.


Src/AutoFixture/RangedNumberGenerator.cs
```diff
@@ -149,7 +149,7 @@ private void CreateAnonymous(RangedNumberRequest range, IComparable value)
 
         private static object Add(object a, object b)
         {
-            switch (Type.GetTypeCode(a.GetType()))
+            switch (TypeExtensions.GetTypeCode(a))
```
adamchester: Is this a mistake?

lbargaoanu: No :) Why?

Src/AutoFixture/Kernel/TypeExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static TypeCode GetTypeCode(this Type type)
+        {
+            if(!type.IsValueType)
+            {
+                return TypeCode.Object;
+            }
+            return GetTypeCode(Activator.CreateInstance(type));
+        }
+
+        public static TypeCode GetTypeCode(object request)
```
adamchester: Why is there a non-extension method in a class that appears to be named for the extensions?

lbargaoanu: It's just a static class. I could change the name I suppose.

Src/AutoFixture/Kernel/TypeExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    internal static class TypeExtensions
+    {
+        public static TypeCode GetTypeCode(this Type type)
+        {
+            if(!type.IsValueType)
+            {
+                return TypeCode.Object;
+            }
+            return GetTypeCode(Activator.CreateInstance(type));
```
adamchester: I realise this is in an internal class, but why is it safe to assume the type can be created like this?

lbargaoanu: Because it's a value type.

adamchester: :+1:

ploeh:

ploeh:
What is the purpose of this pull request?

lbargaoanu:
Type.GetTypeCode is missing in .NET Core.

ploeh:
> Type.GetTypeCode is missing in .NET Core.

Thank you for explaining that to me.

I understand that we need to work around the limitations in .NET Core, but the resulting code is a bit... strange. Here I'm mainly thinking about creating an instance of a value from its `Type` in order to get its `TypeCode`. It may be necessary, but seems unwarranted, if a reader isn't aware of the context.

At the moment, we aren't compiling AutoFixture against .NET core, so given the current pull request, there's no mechanism in place to ensure that the code stays that way. If this truly is the right way of getting the `TypeCode` from a `Type` value on .NET Core, shouldn't we protect it from being refactored back to how it was?

lbargaoanu:
I thought the idea was to split in smaller PR-s :) Of course they don't make sense in pieces. I'm close enough to building. [This](https://github.com/AutoFixture/AutoFixture/issues/404#issuecomment-171943540) is blocking me. But after this one, I can make another PR with the project.json and that will allow building for .NET Core.

ploeh:
Fair enough.

I wonder why we even have the need for testing on `TypeCode` in the code base in the first place. Given that we start out with a `Type` instance, I think it ought to be possible to test whether these values are equal to `typeof(int)`, `typeof(byte)`, `typeof(double)`, etc.

I dug through the commit history for each of the files that switch on `TypeCode`, and the original authors of the files in question are @moodmosaic, @ecampidoglio, and @dlongest. Perhaps they can shed some light on this.

lbargaoanu:
It is possible, but it seemed too big of a change and the code would look less compact as a result (if we were to replace the switch with ifs).

moodmosaic:
It looks like Type.GetTypeCode is [currently used in a couple of places](https://github.com/AutoFixture/AutoFixture/search?utf8=%E2%9C%93&q=Type.GetTypeCode) in the source code. :confused:

#518 proposes an alternative of doing that.

ploeh:
> but it seemed too big of a change

Could it be done in multiple pull requests?

Apart from that, instead of a `switch` or a list of `if` statements, couldn't some of these code blocks be refactored to use [`Convert.ChangeType`](https://msdn.microsoft.com/en-us/library/dtb69x08)?

lbargaoanu:
I've tried that and got some overflow exceptions. There is no TryConvert and the performance suffered. Also I'm not sure whether it is supported on .NET Core. It was weird enough that I gave it up, but if somebody can do it, the code should look much better.

adamchester:
https://github.com/dotnet/corefx/issues/5381

lbargaoanu:
It's RC, they won't change public APIs. Maybe in the next version...

adamchester:
It's marked for the rc2 milestone currently.

lbargaoanu:
I see a comment "Can't be added to TypeInfo in the current contract version".

lbargaoanu:
I gave ChangeType another try and it works. But it's tricky :) Maybe it can be improved... If we reach an acceptable solution, I presume all the switches could be replaced.

lbargaoanu:
I've compiled with checked and I get the same errors as with ChangeType. The generated values are out of range. And this makes sense, because the chosen values don't seem to depend upon the requested type.

moodmosaic:
Fair enough. Though, how is that related to replacement of `Type.GetTypeCode`? Couldn't be also addressed separately?

lbargaoanu:
I've stumbled into it trying to replace the switches for TypeCode with Convert.ChangeType, that's the connection.

moodmosaic:
>I've stumbled into it trying to replace the switches for TypeCode with Convert.ChangeType, that's the connection.

Good finding :+1:

Src/AutoFixture/NumericSequenceGenerator.cs
```diff
@@ -31,34 +32,18 @@ public object Create(object request, ISpecimenContext context)
...
-                    return (float)this.GetNextNumber();
-                case TypeCode.UInt16:
-                    return (ushort)this.GetNextNumber();
-                case TypeCode.UInt32:
-                    return (uint)this.GetNextNumber();
-                case TypeCode.UInt64:
-                    return (ulong)this.GetNextNumber();
-                default:
-                    return new NoSpecimen();
+                return Convert.ChangeType(0, request, CultureInfo.InvariantCulture);
```
ploeh: I do realise that it was my suggestion to investigate whether `Convert.ChangeType` is an option, but this implementation changes the behaviour of the algorithm.

When I suggested looking into using `Convert.ChangeType`, I wasn't aware that it'd throw an exception if the input value was outside the acceptable range for that type.

That's not what happened before, because the explicit cast would simply overflow the value. As an example, for `byte` values, that means that when we've exhausted [0..255], the algorithm simply loops back and starts over. As an example, `(byte)257` returns `1`. This is exactly the behaviour we want.

With the proposed `try/catch` behaviour, what instead happens is that once the maximum for a type is reached, all further numbers are all `0`.

This FSI session demonstrates this:

```F#
> let nsg = NumericSequenceGenerator ();;
val nsg : NumericSequenceGenerator

> [1..300] |> List.map (fun _ -> nsg.Create(typeof<byte>, null)) |> List.skip 250;;
val it : obj list =
  [251uy; 252uy; 253uy; 254uy; 255uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy]
```

This isn't the behaviour we want. Perhaps we should have a test case that covers that scenario.

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -104,54 +105,18 @@ private static void ValidateThatLimitsAreStrictlyAscending(long[] limits)
...
+                return new NoSpecimen();
+            }
+            var nextRandom = GetNextRandom();
+            try
+            {
+                return Convert.ChangeType(nextRandom, request, CultureInfo.InvariantCulture);
+            }
+            catch(OverflowException)
+            {
+                return Convert.ChangeType(0, request, CultureInfo.InvariantCulture);
```
ploeh: See [the above comment](#r51185435).

The only difference is that, for reasons I haven't investigated, it looks like the maximum is reached sooner than 256 iterations:

```F#
> let rnsg = RandomNumericSequenceGenerator ();;
val rnsg : RandomNumericSequenceGenerator

> [1..300] |> List.map (fun _ -> rnsg.Create(typeof<byte>, null)) |> List.skip 100;;
val it : obj list =
  [43uy; 83uy; 29uy; 94uy; 5uy; 18uy; 65uy; 91uy; 88uy; 22uy; 61uy; 31uy; 7uy;
   55uy; 67uy; 97uy; 75uy; 36uy; 76uy; 89uy; 50uy; 92uy; 53uy; 26uy; 44uy;
   101uy; 45uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy; 0uy;
   0uy; ...]
```

ploeh: > for reasons I haven't investigated

Oh, I see now that the first range point was changed to `SByte.MaxValue`, which is 127, so that explains that...

Src/AutoFixture/RandomNumericSequenceGenerator.cs
```diff
@@ -23,7 +24,7 @@ public class RandomNumericSequenceGenerator : ISpecimenBuilder
         /// with the default limits, 255, 32767, and 2147483647.
         /// </summary>
         public RandomNumericSequenceGenerator()
-            : this(1, Byte.MaxValue, Int16.MaxValue, Int32.MaxValue)
+            : this(1, SByte.MaxValue, Int16.MaxValue, Int32.MaxValue)
```
ploeh: Why change to `SByte`?

ploeh:
Thank you for looking into this issue, including investigating whether `Convert.ChangeType` is a viable option. It looks like this isn't the case.

Since what remains of this pull request is mainly an effort to use `Convert.ChangeType`, I'm going to close it. It may be possible to further refactor the code in this pull request so that it correctly solves the problem, but that's going to lead to an even longer commit history, and I don't see the value in that.

I still think the problem we're attempting to address here is relevant, so I'd still be interested in a pull request, but I think I'd be cleaner and easier to start again, with a new branch off *master*.

I understand that it's never fun to have a pull request declined (I've tried it myself more than once), so it's important for me to underscore that I *do* appreciate the time and effort that has gone into this. I don't think that time and effort has been wasted, because you've helped me understand the problem much better, and I think it's now much clearer what the appropriate solution is: `if request == typeof(byte)...else if...else if...else`.

I welcome a new pull request in that direction, should you still be interested.

PR #514, opened 2016-01-15
We don't need ArrayList

lbargaoanu:


ploeh:

ploeh:
What is the purpose of this change?

lbargaoanu:
ArrayList is an old class :) In the full framework is resides in mscorlib, so it doesn't really matter. In .NET Core it lives in a separate assembly offered for backward compatibility. It's used only in this place in the codebase, so we would be loading that assembly just for this usage.

ploeh:
Thank you for the explanation.

Given that this pull request adds more lines of code, someone might pass by in the future and think: *This is too complicated. I'll refactor it to use ArrayList!*

How do we prevent that from happening?

lbargaoanu:
First, I doubt someone would say that :) Second, we would build for at least .NET Core and the full framework. The .NET Core configuration wouldn't build at first, so I'm hoping that would be enough deterrent. Also, the old implementation allocates the ArrayList just to use that method. That's wasteful. This version allocates just the result.

lbargaoanu:
An extension method might help:

    Array ToArray(this IEnumerable enumerable, Type elementType)

ploeh:
How does an extension method help?

lbargaoanu:
The code reads cleaner. You wouldn't think twice about it, check what ToArray is that, is it in the framework or not, how it's implemented.

Src/AutoFixture/Kernel/ArrayRelay.cs
```diff
@@ -48,17 +49,21 @@ public object Create(object request, ISpecimenContext context)
...
-            return ArrayRelay.ToArray(elements, elementType);
+            return ToArray(elements, elementType);
         }
 
-        private static object ToArray(IEnumerable e, Type elementType)
+        private static object ToArray(IEnumerable elements, Type elementType)
         {
-            var al = new ArrayList();
-            foreach (var element in e)
+            var collection = elements as ICollection;
```
ploeh: What's the motivation for this cast attempt?

lbargaoanu: To optimize getting the Count for collections, arrays, lists etc. O(1) vs O(n) in the general case.

ploeh: `Enumerable.Count` [already does that.](http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,1191)

lbargaoanu: Yes, but it is that doing in O(n), by counting one by one :)

lbargaoanu: What I meant is that when I do the Cast<object>, the enumerable is no longer an ICollection, unless I'm missing something.

adamchester: :+1: I think it's worth the extra line or two here to use the cast and `Count` property.

ploeh: You may be right. Let's leave it in :+1:

ploeh:

Please accept my apologies for being difficult; apparently, I'm dense these days :flushed:

lbargaoanu:
I think I might have something to do with that too :) No worries.

Issue #515, opened 2016-01-18
Idioms.ConstructorInitializedMemberAssertion: not working for enums


Tungsten78:
First off, this is a fantastic library.  

I've been just getting my feet wet and have run across some odd behavior regarding enum type arguments.

Here are the failed tests:
```        
        [Fact]
        public void VerifyIllBehavedEnumConstructorArgumentsAreExposedAsFieldsDoesThrow()
        {
            // Fixture setup
            var dummyComposer = new Fixture();
            var sut = new ConstructorInitializedMemberAssertion(dummyComposer);
            var type = typeof(ReadOnlyFieldIncorrectlyInitializedViaConstructor<TestEnum>);

            // Exercise system and verify outcome
            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(type));
            // Teardown
        }

        [Fact]
        public void VerifyIllBehavedEnumConstructorArgumentsAreExposedAsPropertyDoesThrow()
        {
            // Fixture setup
            var dummyComposer = new Fixture();
            var sut = new ConstructorInitializedMemberAssertion(dummyComposer);
            var type = typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestEnum>);

            // Exercise system and verify outcome
            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(type));
            // Teardown
        }

        class ReadOnlyFieldIncorrectlyInitializedViaConstructor<T>
        {
            public readonly T Field;

            public ReadOnlyFieldIncorrectlyInitializedViaConstructor(T field)
            {
            }
        }

        class ReadOnlyPropertyIncorrectlyInitializedViaConstructor<T>
        {
            public ReadOnlyPropertyIncorrectlyInitializedViaConstructor(T property)
            {
            }

            public T Property { get; private set; }
        }
```

Tungsten78:
I've narrowed this issue to be due to the EnumGenerator generating a value that matches the default for the enum.  If the parameter were in any other position other than the first the result is as expected.

Would anyone object to the RoundRobinEnumEnumerable starting at the 2nd element?
It fixes the problem, but results in Enum tests to fail since the change is to the expectation.

```
        private class RoundRobinEnumEnumerable : IEnumerable
        {
            private readonly IEnumerable<object> values;

            internal RoundRobinEnumEnumerable(Type enumType)
            {
                if (enumType == null)
                {
                    throw new ArgumentNullException("enumType");
                }

[...]

PR #516, opened 2016-01-19
Issue #515 - Idioms.ConstructorInitializedMemberAssertion - fixing enum handling

Tungsten78:
The root cause is that CreateAnonymous can return the default enum value when setting up the specimen.  This leads to a false positive when asserting that the ctor value vs proptery/field are equal.

The approach is two fold:
* BuildSpecimenFromConstructor() skips the default value for enums
* In the event we're verifying an enum type with a single element associated to 0 (default value), ConstructorInitializedMemberException is always thrown with custom message due to our inability to discern whether initialization actually occurred

### DIFF
--- a/Src/Idioms/ConstructorInitializedMemberAssertion.cs
+++ b/Src/Idioms/ConstructorInitializedMemberAssertion.cs
@@ -180,7 +180,7 @@ namespace Ploeh.AutoFixture.Idioms
                 throw new ArgumentNullException("propertyInfo");
 
             var matchingConstructors = GetConstructorsWithInitializerForMember(propertyInfo).ToArray();
-            
+
             if (!matchingConstructors.Any())
             {
                 if (IsMemberThatRequiresConstructorInitialization(propertyInfo))
@@ -192,6 +192,11 @@ namespace Ploeh.AutoFixture.Idioms
                 // For writable properties or fields, having no constructor parameter that initializes
                 // the member is perfectly fine.
                 return;
+            }
+
+            if (CouldLeadToFalsePositive(propertyInfo.PropertyType))
+            {
+                throw BuildExceptionDueToPotentialFalsePositive(propertyInfo);
             }
 
             var expectedAndActuals = matchingConstructors
@@ -240,6 +245,11 @@ namespace Ploeh.AutoFixture.Idioms
                 return;
             }
 
+            if (CouldLeadToFalsePositive(fieldInfo.FieldType))
+            {
+                throw BuildExceptionDueToPotentialFalsePositive(fieldInfo);
+            }
+
             var expectedAndActuals = matchingConstructors
                 .Select(ctor => BuildSpecimenFromConstructor(ctor, fieldInfo));
 
@@ -250,11 +260,55 @@ namespace Ploeh.AutoFixture.Idioms
             }
         }
 
+        private static bool CouldLeadToFalsePositive(Type type)
+        {
+            if (type.IsEnum)
+            {
+                //Check for a default value only enum
+                var values = Enum.GetValues(type);
+                return values.Length == 1 && (int)values.GetValue(0) == default(int);
+            }
+
+            return false;
+        }
+
+        private static ConstructorInitializedMemberException BuildExceptionDueToPotentialFalsePositive(MemberInfo propertyOrField)
+        {
+            var message = string.Format(CultureInfo.CurrentCulture, "Unable to properly detect a successful initialization due to {0} being of type enum having a single default value.{3}Declaring type: {1}{3}Reflected type: {2}{3}", propertyOrField.Name, propertyOrField.DeclaringType.AssemblyQualifiedName, propertyOrField.ReflectedType.AssemblyQualifiedName, Environment.NewLine);
+
+            var field = propertyOrField as FieldInfo;
+            var property = propertyOrField as PropertyInfo;
+            if (field != null)
+            {
+                return new ConstructorInitializedMemberException(field, message);
+            }
+            else if (property != null)
+            {
+                return new ConstructorInitializedMemberException(property, message);
+            }
+            else
+            {
+                throw new ArgumentException("Must be a property or field", "propertyOrField");
+            }
+        }
+
         private ExpectedAndActual BuildSpecimenFromConstructor(
             ConstructorInfo ci, MemberInfo propertyOrField)
         {
             var parametersAndValues = ci.GetParameters()
-                .Select(pi => new {Parameter = pi, Value = this.builder.CreateAnonymous(pi)})
+                .Select(pi =>
+                {
+                    var value = this.builder.CreateAnonymous(pi);
+                    
+                    //ensure enum isn't getting the default value, otherwise we won't be able to determine whether initialization occurred
+                    if (pi.ParameterType.IsEnum && (int)value == default(int)) value = this.builder.CreateAnonymous(pi);
+
+                    return new
+                    {
+                        Parameter = pi,
+                        Value = value
+                    };
+                })
                 .ToArray();
 
             // Get the value expected to be assigned to the matching member
@@ -276,7 +330,7 @@ namespace Ploeh.AutoFixture.Idioms
                 var propertyInfo = propertyOrField as PropertyInfo;
                 actual = propertyInfo.CanRead
                     ? propertyInfo.GetValue(specimen, null)
-                    : expectedValueForMember;
+                                    : expectedValueForMember;
             }
             else
             {
--- a/Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
+++ b/Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
@@ -5,12 +5,12 @@ using System.Collections.ObjectModel;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
-using System.Text;
 using Ploeh.Albedo;
 using Ploeh.AutoFixture.Idioms;
 using Ploeh.AutoFixture.Kernel;
 using Ploeh.TestTypeFoundation;
 using Xunit;
+using Xunit.Extensions;
 
 namespace Ploeh.AutoFixture.IdiomsUnitTest
 {
@@ -475,6 +475,51 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
             // Teardown
         }
 
+        [Theory]
+        [InlineData(typeof(ReadOnlyFieldInitializedViaConstructor<TestEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyInitializedViaConstructor<TestEnum>))]
+        [InlineData(typeof(ReadOnlyFieldInitializedViaConstructor<TestSingleNonDefaultEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyInitializedViaConstructor<TestSingleNonDefaultEnum>))]
+        public void VerifyWellBehavedEnumInitializersDoNotThrow(Type type)
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var sut = new ConstructorInitializedMemberAssertion(fixture);
+
+            // Exercise system and verify outcome
+            Assert.DoesNotThrow(() => sut.Verify(type));
+        }
+
+        [Theory]
+        [InlineData(typeof(ReadOnlyFieldIncorrectlyInitializedViaConstructor<TestEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestEnum>))]
+        [InlineData(typeof(ReadOnlyFieldIncorrectlyInitializedViaConstructor<TestSingleNonDefaultEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestSingleNonDefaultEnum>))]
+        public void VerifyIllBehavedEnumInitializersDoThrow(Type type)
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var sut = new ConstructorInitializedMemberAssertion(fixture);
+
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(type));
+        }
+
+        [Theory]
+        [InlineData(typeof(ReadOnlyFieldInitializedViaConstructor<TestDefaultOnlyEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyInitializedViaConstructor<TestDefaultOnlyEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestDefaultOnlyEnum>))]
+        [InlineData(typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestDefaultOnlyEnum>))]
+        public void VerifyDefaultOnlyEnumDoesThrowBecauseOfPotentialFalsePositive(Type type)
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var sut = new ConstructorInitializedMemberAssertion(fixture);
+
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(type));
+        }
+
         static void AssertExceptionPropertiesEqual(ConstructorInitializedMemberException ex, ConstructorInfo ctor, ParameterInfo param)
         {
             Assert.Equal(param, ex.MissingParameter);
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ build/
 [Oo]bj/
 [Hh]elp/
 NuGetPackages/
+[pP]ackages
```
adamchester: Why is this added?

adamchester:
Thank you for the investigation and PR on this issue! It looks like the build failed due to a problem out of our control with AppVeyor/Google. Are you able to force another build?

I will take a look in next couple of days.

Tungsten78:
@adamchester, I'd be happy to force a build but unfortunately am not sure how to do this.

.gitignore
Tungsten78: I meant to annotate this in my PR as an inquiry - sorry for the oversight.  The nuget packages were showing up as needing to be added following my first compilation of the fork. What is the accepted way to handle this?

Tungsten78: This was removed, the issue is that newer versions of NuGet are ignoring the repositoryPath redirect.  This happens to occur if using VS 2015 (Nuget 3.3) with restore packages on.  My apologies for not seeing the packages having been committed at the root level of the repo.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
             // Teardown
         }
 
+        [Fact]
+        public void VerifyWellBehavedEnumFieldDoesNotThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            // Exercise system and verify outcome
+            Assert.DoesNotThrow(() => sut.Verify(fieldInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyIllBehavedEnumFieldDoesThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(fieldInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyWellBehavedEnumPropertyDoesThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            // Exercise system and verify outcome
+            Assert.DoesNotThrow(() => sut.Verify(propertyInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyIllBehavedEnumPropertyDoesThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(propertyInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyDefaultOnlyEnumFieldDoesThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(fieldInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyDefaultOnlyEnumPropertyDoesThrow()
+        {
+            // Fixture setup
+            var dummyComposer = new Fixture();
```
ploeh: Why `dummyComposer`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            var sut = new ConstructorInitializedMemberAssertion(dummyComposer);
+            var propertyInfo = typeof(ReadOnlyPropertyIncorrectlyInitializedViaConstructor<TestEnum>).GetProperty("Property");
+
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(propertyInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyDefaultOnlyEnumFieldDoesThrow()
```
ploeh: Could this test be refactored to a Parametrised Test and combined with `VerifyIllBehavedEnumFieldDoesThrow`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -475,6 +474,84 @@ public void VerifyTypeWithPublicReadOnlyFieldsIncorrectlyInitialisedViaConstruct
...
+            var sut = new ConstructorInitializedMemberAssertion(dummyComposer);
+            var fieldInfo = typeof(ReadOnlyFieldIncorrectlyInitializedViaConstructor<TestDefaultOnlyEnum>).GetField("Field");
+
+            // Exercise system and verify outcome
+            Assert.Throws<ConstructorInitializedMemberException>(() => sut.Verify(fieldInfo));
+            // Teardown
+        }
+
+        [Fact]
+        public void VerifyDefaultOnlyEnumPropertyDoesThrow()
```
ploeh: Could this test be refactored to a Parametrised Test and combined with `VerifyIllBehavedEnumPropertyDoesThrow`?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
```diff
@@ -705,5 +802,11 @@ int IEqualityComparer.GetHashCode(object obj)
                 throw new NotImplementedException();
             }
         }
+
+        enum TestEnum { none = 0, one, two, three };
+
+        enum TestDefaultOnlyEnum { none = 0 };
+
+        enum TestSingleNonDefaultEnum { none = 1 };
```
ploeh: Where is this used?

ploeh:
Thank you for your interest in contributing to AutoFixture. This pull request looks promising :+1: 

The few questions I had shouldn't be a major obstacle, I hope.

adamchester:
It looks good.

I wonder have we thought if there are any other types that might have a similar issue? Types where we might not reliably be able to generate a non-default value? Enum is the only case I can think of.

Assuming the user actually has an enum with a single default value, now what can they do about it? I assume, given that we throw, they can no longer verify that class constructor automatically, and they have to manually test them? Would that manual test have to skip that parameter?

Tungsten78:
Thanks both for the review.  I'll address each comment individually.

@adamchester - I too reflected on what other types might suffer this same fate and came up empty.  In regards to the single default value enum, I myself can't see the point of using an enum like this.  I felt the guard was important to prevent a false positive, but this seemed like a pretty rare scenario.  I agree with your solutions and would add another being possibly changing the value assignment if sticking to a single element was a requirement.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: These are used as type arguments in the arrange phase of the newly added [unit tests](https://github.com/Tungsten78/AutoFixture/blob/905a836fe4248d40447e11b758ccfd01ae4f1ae6/Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs#L535).

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Great suggestion, I'll give it a try.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Great suggestion, I'll give it a try.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: To be honest, I was following the convention laid out by existing unit tests in this source file.  

Thinking about it, "dummy" might suggest a [dummy object](http://xunitpatterns.com/Dummy%20Object.html) which is not used by the test.  In this case, it is a critical stand-in for `ISpecimenBuilder`.

Any objections to `fixture` instead?   Shall I change all pre-existing instances as well? 
I assume we frown on inconsistency, but so to is mucking around too far outside the scope of a PR.

adamchester: Looks like `dummyComposer` was my fault, and I think @ploeh was sick of my crappy PRs at the time and decided to allow it :)

Suggest we fix the names in a separate PR all together.

Src/Idioms/ConstructorInitializedMemberAssertion.cs
```diff
@@ -250,11 +260,53 @@ public override void Verify(FieldInfo fieldInfo)
...
             ConstructorInfo ci, MemberInfo propertyOrField)
         {
             var parametersAndValues = ci.GetParameters()
-                .Select(pi => new {Parameter = pi, Value = this.builder.CreateAnonymous(pi)})
+                .Select(pi =>
+                {
+                    var value = this.builder.CreateAnonymous(pi);
+                    
+                    //ensure enum isn't getting the default value, otherwise we won't be able to determine whether initialization occurred
+                    if (pi.ParameterType.IsEnum && (int)value == default(int)) value = this.builder.CreateAnonymous(pi);
```
Tungsten78: I'd like to discuss this.

I was a bit uncomfortable doing this here thinking there might be a more declarative approach.  Perhapsa customization which overrides `EnumGenerator`'s behavior - say `EnumGeneratorNoDefault`.  Trouble is I wasn't comfortable moving this towards 'IFixture'.  Maybe there's a better way?

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Sorry, am not used to the comment vs line placement.
Thanks, `TestSingleNonDefaultEnum` is indeed missed in my tests.

ploeh:
> I wonder have we thought if there are any other types that might have a similar issue?

Good question. `bool`?

I must admit that I can't even remember what the current algorithm is for `bool`, and I'm typing this on my way to another thing, so don't have time to look it up right now :flushed:

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
ploeh: > Looks like `dummyComposer` was my fault, and I think @ploeh was sick of my crappy PRs at the time and decided to allow it :)

:laughing: 

> Suggest we fix the names in a separate PR all together.

:+1:

Src/Idioms/ConstructorInitializedMemberAssertion.cs
ploeh: I looked at that line of code for quite a few minutes :smile: 

Then I thought: *oh, well, it's an implementation detail... If it's a problem, we can change it later.*

Not that I'm a proponent of not thinking things through, etc. but there's also the saying that *perfect is the enemy of good*. If you can think of a better way *right now*, then feel free to update the PR. If not, but you think of something next month, feel free to send another pull request :smile:

.gitignore
Tungsten78: Pull Request #517 should help mitigate this in the future

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Addressed, now `fixture` in refactored tests.  

I will submit a PR that covers `dummyComposer` -> `fixture` in the context of this test file.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Tests were refactored and are now relying on parameterised input.  The result is very nice.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: Tests were refactored and are now relying on parameterised input. The result is very nice.

Src/IdiomsUnitTest/ConstructorInitializedMemberAssertionTest.cs
Tungsten78: `TestSingleNonDefaultEnum` now incorporated into the parameterised tests

Src/Idioms/ConstructorInitializedMemberAssertion.cs
Tungsten78: :thumbsup:

Tungsten78:
I've been force pushing the PR branch to keep the commit clean.  I hope this is reasonable and not irritating.

Thanks for the feedback!

adamchester:
:+1:

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.39.0](https://www.nuget.org/packages/AutoFixture/3.39.0).

PR #518, opened 2016-01-23
Get rid of Type.GetTypeCode

moodmosaic:
It looks like Type.GetTypeCode is currently missing in .NET Core, and so it makes our code less portable.

See discussion at #513.

### DIFF
--- a/Src/AutoFixture/RangedNumberGenerator.cs
+++ b/Src/AutoFixture/RangedNumberGenerator.cs
@@ -155,42 +155,53 @@ namespace Ploeh.AutoFixture
 
         private static object Add(object a, object b)
         {
-            var candidateType = a.GetType();
+            switch (GetTypeCode(a))
+            {
+                case TypeCode.Int16:
+                    return (short)((short)a + (short)b);
 
-            if (candidateType == typeof(short))
-                return (short)((short)a + (short)b);
+                case TypeCode.UInt16:
+                    return (ushort)((ushort)a + (ushort)b);
 
-            if (candidateType == typeof(ushort))
-                return (ushort)((ushort)a + (ushort)b);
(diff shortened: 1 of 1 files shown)


lbargaoanu:
Here you have an object and for object the implementation is [trivial] (https://github.com/AutoFixture/AutoFixture/pull/513/files#diff-03b8bf48a2b8c724baa0216d538313e9R22). The problem is that when you have only the Type you have to manufacture an instance.

moodmosaic:
One step at a time :smiley:

ploeh:
As @lbargaoanu points out, since we already have a value, and we're only testing for the types distinguishable with `TypeCode`, wouldn't it be less indirect to pull it out of the value using the `IConvertible` interface?

moodmosaic:
Very nice. I think @lbargaoanu's approach is more efficient and also makes the code easier to read.

Src/AutoFixture/RangedNumberGenerator.cs
```diff
@@ -194,5 +194,14 @@ private static object Add(object a, object b)
                     throw new InvalidOperationException("The underlying type code of the specified types is not supported.");
             }
         }
+
```
moodmosaic: We can move this to a separate file but, according to the [rule of three](http://c2.com/cgi/wiki?RuleOfThree), we can wait.

moodmosaic:
I've just updated the Pull Request. Notice how the diff-output looks cleaner now :smiley:

ploeh:

I didn't release a new version of AutoFixture, since this doesn't change any behaviour.

moodmosaic:
Thank you for the great feedback @lbargaoanu and @ploeh.

PR #519, opened 2016-01-24
Get rid of another Type.GetTypeCode usage

moodmosaic:
Similar to (merged) Pull Request #518. See also the discussion at #513.

### DIFF
--- a/Src/AutoFixture/RandomRangedNumberGenerator.cs
+++ b/Src/AutoFixture/RandomRangedNumberGenerator.cs
@@ -99,9 +99,7 @@ namespace Ploeh.AutoFixture
         /// <returns></returns>
         private static long ConvertLimit(object limit)
         {
-            var limitType = limit.GetType();
-
-            switch (Type.GetTypeCode(limitType))
+            switch (GetTypeCode(limit))
             {
                 case TypeCode.Byte:                   
                         return (long)(byte)limit;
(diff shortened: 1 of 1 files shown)


Src/AutoFixture/RandomRangedNumberGenerator.cs
```diff
@@ -135,13 +133,18 @@ private static long ConvertLimit(object limit)
...
-
-               
             }
 
             throw new ArgumentException("Limit parameter is non-numeric ", "limit");         
-        }      
-       
+        }
+
+        private static TypeCode GetTypeCode(object request)
```
moodmosaic: We can move this to a separate file but, according to the [rule of three](http://c2.com/cgi/wiki?RuleOfThree), we can wait.

Src/AutoFixture/RandomRangedNumberGenerator.cs
```diff
@@ -135,13 +133,18 @@ private static long ConvertLimit(object limit)
 
                 case TypeCode.UInt64:
                         return (long)(ulong)limit;
-
-               
             }
 
             throw new ArgumentException("Limit parameter is non-numeric ", "limit");         
-        }      
-
```
moodmosaic: I removed a few unnecessary blank lines here and there, by following [the boy scout rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule).

ploeh:
Thank you for your contribution! Since there's no changed behaviour, I haven't released a new version of AutoFixture.

moodmosaic:
Sure, that makes sense. :+1:

lbargaoanu:
A little late, but there is actually Convert.GetTypeCode(object) for both frameworks.

ploeh:
I wouldn't mind a pull request with a further refactoring towards `Convert.GetTypeCode(object)`, if anyone feels like it :smile:

Issue #522, opened 2016-02-04
Connect AutoFixture with NUnit3


hackle:
The much endeared [AutoData] feature of AutoFixture has yet to be connected to the shiny new NUnit 3, introducing a unnecessary dilemma for many of us:  AutoFixture or NUnit3?

In order to bridge the gap I have ventured a solution:

https://github.com/hackle/AutoFixture.NUnit3.AutoDataConnector

It'll be awesome if with the help of AutoFixture supporters, this somewhat naive attempt could develop into a new component of AutoFixture to enable it to stay up-to-date.

I've also particularly tested that it works under Resharper 10.0.2 + NUnit 3, hopefully this puts an end to the woeful support from Resharper 9 + NUnit 2.6.*

ploeh:
This could definitely be a welcome addition to the AutoFixture family. When the _AutoFixture.NUnit2_ package was created, we were fully aware that NUnit 2's extensibility model would make it awkward to use at best. We also expected that this would become better in NUnit 3, but that it would involve breaking changes. This is the reason we deliberately named the Glue Library _AutoFixture.NUnit2_, instead of simply _AutoFixture.NUnit_.

This means there's room for a brand-new, independently developed _AutoFixture.NUnit3_ package.

For anyone interested in contributing to a new _AutoFixture.NUnit3_ package, I think that @hackle's code can serve as a prototype. It's probably going to be easier to re-implement it than try to move the code wholesale; if for no other reason, then simply for the reason that I'm not going to review a big pull request in one go (more details [here](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests)).

The situation is similar to when the _AutoFixture.XUnit2_ package was created. After [an initial attempt with a big pull request](https://github.com/AutoFixture/AutoFixture/pull/294), we succeeded by instead breaking it up in several small pull requests, adding features as we went along.

Particularly, I'd like to draw attention to [this comment that explains the motivation for working incrementally](https://github.com/AutoFixture/AutoFixture/pull/294#issuecomment-82832126), and [this comment outlining a possible plan](https://github.com/AutoFixture/AutoFixture/pull/294#issuecomment-82385218). In the end, I think the progress of pull requests deviated a bit from the original plan, but they're all visible (among other pull requests) if you [search for them](https://github.com/AutoFixture/AutoFixture/pulls?page=2&q=is%3Apr+is%3Aclosed+xunit). I believe the first one was #373.

[...]

PR #524, opened 2016-02-09
Autodata glue for nunit3 (issue #522)

hackle:
This is ported over from https://github.com/hackle/AutoFixture.NUnit3.AutoDataConnector following issue: https://github.com/AutoFixture/AutoFixture/issues/522

In short it introduces AutoDataAttribute which supports FrozenAttribute, with an effort at making everything as testable as possible while keeping code ported over from NUnit (in particular TestCaseAttribute).

Points / concerns:

1. No nuget spec yet - NUnit3 has different packages for different versions of .NET, which may require corresponding design, but that can complicate this first pull request. The current project (and therefore resultant dll) only targets .NET4.0

2. No changes to the build script for the same reason

Thanks.


BuildRelease.msbuild
```diff
@@ -107,7 +107,7 @@
         <ItemGroup>
           <XunitTestAssemblies
               Include="Src\*Test\bin\$(Configuration)\*Test.dll"
-              Exclude="Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
+              Exclude="Src\AutoFixture.NUnit*.*Test\bin\$(Configuration)\*.dll" />
```
moodmosaic: Is this change necessary?

Src/AutoFixture.NUnit3/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,41 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("AutoFixture.NUnit3")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Ploeh")]
+[assembly: AssemblyProduct("AutoFixture.NUnit3")]
+[assembly: AssemblyCopyright("Copyright © Ploeh 2011")]
```
moodmosaic: Is this line accurate enough? :confused:

Src/AutoFixture.NUnit3/ITypedValueProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    public interface ITypedValueProvider
```
moodmosaic: What's the purpose of `ITypedValueProvider`?

moodmosaic:

AFAICT, there are public types introduced in this Glue Library that don't exist in AutoFixture.NUnit2. Do we really need to support and maintain them? :confused: 

Unless @ploeh, @adamchester, or @ecampidoglio have a different opinion, can this Pull Request be broken into separate smaller ones? ([Similar to what we did with AutoFixture.Xunit2](https://github.com/AutoFixture/AutoFixture/pull/294#issuecomment-82385218).)

BuildRelease.msbuild
hackle: This is to exclude AutoFixture.NUnit3 unit tests from breaking the build.

Src/AutoFixture.NUnit3/Properties/AssemblyInfo.cs
hackle: Not enough I guess I'll update it. Missed it when copying from AutoFixture.NUnit2...

Src/AutoFixture.NUnit3/ITypedValueProvider.cs
hackle: Hmm so this might be a bad name. The purpose of it is to expose capacities to provide values with different purposes - hence the methods CreateValue(Type) and CreateFrozenValue(Type).

I agree on the concern of exposure of public interfaces which may cause confusion.

hackle:
Thanks for your timely review @moodmosaic I just skimped through the AutoFixture.Xunit2 thread and think that makes sense. (@ploeh actually referenced it in issue #522 so my bad to have not followed up) Is it as easy as following the same path and starting with small requests?

moodmosaic:
@hackle :+1:

>Is it as easy as following the same path and starting with small requests?

This way it was easier with AutoFixture.Xunit2, so it should also be easier with AutoFixture.NUnit3. In case it helps, #377 and #378 were the first merged Pull Requests towards AutoFixture.Xunit2. :wink:

ploeh:
Thank you for your interest in contributing to AutoFixture :+1: 

As @moodmosaic has already suggested, I'd appreciate if this pull request could be split into several smaller pull requests. If not, I'm afraid the review may miss important points, and instead devolve into [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality). The addition of support for NUnit 3 is, I think, too important for that.

Could the first step be to add a new, empty Visual Studio solution?

For the reasons outlined above, I'm going to close this pull request. It's important to me, however, to underscore that I welcome the contribution, and that I think it can become a valuable addition to AutoFixture. I'm closing this pull request for the sole reason that it's too big. Apart from that, you should read no judgement into this action, as I haven't otherwise looked at the code yet.

Please don't be discouraged that this pull request is closed. We still eagerly await your contribution to this new feature; we only need them in smaller instalments.

PR #526, opened 2016-02-12
Autodata glue for nunit3

hackle:
Ported over configs from AutoFixture.NUni2, including

- TreatWarningsAsErrors
- UseVSHostingProcess
- DocumentationFile
- CodeAnalysisRuleSet and other settings for "Verify"

I hope this is not big enough to breach any protocols? Otherwise I could go littler by littler.


ploeh:

hackle:
Thanks. Any pointers what should come next?

ploeh:
When reviewing this, I noticed that both assembly name and default namespace is _AutoFixture.NUnit3_.

Both should be _Ploeh.AutoFixture.NUnit3_, in order to stay consistent with all other naming in AutoFixture.

PR #523, opened 2016-02-07
[WIP] Fake build windows

ErikSchierboom:
This PR adds a FAKE script that has the exact same functionality as the existing MSBuild file. This will allow for C# 6 features to be used (see #500). It also opens up the possibility for cross-platform building. Hopefully, I managed to get the dependencies right, which is why I marked this as a WIP.

It is also marked as WIP as I haven't been able to get the cross-platform stuff working. While everything builds fine on Windows, on my Mac it got 24 compile errors. It turns out the Mono compiler did not like some of the `<seecref>` comment references. To fix this, the types references needed to be prefixed with "T:" (as per [this link](http://mono.1490590.n4.nabble.com/Errors-with-XML-Comments-during-build-td3009482.html)). This solved the problem of the 24 compile errors, but now the next problem was that the xUnit runner threw an exception:

<pre>
ystem.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object

Server stack trace: 

  at Xunit.TestCaseMessage..ctor (ITestCase testCase) <0x31bcc00 + 0x0000c> in <filename unknown>:0 

  at Xunit.TestMessage..ctor (ITest test) <0x31c5f48 + 0x00027> in <filename unknown>:0 

  at Xunit.TestStarting..ctor (ITest test) <0x31c5f20 + 0x00017> in <filename unknown>:0 

  at (wrapper remoting-invoke-with-check) Xunit.TestStarting:.ctor (Xunit.Abstractions.ITest)

  at Xunit.TestClassCallbackHandler.OnStart (System.Xml.XmlNode xml) <0x31c5190 + 0x0012f> in <filename unknown>:0 

  at Xunit.TestClassCallbackHandler.OnXmlNode (System.Xml.XmlNode node) <0x31c50d8 + 0x0006a> in <filename unknown>:0 

  at Xunit.XmlNodeCallbackHandler.System.Web.UI.ICallbackEventHandler.RaiseCallbackEvent (System.String eventArgument) <0x304ec08 + 0x0008b> in <filename unknown>:0 

  at (wrapper managed-to-native) System.Runtime.Remoting.RemotingServices:InternalExecute (System.Reflection.MethodBase,object,object[],object[]&)

  at System.Runtime.Remoting.RemotingServices.InternalExecuteMessage (System.MarshalByRefObject target, IMethodCallMessage reqMsg) <0x1a9a540 + 0x00234> in <filename unknown>:0 

[...]

PR #528, opened 2016-02-14
add NUnit3 nuget package; add unit test project

hackle:
This PR may look a bit oversized because of reference to nuget package NUnit3, other than that:

1. Added a unit test project
2. Ignored the new unit test project in BuildRelease.msbuild to prevent build from breaking

**Points of interest**

1. NUnit3 has different builds targetting .Net4 and .Net4.5, would that have impact on how AutoFixture.NUnit3 is packaged for Nuget? If so what kind of?
2. There seems to be a lack of msbuild task for NUnit3?


Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    [TestFixture]
+    public class PilotTest
+    {
+        [Test]
+        public void AssertTrue()
+        {
+            Assert.IsTrue(true);
```
moodmosaic: Is this test supposed to be discovered? I've changed it `false` to see if it fails, run `build-release.sh` and it succeeded.

BuildRelease.msbuild
```diff
@@ -107,7 +107,7 @@
         <ItemGroup>
           <XunitTestAssemblies
               Include="Src\*Test\bin\$(Configuration)\*Test.dll"
-              Exclude="Src\AutoFixture.NUnit2.*Test\bin\$(Configuration)\*.dll" />
+              Exclude="Src\AutoFixture.NUnit*.*Test\bin\$(Configuration)\*.dll" />
```
moodmosaic: Do we know if this change works? See [comment](https://github.com/AutoFixture/AutoFixture/pull/528/files#r52838572) below.

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
ploeh: Like @moodmosaic, I'm wondering what the purpose is of this test.

BuildRelease.msbuild
hackle: This prevents the single test under AutoFixture.NUnit3.UnitTest from running. Without this change the build would fail as xUnit runners will include it in assemblies.

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
hackle: This is a dummy test to make sure that NUnit3 is referenced and works correctly. However in the build process I could not find a way to create a NUnit task that supports NUnit 3. So this unit test assembly now serves only for local development, not for continuous integration (as it's excluded from xUnit assemblies - see below - but not run with a NUnit build task, as is assembly AutoFixture.NUnit2.UnitTest)

BuildRelease.msbuild
hackle: https://ci.appveyor.com/project/AutoFixture/autofixture/build/1.0.131-ftldqexm
(line 8668 - 8696)
The build log shows that all xUnit tests are run, as well as Ploeh.AutoFixture.NUnit2.UnitTest.dll, but not Ploeh.AutoFixture.NUnit3.UnitTest.dll which is not set up with a NUnit3 build task and should be excluded at this moment.

Does that makes sense?

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
ploeh: Is this file intended to stay in the repository?

ploeh:
> There seems to be a lack of msbuild task for NUnit3?

We'll need to find a solution to this problem, but I think it doesn't have to be in this pull request.

Have you tried asking on whatever support channels exist for NUnit, whether an MSBuild task exists?

If none exists, could we solve the problem by shelling out?

In any case, perhaps we're not even going to need an MSBuild task for this, as @ErikSchierboom is working on [porting the build script to Fake](https://github.com/AutoFixture/AutoFixture/pull/523). If so, we'd like to use Fake's built-in support for running tests. If Fake doesn't already have an NUnit 3 runner, we'd need to contribute one.

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
hackle: No, it's here because there is nothing else to write a test against to show that the project is set up properly. So it should disappear with the appearance first meaningful unit test.

ploeh: That makes sense, but now that you've seen that it works, couldn't you remove it?

(I trust that unit tests will work - particularly if you tell me that you've seen it work. I don't need a dummy test to prove that.)

hackle:
All the better if we don't need one.

The closest I could find was this pull request here targetting NUnit3 support for msbuild.
https://github.com/loresoft/msbuildtasks/pull/205

Either way, is it correct that we still need to exclude this new Unit Test project from the msbuild script, for the time being?

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
ploeh: It's not that it's a big deal, but I prefer only committing code to the repository if it has a long-term purpose.

I do realise that it's temporary, but the problem with committing temporary code is that it may become permanent.

While _you_ may know that you'd like to remove this file again later, other people may not be aware of this. It's probably going to be fine, but what _could_ happen is that after you've committed this file, something happens to you that sidetracks you:

- You may become very busy at work
- You win the lottery, and decide to travel the world instead of contributing to open source
- You meet a nice person (if you're not already in a relationship)
- You're in an accident
- Etc.

If we _can't avoid_ temporary things, we'll take them, but if they _can_ be avoided, I'd prefer to do so, as it minimises the items that everyone needs to keep track of.

If a temporary item _can't_ be avoided, I'd prefer a big fat comment to that effect in the source file, if at all possible.

ploeh:
> Either way, is it correct that we still need to exclude this new Unit Test project from the msbuild script, for the time being?

I'd prefer _including_ it as soon as possible.

We're still in a phase where we aren't _publishing_ the new package, so we _can_ wait a bit. I don't want to publish anything if it isn't covered by the build script, though.

hackle:
:) nice comments. Thanks for the writing.

I have removed the temporary test. Next I will try adding Nunit 3 console command line to the build script, if that's desirable at all.

Src/AutoFixture.NUnit3.UnitTest/PilotTest.cs
moodmosaic: Nice comment! FWIW, if *I* ever win the lottery I'd contribute *only* to open-source (while also traveling around I guess)! 😀

ErikSchierboom:
There is an NUnit 3 runner in FAKE: http://fsharp.github.io/FAKE/apidocs/fake-testing-nunit3.html :)

ploeh:

PR #462, opened 2015-11-05
Make some InlineAutoDataAttribute constructor overloads protected

moodmosaic:
Related to the issue described in #458.

Issue #458, opened 2015-10-31
Make some InlineAutoDataAttribute constructor overloads protected


ploeh:
`InlineAutoDataAttribute` has a constructor overload that takes an `AutoDataAttribute` as an argument.

As #457 implicitly points out, there's no way to use this constructor overload when using the `[InlineAutoData]` attribute to annotate tests.

This constructor overload only exists to enable a derived class to supply a custom, derived instance of `AutoDataAttribute`, thereby customizing the behaviour of the derived attribute. Since this constructor overload only exists to enable inheritance, we should consider changing its accessibility level to `protected`.

This is a breaking change, so can only be done in AutoFixture 4.

Note that this ought to be done for both *AutoFixture.Xunit* and *AutoFixture.Xunit2*.

moodmosaic:
I'm going to take a look at this, and (hopefully) open a Pull Request over the next couple of days.

moodmosaic:
I think we can now close this, since #462 is merged.

ploeh:
Addressed by 5ac109164aa212c5963d620d3a4a166ebbe1096f

PR #529, opened 2016-02-15
Add AutoDataAttribute

hackle:
Added NUnit.Console.3.0.1 so it may look a bit oversized.

1. Added AutoDataAttribute and tests.
2. Added AutoFixture.NUnit3.UnitTest project into the build script, with an Exec task taking a commandline - tell me if anything is silly as I am no expert in msbuild.


Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Construct a <see cref="AutoDataAttribute"/> with an <see cref="IFixture"/>
+        /// </summary>
+        protected AutoDataAttribute(IFixture fixture)
+        {
+            _fixture = fixture;
+            this.RunState = RunState.Runnable;
+            this.Properties = new PropertyBag();
+        }
```
ploeh: Does this constructor need to be here?

If I delete it, and slightly modify the parameterless constructor to the following implementation, the build still succeeds.

```C#
public AutoDataAttribute()
{
    this._fixture = new Fixture();
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+                var specimenBuilder = new SpecimenContext(this._fixture);
+                var parameterValues = method.GetParameters()
+                    .Select(p => specimenBuilder.Resolve(p.ParameterType))
+                    .ToArray();
+                return new TestCaseParameters(parameterValues);
+            }
+            catch (Exception ex)
+            {
+                return new TestCaseParameters(ex);
+            }
```
ploeh: Why is the `try/catch` block there? If I remove it, the build still succeeds.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+            yield return test;
+        }
+
+        private TestCaseParameters GetParametersForMethod(IMethodInfo method)
+        {
+            try
+            {
+                var specimenBuilder = new SpecimenContext(this._fixture);
+                var parameterValues = method.GetParameters()
+                    .Select(p => specimenBuilder.Resolve(p.ParameterType))
```
ploeh: Wouldn't `.Select(p => specimenBuilder.Resolve(p.ParameterInfo))` be more flexible?

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+        }
+
+        #region ITestData Members
+
+        /// <summary>
+        ///     Gets or sets the name of the test.
+        /// </summary>
+        /// <value>The name of the test.</value>
+        public string TestName { get; set; }
```
ploeh: If I replace this with the following, the build still succeeds.

```C#
public string TestName
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        ///     Gets or sets the RunState of this test case.
+        /// </summary>
+        public RunState RunState { get; private set; }
+
+        /// <summary>
+        ///     Gets the list of arguments to a test case
+        /// </summary>
+        public object[] Arguments { get; private set; }
```
ploeh: If I replace this with the following, the build still succeeds.

```C#
public object[] Arguments { get { return null; } }
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        ///     Gets the properties of the test case
+        /// </summary>
+        public IPropertyBag Properties { get; private set; }
```
ploeh: If I replace this with the following, the build still succeeds.

```C#
public IPropertyBag Properties { get { return null; } }
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+            get
+            {
+                return this._expectedResult;
+            }
+            set
+            {
+                this._expectedResult = value;
+                this.HasExpectedResult = true;
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public object ExpectedResult
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        private object _expectedResult;
+
+        /// <summary>
+        ///     Returns true if the expected result has been set
+        /// </summary>
+        public bool HasExpectedResult { get; private set; }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public bool HasExpectedResult { get { return false; } }
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            get
+            {
+                return this.Properties.Get(PropertyNames.Description) as string;
+            }
+            set
+            {
+                this.Properties.Set(PropertyNames.Description, value);
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string Description
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            get
+            {
+                return this.Properties.Get(PropertyNames.Author) as string;
+            }
+            set
+            {
+                this.Properties.Set(PropertyNames.Author, value);
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string Author
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+            get
+            {
+                return this._testOf;
+            }
+            set
+            {
+                this._testOf = value;
+                this.Properties.Set(PropertyNames.TestOf, value.FullName);
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public Type TestOf
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            get
+            {
+                return this.IgnoreReason;
+            }
+            set
+            {
+                this.IgnoreReason = value;
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string Ignore
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            get
+            {
+                return this.RunState == RunState.Explicit;
+            }
+            set
+            {
+                this.RunState = value ? RunState.Explicit : RunState.Runnable;
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public bool Explicit
{
    get { return false; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            get
+            {
+                return this.Properties.Get(PropertyNames.SkipReason) as string;
+            }
+            set
+            {
+                this.Properties.Set(PropertyNames.SkipReason, value);
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string Reason
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+            get
+            {
+                return this.Reason;
+            }
+            set
+            {
+                this.RunState = RunState.Ignored;
+                this.Reason = value;
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string IgnoreReason
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+#if !PORTABLE
+
+        /// <summary>
+        ///     Comma-delimited list of platforms to run the test for
+        /// </summary>
+        public string IncludePlatform { get; set; }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string IncludePlatform
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        ///     Comma-delimited list of platforms to not run the test for
+        /// </summary>
+        public string ExcludePlatform { get; set; }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string ExcludePlatform
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+                return this.Properties.Get(PropertyNames.Category) as string;
+            }
+            set
+            {
+                foreach (var cat in value.Split(','))
+                {
+                    this.Properties.Add(PropertyNames.Category, cat);
+                }
+            }
+        }
```
ploeh: If I replace this with the following implementation, the build still succeeds.

```C#
public string Category
{
    get { return null; }
    set { }
}
```

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+        private object _expectedResult;
+
+        /// <summary>
+        ///     Returns true if the expected result has been set
+        /// </summary>
+        public bool HasExpectedResult { get; private set; }
+
+        #endregion
+
+        #region Other Properties
```
ploeh: Why are the properties in this region present on this class?

I tried to delete them all, and the build still succeeds.

Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            Assert.That(new AutoDataAttribute(), Is.InstanceOf<NUnitAttribute>());
+        }
+
+        [Test]
+        public void ImplementsITestBuilderITestCaseDataAndIImplyFixture()
+        {
+            var autoDataFixture = new AutoDataAttribute();
+            Assert.That(autoDataFixture, Is.InstanceOf<ITestBuilder>());
+            Assert.That(autoDataFixture, Is.InstanceOf<ITestCaseData>());
```
ploeh: Why is it important that the class implements `ITestCaseData`?

I tried removing the interface implementation from `AutoDataAttribute`, and commented out this line here, and the build still succeeds.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,275 @@
...
+                    .ToArray();
+                return new TestCaseParameters(parameterValues);
+            }
+            catch (Exception ex)
+            {
+                return new TestCaseParameters(ex);
+            }
+        }
+
+        #region ITestData Members
```
ploeh: Why are these properties here? What is `ITestData`?

If I delete the properties in this region, the build still succeeds.

Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
```diff
@@ -0,0 +1,50 @@
...
+            Assert.That(autoDataFixture, Is.InstanceOf<IImplyFixture>());
```
ploeh: Why is it important that the class implements `IImplyFixture`?

I tried removing the interface implementation from `AutoDataAttribute`, and commented out this line here, and the build still succeeds.

Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    [TestFixture]
+    public class AutoDataAttributeTest
+    {
+        [Test]
+        public void ExtendsNUnitAttribute()
+        {
+            Assert.That(new AutoDataAttribute(), Is.InstanceOf<NUnitAttribute>());
```
ploeh: Why is it important that the class derives from `NUnitAttribute`?

I tried changing `AutoDataAttribute` so that it derives directly from `Attribute`, and commented out this line here, and the build still succeeds.

ploeh:
Thank you for this pull request! I used the _Devil's Advocate_ review technique, and based on what the build script checks, I can strip down `AutoDataAttribute` to this implementation, and still pass all tests:

```C#
/// <summary>
/// This attribute uses AutoFixture to generate values for unit test parameters. 
/// This implementation is based on TestCaseAttribute of NUnit3
/// </summary>
[AttributeUsage(AttributeTargets.Method)]
public class AutoDataAttribute : Attribute, ITestBuilder
{
    private readonly IFixture _fixture;

    /// <summary>
    /// Construct a <see cref="AutoDataAttribute"/>
    /// </summary>
    public AutoDataAttribute()
    {
        this._fixture = new Fixture();
    }

    /// <summary>
    ///     Construct one or more TestMethods from a given MethodInfo,
    ///     using available parameter data.
    /// </summary>
    /// <param name="method">The MethodInfo for which tests are to be constructed.</param>
    /// <param name="suite">The suite to which the tests will be added.</param>
    /// <returns>One or more TestMethods</returns>
    public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
    {
        var test = new NUnitTestCaseBuilder().BuildTestMethod(method, suite, this.GetParametersForMethod(method));

        yield return test;
    }

    private TestCaseParameters GetParametersForMethod(IMethodInfo method)
    {
        var specimenBuilder = new SpecimenContext(this._fixture);
        var parameterValues = method.GetParameters()
            .Select(p => specimenBuilder.Resolve(p.ParameterInfo))
            .ToArray();
        return new TestCaseParameters(parameterValues);
    }
}
```

hackle:
That's some wicked review @ploeh!!!

So I look at the source code of NUnit, it seems NUnitAttribute is just a marker, ITestCaseData is for test method with ExpectedResult, and use of IImplyFixture is downright advised against.

So I took the shortened version and re-visited the protected constructor and the try catch blocks, and added some unit tests so hopefully it make sense to keep them.

Thanks!

moodmosaic:
I think [this](https://github.com/AutoFixture/AutoFixture/pull/529#issuecomment-184227135) shorter version shown by @ploeh should be just fine for all the mainstream scenarios.

BuildRelease.msbuild
```diff
@@ -120,6 +121,7 @@
 
         <NUnit ToolPath="$(NUnitToolsFolder)" Assemblies="Src\AutoFixture.NUnit2.UnitTest\bin\Release\Ploeh.AutoFixture.NUnit2.UnitTest.dll" ContinueOnError="true" />
         <NUnit ToolPath="$(NUnitToolsFolder)" Assemblies="Src\AutoFixture.NUnit2.Addins.UnitTest\bin\Release\Ploeh.AutoFixture.NUnit2.Addins.UnitTest.dll" OutputXmlFile="TestResult-Addins.xml" ContinueOnError="true" />
+		<Exec Command="$(NUnit3ToolsFolder)\nunit3-console.exe --framework=v4.0 --result=TestResult-NUnit3.xml Src\AutoFixture.NUnit3.UnitTest\bin\Release\Ploeh.AutoFixture.NUnit3.UnitTest.dll" ContinueOnError="true" />
```
ploeh: `ContinueOnError="true"` causes this step to only issue a warning instead of breaking the build. This means that even if a test fails, the build continues. This ought to instead be `ContinueOnError="false"`, so that a failing test breaks the build.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+        public AutoDataAttribute()
+            : this(new Fixture())
+        {
+        }
+
+        /// <summary>
+        /// Construct a <see cref="AutoDataAttribute"/> with an <see cref="IFixture"/> 
+        /// </summary>
+        /// <param name="fixture"></param>
+        protected AutoDataAttribute(IFixture fixture)
```
ploeh: Is there any reason for _not_ making this `public`?

Src/AutoFixture.NUnit3.UnitTest/DummyFixture.cs
```diff
@@ -0,0 +1,82 @@
+using System;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    /// <summary>
+    /// A <see cref="IFixture"/> for the benefit of creating stubs of <see cref="AutoDataAttribute"/> for unit testing
+    /// </summary>
+    public class DummyFixture : IFixture
```
ploeh: AFAICT, this class is specifically used in a single test case, so I'd prefer moving it to the same file as a nested, private class. This keeps it close to where it's used. It also makes it clearer that if the test is ever deleted for some reason, we can also delete the `DummyFixture` class.

Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            Assert.That(stub, Is.AssignableTo<AutoDataAttribute>());
+        }
+
+        [Test]
+        public void IfCreateParametersThrowsExceptionThenReturnsNotRunnableTestMethodWithExceptionInfoAsSkipReason()
+        {
+            // Arrange
+            // DummyFixture is set up to throw DummyException when invoked by AutoDataAttribute
+            var autoDataAttributeStub = new AutoDataAttributeStub(new DummyFixture());
```
ploeh: The way `DummyFixture` is used here, the name is misleading. According to the terminology of [xUnit Test Patterns](http://bit.ly/xunitpatterns) (which is what we use in the AutoFixture code base), a _dummy_ is "a placeholder object that is passed to the SUT as an argument [...] but is never actually used."

The `DummyFixture` here, on the contrary, is used to provoke an exception; a _stub_ would be a more appropriate designation. I'd prefer calling it `ThrowingStubFixture` or something like that, in order to signify that it's going to throw an exception when used.

ploeh:
It's looking good now :+1:  The last few comments I left should hardly present any obstacles.

Adding a test step to `BuildRelease.msbuild` is correct, but short-lived. You weren't expected to be aware of this, but I've recently pulled #523 which replaces the MSBuild-based build scripts with FAKE-based build scripts.

As I wrote in my final comment to that pull request, I'd like to keep the MSBuild-based scripts around for one or two releases yet, to be on the safe side. Thus, it's great that you've added the test step there.

Since the FAKE-based build script `Build.fsx` is taking over, we should add the NUnit 3 build step there as well. If you need help with that, I got the impression that @ErikSchierboom already knows how to do that, but else, I'll be happy to help as well.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
moodmosaic: @ploeh, according to #458, isn't this (protected constructor) the right thing to do though?

hackle: Similar reason. To leave it protected for extension. Besides a public constructor is not CLS compliant for an attribute.

ErikSchierboom:
As @ploeh mentioned, I already had something working. To verify that it worked, I created a local copy of this PR's branch, rebased it based on the master and then modified the FAKE build script.

As this PR already adds the NUnit 3 console, you only need to add the following lines to the end of the `"TestOnly"`target;

```fsharp
let nunit3TestAssemblies = !! (sprintf "Src/AutoFixture.NUnit3.*Test/bin/%s/*Test.dll" configuration)

nunit3TestAssemblies
|> NUnit3 (fun p -> { p with StopOnError = false })
```

See: https://github.com/ErikSchierboom/AutoFixture/commit/edaa84b1a88e75993cff190899a4a24872b6a21f

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
ploeh: Ah, good point. Should we do something similar for `[AutoData]` in _AutoFixture.Xunit2_, then?

moodmosaic: This is already in master branch, so I think it should be done in v4 for AutoFixture.Xunit2.

(We could also update #458.)

ploeh: > This is already in master branch

Is it? [It doesn't look like that to me](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/AutoDataAttribute.cs#L58)...

moodmosaic: Oh, I'm sorry. — What I mean is that, for AutoFixture.Xunit2, we should make the contructor protected in `v4` because it'll be a breaking change if we do in that `master`. We did something similar already in Pull Request #462.

hackle:
Thanks @ploeh @moodmosaic @ErikSchierboom I'll have a play with the FAKE script and make a PR of it later if OK with all you guys.

moodmosaic:
@hackle :+1:

hackle:
Added to the Fake build script. Thanks @ErikSchierboom, that was super easy!

ploeh:
The commit history looks strange now. It's like all commits appear twice, and include @ErikSchierboom's recent commits as well. Did you perform a rebase?

Can you please clean it up? You're probably going to need to do a force push against your own fork...

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
ploeh: Sorry, I misunderstood, but I think I now understand what you meant.

I've created #532 to track this issue.

hackle:
@ploeh I believe it's now cleaned up. Yes I did a rebase so I can squeeze in the addition to the FAKE build script however I somehow messed up the history. :-(
Please see if it's now OK?

ploeh:

Now that we have a working `[AutoData]` attribute for NUnit 3, I'm wondering if the next step ought to be publication of an _AutoFixture.NUnit3_ NuGet package?

moodmosaic:
I'm just wondering whether the users of AutoFixture.NUnit3 would expect the other attributes are in-place as well.

CompositeDataAttribute, CustomizeAttribute, FrozenAttribute, and so on...

ploeh:
Perhaps, but IIRC, we released _AutoFixture.Xunit2_ that way as well...

If we have something useful, why not release it?

moodmosaic:
Oh, yes. Then it's not a bad idea to publish an AutoFixture.NUnit3 NuGet package.

hackle:
As much as I am excited about publishing in NuGet, could I suggest that we include the Frozen attribute first? It could be bare minimum, e.g. without Matching. From personal experience I think it can cover the vast majority of usages and prove very practical for users who are looking to upgrade from NUnit2 to NUnit3.

ploeh:
@hackle, in the end, you decide. You're the main contributor on this feature, and I'm not going to dictate what you do. I only appreciate your voluntary efforts :smile: The following, therefore, is only an attempt to explain my opinion on the topic.

In the spirit of _continuous delivery_, I think it'd be valuable to publish a package already. This would enable you to get early feedback on the basic capabilities of the package. It might be, for example, that there are environment-related issues that we haven't already unearthed.

I don't think a `[Frozen]` attribute is required in order to make the `[AutoData]` attribute useful. As an example, I decreasingly rely on it myself. Instead, I often prefer to use Inspection Properties to get at the injected mocks. Here's an example:

```C#
[Theory, AutoTennisData]
public void PlayerOneWinsReturnsResultWithCorrectScoreForPlayerOne([Greedy]Game sut, IPoints newPoints)
{
    Mock.Get(sut.PlayerTwoScore).Setup(p => p.Accept(sut.PlayerOneScore)).Returns(newPoints);
    var result = sut.PlayerOneWinsBall();
    Assert.Equal(newPoints, result.PlayerOneScore);
}
```

The advantage of using `Mock.Get` instead of relying on a `[Frozen]` attribute is that it's possible to distinguish between multiple injected Services of the same type:

```C#
[Theory, AutoGameData]
public void SendSendsOnAllChannels(CompositeChannel<T> sut, T cmd)
{
    sut.Send(cmd);
    var mocks = from c in sut.Channels
                select Mock.Get(c);
    mocks.ToList().ForEach(m =>
        m.Verify(c =>
            c.Send(cmd)));
}
```

When using Moq, this technique requires the use of `Mock.Get`, which can seem a bit awkward, but actually takes up about the same number of characters as `[Frozen]`.

When using other dynamic mock libraries that instead rely on extension methods (IIRC NSubstitute and/or FakeItEasy), it's more natural to access and configure the test doubles via Inspection Properties. From discussing such features with various users, it's my impression that NSubstitute and/or FakeItEasy users don't use `[Frozen]` attributes much, if at all.

hackle:
Totally Agreed (almost)! However that reminds me that we may need to tackle the NUnut3 build versioning issue (for .Net4.5 and .Net4 different builds of NUnit3 will be referenced) sooner rather than later. I'll get on with the NuGet package then!

PR #530, opened 2016-02-17
[ON HOLD] C# 6: nameof (part 1)

ErikSchierboom:
Now that #500 is fixed by #523, we are ready to use C# 6 features. This PR is part 1 of a series of PR's to use C# 6 features in the existing code base.

This PR is parts of a series that converts parameter string names to `nameof` calls. There are 161 modified files in total. As I remember, having multiple small PR's was the preferred method, so that is what I'll do. The only question I have is: how many modified files should each PR contain? For the moment, I have opted for 10 modified files, would that suffice or is too much or too little?

### DIFF
--- a/Src/AutoFixture/AutoPropertiesTarget.cs
+++ b/Src/AutoFixture/AutoPropertiesTarget.cs
@@ -38,7 +38,7 @@ namespace Ploeh.AutoFixture
         public AutoPropertiesTarget(ISpecimenBuilder builder)
         {
             if (builder == null)
-                throw new ArgumentNullException("builder");
+                throw new ArgumentNullException(nameof(builder));
 
             this.builder = builder;
         }
--- a/Src/AutoFixture/BehaviorRoot.cs
+++ b/Src/AutoFixture/BehaviorRoot.cs
@@ -38,7 +38,7 @@ namespace Ploeh.AutoFixture
         public BehaviorRoot(ISpecimenBuilder builder)
         {
             if (builder == null)
-                throw new ArgumentNullException("builder");
+                throw new ArgumentNullException(nameof(builder));
 
             this.builder = builder;
         }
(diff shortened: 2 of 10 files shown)


moodmosaic:
Looks great to me :+1:

ploeh:
Looks good to me too :+1: 

I'd like to wait merging this until I've seen that I can publish new releases of AutoFixture with the new FAKE-based script. While I certainly expect that this is going to be possible, I'd also like to get verification.

Merging this pull request right now effectively prevents us from using the old MSBuild-based script, and if it turns out that the FAKE-based script somehow doesn't work out, we'd be in a bit of a bind.

ErikSchierboom:
That totally makes sense. I'll wait for the next release and if everything works as intended, I'll redo the PR.

As for the number of files per PR, how many would you like?

ploeh:
Unless this pull request becomes unmergeable, there's no reason to redo it. We'll just let it sit here, and then I can pull it once I feel confident that the FAKE-based build scripts hold no surprises.

> As for the number of files per PR, how many would you like?

The answer to this question really depends on what changed. This pull request, for instance, has a razor-sharp focus: all the changes are identical, so it's easy to review. In this case, I can easily handle triple or quadruple the amount of changes (but please just let this pull request stays as it is).

For other types of changes, I might prefer only 2-3 files changed... The point is that I don't think I can give a general answer to this question.

ErikSchierboom:
Thanks for the clarification. I'll prepare the other PR's, but won't yet submit them.

ploeh:
Current status is that I've published versions 3.40.1 and 3.41.0 with the new FAKE-based script. I wasn't entirely happy with 3.40.1, because it included the source code instead of publishing the source code in a separate _symbols_ package, but this has been fixed by #546.

I'd like to see _one_ more successful deployment with the new script. If the next publication also goes off without problems, I feel confident that we can complete the switch.

ErikSchierboom:
I completely understand. Thanks for the heads-up.

ErikSchierboom:
@ploeh Seeing that this is merged, shall I submit the other PR's?

ploeh:
Please do :+1:

PR #535, opened 2016-02-21
Addressed compiler warnings related to CA.

ploeh:
These warnings did not cause compilation errors, because the Code Analysis
settings were only incorrect in build configurations that don't run Code
Analysis.


moodmosaic:
Good catch! :+1:

PR #536, opened 2016-02-22
Autodata nunit3 nuget packaging

hackle:
- Created NuGet spec file
- Created another project to build dll targeting .NET4.5 (Which I am not sure if is the best way to go about).


ploeh:
What happens if we _don't_ make a .NET 4.5-specific build, and _don't_ put anything in any `net45` directory?

hackle:
If we don't put anything under net45, then projects targeting .Net framework won't be able to reference the package correctly.

1. AutoFixture.NUnit3 reference the .Net4 build of NUnit3
2. However because the hosting project targets .Net4.5, NuGet will install the .Net4.5 build of NUnit3
3. Now AutoFixture.NUnit3 is confused as it cannot find the correct build of NUnit3

ploeh:
What happens if you add a binding redirect, then?

hackle:
Ah good idea - I will try that out.

hackle:
I struggled to get binding redirect to work, while I tried packing with the Install.ps1 script, or added binding redirect with either NuGet command or manually, I still got build errors. Maybe you can give it a try @ploeh?
Either way I am keen to however have a separate build for .NET4.5, it seems more work but it's cleaner (debatably?) 
Because I suddenly remember when I tried to get AutoFixture.NUnit2 to work on a instance of TeamCity with an agent that runs NUnit runner 2.6.3 (after getting it to work locally), and binding redirect was no easy thing to accomplish and I ended up with a twisted workaround.
While this is not necessarily the case for AutoFixture.NUnit3, I think user experience is best is users don't have to
1. Manually add a binding redirect, as I sometimes get with AutoFixture.NUnit2, with a sweet readme.txt, or
2. Make workaround so NUnit test runners can pick up the binding redirect (or not)
Otherwise I would suggest we rather do more work on our side instead, it'll be worthwhile.

ploeh:
I've created this issue for NUnit: https://github.com/nunit/nunit/issues/1295

hackle:
This is a nasty gotcha. should we still package to different versions? when nunit fixes that issue we can adapt.



Sent from my Mi phoneOn Mark Seemann <notifications@github.com>, Feb 29, 2016 1:45 AM wrote:I've created this issue for NUnit: nunit/nunit#1295

ploeh:
My preference is to keep it simple, so I'd like to start with only building on .NET 4.0 for now. We can release that and see what happens.

It's harder to back out of a situation where we're already publishing two different builds.

hackle:
@ploeh Could you help me understand the implication a bit better? My understanding is if we adapt to NUnit 3.2 when it's available, we can increase the version of the NuGet package to override old versions and remove one of the builds. With the way the dlls are referenced, users won't have to work around anything, correct?

My concerns with publishing under either .NET4 or .NET4.5 is that due to the NUnit inconsistency, our package can cause frustrations to users. It gives me blues as I know how it can feel like :(

I agree It's indeed a bit awkward maintaining 2 builds, and I think it's up to a debate between maintainability vs usability?

Could I also make an argument that if we do end up supporting only one version of .NET, maybe we can target .NET 4.5 as I presume (lacking any stats) it would be more popular than 4.0?

ploeh:
Please accept my apologies that I'm slow to respond, but I don't have any ready answers; I needed to explore in order to be able to answer.

In order to investigate, I changed the configuration of the existing _AutoFixture.NUnit3_ project, so that it compiles against _NUnit_ 3.0.0, instead of 3.0.1.

I then created a .NET 4.5 test project, and added the _NUnit 3.0.1_ package _before_ adding _AutoFixture.NUnit3_, as outlined in https://github.com/nunit/nunit/issues/1295.

Because that custom build of _AutoFixture.NUnit3_ is compiled against _NUnit 3.0.0_, and that .NET 4.5 test project has added _NUnit 3.0.1_, there's no version conflict. It compiles and the test runs.

(@moodmosaic, @adamchester, and @ecampidoglio are welcome to chime in on this discussion, if they want to.)

hackle:
Thank you very much @ploeh and everybody for helping out.

Based on your investigation I figure why don't I just reference the correct revisions of nunit 3 directly? So I have removed the reference to nunit in nuspec, and added the nunit.framework.dll / xml to the release folder and the build scripts.

This is tested under .Net 4.0 / 4.5 / 4.6, all seem to work.

Any concerns though?

ploeh:
Doesn't that go against the entire idea of using NuGet for dependency management?

Doesn't it make it opaque that _AutoFixture.NUnit3_ depends on NUnit? What happens if the user has another library with an NUnit dependency (e.g. [FsUnit](https://www.nuget.org/packages/FsUnit/))?

hackle:
Of course! My bad for not having investigated your solution better! Now that I have and everything works, pure gold @ploeh

hackle:
Build fails after rebasing to the new default, although no extra changes can be found from the file diff.

It seems under .\NuGetPackages in file Ploeh.AutoFixture.AutoMoq.nuspec, "@build.number@" is not transformed in this line:
      <dependency id="AutoFixture" version="@build.number@" />

However the line above is transformed:
    <id>AutoFixture.AutoMoq</id>
    <version>3.40.1</version>

And the FAKE script builds successfully. All seems a bit strange to me... Anybody has any ideas?

Nuget/Ploeh.AutoFixture.NUnit3.nuspec
```diff
@@ -0,0 +1,26 @@
...
+        <version>@build.number@</version>
+        <title>AutoFixture with NUnit3</title>
+        <authors>Hackle Wayne, AutoFixture</authors>
+        <owners>AutoFixture</owners>
+        <dependencies>
+            <dependency id="autofixture" version="@build.number@" />
+            <dependency id="nunit" version="3.0.1" />
+        </dependencies>
+        <summary>Extension to AutoFixture that integrates it with NUnit3.</summary>
+        <description>By leveraging the some features of NUnit3, this extension turns AutoFixture into a declarative framework for writing unit tests. In many ways it becomes a unit testing DSL (Domain Specific Language).</description>
```
ploeh: Is there a rogue _the_ here?

Sorry to be nitpicking, but this text goes on the NuGet package page, so need to be as descriptive and trustworthy as possible.

ploeh:
Don't worry about the failing build server for the moment. #523 introduced some changes to the build process, but we're still evaluating those changes, and have yet to update the build server configuration.

hackle:
Thanks @ploeh that was a good spot, shame that I missed it...

ploeh:
Looking good, thank you :+1: 

Should we publish the package?

I haven't done it yet, but feel inclined to do so.

hackle:
I'll look into porting over other classes from AutoFixture.NUnit2 in fair bunches if no objections.

ploeh:
Why not release what we already have? While I understand that it isn't 'done' yet, couldn't someone get value out of what we already have?

Once a pull request is accepted, it takes me about five minutes to publish a new release, so we can release new features continuously, as we add them.

hackle:
@ploeh I totally agree with you on releasing now - by that I was just planning some work ahead. I am full in for the release, hurrah!

ploeh:
It's now live as [AutoFixture.NUnit3 3.41.0](https://www.nuget.org/packages/AutoFixture.NUnit3/3.41.0)!

@hackle, I think this is tweet-worthy, but I couldn't find a Twitter profile that was obviously yours. If you have a Twitter account, let me know, and I'll credit you in the tweet.

hackle:
@ploeh Exciting! My twitter handle is "hacklew" - however I should really be using it...

ploeh:
You don't have to use Twitter, but you may get yourself a few followers from [this tweet](https://twitter.com/ploeh/status/708562122347847680) :wink:

Issue #539, opened 2016-03-05
Augment the IFixture interface with the Repeat function from the Fixture class


SeanFarrow:
It would be good if the IFixture interface provided access to the repeat function from the fixture class..

I'm using JustBehave and they've recently changed there CreateFixture method to take an IFixture instead of a Fixture so now the repeat method from the Fixture class is unavailable. I'm happy to do the work and provide a pull request.
Any thoughts/objections to this?

moodmosaic:
I haven't heard about JustBehave, though [there is a RepeatCount member](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/IFixture.cs#L35) on the `IFixture` interface...

SeanFarrow:
Hi,

 

I am aware of RepeatCount, but I’m looking to add the IEnumerable<T Repeat(Func<T> function) member to the IFixture interface.

Cheers

Sean.

moodmosaic:
Isn't that already possible with `Generator<T>`? An example can be found [here](http://stackoverflow.com/a/17170033).

ploeh:
Thank you for writing, and for your interest in AutoFixture. It's always wonderful to hear that someone is using the library, even if we often learn about it when things work less than stellar :smile: 

Unfortunately, changing an interface (including adding new members) is a breaking change, so we can't do that in AutoFixture 3. It'd be possible to do that in [AutoFixture 4](https://github.com/AutoFixture/AutoFixture/milestones/4.0), but I'd be inclined to think that the change isn't warranted.

I've had my eye on the entire `IFixture` interface for years, because I find it crude, but I must admit that it seems to work. Still, I'd be reluctant to _add_ more members to it.

If you consider the implementation of the particular `Repeat` method, however, it wouldn't have to be specific to the `Fixture` class:

```C#
public IEnumerable<T> Repeat<T>(Func<T> function)
{
    return from f in Enumerable.Repeat(function, this.RepeatCount)
            select f();
}
```

Would it be possible, instead, to move it to an extension method like the following?

```C#
public static IEnumerable<T> Repeat<T>(this IFixture fixture, Func<T> function)
{
    return from f in Enumerable.Repeat(function, fixture.RepeatCount)
            select f();
}
```

I haven't tried the above, so it's possible that there's a flaw in my reasoning.

All that said, though, I can't recall ever having used that function, so I wonder why it's even there... Why do you need it?

SeanFarrow:
[...]

PR #540, opened 2016-03-06
Move Repeat<t> to an extension method

SeanFarrow:
This PR solves issue #539.

### DIFF
--- a/Src/AutoFixture/IFixtureExtensions.cs
+++ b/Src/AutoFixture/IFixtureExtensions.cs
@@ -13,8 +13,9 @@ namespace Ploeh.AutoFixture
             {
                 throw new ArgumentNullException("fixture");
             }
-            
-            throw new NotImplementedException();
+
+            return from f in Enumerable.Repeat(function, fixture.RepeatCount)
+                   select f();
         }
     }
 }
--- a/Src/AutoFixture/IFixtureExtensions.cs
+++ b/Src/AutoFixture/IFixtureExtensions.cs
@@ -7,6 +7,25 @@ namespace Ploeh.AutoFixture
 {
     public static class IFixtureExtensions
     {
+        /// <summary>
+        /// Repeats a function many times.
+        /// </summary>
+        /// <typeparam name="T">
+        /// The type of object that <paramref name="function"/> creates.
+        /// </typeparam>
+        /// <param name="fixture">
+        /// The <see cref="IFixture"/>to use
+        /// </param>
+        /// <param name="function">
+        /// A function that creates an instance of <typeparamref name="T"/>.
+        /// </param>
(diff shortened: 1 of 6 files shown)


Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -1847,63 +1847,7 @@ public void DefaultRepeatCountIsThree()
             Assert.Equal<int>(expectedRepeatCount, result);
             // Teardown
         }
-
```
moodmosaic: Wouldn't it be possible to avoid changing/moving these tests?

AFAICT, this change is a pure refactoring, it shouldn't be necessary to modify existing tests. This also makes it possible to verify that no breaking changes are introduced.

Since `Repeat` is extracted into an `IFixture` extension method, `sut` should still be able to access it, so these tests should still compile and pass.

What do I probably missed? :flushed:

Src/AutoFixture/IFixtureExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Ploeh.AutoFixture
+{
+    public static class IFixtureExtensions
```
moodmosaic: Inspired by Haskell's [replicate](http://zvon.org/other/haskell/Outputprelude/replicate_f.html), how about naming this `FixtureReplicator`, `Replicator`, or similar? Or maybe we could name the class `FixtureRepeater`, `Repeater`, and so on.

FWIW, we have another class with extension methods, and it's called [`FixtureFreezer`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/FixtureFreezer.cs) instead of the general purpose XyzExtensions name.

moodmosaic: I think we need some XML documentation there, otherwise the Code Analysis treats this as a warning (which then gets treated as an error during the build process).

moodmosaic:
Thank you for opening a Pull Request!  :smiley: :+1:

I just had a look and added a few comments there.

SeanFarrow:
Agree with all your comments, will fix later.

SeanFarrow:
Ok, All comments have been addressed.

I’ve moved this to a new branch FixtureRepeater to more closely align with what the pr does.

ploeh:
@SeanFarrow, you closed this pull request by yourself, but from the comments I get the impression that you've addressed the review somewhere else. If you'd like us to review and ultimately pull those changes, you'll have to open a new pull request.

If you do that, I'll be happy to consider it.

SeanFarrow:
I’ve opened a new pr!

PR #542, opened 2016-03-07
Update Ploeh.AutoFixture.AutoMoq.nuspec

davidpeden3:
Fix casing of dependencies to work with `dnu build/restore`


ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.40.1](https://www.nuget.org/packages/AutoFixture/3.40.1).

PR #543, opened 2016-03-07
Update Ploeh.AutoFixture.AutoMoq.nuspec

davidpeden3:
update to latest version of `Moq`

### DIFF
--- a/Nuget/Ploeh.AutoFixture.AutoMoq.nuspec
+++ b/Nuget/Ploeh.AutoFixture.AutoMoq.nuspec
@@ -7,8 +7,8 @@
         <authors>Mark Seemann</authors>
         <owners>Mark Seemann</owners>
         <dependencies>
-            <dependency id="autofixture" version="@build.number@" />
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for your interest in contributing to AutoFixture. What is the motivation for the proposed change?

davidpeden3:
AutoFixture.AutoMoq is shipping with a two year old version of Moq. With the other pull request that you merged and published, this pull request is not strictly needed.

ploeh:
_AutoFixture.AutoMoq_ deliberately depends on the minimum version it needs in order to work. This is [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) applied to package dependency management. The _AutoFixture.AutoMoq_ package can accept all versions of Moq it needs in order to perform its duties.

Why restrict the set of useful versions?

If you want to use a newer version of Moq, you can: just install (or update to) the latest version of Moq, and add a binding redirect. I've been doing this for years, and never had any problems.

PR #549, opened 2016-03-12
Get rid of build warnings from misconfigured code analysis

ploeh:


ploeh:
Note to reviewers: the reason the build is failing is because the changes made in #523 means that the old build script no longer works; it fails when packaging NuGet packages.

As explained in #530, I'm currently evaluating the new FAKE-based build script. Once I'm satisfied that there are no issues with the new build script, we should update the build server accordingly.

adamchester:
:shipit:

PR #564, opened 2016-03-14
Add Repeat to the IFixture interface issue #540

SeanFarrow:
Reopening the pr as I inadvertently closed it!

### DIFF
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -1847,63 +1847,7 @@ namespace Ploeh.AutoFixtureUnitTest
             Assert.Equal<int>(expectedRepeatCount, result);
             // Teardown
         }
-
-        [Fact]
-        public void RepeatWillPerformActionTheDefaultNumberOfTimes()
-        {
-            // Fixture setup
-            Fixture sut = new Fixture();
-            int expectedCount = sut.RepeatCount;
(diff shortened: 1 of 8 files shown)


Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -1847,12 +1847,11 @@ public void DefaultRepeatCountIsThree()
...
             // Teardown
         }
-
-        [Fact]
+            [Fact]
         public void RepeatWillPerformActionTheDefaultNumberOfTimes()
         {
             // Fixture setup
-            Fixture sut = new Fixture();
+            IFixture sut = new Fixture();
```
ploeh: Why was this test changed to work on `IFixture`? Doesn't it work on `Fixture` any longer?

ploeh:
Looking good :+1: 

It looks like there are a few issues to address, but they look minimal to me.

SeanFarrow:
I think there all fixed, I’ve reverted the tests to use the Fixture class and fixed all the documentation warnings.

Cheers

Sean.

ploeh:
Almost there, but the `FixtureTest` code looks funny, now.

Basically, I'd expect that that file wasn't touched at all...

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.43.2](https://www.nuget.org/packages/AutoFixture/3.43.2).

PR #565, opened 2016-03-15
C# 6: getter-only properties (pt. 1)

ErikSchierboom:
This PR converts getter-only properties with readonly backing-fields to C# 6's auto-implemented getter-only properties.

Before:
```charp
private readonly ISpecimenBuilderNode node;	

public ISpecimenBuilderNode Node
{		
    get { return this.node; }		
}
```

After:
```charp	
public ISpecimenBuilderNode Node { get; }
```

Functionally, they are equivalent, as the auto-implemented getter-only property also receives a `readonly` backing field under the hood ([description on the Roslyn wiki](https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6#getter-only-auto-properties)).

Is this something you'd be in favor of? If so, I'd be happy to also convert all other such properties.

### DIFF
--- a/Src/AutoFixture/DataAnnotations/MinimizationOperations.cs
+++ b/Src/AutoFixture/DataAnnotations/MinimizationOperations.cs
@@ -539,24 +539,15 @@ namespace Ploeh.AutoFixture.DataAnnotations
 
         private sealed class IntPair
         {
-            private readonly int n1;
-            private readonly int n2;
-
             internal IntPair(int n1, int n2)
             {
-                this.n1 = n1;
-                this.n2 = n2;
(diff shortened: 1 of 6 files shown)


moodmosaic:
Looking good!  — [43 additions and 84 deletions](https://github.com/AutoFixture/AutoFixture/pull/565/files) without breaking changes can be a good thing :+1:

ploeh:
Thank you :+1: 

Since this doesn't change the behaviour of the system, no release is associated with these changes.

ErikSchierboom:
Great. I'll submit the rest tomorrow.

PR #570, opened 2016-03-16
Convert to auto-implemented getter-only properties (pt. 6)

ErikSchierboom:


moodmosaic:
:+1: This is looking good.

PR #583, opened 2016-03-16
Convert to auto-implemented getter-only properties (pt. 19)

ErikSchierboom:


Src/TestTypeFoundation/GenericType.cs
```diff
@@ -13,9 +11,9 @@ public GenericType(T t)
                 throw new ArgumentNullException(nameof(t));
             }
 
-            this.t = t;
+            this.Value = t;
         }
 
-        T Value { get { return this.t; }}
+        private T Value { get; }
```
moodmosaic: Good catch that this one should be marked as `private` :+1:

moodmosaic:
:+1: This is looking good.

PR #586, opened 2016-03-19
Port over frozen attribute to autofixture nunit3

hackle:
Included only what's needed for [Frozen].


ploeh:
What does the `Src/AutoFixture.NUnit3.UnitTest/AutoFixture.NUnit3.UnitTest.lnk` file do?

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -52,16 +52,41 @@ private TestCaseParameters GetParametersForMethod(IMethodInfo method)
...
+        private IFixture CustomizeFixtureByParameter(IFixture fixture, IParameterInfo parameter)
+        {
+            var customizeAttributes = parameter.GetCustomAttributes<CustomizeAttribute>(false);
+            foreach (var ca in customizeAttributes)
+            {
+                var customization = ca.GetCustomization(parameter.ParameterInfo);
+                fixture.Customize(customization);
+            }
+
+            return fixture;
```
ploeh: Why does this method return the fixture? Is the method [a Command or a Query](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation)?

ploeh:
Looking good, although I'd like to learn why that `.lnk` file is there.

@ecampidoglio, since you wrote the original implementation of the declarative matching DSL, if you have any comments regarding this pull request, please share :smile:

hackle:
Thanks @ploeh good spot - the query should be a command, and lnk file is a bad accident. Fixed and pushed.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit3 3.42.0](https://www.nuget.org/packages/AutoFixture.NUnit3/3.42.0).

ploeh:
BTW, you may feel that that [CQS](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation) comment was a bit of [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality), but I had to look at that code quite a few times before I was convinced that it was correct af all. In all, I think I probably spent 5-10 minutes on that particular issue, because I ended up loading up _AutoFixture.NUnit2_ and _AutoFixture.Xunit2_ for comparison.

I'm not writing this to make you feel bad, or in order to complain. The reason I write this is because I think it's important to communicate how, sometimes, seemingly small things can dramatically slow down a code reader's ability to comprehend source code. Most programmers seem to not realise this, which is why I try to point it out whenever I have trouble understanding a piece of code.

ecampidoglio:
> @ecampidoglio, since you wrote the original implementation of the declarative matching DSL, if you have any comments regarding this pull request, please share

Looks good to me :+1:

hackle:
@ploeh I think that's a great call! Totally happy to turn that into a command, makes much more sense.

PR #563, opened 2016-03-14
Separate NUnit test results 

ErikSchierboom:
This PR makes the NUnit 2 test runner output its results to `NUnit2TestResult.xml`, whereas the NUnit 3 test runner outputs its results to `NUnit3TestResult.xml`, in NUnit 2 format. 

It also adds those two NUnit result files to the `.gitignore` file and removes the build step that removed the test runner results.

PR #588, opened 2016-03-25
Port over favors greedy and modest to autofixture nunit3

hackle:
Question - should [FavorLists] and [FavorEnumberables] be included in the Scenarios?


Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -65,6 +65,33 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+
+        [Test, AutoData]
+        public void GreedyCreatesParameterWithGreedyConstructor([Greedy]MultiUnorderedConstructorType p)
+        {
+            Assert.False(string.IsNullOrEmpty(p.Text));
+            Assert.AreNotEqual(0, p.Number);
+        }
+
+        [Test, AutoData]
+        public void BothFrozenAndGreedyAttributesCanBeAppliedToSameParameter([Frozen][Greedy]MultiUnorderedConstructorType p1, MultiUnorderedConstructorType p2)
```
ploeh: In .NET, the order of attributes are not guaranteed. What happens if `[Frozen]` is evaluated and executed _before_ `[Greedy]`? Could that make the test fail?

hackle:
Added another unit test with [Frozen] and [Greedy] in reverse order - however that relies on .NET being consistent for one execution, e.g. it's not doing a Random() or Shuffle() for the attributes, which even true I doubt should be covered.

ploeh:
Why is `TestResult.xml` checked in? Where does it come from?

I'd have hoped that #563 would have ignored such a file, but I can see that it doesn't match that particular pattern.

hackle:
My bad, I think by mistake I must have checked in the TestResult.xml which comes out of a manual run of the NUnit3 Console. The build script specified a result file name so there is nothing wrong with it.

Thanks @ploeh

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit3 3.43.0](https://www.nuget.org/packages/AutoFixture.NUnit3/3.43.0).

PR #589, opened 2016-03-28
Inline autodata attribute autofixture nunit3

hackle:
I had struggled to flesh out **CompositeDataAttribute**. If I understood things correctly, it seems not easy to compose multiple **ITestBuilder** that is shared across **TestCaseAttribute** and **AutoDataAttribute**, which only exposes a method
 `IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)`
It seems in order to compose, I need to merge 2 or more sets of `IEnumerable<TestMethod>`. I don't know if this is why **InlineAutoData** is not supported in NUnit2?

However I implemented **InlineAutoDataAttribute** in a different approach. Since it breaks away from the existing approach I guess much can be left to debate.


Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -0,0 +1,36 @@
...
+            Assert.That(new InlineAutoDataAttribute(), 
+                Is.AssignableTo(typeof(AutoDataAttribute)));
+        }
+
+        [Test]
+        public void InlineAutoDataCanBeExtendedWithImplementationOfIFixture()
+        {
+            var extended = new InlineAutoDataAttributeStub();
+
+            Assert.That(extended, Is.AssignableTo(typeof(InlineAutoDataAttribute)));
```
moodmosaic: What does it mean if `extended` which, is an instance of the SUT, is assignable to the SUT itself?

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,52 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+        {
+            Assert.That(1, Is.EqualTo(p1));
+            Assert.That(2, Is.EqualTo(p2));
+            Assert.That(3, Is.EqualTo(p3));
+        }
+
+        [Theory]
+        [InlineAutoData]
+        public void InlineAutoDataProvidesParameterValuesWhenNoneGiven(int p1, int p2, int p3)
+        {
```
moodmosaic: Isn't an implementation detail that (currently) integers aren't all `0`s by default?

What happens if the implementation of random numbers changes in future versions of AutoFixture?

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,52 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+            Assert.That(0, Is.Not.EqualTo(p3));
+        }
+
+        [Theory]
+        [InlineAutoData(1, 2)]
+        public void InlineAutoDataProvidesParameterValuesWhenMissing(int p1, int p2, int p3)
+        {
+            Assert.That(1, Is.EqualTo(p1));
+            Assert.That(2, Is.EqualTo(p2));
+            Assert.That(0, Is.Not.EqualTo(p3));
```
moodmosaic: [Same](https://github.com/AutoFixture/AutoFixture/pull/589/files#r57543945) here as well.

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,52 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+        [Theory]
+        [InlineAutoData(1, 2)]
+        [InlineAutoData(4)]
+        [InlineAutoData]
+        public void InlineAutoDataCanBeUsedInMultiple(int p1, int p2, int p3)
+        {
+            Assert.That(0, Is.Not.EqualTo(p1));
+            Assert.That(0, Is.Not.EqualTo(p2));
+            Assert.That(0, Is.Not.EqualTo(p3));
+        }
```
moodmosaic: [Same](https://github.com/AutoFixture/AutoFixture/pull/589/files#r57543945) here as well.

In addition to that, how do we know that there are multiple test-runs? Shouldn't NUnit3 be responsible for this outcome?

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,52 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+
+        [Theory]
+        [InlineAutoData(1, 2)]
+        [InlineAutoData(4)]
+        [InlineAutoData]
+        public void InlineAutoDataCanBeUsedWithFrozen(int p1, int p2, [Frozen]string p3, string p4)
+        {
+            Assert.That(p3, Is.EqualTo(p4));
```
moodmosaic: :+1: Nice demo of `[Frozen]` in combination with the proposed `[InlineAutoData]`.

Do we need more than one `[InlineAutoData]` attributes applied to the test case?

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
hackle: Mistake - will fix.

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
hackle: I had doubts too when I basically copied the very 1st test from the Scenarios

`
        [Test, AutoData]
        public void AutoTestCaseProvidesCorrectInteger(int primitiveValue)
        {
            Assert.AreNotEqual(0, primitiveValue);
        }

`

Any clarifications?

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
hackle: I think this is a naming confusion - by **InMultiple** I mean being used like [TestCase], so one test methods can have multiple cases.

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
hackle: Good point - will fix.

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
hackle: Have changed to using **string** parameters instead.

hackle:
Thanks for the review @moodmosaic have just made changes.

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,50 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+        }
+
+        [Theory]
+        [InlineAutoData]
+        public void InlineAutoDataProvidesParameterValuesWhenNoneGiven(string p1, string p2, string p3)
+        {
+            Assert.That(p1, Is.Not.Null);
+            Assert.That(p2, Is.Not.Null);
+            Assert.That(p3, Is.Not.Null);
+        }
```
moodmosaic: :+1:

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,50 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+
+        [Theory]
+        [InlineAutoData("alpha", "beta")]
+        public void InlineAutoDataProvidesParameterValuesWhenMissing(string p1, string p2, string p3)
+        {
+            Assert.That(p1, Is.EqualTo("alpha"));
+            Assert.That(p2, Is.EqualTo("beta"));
+            Assert.That(p3, Is.Not.Null);
+        }
```
moodmosaic: :+1:

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
```diff
@@ -295,5 +295,50 @@ public void FreezeSecondParameterOnlyFreezesSubsequentParameters(Guid g1, [Froze
...
+
+        [Theory]
+        [InlineAutoData("alpha", "beta")]
+        [InlineAutoData("charlie")]
+        [InlineAutoData]
+        public void InlineAutoDataCanBeUsedInMultiple(string p1, string p2, string p3)
```
moodmosaic: If I change the attributes to

```
[Theory]
[InlineAutoData]
[InlineAutoData]
[InlineAutoData]
```

this test should still pass.

Src/AutoFixture.NUnit3.UnitTest/ThrowingStubFixture.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    /// <summary>
+    /// A <see cref="IFixture"/> for the benefit of creating stubs of <see cref="AutoDataAttribute"/> for unit testing
+    /// </summary>
+    public class ThrowingStubFixture : IFixture
```
moodmosaic: Most other hand-written [Test Doubles](http://xunitpatterns.com/Test%20Double.html) in the AutoFixture codebase are named like *DelegatingXyz*, so it might help decreasing the cognitive load if we just name this as `DelegatingFixture` or similar.

What do you think? :smiley:

Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
hackle: True very true!

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        [Test]
+        public void InlineAutoDataCanBeExtendedWithImplementationOfIFixture()
+        {
+            var extended = new InlineAutoDataAttributeStub();
+
+            Assert.That(extended, Is.AssignableTo(typeof(AutoDataAttribute)));
+        }
+
+        private class InlineAutoDataAttributeStub : InlineAutoDataAttribute
```
moodmosaic: AFAICT, this test is here in order to prevent someone, in the future, from removing the `.ctor` that takes an `IFixture` instance as parameter.

If this is the case, I'm wondering if *DerivedInlineAutoDataAttribute* would be a slightly better name instead of InlineAutoDataAttributeStub.


What do you think? :smiley:

Src/AutoFixture.NUnit3.UnitTest/ThrowingStubFixture.cs
hackle: I think the original one being named **ThrowingStubFixture ** actually makes sense as it describes its purpose. However for **InlineAutoData** I am not really sure. Maybe I can just create an inline **Stub **as it doesn't really do anything?

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
hackle: Sure good point!

hackle:
I am actually not sure about making 2 methods in **AutoDataAttribute** so it can be extended by **InlineAutoDataAttribute** - there is not much semantic value other than the convenience it offers.

Src/AutoFixture.NUnit3.UnitTest/ThrowingStubFixture.cs
ploeh: The reason that most Test Doubles are called `DelegatingXyz` is because they _delegate_ their implementation to a (configurable) Func or Action.

AFAICT, this `ThrowingStubFixture` doesn't delegate its implementation. It only throws exceptions upon being called. Isn't the name an accurate description of that? Am I missing something?

moodmosaic: >AFAICT, this `ThrowingStubFixture` doesn't delegate its implementation.

@ploeh Going point.

Based on that, `ThrowingStubFixture` has actually a very precise name. And I shouldn't have proposed to rename it in the first place. @hackle I apologize for this.

moodmosaic:
>I am actually not sure about making 2 methods in AutoDataAttribute protected [..] there is not much semantic value other than the convenience it offers.

The object model in this particular case is based on inheritance (which I think isn't *that* bad given the scope of Glue Libraries and the programming language itself).

So, it shouldn't matter if we can take advantage of this and reduce code duplication by marking those two methods as `protected`.

All in all, I don't see why not, though it'd be great if we have also @ploeh's opinion on this matter.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -64,12 +64,18 @@ private TestCaseParameters GetParametersForMethod(IMethodInfo method)
             }
         }
 
-        private IEnumerable<object> GetParameterValues(IEnumerable<IParameterInfo> parameters)
```
moodmosaic: FWIW, this change is related to what @hackle commented [here](https://github.com/AutoFixture/AutoFixture/pull/589#issuecomment-202308310).

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -64,12 +64,18 @@ private TestCaseParameters GetParametersForMethod(IMethodInfo method)
...
+        /// <summary>
+        /// Get values for a collection of <see cref="IParameterInfo"/>
+        /// </summary>
+        protected virtual IEnumerable<object> GetParameterValues(IEnumerable<IParameterInfo> parameters)
         {
-            return parameters.Select(Resolve);
+            return parameters.Select(GetValueForParameter);
         }
 
-        private object Resolve(IParameterInfo parameterInfo)
```
moodmosaic: FWIW, this change is related to what @hackle commented [here](https://github.com/AutoFixture/AutoFixture/pull/589#issuecomment-202308310).

hackle:
Thanks for the review @moodmosaic all very healthy discussions so no need to be sorry.

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    /// <summary>
+    /// This attribute acts as a TestCaseAttribute but allow incomplete parameter values, 
+    /// which will be provided by AutoFixture. 
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    [CLSCompliant(false)]
+    public class InlineAutoDataAttribute : AutoDataAttribute
```
hackle: Just an idea - I can actually reverse the inheritance here, so **AutoDataAttribute** is just a special case of **InlineAutoDataAttribute** that has zero parameters!
Any thoughts @moodmosaic @ploeh

moodmosaic: In the other Glue Libraries, InlineAutoDataAttribute *is-a* CompositeDataAttribute.

It's just that CompositeDataAttribute isn't implemented yet, and so InlineAutoDataAttribute for the time being may just be an AutoDataAttribute (the base of all auto data-theory attributes).

Once the CompositeDataAttribute is available, I could imagine that this proposed implementation could be refactored and simplified...

ploeh:
(Please accept my apologies that it took me so long even to respond, but I've been on vacation :wine_glass:)

Would it not be possible to implement a `CompositeDataAttribute` like in _AutoFixture.Xunit2?_ It's probably only my own bias speaking, but I consider that library the canonical implementation of the various test framework Glue Libraries that AutoFixture offers. I'd also like to keep API parity if possible.

Looking at the API, however, I can see that it's different because NUnit defines the pivotal method with a return type of `IEnumerable<TestMethod>`. The question, then, is whether or not `TestMethod` is a composable type?

It seems to be a concrete, but non-sealed class, so it may be possible to override appropriate methods in order to achieve the desired behaviour. On the other hand, it may not be. I haven't investigated.

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
ploeh: Wouldn't that be a breaking change?

moodmosaic: I thought that the opposite would be a breaking change.

ploeh: Perhaps I misunderstood your suggestion, but here's what I understood: the current proposal is `InlineAutoDataAttribute : AutoDataAttribute`, which you suggest could later be changed to `InlineAutoDataAttribute : CompositeDataAttribute`.

moodmosaic: I gave this a second thought and you're right, it will be a breaking change.

(The reason is that the CompositeDataAttribute is a *DataAttribute*. I had in mind that it was an AutoDataAttribute...)

hackle: Oh I was suggesting AutoDataAttribute:InlineAutoDataAttribute. But that is  not consistent with the current implementation.

I figure it WilL be possible to compose testcases through reflection? Or better that AutoData returns a derived type of TestCase that somehow exposes certain values for composition. However I need to look harder and experiment.

hackle:
Thanks @ploeh for looking at this on your holiday!

Indeed **TestMethod** is concrete and the best idea to compose I have is to inspect its **Arguments** through reflection:
`internal object[] Arguments`
Although I am not sure if this is a good idea after all.

ploeh:
Agreed; depending on reflection isn't a good idea.

ploeh:
It's entirely possible that `TestMethod` isn't sufficiently composable to make `CompositeDataAttribute` a viable option.

That shouldn't prevent us from supplying an `InlineAutoDataAttribute` for _AutoFixture.NUnit3_, but instead of deriving it from `AutoDataAttribute`, couldn't we derive it directly from `object` instead?

Truth be told, `CompositeDataAttribute` was an interesting idea, but I suspect it may have been over-engineering. At least, I've never used it directly, but only ever used `InlineAutoDataAttribute`. There may be people out that that use it for other purposes, but I suspect they are rare.

If we define `InlineAutoDataAttribute` by not deriving from any class specifically (thus, effectively deriving from `object`), and then later find a way to implement `CompositeDataAttribute`, we can always insert that into the inheritance hierarchy at that point.

hackle:
@ploeh do you mean **InlineAutoDataAttribute** should implement **ITestBuilder** directly?
Maybe I have missed something, but I think your point is somewhat paradoxical - if **CompositeDataAttribute** is considered over-engineering, then why change the structure of inheritance (regardless of whether the inheritance itself is good or not) to leave room for something uncertain in the future? Doesn't that seem like a case of YAGNI?
Any thoughts?

hackle:
Regarding composition - would it be possible to factor out the parameter-generation to a separate interface called like `ITestMethodParametersBuilder` with a signature like `IParameterInfo[] -> object[]` and take things from there?

ploeh:
> do you mean InlineAutoDataAttribute should implement ITestBuilder directly?

Yes, I suppose that was what I meant.

> if **CompositeDataAttribute** is considered over-engineering, then why change the structure of inheritance (regardless of whether the inheritance itself is good or not) to leave room for something uncertain in the future? Doesn't that seem like a case of YAGNI?

Good point. I'm not sure...

As a general rule, I hardly ever use inheritance for reuse. (In fact, I rarely use inheritance for anything.) There are better options for reuse, such as composition.

One exception from this rule regards .NET attributes. You can't use Dependency Injection with attributes, so the most desirable option for applying the [Open Closed Principle](https://en.wikipedia.org/wiki/Open/closed_principle) (OCP) is already out of the question.

The only remaining alternative for applying the OCP, then, is through inheritance (at least, I can't think of other options).

The original intent of `CompositeDataAttribute` wasn't reuse, but enabling extension. As far as I can tell, few people extend `CompositeDataAttribute`, and that's why I consider it over-engineering.

ploeh:
> Regarding composition - would it be possible to factor out the parameter-generation to a separate interface called like `ITestMethodParametersBuilder` with a signature like `IParameterInfo[] -> object[]` and take things from there?

(Ooh, F# type syntax :+1: )

May I suggest, instead, that we start with a copy-pasted implementation?

There are [disadvantages as well as advantages to DRY](http://blog.ploeh.dk/2014/08/07/why-dry), and I'm not sure we gain anything from making things more complicated than they have to be.

If we start with a copy-pasted implementation (covered by tests), we can always look at the result and then refactor in the future if we see an obvious route.

hackle:
@ploeh copy & paste done along some minors changes.

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -0,0 +1,120 @@
...
+            var usageAttribute = typeof(InlineAutoDataAttribute)
+                .GetCustomAttributes(false)
+                .FirstOrDefault(c => c.GetType() == typeof(AttributeUsageAttribute)) as AttributeUsageAttribute;
+
+            Assert.That(usageAttribute, Is.Not.Null);
+            Assert.That(usageAttribute.AllowMultiple, Is.True);
+        }
+
+        [Test]
+        public void InlineAutoDataCanBeExtendedWithImplementationOfIFixture()
```
ploeh: What purpose does this test serve?

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        [Test]
+        public void InlineAutoDataIsAttribute()
+        {
+            Assert.That(new InlineAutoDataAttribute(), 
+                Is.AssignableTo(typeof(Attribute)));
+        }
+
+        [Test]
+        public void InlineAutoDataCanBeUsedInMultiple()
```
ploeh: Isn't this covered by `Scenario.InlineAutoDataCanBeUsedInMultiple`?

Src/AutoFixture.NUnit3.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    [TestFixture]
+    public class InlineAutoDataAttributeTest
+    {
+        [Test]
+        public void InlineAutoDataIsAttribute()
```
ploeh: Isn't this implicitly covered by the various `Scenario` tests?

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+                return new TestCaseParameters(ex);
+            }
+        }
+
+        /// <summary>
+        /// Get values for a collection of <see cref="IParameterInfo"/>
+        /// </summary>
+        private IEnumerable<object> GetParameterValues(IEnumerable<IParameterInfo> parameters)
+        {
+            return this._existingParameterValues.Concat(this.GetMissingValues(parameters));
```
ploeh: What happens if `_existingParameterValues` is `null`?

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            try
+            {
+                var parameters = method.GetParameters();
+
+                var parameterValues = this.GetParameterValues(parameters);
+
+                return new TestCaseParameters(parameterValues.ToArray());
+            }
+            catch (Exception ex)
```
ploeh: Is this branch covered by any tests?

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            return parametersWithoutValues.Select(this.GetValueForParameter);
+        }
+
+        /// <summary>
+        /// Get value for an <see cref="IParameterInfo"/>
+        /// </summary>
+        private object GetValueForParameter(IParameterInfo parameterInfo)
+        {
+            CustomizeFixtureByParameter(this._fixture, parameterInfo);
```
ploeh: Why pass a class field as a parameter to another instance method on the same class?

hackle:
@ploeh Thanks for the review! Have just made changes accordingly.

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        /// <summary>
+        /// Get value for an <see cref="IParameterInfo"/>
+        /// </summary>
+        private object GetValueForParameter(IParameterInfo parameterInfo)
+        {
+            CustomizeFixtureByParameter(parameterInfo);
+
+            return new SpecimenContext(this._fixture)
```
ploeh: What happens if `this._fixture` is `null`?

Src/AutoFixture.NUnit3/InlineAutoDataAttribute.cs
```diff
@@ -0,0 +1,112 @@
...
+                .Resolve(parameterInfo.ParameterInfo);
+        }
+
+        private void CustomizeFixtureByParameter(IParameterInfo parameter)
+        {
+            var customizeAttributes = parameter.GetCustomAttributes<CustomizeAttribute>(false);
+            foreach (var ca in customizeAttributes)
+            {
+                var customization = ca.GetCustomization(parameter.ParameterInfo);
+                this._fixture.Customize(customization);
```
ploeh: What happens if `this._fixture` is `null`?

hackle: I thought about this and my reason for not having a null check for fixture is that the public constructor takes no fixture, and any inheriting class should be responsible for null checking before passing a fixture to the protected constructor.

ploeh: Isn't the presence of an `IFixture` object a precondition for this class to work?

hackle: Makes sense, changes made.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit3 3.44.0](https://www.nuget.org/packages/AutoFixture.NUnit3/3.44.0).

PR #590, opened 2016-03-29
Reorder construction attributes so that [Frozen] is always applied last

sergeyshushlyapin:
Addresses issue #420 simply moving the `[Frozen]` attribute to the end of the custom attribute list.

### DIFF
--- a/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
@@ -75,6 +75,7 @@
     <Compile Include="InlineAutoDataAttributeTest.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Scenario.cs" />
+    <Compile Include="TypeWithCustomizationAttributes.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\AutoFixture.xUnit.net2\AutoFixture.xUnit.net2.csproj">
--- a/Src/AutoFixture.xUnit.net2.UnitTest/TypeWithCustomizationAttributes.cs
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/TypeWithCustomizationAttributes.cs
@@ -4,11 +4,51 @@ namespace Ploeh.AutoFixture.Xunit2.UnitTest
 {
     internal class TypeWithCustomizationAttributes
     {
+        public void CreateWithFrozenAndFavorArrays([Frozen, FavorArrays] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFavorArraysAndFrozen([FavorArrays, Frozen] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFrozenAndFavorEnumerables([Frozen, FavorEnumerables] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFavorEnumerablesAndFrozen([FavorEnumerables, Frozen] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFrozenAndFavorLists([Frozen, FavorLists] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFavorListsAndFrozen([FavorLists, Frozen] ConcreteType sut)
+        {
+        }
+
         public void CreateWithFrozenAndGreedy([Frozen, Greedy] ConcreteType sut)
         {
         }
 
         public void CreateWithGreedyAndFrozen([Greedy, Frozen] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithFrozenAndModest([Frozen, Modest] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithModestAndFrozen([Modest, Frozen] ConcreteType sut)
+        {
+        }
+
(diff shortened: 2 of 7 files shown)


ploeh:
(Please accept my apologies that it took me so long even to respond, but I've been on vacation :wine_glass:)

This suggested fix to the problem is quite a Columbus egg. It surprises me that no one (myself included) has thought of it before.

I'm inclined to pull the code as is. My only reservation is that this solution is so unexpected that feel a bit blind-sided by it, so I wonder if there are issues that I haven't considered. Therefore, I'd like to solicit comments from other reviewers (particularly @moodmosaic, @adamchester, and @ecampidoglio), but if no one has any reservations, I'll pull this with gratitude :smile:

moodmosaic:
I couldn't imagine this being an issue but I think it makes sense — the order of attributes isn't guaranteed. (Also, I can't remember myself using the FrozenAttribute together with other attributes.)

That being said, it shouldn't hurt if we move on and review these changes :smiley:

Src/AutoFixture.xUnit.net2.UnitTest/TypeWithCustomizationAttributes.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    internal class TypeWithCustomizationAttributes
+    {
+        public void CreateWithFrozenAndGreedy([Frozen, Greedy] ConcreteType sut)
+        {
+        }
+
+        public void CreateWithGreedyAndFrozen([Greedy, Frozen] ConcreteType sut)
+        {
+        }
```
moodmosaic: Wouldn't it be safer if we tested with more than 2 attributes? For example Frozen, Greedy, and FavorArrays attributes.

sergeyshushlyapin:
Thanks @ploeh, I'm glad to hear that :)

The only scenario that hasn't been impacted by the fix is the combination of `[Substitute]` and `[Frozen]` attributes:

```csharp
[Theory, AutoNSubstituteData]
public void FrozenSubstituteTest([Substitute][Frozen] ConcreteType sut)
{
  Assert.IsAssignableFrom<ConcreteType>(sut);
  Assert.IsNotType<ConcreteType>(sut); // Fails. `sut` is a concrete type rather than a mock.
}
```

That is because the `SubstituteAttribute` is not inherited from the xUnit's `AutoDataAttibute`. It's based on the  _Residue Collectors_, so I'd treat it as an independent issue.

Src/AutoFixture.xUnit.net2.UnitTest/TypeWithCustomizationAttributes.cs
sergeyshushlyapin: Make sense to me. Added https://github.com/AutoFixture/AutoFixture/pull/590/commits/1e62a5ea36f289a9c3399236d8dab83b1f833ce9.

sergeyshushlyapin:
I applied the fix to the [xUnit1](https://github.com/sergeyshushlyapin/AutoFixture/commit/0f9791fd6feda4ecd56cd5ec2d0baa200dee3759) and [NUnit2](https://github.com/sergeyshushlyapin/AutoFixture/commit/c54afc68eb0bf43dc0458136d9e3d2f2bbe148f0) glue libraries (NUnit3 is [not ready](https://github.com/sergeyshushlyapin/AutoFixture/commit/cb0c80a6fe35ce3f3c40a6f661328b6c138753a4) to be fixed so far). Does it make sense to merge them into this PR?

ploeh:
Let's keep those as separate pull requests. I'll give @adamchester and @ecampidoglio another day to comment or object. If we hear nothing in that time, I'll interpret this as tacit agreement and pull this pull request.

The changes for the other glue libraries we can merge soon thereafter.

Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
```diff
@@ -93,6 +94,9 @@
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
ecampidoglio: This line jumped out to me. I'm not sure whether it was intentional or not, but it seems it was [automatically added by Visual Studio 2013](https://connect.microsoft.com/VisualStudio/feedback/details/800245/vs2013rc-adds-to-vs2012-c-project-section-itemgroup) to tell Test Explorer that this project contains unit tests.

Either way, I don't think this line is related to the rest of the patch in this commit, so it should either be moved to its own commit to explain why it's there or maybe removed altogether. What do you think?

ecampidoglio:
A part from [one minor issue](https://github.com/AutoFixture/AutoFixture/pull/590/files#r59133349) this PR looks like it's ready to be merged 👍

Thank you for providing such an elegant solution to the problem, @sergeyshushlyapin 💯

Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
sergeyshushlyapin: @ecampidoglio, you're right, this line is automatically added every time I run tests in VS Test Explorer. I shouldn't have mixed it with all the other changes, so I'm fine with removing it from the commit.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit2 3.43.1](https://www.nuget.org/packages/AutoFixture.Xunit2/3.43.1).

Issue #19, opened 2012-10-18
Support for value types with default constructors


ploeh:
AutoFixture should optionally support value types with default constructors. See [this discussion](http://stackoverflow.com/q/12930963/126014) for more details.

Since I believe in the [GOOS](http://www.amazon.com/gp/product/0321503627/ref=as_li_ss_tl?ie=UTF8&camp=1789&creative=390957&creativeASIN=0321503627&linkCode=as2&tag=ploeh-20) paradigm, I think it's a beneficial feature that AutoFixture OOB chokes on such constructs, but we should consider elaborating on this potential issue in the error message. E.g. instead of saying

> AutoFixture was unable to create an instance from Foo, most likely because it has no public constructor, is an abstract or non-public type.

we should consider adding more information - e.g.

> AutoFixture was unable to create an instance from Foo, since it's a value type with no explicit, parameterized constructors. Are you attempting to create an instance of a mutable value type? If so, you should strongly consider changing the design of the value type. However, if you are unable to do so, you can add the SupportValueTypeDefaultConstructors customizations to your Fixture instance:
> 
> var fixture = new Fixture().Customize(new SupportValueTypeDefaultConstructors());

This would obviously also require us to add said SupportValueTypeDefaultConstructors customization to the AutoFixture code base.

PR #596, opened 2016-04-06
Handle enum types with no values

### LINKED ISSUE #594: AutoFixture hangs when creating an enum that has no defined values
Given the following:
```
public class MyClass
{
    public EmptyEnum MyProperty { get; set; }
}

public enum EmptyEnum
{
    //having no values here will break AutoFixture
}
```

When creating an instance of MyClass using 
```
var result = new Fixture().Create<MyClass>();
```

AutoFixture will hang until the process is manually killed.

----

The issue lies within the EnumGenerator class; the RoundRobinEnumEnumerable.GetEnumerator method will continually loop without ever hitting the yield return statement.

Question here is what would be the best thing to do in such instances - I'm thinking the RoundRobinEnumEnumerable constructor could throw an InvalidOperationException if an enum has no values?  If an appropriate exception message was thrown then would allow the caller to fix their code, perhaps something like:

> Cannot create a value for EmptyEnum as the enum has no values defined

lowds:
fixes #594 
This stops fixture creation hanging when a type contains a field with an empty enum - field values are initialised to 0 (zero) which appears to be in line with the C# language specification

### DIFF
--- a/Src/AutoFixture/EnumGenerator.cs
+++ b/Src/AutoFixture/EnumGenerator.cs
@@ -96,11 +96,14 @@ namespace Ploeh.AutoFixture
 
             public IEnumerator GetEnumerator()
             {
-                while (true)
+                if (this.values.Any())
                 {
-                    foreach (var obj in this.values)
+                    while (true)
                     {
-                        yield return obj;
+                        foreach (var obj in this.values)
+                        {
+                            yield return obj;
+                        }
                     }
                 }
             }
--- a/Src/AutoFixture/EnumGenerator.cs
+++ b/Src/AutoFixture/EnumGenerator.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using Ploeh.AutoFixture.Kernel;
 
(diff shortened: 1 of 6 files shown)


ploeh:
I haven't consulted the C# language specification regarding this issue, but IIRC a normal `enum` is an `int` at run-time, so the default value is naturally `0`. From that perspective, the proposed behaviour makes sense.

I prefer the suggestion from #594 of throwing an exception, though.

Is there ever a scenario where an empty `enum` represents a proper design choice?

lowds:
I found [this on stack overflow](http://stackoverflow.com/questions/23794948/why-does-the-c-sharp-compiler-allow-empty-enums) when I was looking into it as my first thought was that the compiler should not even allow it

From a testing perspective an argument could be made for having the enum there as a place-holder with no values yet while you're developing the code around it, which is what happened for me and how I came across this, there again could just as easily add a dummy value and replace it later.

Out of the two approaches for this issue ("have it default to zero" or "throw an exception") either would have been fine for me while I was testing my code originally

I could change this pull request to throw an exception if that would be preferable(?)

ploeh:
I'd prefer throwing an exception with a descriptive error message. There's precedence for that in AutoFixtrure: #19.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.43.2](https://www.nuget.org/packages/AutoFixture/3.43.2).

PR #599, opened 2016-04-11
Reorder construction attributes so that [Frozen] is always applied last (xUnit.net)

sergeyshushlyapin:
Fixes issue #420 for [xUnit.net](https://www.nuget.org/packages/AutoFixture.Xunit) glue library.

### DIFF
--- a/Src/AutoFixture.xUnit.net.UnitTest/DelegatingFixture.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/DelegatingFixture.cs
@@ -57,7 +57,8 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
 
         public IFixture Customize(ICustomization customization)
         {
-            throw new NotImplementedException();
+            this.OnCustomize(customization);
+            return this;
         }
 
         public void Customize<T>(Func<Ploeh.AutoFixture.Dsl.ICustomizationComposer<T>, ISpecimenBuilder> composerTransformation)
@@ -101,5 +102,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
         }
 
         internal Func<object, ISpecimenContext, object> OnCreate { get; set; }
+
+        internal Action<ICustomization> OnCustomize { get; set; }
     }
 }
--- a/Src/AutoFixture.xUnit.net.UnitTest/DelegatingFixture.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/DelegatingFixture.cs
@@ -57,8 +57,7 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
 
         public IFixture Customize(ICustomization customization)
         {
-            this.OnCustomize(customization);
-            return this;
+            return this.OnCustomize(customization);
         }
 
         public void Customize<T>(Func<Ploeh.AutoFixture.Dsl.ICustomizationComposer<T>, ISpecimenBuilder> composerTransformation)
@@ -103,6 +102,6 @@ namespace Ploeh.AutoFixture.Xunit.UnitTest
 
         internal Func<object, ISpecimenContext, object> OnCreate { get; set; }
 
-        internal Action<ICustomization> OnCustomize { get; set; }
+        internal Func<ICustomization, IFixture> OnCustomize { get; set; }
     }
 }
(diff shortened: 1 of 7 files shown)


Src/AutoFixture.xUnit.net.UnitTest/DelegatingFixture.cs
```diff
@@ -101,5 +102,7 @@ public object Create(object request, ISpecimenContext context)
         }
 
         internal Func<object, ISpecimenContext, object> OnCreate { get; set; }
+
+        internal Action<ICustomization> OnCustomize { get; set; }
```
ploeh: Why is this an `Action`, and not a `Func`?

sergeyshushlyapin: I might be missing something, but I'm not sure I understand how `Func` can be used here. I need to track that `sut.GetData()` calls the `fixture.Customize()` method twice with proper customizations in proper order. I store the actual *calls* in the `customizationLog` variable.

If I try to use a `Func`, I end up with smth like this:

```csharp
var customizationLog = new List<ICustomization>();
var fixture = new DelegatingFixture();
fixture.OnCustomize = c =>
{
  customizationLog.Add(c);
  return fixture;
};

var sut = new AutoDataAttribute(fixture);
// Exercise system
sut.GetData(method, parameterTypes);
// Verify outcome
Assert.False(customizationLog[0] is FreezeOnMatchCustomization);
Assert.True(customizationLog[1] is FreezeOnMatchCustomization);
// Teardown
```

Actually I still use it as an `Action` but now it takes even more space... Any suggestions how to simplify it? Thanks!

sergeyshushlyapin: If I understand the idea with `Func` correctly, the test could look like the following:
```csharp
var fixture = new DelegatingFixture();
fixture.OnCustomize = c =>
{
    c.Customize(fixture);
    return fixture;
};
//...
Assert.False(fixture.Customizations[0] is FreezeOnMatchCustomization);
Assert.True(fixture.Customizations[1] is FreezeOnMatchCustomization);
```

I tried to build it this way originally but then I realized that the `fixture.Customizations` property is of type `IList<ISpecimenBuilder>` rather than `IList<ICustomization>`. And the order of the *Specimen Builder*'s might not correspond to the original *Customizations* order. That is why I decided to introduce the `customizationLog` list.

ploeh: The method being implemented has the signature `public IFixture Customize(ICustomization customization)`. It returns an object.

The intention of all `DelegatingXyz` classes in the AutoFixture test code base is that they serve as general-purpose Test Doubles. Someone else may, in the future, need to configure the `Customize` method to return a particular object. That isn't going to be possible with the current design, and I'd think it'd be a shame if we make something like that too difficult in the future.

Granted, it makes writing the current test a bit more difficult, but it's more principled, and I don't think that the cost is _that_ big.

You could argue that we ought to apply the YAGNI principle and keep it as you already wrote it, and you'd have a good point. On the other hand, I've seen examples of code bases where a too unprincipled approach to test API design could lead to code that was hard to change. My scars are showing, I know :sweat_smile: 

I can see that I failed to notice this in [your previous pull request](https://github.com/AutoFixture/AutoFixture/pull/590), but I'll go and address that myself.

sergeyshushlyapin: Thanks @ploeh! That definitely make sense!

To be honest it took me a while to figure out how to test this scenario. The first thing that came into my mind was to inject an `IFixture` mock and check if the `Customize()` method is called, but the tests are mocking framework agnostic! I'm just wondering what is your motivation to design tests in such *fundamental* way?

ploeh: Various Glue Libraries (_AutoFixture.AutoMoq_, _AutoFixture.AutoNSubstitute_, etcetera) each rely on their respective dynamic mock libraries (_Moq_, _NSubstitute_, etcetera), so I was concerned that using one of those in other parts would create dependency conflicts. Perhaps it was overkill...

We use _Moq_ in the _AutoFixture.AutoMoq_ codebase, _FakeItEasy_ in the _AutoFixture.AutoFakeItEasy_ code base, and so on.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Xunit 3.43.3](https://www.nuget.org/packages/AutoFixture.Xunit/3.43.3).

Issue #598, opened 2016-04-11
Fix code analysis configuration


sergeyshushlyapin:
I was about to push the issue #420 fix for [xUnit.net](https://github.com/sergeyshushlyapin/AutoFixture/commit/8ed896363aa7be0fcb01b025eab669410ee5e69f) library but it appeared that the build fails if I run `build.cmd` (everything is OK in VS).

After some research I found out that the Code Analysis rules are enabled on the [xUnit.net](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/AutoFixture.xUnit.net.csproj) project:

![image](https://cloud.githubusercontent.com/assets/4719165/14423882/c6e33e62-ffe6-11e5-9431-d11f805de925.png)

... but disabled on the xUnit.net2:
![image](https://cloud.githubusercontent.com/assets/4719165/14423859/ae191c8a-ffe6-11e5-9db7-7eed28e37dc1.png)

Before moving forward with the #420 fixes, I'd like to fix the code analysis configuration. There are couple of questions:

1. Does it make sense to enable code analysis on build (now disabled) so that it reports the same set of errors in VS and Console
![image](https://cloud.githubusercontent.com/assets/4719165/14423444/81ac62b2-ffe4-11e5-8c55-1be60c299774.png)

2. There are two `AutoFixture.ruleset` files, both created about 5 years ago:
  - Src\AutoFixture.ruleset
  - Src\AutoFixture\AutoFixture.ruleset

  The first one is almost empty and used in the `AutoNSubstitute` project only. The second one contains full set of rules configured and used everywhere (except xUnit.net2 ;)). I'd nuke the first one.

sergeyshushlyapin:
In order to fix issue #420 for xUnit.net (PR #599) I had to change the [CustomizeAttributeComparer](https://github.com/AutoFixture/AutoFixture/pull/599/commits/7ae22322ca4d3212de14d658ad1bc9031fe2b2cb) a bit to avoid the following code analysis error:

```
  1) Building S:\autofixture\Src\AutoFixture.xUnit.net.sln failed with exitcode 1.
[...]

PR #603, opened 2016-04-13
Fix code analysis configuration for NUnit2 (#598)

sergeyshushlyapin:
Addresses issue #598 for NUnit2 Glue Library.


Src/AutoFixture.NUnit2/GlobalSuppressions.cs
```diff
@@ -0,0 +1,11 @@
+﻿[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1702:CompoundWordsShouldBeCasedCorrectly", MessageId = "NUnit", Scope = "namespace", Target = "Ploeh.AutoFixture.NUnit2")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1702:CompoundWordsShouldBeCasedCorrectly", MessageId = "NUnit")]
```
ploeh: Why does this suppression appear twice?

sergeyshushlyapin:
There are two errors with the same code but slightly different descriptions reported (assembly and namespace valization):
```
  1) Building S:\autofixture\Src\AutoFixture.NUnit2.sln failed with exitcode 1.
  2) CA1702: (-1,0): Microsoft.Naming : The compound word 'NUnit' in namespace name 'Ploeh.AutoFixture.NUnit2' exists as a discrete term. If your usage is intended to be single word, case it as 'Nunit'.
  3) CA1702: (-1,0): Microsoft.Naming : The compound word 'NUnit' in assembly name 'Ploeh.AutoFixture.NUnit2.dll' exists as a discrete term. If your usage is intended to be single word, case it as 'Nunit'.
```

The suppressions have the same code and name but slightly different *auto-generated* configuration. Removing either of them leads to failure.

ploeh:
Thank you :+1:

PR #604, opened 2016-04-13
Fix: NPE is thrown when without is called on a base class's field.

Mykezero:
I'm not sure if this is desired behavior, but it looks like an exception is raised by `EqualRequestSpecification` when using `Without` on a field that exists in a class's base class. 

Here's the simplified example demonstrating this behavior:

```
public class CoolPerson : CoolPersonality
{            
}
```
```
public class CoolPersonality
{
    public bool IsCool = true;
}
```
```
[Fact]
public void ThrowsNullPointerException_WhenWithoutIsCalledOnABaseClassesField()
{
    Fixture f = new Fixture();
    Assert.Throws<ArgumentNullException>(() => f.Build<CoolPerson>().Without(x => x.IsCool).Create());
}
```

Since the `NodeComposer` doesn't check that the reflected type could potentially be a field, null is passed down the line to `EqualRequestSpecification` where a NPE is raised. 

This was discovered when I was trying to 'mock out' a statically assigned field in a base class to see if AutoFixture + Moq could help me out. 

Sorry for the lack of tests but since I don't understand the architecture of Autofixture that well, it was hard for me to gauge how this change would impact other classes. 

I just wanted to quickly put up this merge request so that I wouldn't forget about it; I'll try to write some tests this weekend!

### DIFF
--- a/Src/AutoFixture/Dsl/NodeComposer.cs
+++ b/Src/AutoFixture/Dsl/NodeComposer.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text;
 using Ploeh.AutoFixture.Kernel;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace Ploeh.AutoFixture.Dsl
 {
@@ -410,7 +411,9 @@ namespace Ploeh.AutoFixture.Dsl
         {
             var m = propertyPicker.GetWritableMember().Member;
             if (m.ReflectedType != typeof(T))
-                m = typeof(T).GetProperty(m.Name);
+            {
+                m = typeof(T).GetProperty(m.Name) ?? (MemberInfo) typeof(T).GetField(m.Name);
+            }
 
             return (NodeComposer<T>)this.ReplaceNodes(
                 with: n => n.Compose(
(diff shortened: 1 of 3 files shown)


ploeh:
Thank you for reporting this :+1: That's quite a surprising omission :flushed: 

I agree that we should also have at least one test that reproduces the defect, but I'd be perfectly happy with a [Facade Test](http://blog.ploeh.dk/2012/06/27/FacadeTest) like the one sketched in the PR comment.

ploeh:
I can see that you've pushed new commits to this pull request. I will commence a review shortly.

TIP: Contrary to most other activities, GitHub doesn't send out notifications when pull requests are updated. This means that it's easy for me (and other maintainers) to forget about it.

Another time, in order to ensure that the reviewer gets a notification about updates to a pull request, you are welcome to leave a comment to that effect. Adding a comment will trigger a notification to all subscribers.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.44.1](https://www.nuget.org/packages/AutoFixture/3.44.1).

Mykezero:
Much appreciated!  :smile:

PR #605, opened 2016-04-14
Reorder construction attributes so that [Frozen] is always applied last (NUnit2)

sergeyshushlyapin:
Fixes issue #420 for NUnit2 Glue Library.


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit2 3.43.4](https://www.nuget.org/packages/AutoFixture.NUnit2/3.43.4).

PR #606, opened 2016-04-14
Fix code analysis configuration for NUnit3 (#598) 

sergeyshushlyapin:
Addresses issue #598 for NUnit3 Glue Library.

When I fixed the code analysis configuration I got the following errors:
```
  1) Building S:\autofixture\Src\AutoFixture.NUnit3.sln failed with exitcode 1.
  2) CA1704: (-1,0): Microsoft.Naming : Correct the spelling of 'Ploeh' in namespace name 'Ploeh.AutoFixture.NUnit3'.
  3) CA1704: (-1,0): Microsoft.Naming : Correct the spelling of 'Ploeh' in assembly name 'Ploeh.AutoFixture.NUnit3.dll'.
  4) CA1813: (-1,0): Microsoft.Performance : Seal 'AutoDataAttribute', if possible.
  5) CA1822: S:\autofixture\Src\AutoFixture.NUnit3\AutoDataAttribute.cs(82,0): Microsoft.Performance : The 'this' parameter (or 'Me' in Visual Basic) of 'AutoDataAttribute.CustomizeFixtureByParameter(IFixture, IParameterInfo)' is never used. Mark the member as static (or Shared in Visual Basic) or use 'this'/'Me' in the method body or at least one property accessor, if appropriate.
  6) CA1031: S:\autofixture\Src\AutoFixture.NUnit3\AutoDataAttribute.cs(63,0): Microsoft.Design : Modify 'AutoDataAttribute.GetParametersForMethod(IMethodInfo)' to catch a more specific exception than 'Exception' or rethrow the exception.
  7) CA1704: (-1,0): Microsoft.Naming : Correct the spelling of 'Enumerables' in type name 'FavorEnumerablesAttribute'.
```

I suppressed the error `CA1031: Do not catch general exception types` in the `AutoDataAttribute.GetParametersForMethod()` method. AFAICS it's by design, but we'd better ask @hackle to confirm this.

### DIFF
--- a/Src/AutoFixture.NUnit3.sln
+++ b/Src/AutoFixture.NUnit3.sln
@@ -1,11 +1,13 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2012
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture", "AutoFixture\AutoFixture.csproj", "{400AC174-9A4A-4C7D-815B-F2A21130A0D1}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixtureUnitTest", "AutoFixtureUnitTest\AutoFixtureUnitTest.csproj", "{01C86B79-C79F-4EC4-B32E-DA5467DBBD31}"
+EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.NUnit3", "AutoFixture.NUnit3\AutoFixture.NUnit3.csproj", "{6539DC8B-7829-478E-A601-7625B0C814E1}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.NUnit3.UnitTest", "AutoFixture.NUnit3.UnitTest\AutoFixture.NUnit3.UnitTest.csproj", "{1E512431-ADC0-4218-873E-E4DF40C97268}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture", "AutoFixture\AutoFixture.csproj", "{400AC174-9A4A-4C7D-815B-F2A21130A0D1}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestTypeFoundation", "TestTypeFoundation\TestTypeFoundation.csproj", "{7FF65787-F462-4204-BB9D-60B0D1BB6CC1}"
 EndProject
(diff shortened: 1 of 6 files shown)


Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -13,6 +13,7 @@ namespace Ploeh.AutoFixture.NUnit3
     /// This implementation is based on TestCaseAttribute of NUnit3
     /// </summary>
     [AttributeUsage(AttributeTargets.Method)]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", Justification = "This attribute is the root of a potential attribute hierarchy.")]
```
ploeh: > This attribute is the root of a potential attribute hierarchy.

Is it?

ploeh: Yes, it is. My mistake :flushed: Please disregard my previous comment.

Src/AutoFixture.NUnit3/AutoDataAttribute.cs
```diff
@@ -48,6 +49,7 @@ public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
             yield return test;
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
```
ploeh: According to [its documentation](https://msdn.microsoft.com/en-us/library/ms182137), this particular rule mustn't be suppressed, but I agree that it still should, in this particular case. I would like, however, a `Justification` text.

sergeyshushlyapin: Good point. Added.

Src/AutoFixture.NUnit3.sln
```diff
@@ -1,13 +1,13 @@
...
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.NUnit3", "AutoFixture.NUnit3\AutoFixture.NUnit3.csproj", "{6539DC8B-7829-478E-A601-7625B0C814E1}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture", "AutoFixture\AutoFixture.csproj", "{400AC174-9A4A-4C7D-815B-F2A21130A0D1}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.NUnit3.UnitTest", "AutoFixture.NUnit3.UnitTest\AutoFixture.NUnit3.UnitTest.csproj", "{1E512431-ADC0-4218-873E-E4DF40C97268}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixtureUnitTest", "AutoFixtureUnitTest\AutoFixtureUnitTest.csproj", "{01C86B79-C79F-4EC4-B32E-DA5467DBBD31}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture", "AutoFixture\AutoFixture.csproj", "{400AC174-9A4A-4C7D-815B-F2A21130A0D1}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture.NUnit3", "AutoFixture.NUnit3\AutoFixture.NUnit3.csproj", "{6539DC8B-7829-478E-A601-7625B0C814E1}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestTypeFoundation", "TestTypeFoundation\TestTypeFoundation.csproj", "{7FF65787-F462-4204-BB9D-60B0D1BB6CC1}"
```
ploeh: What happened to `TestTypeFoundation`?

sergeyshushlyapin: Oops.. I simply lost it.. Just wondering why nothing has failed.

sergeyshushlyapin: Added, thanks for pointing to this! BTW, just noticed that the `AutoFixtureUnitTests` project was originally missing in the solution. I added it mechanically :)

ploeh: > wondering why nothing has failed.

I can explain that :smile: 

The reason is that when you run `build.sh`, or in general just compile any of the other solutions in this repository, `TestTypeFoundation/bin/Release/Ploeh.TestTypeFoundation.dll` is created, and it remains on disk.

Even when the project is missing from the _solution_, it's still there as a project reference, and while the project isn't part of the solution, the `.dll` is still there. This means that you can still compile the project.

If you clean the repository with `git clean -xdf` and then attempt to build _only_ `AutoFixture.NUnit3.sln`, it will fail.

When you run `build.sh`, however, everything still succeeds because other solutions are build before `AutoFixture.NUnit3.sln`, leaving `Ploeh.TestTypeFoundation.dll` behind.

ploeh:

Src/AutoFixture.NUnit3.sln
sergeyshushlyapin: Thanks @ploeh.

>If you clean the repository with git clean -xdf and then attempt to build only AutoFixture.NUnit3.sln, it will fail.

I was thinking about this scenario as well. It appeared that the solutions can *impact* each other during a build and it might be hard to track if smth goes wrong (in this particular case only code review helped ;)).

Shouldn't the `Clean` target be called before compiling _each_ of the solutions?

ploeh: Calling the `Clean` target before compiling each solution isn't going to help, because a solution only cleans itself. If a reference isn't part of a solution, the files will not be removed.

The only solution I can think of is to run `git clean -xdf` before a build, but that has the other drawback that it removes _everything_ you may currently be experimenting with.

PR #610, opened 2016-04-20
Removed old build scripts,

ploeh:
since the AutoFixture project now uses FAKE-based build scripts. The new
scripts have been tested, and been used to deploy new versions of
AutoFixture for more than a handful of versions now.

The old scripts are no longer maintained, and don't work, so it's
confusing for new contributors that they are still around.


moodmosaic:
:+1:

PR #611, opened 2016-04-24
No autoproperty attribute

hackle:
This is a simple port over from AutoFixture.Xnunit2 of NoAutoPropertiesAttribute along with

1. with an extra test in Scenario
2. with DependencyConstraints

AFAIK this should be the last pull request to complete parity with other glue libraries.

### DIFF
--- a/Src/AutoFixture.NUnit3.UnitTest/AutoFixture.NUnit3.UnitTest.csproj
+++ b/Src/AutoFixture.NUnit3.UnitTest/AutoFixture.NUnit3.UnitTest.csproj
@@ -57,6 +57,7 @@
     <Compile Include="ModestAttributeTest.cs" />
     <Compile Include="InlineAutoDataAttributeTest.cs" />
     <Compile Include="MyClass.cs" />
+    <Compile Include="NoAutoPropertiesAttributeTest.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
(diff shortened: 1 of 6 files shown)


ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit3 3.45.0](https://www.nuget.org/packages/AutoFixture.NUnit3/3.45.0).

hackle:
Thanks @ploeh I think the NUnit3 glue library is now complete.

However I think at some point we can look at NUnit 3.2 to see if an upgrade is due? What do you think?

ploeh:
Does NUnit 3.2 add capabilities that would enable us to add features to _AutoFixture.NUnit3?_

hackle:
Sorry for the late reply. Not much really. I was more speaking of the version conflict we had to fight and I am under the impression that this might have been solved in NUnit 3.2

However we do have a working solution now so I don't know if this is absolutely necessary.

ploeh:
When I help steward a code base, I use [Postel's law](https://en.m.wikipedia.org/wiki/Robustness_principle) as a principle for dependency management. If *AutoFixture.NUnit3* doesn't need *NUnit 3.2* to implement the desired features, there's no reason to artificially restrict clients' options. Clients who *do* wish to use *NUnit 3.2* can do so, and add a binding redirect.

Issue #532, opened 2016-02-20
Make some AutoData constructor overloads protected


ploeh:
Like originally discussed in #458, some constructor overloads on the `[AutoData]` attributes are only meaningful if you derive from the attribute. This is the case for the constructor overloads that take an `IFixture` instance as argument:

```C#
public AutoDataAttribute(IFixture fixture)
```

Since you can't create an instance of `IFixture` when adorning a type or member with an attribute, this constructor overload is (close to) useless.

While it's fairly harmless that these constructor overloads are `public`, it may cause confusion with users. Making these constructor overloads `protected` would clearly signal that these overloads are only intended for use when deriving from the attribute.

This is a breaking change, so should only be done in AutoFixture 4. It can, however, be completed already today, by committing only to the *v4* branch.

This change should be made for both _AutoFixture.Xunit_, _AutoFixture.Xunit2_, and _AutoFixture.NUnit2_.

ploeh:
Done in #534

PR #613, opened 2016-04-25
Reorder construction attributes so that [Frozen] is always applied last (NUnit3)

sergeyshushlyapin:
Fixes issue #420 for NUnit3 Glue Library.

Question: I noticed that the `AutoDataAttribute(IFixture fixture)` constructor is *protected* in the [NUnit3](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.NUnit3/AutoDataAttribute.cs#L33) library but it's public in all the [other](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/AutoDataAttribute.cs#L59) [Glue](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.NUnit2/AutoDataAttribute.cs#L56) [Libraries](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/AutoDataAttribute.cs#L56). Is there any reason for this?

### DIFF
--- a/Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/AutoDataAttributeTest.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using Ploeh.TestTypeFoundation;
 using NUnit.Framework;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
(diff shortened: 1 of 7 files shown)


moodmosaic:
>Question: I noticed that the `AutoDataAttribute(IFixture fixture)` constructor is *protected* in the [NUnit3](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.NUnit3/AutoDataAttribute.cs#L33) library but it's public in all the [other](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/AutoDataAttribute.cs#L59) [Glue](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.NUnit2/AutoDataAttribute.cs#L56) [Libraries](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/AutoDataAttribute.cs#L56). Is there any reason for this?

:+1: 

It'll be a breaking change on the `master` branch, but they are already changed in the `v4` branch. See also https://github.com/AutoFixture/AutoFixture/issues/532 and https://github.com/AutoFixture/AutoFixture/pull/534.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.NUnit3 3.45.1](https://www.nuget.org/packages/AutoFixture.NUnit3/3.45.1).

PR #616, opened 2016-05-03
set CodeAnalysisCulture to en-US for AutoFixture.xUnit.net2

adamralph:
I originally fixed this for all the projects in #200 but I guess AutoFixture.xUnit.net2 was added since then.

Before this fix, the build failed on my machine (which has en-GB set as the OS language) because it detects "favor" as a spelling mistake (the British spelling is "favour").


moodmosaic:
Good catch! Thank you :+1:

Merging this so that we're consistent with the rest of the changes made in #200.

PR #617, opened 2016-05-03
constrain FakeItEasy version to less than 2.0

adamralph:
FakeItEasy 2.0 (as the version implies) has breaking changes.

Unfortunately, the current version of AutoFixture.AutoFakeItEasy is incompatible with FakeItEasy 2.0 because of these two lines of code

* https://github.com/AutoFixture/AutoFixture/blob/98d7b4d00ba05e406796e32b560b5975f1f9af78/Src/AutoFakeItEasy/FakeItEasyBuilder.cs#L76
* https://github.com/AutoFixture/AutoFixture/blob/98d7b4d00ba05e406796e32b560b5975f1f9af78/Src/AutoFakeItEasy/FakeItEasyRelay.cs#L85

This is because `FakeItEasy.Configuration.IHideObjectMembers` has been renamed to `FakeItEasy.IHideObjectMembers`.

This is, of course, easily fixed, but I believe the best course of action, given that the FakeItEasy 2.0 release is imminent, is to, as soon as possible, release a patch of AutoFixture.AutoFakeItEasy which restricts the FakeItEasy dependency to less than 2.0.

After FakeItEasy 2.0 is released, a version of AutoFixture.AutoFakeItEasy may be released with those two lines of code fixed, and with the FakeItEasy dependency restricted to greater than 2.0. (If you are following SemVer then I guess this would also imply a major version bump of AutoFixture.AutoFakeItEasy since the upgrade would force a major upgrade of FakeItEasy.)

### DIFF
--- a/Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
+++ b/Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
@@ -8,7 +8,7 @@
         <owners>Nikos Baxevanis</owners>
         <dependencies>
             <dependency id="autofixture" version="@build.number@" />
-            <dependency id="FakeItEasy" version="1.7" />
+            <dependency id="FakeItEasy" version="[1.7,2)" />
         </dependencies>
         <summary>Turns AutoFixture into an Auto-Mocking Container based on FakeItEasy.</summary>
         <description>This extension turns AutoFixture into an Auto-Mocking Container. The mock instances are created by FakeItEasy. To use it, add the AutoFakeItEasyCustomization to your Fixture instance. Read more at http://autofixture.codeplex.com/</description>


moodmosaic:
Thanks for the heads up :+1:

I think for now the best thing we can do is merge the changes proposed in this Pull Request.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoFakeItEasy 3.45.2](https://www.nuget.org/packages/AutoFixture.AutoFakeItEasy/3.45.2).

blairconrad:
This has long since been resolved, but the release of AutoFakeItEasy2 brought it to my attention. Why does the code cast to `IHideObjectMembers`? It seems to be some sort of safeguard. Is the concern that someone will have registered an `ISpecimanBuilder` that returns something other than a new `Fake<T>`?
If so, why trust that anything implementing `IHideObjectMembers` would be satisfactory? I'd've thought that checking `fake != null && fake.GetType().IsFake()` could've replaced the cast to `IHideObjectMembers`. Perhaps I've missed something?

moodmosaic:
>I'd've thought that checking `fake != null && fake.GetType().IsFake()` could've replaced the cast to `IHideObjectMembers`

Feel free to give it a shot :smiley:

It'd be nice if we can do it in both AutoFixture.AutoFakeItEasy and AutoFixture.AutoFakeItEasy2 Glue Libraries, if possible.

blairconrad:
@moodmosaic, I can give it a go, but amn't sure how to trigger the right failure conditions. I'll at least try and run the tests.

> It'd be nice if we can do it in both AutoFixture.AutoFakeItEasy and AutoFixture.AutoFakeItEasy2 Glue Libraries, if possible.

If it works, we won't need AutoFakeItEasy2

PR #620, opened 2016-05-11
Fixing InvalidCastException in ConstructorInitializedMemberAssertion

Tungsten78:
In a previous PR (#516), I incorrectly assumed the value would be castable to int.   As a result, when dealing with enums of type byte an exception would result.

Fix is now type agnostic.

### DIFF
--- a/Src/Idioms/ConstructorInitializedMemberAssertion.cs
+++ b/Src/Idioms/ConstructorInitializedMemberAssertion.cs
@@ -266,7 +266,7 @@ namespace Ploeh.AutoFixture.Idioms
             {
                 //Check for a default value only enum
                 var values = Enum.GetValues(type);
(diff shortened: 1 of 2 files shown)


adamchester:
Thank you! It looks good.

Although unlikely, it's possible that we accidentally change the implementation in the future to only support `int` and `byte` underlying types. Therefore, I believe we should add tests for all possible underlying types of an enum. According to [MSDN](https://msdn.microsoft.com/en-us/library/sbbt4032.aspx?f=255&MSPPError=-2147217396), they are:

> The approved types for an enum are byte, sbyte, short, ushort, int, uint, long, or ulong.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.Idioms 3.45.3](https://www.nuget.org/packages/AutoFixture.Idioms/3.45.3).

PR #621, opened 2016-05-12
Introduce Lambda Expression Tree generator

sergeyshushlyapin:
This PR enables AutoFixture to generate expression trees:
```csharp
[Theory, AutoData]
public void GenerateFunc(Expression<Func<bool>> expression)
{
}

[Theory, AutoData]
public void GenerateAction(Expression<Action> expression)
{
}
```

### DIFF
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -100,6 +100,7 @@
     <Compile Include="DomainName.cs" />
     <Compile Include="DomainNameGenerator.cs" />
     <Compile Include="ElementsBuilder.cs" />
+    <Compile Include="LambdaExpressionGenerator.cs" />
     <Compile Include="FixtureRepeater.cs" />
     <Compile Include="InvariantCultureGenerator.cs" />
     <Compile Include="EmailAddressLocalPart.cs" />
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -72,11 +72,11 @@ namespace Ploeh.AutoFixture
                                         this),
                                     new OrRequestSpecification(
                                         new ExactTypeSpecification(
-                                            typeof (Fixture)),
+                                            typeof(Fixture)),
                                         new ExactTypeSpecification(
-                                            typeof (IFixture)),
+                                            typeof(IFixture)),
                                         new ExactTypeSpecification(
-                                            typeof (ISpecimenBuilder)))),
+                                            typeof(ISpecimenBuilder)))),
                                 new StableFiniteSequenceRelay(),
                                 new FilteringSpecimenBuilder(
                                     new Postprocessor(
@@ -108,6 +108,7 @@ namespace Ploeh.AutoFixture
                                 new StringLengthAttributeRelay(),
                                 new RegularExpressionAttributeRelay(),
                                 new EnumGenerator(),
+                                new LambdaExpressionGenerator(),
                                 CreateDefaultValueBuilder(CultureInfo.InvariantCulture),
                                 CreateDefaultValueBuilder(Encoding.UTF8),
                                 CreateDefaultValueBuilder(IPAddress.Loopback))),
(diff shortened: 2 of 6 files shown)


moodmosaic:
This looks good! :+1:

ploeh:
Thank you for submitting this pull request. This looks interesting.

As far as I can tell, the generated actions work:

```C#
[Fact]
public void UseActionExpression()
{
    var fixture = new Fixture();
    var actual = fixture.Create<Expression<Action<string, int>>>();
    Assert.DoesNotThrow(() => actual.Compile()("foo", 42));
}
```

On the other hand, it looks like the functions don't work:

```C#
[Fact]
public void CreateFuncExpression()
{
    var fixture = new Fixture();
    var actual = fixture.Create<Expression<Func<object>>>();
    Assert.NotNull(actual.Compile()());
}
```

This test throws the following exception:

````
Test 'Ploeh.AutoFixtureUnitTest.FixtureTest.CreateFuncExpression' failed: System.InvalidOperationException : variable '' of type 'System.Object' referenced from scope '', but it is not defined
	at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
	at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
	at System.Linq.Expressions.ParameterExpression.Accept(ExpressionVisitor visitor)
	at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
	at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
	at System.Linq.Expressions.Compiler.VariableBinder.VisitLambda[T](Expression`1 node)
	at System.Linq.Expressions.Expression`1.Accept(ExpressionVisitor visitor)
	at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
	at System.Linq.Expressions.Compiler.VariableBinder.Bind(LambdaExpression lambda)
	at System.Linq.Expressions.Compiler.LambdaCompiler.Compile(LambdaExpression lambda, DebugInfoGenerator debugInfoGenerator)
	FixtureTest.cs(5879,0): at Ploeh.AutoFixtureUnitTest.FixtureTest.CreateFuncExpression()
```

I don't know what the intent of this feature is, but I'm quite certain that someone will attempt to do what I just did, and file a bug on funcs.

sergeyshushlyapin:
Hm.. My requirement was to get _some_ expression tree :)
Good catch, I'll take a look! Thanks!

ploeh:
I don't know if this is relevant, but AutoFixture already has [a Builder for actions and funcs](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/DelegateGenerator.cs). Perhaps you can leverage that, or some of its code.

sergeyshushlyapin:
Yes, I was looking for that, thanks!

sergeyshushlyapin:
Fixed `Func`.
Will try to figure out how to use the `DelegatingSpecimenContext` [here](https://github.com/AutoFixture/AutoFixture/pull/621/commits/4130a42406b77ac12acbd9e0614062b0add9a55b#diff-5055be5b604f21a7d499f7fbe6d9ecbdR75) instead of the real `Fixture` instance.

sergeyshushlyapin:
Looks like using the real `SpecimenContext` is more robust than trying to configure `DelegatingSpecimenContext` in the [CreateWithExpressionRequestReturnsCorrectResult](https://github.com/sergeyshushlyapin/AutoFixture/blob/4130a42406b77ac12acbd9e0614062b0add9a55b/Src/AutoFixtureUnitTest/LambdaExpressionGeneratorTest.cs#L75) test.. If there are no objections, I will leave the current implementation.

ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.46.0](https://www.nuget.org/packages/AutoFixture/3.46.0).

sergeyshushlyapin:
Great :)

I'm just a couple minutes late with my tiny test improvement. Please take a look on [this](https://github.com/sergeyshushlyapin/AutoFixture/commit/9c6f4ea44afefb934306ead89992e6b40f428aaf) commit which replaces the real `new SpecimenContext(new Fixture())` call with the `DelegatingSpecimenContext`. If it worth to be merged, I'll send a new PR.

ploeh:
In fact, I don't mind the [Facade Test](http://blog.ploeh.dk/2012/06/27/FacadeTest).

sergeyshushlyapin:
Ok, definitely may sense to me. With the facade test the code looks cleaner.
Then we're totally done with this task. Thanks for review and valuable comments!

PR #626, opened 2016-05-25
Add auto-generated <Service /> item in order to allow VS Test Explorer identify unit tests

sergeyshushlyapin:
This PR contains auto-generated code that appears in the `AutoFixture.xUnit.net2.UnitTest.csproj` file every time one runs unit tests using VS Test Explorer. This question is well described on [SO](http://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for).


ploeh:
Thank you for your contribution. Since there's no change in behaviour associated with this pull request, I haven't released a new version of AutoFixture.

PR #627, opened 2016-05-25
Fix [Substritute] parameter creating when used in combination with [Frozen] attribute

sergeyshushlyapin:
As it was stated [earlier](https://github.com/AutoFixture/AutoFixture/pull/590#issuecomment-207034831), the combination of the `[Frozen]` and `[Substitute]` parameter attributes does not play well. The following test fails:
```csharp
[Theory, AutoNSubstituteData]
public void FreezeConcreteTypeSubstitute([Frozen, Substitute] ConcreteType ct1, ConcreteType ct2)
{
    Assert.Same(ct1, ct2);
    Assert.True(ct1.GetType().BaseType == typeof(ConcreteType)); // fails
    // A real instance is created instead of a mock.
}
```

This POC addresses that issue but requires a list of significant (but not breaking) changes. I'd appreciate if you could share you oppinion about the following questions:

1. There are no *integration* projects in the AutoFixture code base that would allow me to create a test that uses both `xUnit` and `NSubstitute` Glue Libraries at the same time. I'm almost sure the reason for this is in the enormous amount of the possible combinations (4 test framework libs * 4 mocking libs = 16 potential 'integration test' libraries). For the time being I created new `AutoFixture.xUnit.net.NSubstitute` solution.

2. The [Substitute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/SubstituteAttribute.cs) attribute cannot be an empty 'marker' attribute anymore. It should inherit the `CustomizeAttribute` and apply the [ConcreteSubstituteCustomization](https://github.com/sergeyshushlyapin/AutoFixture/blob/frozen-substitute/Src/AutoFixture.xUnit.net.NSubstitute/FrozenSubstituteTest.cs#L36) customization on fly.

3. There are numerous `CustomizeAttribute` abstract classes located in the [different](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/CustomizeAttribute.cs) [Glue](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net/CustomizeAttribute.cs) [Libraries](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.NUnit2/CustomizeAttribute.cs). In order to accomplish point 2 a single abstraction is needed in the main AutoFixture assembly. Likely all the classes are identical so creating a common abstraction won't be a breaking change.


sergeyshushlyapin:
This [Facade test](https://github.com/sergeyshushlyapin/AutoFixture/blob/frozen-substitute/Src/AutoNSubstituteUnitTest/ConcreteClassNSubstituteCustomizationTest.cs#L43) gives me enough confidence that the issue is fixed. I guess the integration project `AutoFixture.xUnit.net.NSubstitute` can be removed now.

sergeyshushlyapin:
Hi there,
My POC suddenly turned on into a complete fix that you're welcome to start reviewing. Please excuse my resistless desire to clean up the code immediately removing redundant 'using' statements and spaces. I know that is not [recommended](https://github.com/AutoFixture/AutoFixture/blob/master/CONTRIBUTING.md) but the changes are not numerous and are pretty minor so they should not make a lot of 'noise' :)

Regarding the code, the [SubstituteAttributeRelay](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/SubstituteAttributeRelay.cs) class is not required any more. It's still in the code base, I haven't obsolete it, but it won't be used. In order to stop calling it from the [AutoNSubstituteCustomization](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/AutoNSubstituteCustomization.cs#L43) class without any fear to break something we might get back to the idea of creating [integration test](https://github.com/sergeyshushlyapin/AutoFixture/blob/integration-test/Src/AutoFixture.xUnit.net.IntegrationTest/NSubstituteIntegrationTest.cs) project.

Src/AutoFixture/CustomizeAttribute.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Reflection;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Base class for customizing parameters in methods.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = true)]
+    public abstract class CustomizeAttribute : Attribute
```
ploeh: To be direct, I don't think this class belongs in the 'core' AutoFixture library.

Can we think of a single, realistic usage scenario where this class is useful in itself, without being tied to one of the unit testing Glue Libraries?

sergeyshushlyapin: No, I cannot find such a scenario either. I agree that it might look a bit strange.

I noticed two attribute design patterns in AutoFixture:

1. "Lightweight" marker attributes with no logic that rely on *Relays* to create specimens. Except the [SubstituteAtribute](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/SubstituteAttribute.cs#L18), all the Data Annotation customizations are implemented this way. See [RangeAttributeRelay](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DataAnnotations/RangeAttributeRelay.cs), [RegularExpressionAttributeRelay](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DataAnnotations/RegularExpressionAttributeRelay.cs).

2. "Heavy" attributes that perform all the fixture customization logic by themselves. That is the way all the Glue Library attributes like [Greedy](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/GreedyAttribute.cs#L31) or [Frozen](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/FrozenAttribute.cs) are implemented.

My intention was to make the attributes look 'similar' so that I could [manage](https://github.com/AutoFixture/AutoFixture/pull/590) them easily. I achieved this introducing the common `CustomizeAttribute` abstraction.

But it appeared that I probably need to look into the opposite way - getting rid of all the `CustomizeAttribute` classes in the Glue Libraries and replacing them with Relays. That would allow not to introduce new doubtful abstractions and prevent copy pasting attribute logic in GLs.

ploeh: That's a good point; I certainly wouldn't mind moving towards [Passive Attributes](http://blog.ploeh.dk/2014/06/13/passive-attributes) if possible. Normally, however, Passive Attributes rely on an extensible filtering mechanism, like I've described in the linked article. Otherwise, users can't plug in their own Passive Attributes.

So far, the unit testing frameworks targeted by AutoFixture don't have such a filtering mechanism. In both xUnit.net and NUnit, the extensibility points seem to revolve around Active Attributes. I must admit, though, that I haven't studied the extensibility mechanism of either xUnit.net 2 or NUnit 3 enough to be sure. This is certainly an area where I'd love to be proven wrong.

sergeyshushlyapin: It looks I got stuck creating the Greedy Passive Attribute and [GreedyAttributeRelay](https://github.com/sergeyshushlyapin/AutoFixture/blob/passive-attributes/Src/AutoFixture.xUnit.net2/GreedyAttributeRelay.cs#L29). In the Relay I need to resolve the specimen using the original `request` (which contains the specimen type) and `GreedyConstructorQuery`:

```csharp
var methodInvoker = new MethodInvoker(new GreedyConstructorQuery());
return context.Resolve(request /*, methodInvoker*/);
```

Just cannot figure out how to create such kind of specimen. Any help is highly appreciated! Thanks.

sergeyshushlyapin: [Solved](https://github.com/sergeyshushlyapin/AutoFixture/blob/passive-attributes/Src/AutoFixture.xUnit.net2/GreedyAttributeRelay.cs#L29). Moving forward.

sergeyshushlyapin:
Closing this PR in favor of a new one which should solve the issue using Passive Attributes.

Issue #624, opened 2016-05-18
Fill SortedSet


ploeh:
When you ask AutoFixture for a hash set, it provides a non-empty value:

```C#
var fixture = new Fixture();
var actual = fixture.Create<HashSet<string>>();
Assert.NotEmpty(actual); // Succeeds
```

When you request a sorted set, on the other hand, the returned set is empty. This test fails its assertion:

```C#
var fixture = new Fixture();
var actual = fixture.Create<SortedSet<string>>();
Assert.NotEmpty(actual); // Fails
```

Since AutoFixture can already supply non-empty collections, and one of the constructor overloads for `SortedSet<T>` takes a `IEnumerable<T>`, it ought to be fairly easy to add a _relay_ from `SortedSet<T>` to `IEnumerable<T>`.

The implementation could for example be a new `ISpecimenBuilder` that handles requests for any `SortedSet<T>` by resolving `IEnumerable<T>` from the `context`, and then invoking the appropriate constructor overload with the returned value.

Another implementation option is to mimic the way hash sets are already filled. See the combination of `EnumerableFavoringConstructorQuery` and `HashSetSpecification` in the `Fixture` class.

ploeh:
See also #622 and #623.

PR #629, opened 2016-05-28
Add fill sorted set to fixture customizations.

Mykezero:
This pull request addresses issue #624 with the option to fill sorted sets. A new customization has been added to the Fixture class that uses the new SortedSetSpecification to start filling sorted set instances.

### DIFF
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -5130,6 +5130,30 @@ namespace Ploeh.AutoFixtureUnitTest
             // Teardown
         }
 
+        [Fact]
+        public void CreateHashSetReturnsCorrectResult()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            // Exercise system
+            var result = fixture.Create<HashSet<string>>();
+            // Verify outcome
+            Assert.NotEmpty(result);
+            // Teardown
+        }
+
+        [Fact]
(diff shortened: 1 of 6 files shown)


Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -4960,6 +4960,20 @@ public void CreateAnonymousHashSetWithCustomizationsReturnsCorrectResult()
         }
 
         [Fact]
+        public void CreateAnonymousSortedSetWithCustomizationsReturnsCorrectResult()
```
ploeh: Do we need this test?

ploeh:
This is looking good. As you can see from my single comment, my only concern is that there may be code we can delete from this pull request :smile:

Mykezero:
Thanks @ploeh, It looks like that scenario is properly covered by the other tests so it should be safe to remove it. I've updated the PR with that change.

ploeh:
Thank you for your contribution :+1: It's now live as [AutoFixture 3.47.0](https://www.nuget.org/packages/AutoFixture/3.47.0).

When someone contributes a new feature, I usually announce it on Twitter, giving the contributor a mention. If you'd like that, let me know what your Twitter handle is. This is entirely voluntary; it's only if you'd like the publicity :wink:

Mykezero:
That would be pretty cool but I don't have a twitter account. Thanks for letting me help out and definitely keep those Jump In issue coming. :smile:

Issue #630, opened 2016-05-30
AutoConfiguredNSubstituteCustomization could randomly fail due to Task inlining


Zvirja:
__Steps to reproduce:__

- Create interface with non-void method.
- Create interface with method that results interface from point 1:

```csharp
public interface IInterfaceWithMethod
{
    string Method(string s);
}

public interface IInterfaceWithMethodSource
{
    IInterfaceWithMethod Get();
}
```

- Resolve `IInterfaceWithMethodSource` and later call the `Get()` method to get `IInterfaceWithMethod` instance:
```csharp
var fixture = new Fixture().Customize(new AutoConfiguredNSubstituteCustomization());
var interfaceSource = fixture.Create<IInterfaceWithMethodSource>();
var instance = interfaceSource.Get();
```

__Result:__

Sometimes code fails with the following exception:
```
NSubstitute.Exceptions.CouldNotSetReturnException

Could not find a call to return from.
Make sure you called Returns() after calling your substitute (for example: mySub.SomeMethod().Returns(value)).
If you substituted for a class rather than an interface, check that the call to your substitute was on a virtual/abstract member.
Return values cannot be configured for non-virtual/non-abstract members.

   at NSubstitute.Core.SubstitutionContext.LastCallShouldReturn(IReturn value, MatchArgs matchArgs)
```

Issue happens because code that creates `IInterfaceWithMethod` mock (when you call the `IInterfaceWithMethodSource.Get()` method) could be executed on the same thread. As result, mock creation will corrupt the current NSubstitute state and it will be unable to setup result later.

Technically, it happens because [this code](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs#L207) is not always executed on a new thread. It could happen that task is inlined on the current thread. See this [stackoverflow discusstion](http://stackoverflow.com/questions/12245935/is-task-factory-startnew-guaranteed-to-use-another-thread-than-the-calling-thr).

Zvirja:
Hi guys,

Likely, that is not a proper question here, but could you please review the bug and PR? This is rather a critical issue for my solution, because when I run my tests under dotCover, almost 30 tests becomes red and I cannot measure code coverage properly. Therefore, it could be great if this fix could be merged and published to NuGet, so I could update assembly.

If you have any questions or notes regarding the bug/PR - let me know.

ploeh:
[...]

PR #631, opened 2016-05-30
Fix bug 630 by forcing Task to be never inlined

Zvirja:
This PR fixes [issue 630](https://github.com/AutoFixture/AutoFixture/issues/630).

I applied a trick to ensure that nested task is never inlined on current thread.

### DIFF
--- a/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
+++ b/Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using NSubstitute;
 using NSubstitute.Core;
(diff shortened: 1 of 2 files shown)


Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
```diff
@@ -204,9 +205,20 @@ private object Resolve(Type type)
...
                 // that return a circular reference.
                 // See discussion at https://github.com/AutoFixture/AutoFixture/pull/397
-                return Task.Factory
-                    .StartNew(() => Context.Resolve(type))
-                    .Result;
+                var task = Task.Factory.StartNew(() => Context.Resolve(type));
+
+                // It could happen that task above is inlined on the current thread.
+                // As result, the last NSubstitute call router could become empty.
+                // Therefore, we pass cancellation token to the Wait() method to prevent task from being inlined.
```
ploeh: Wouldn't it be simpler to explicitly start a new thread, using `ThreadPool.QueueUserWorkItem` or `Thread.Start`?

ploeh:
Ping @mrinaldi (if you have time) :innocent:

Src/AutoNSubstitute/NSubstituteVirtualMethodsCommand.cs
Zvirja: Definitely makes sense, overlooked that.
It's better to use ThreadPool directly, rather than use tasks and apply tricks to prevent inlining.

ploeh:
Thank you for your contribution :+1: It's now live as [AutoFixture.AutoNSubstitute 3.47.3](https://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.47.3).

mrinaldi:
Very interesting issue.

A little too late, but LGTM.
Thank you @Zvirja for fixing it.

PR #634, opened 2016-06-01
Corrected typo in ObjectCreationException error message

DavidBrower:
On receiving an ObjectCreationException the exception message includes the following text: "if you have a concrete class deriving from the abstractclass". There should be a space between "abstract" and "class". This has been added.


ploeh:
Thank you for your contribution :+1: It's now live as [AutoFixture 3.47.1](https://www.nuget.org/packages/AutoFixture/3.47.1).

Issue #622, opened 2016-05-18
Fill SortedLists


ploeh:
When you ask AutoFixture for a dictionary, it provides a non-empty value:

```C#
var fixture = new Fixture();
var actual = fixture.Create<IDictionary<int, string>>();
Assert.NotEmpty(actual); // Succeeds
```

When you request a sorted list, on the other hand, the returned list is empty. This test fails its assertion:

```C#
var fixture = new Fixture();
var actual = fixture.Create<SortedList<int, string>>();
Assert.NotEmpty(actual); // Fails
```

Since AutoFixture can already supply non-empty dictionaries, and one of the constructor overloads for `SortedList<TKey, TValue>` takes a dictionary, it ought to be fairly easy to add a _relay_ from `SortedList<TKey, TValue>` to `IDictionary<TKey, TValue>`.

The implementation could for example be a new `ISpecimenBuilder` that handles requests for any `SortedList<TKey, TValue>` by resolving `IDictionary<TKey, TValue>` from the `context`, and then invoking the appropriate constructor overload with the returned value.

Another implementation option is to mimic the way concrete dictionaries are already filled. See the combination of `DictionaryFiller` and `DictionarySpecification` in the `Fixture` class.

ploeh:
This issue was prompted by [this Stack Overflow question](http://stackoverflow.com/q/37218084/126014).

ploeh:
See also #623 and #624.

sbrockway:
@ploeh I'd like to take a shot at implementing this.  I'm a newbie to git.  I forked and have it going in VS locally.  I assume I send a pull request when it is ready.  Do you assign the issue to me, or is that not necessary?  I don't see a way to assign it to myself.

adamchester:
@sbrockway announcing you're going to have a go at it should be enough, assigning shouldn't be necessary :+1:

ploeh:
@sbrockway, @adamchester is right: making the announcement is usually enough. AutoFixture is a moderately successful open source project, so it sees a steady trickle of contributions, but I don't think I've ever seen a case where several contributors compete to do the same feature :smile: 

Other, more popular, projects may do things differently, so you did the right thing by asking :+1: 

If you need help with anything, please ask :smile:

sbrockway:
I'm having trouble learning how to use GitHub.  Apologies for the pull request yesterday, I was trying to re-base my fork and it mis-guessed which direction the pull request was moving.  I figured it out and got my repo up to date.
[...]

PR #635, opened 2016-06-01
Add support for SortedList<,> Issue #622

sbrockway:
Implemented a ISpecimenBuilder relay as @ploeh suggested in the StackOverflow question linked to Issue #622 .
I noticed the obsolete NoSpecimen(request), but followed the pattern I saw elsewhere of using it with disabling warning 618.
I can make the switch to the 0 param ctor for my implementation if you'd prefer.

Let me know if I missed anything.  My first OSS try.  I want to learn if I missed a step in process/convention.
Or if I missed something in the implementation.
Thanks for the opportunity with the Jump In label on the issue.


sbrockway:
Thanks as well to @Mykezero.  I learned from, and mimiced what you did for #624

ploeh:
Thank you for your pull request! Looking through the commit history, it looks like you started implementing the feature using a specification, but then changed the implementation to use a relay instead. What was the motivation for doing that?

sbrockway:
@ploeh I started out mimicing what @Mykezero did which was using the specification.  I created a DictionaryFavoringConstructorQuery : IMethodQuery hoping that it would receive and handle the request, but it did not work.  I registered a snippet in the map building section to use MethodInvoker but as I recall, it never invoked my DictionaryFavoringConstructorQuery.  I should have checked it in, so there would be a record to review.

ploeh:
Now that #629 already chose the route of using Specifications, I'd be keen on pursuing that way to solve the issue here as well, if at all possible. Otherwise, I'm a bit concerned that it'll be confusing to maintainers that there are more than one way of doing seemingly related things.

Is there anything I can do to help going in that direction?

sbrockway:
I'll try the DictionaryFavoringConstructorQuery again, and reach out if I get stuck again.

PR #636, opened 2016-06-01
Fixed #633

### LINKED ISSUE #633: RandomBooleanSequenceGenerator is biased towards `false`
I was playing around with `RandomBooleanSequenceCustomization` when I noticed that, more often that not, I was getting a `false` value. Roughly 66% of the time.

So I looked into it, and it turns out that the customization defers to its [generator counterpart (`RandomBooleanSequenceGenerator`)][0], which in turn defers to `RandomNumericSequenceGenerator`. Here are the relevant bits:

```cs
this.randomBooleanNumbers =
    new RandomNumericSequenceGenerator(0, 2);
```

```cs
private bool GenerateBoolean(ISpecimenContext context)
{
    return (int)this.randomBooleanNumbers.Create(typeof(int), context) == 0;
}
```

The problem seems to be that the upper bound passed to `RandomNumericSequenceGenerator` (2) is inclusive, not exclusive. Which means we'll either get 0, 1 or 2. And because 0 is translated to `true`, and 1 and 2 to `false`, the random boolean generator ends up being biased towards `false`.

I don't think this was the intended behavior. Am I correct in assuming this is a bug?

 [0]: https://github.com/AutoFixture/AutoFixture/blob/98d7b4d00ba05e406796e32b560b5975f1f9af78/Src/AutoFixture/RandomBooleanSequenceGenerator.cs

dcastro:
This fixes #633

### DIFF
--- a/Src/AutoFixture/RandomBooleanSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomBooleanSequenceGenerator.cs
@@ -17,7 +17,7 @@ namespace Ploeh.AutoFixture
         public RandomBooleanSequenceGenerator()
         {
             this.randomBooleanNumbers =
-                new RandomNumericSequenceGenerator(0, 2);
+                new RandomNumericSequenceGenerator(0, 1);
         }
 
         /// <summary>


ploeh:
Thank you for your contribution :+1: It's now live as [AutoFixture 3.47.2](https://www.nuget.org/packages/AutoFixture/3.47.2).

PR #639, opened 2016-06-03
Use Task to ensure exceptions don't block thread forever

Zvirja:
This is a fix to PR #631.

It could happen that mock creation throws exception and as result we'll get thread blocked indefinitely.
I decided to revert to task usage, as it provides us with mechanism of exceptions re-throwing.   
As const - we have a trick with cancellation token to suppress inlining.


ploeh:
Thank you for your contribution. It's now live as [AutoFixture.AutoNSubstitute 3.47.4](https://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.47.4).

Zvirja:
Thank you for quick update.

Issue #643, opened 2016-06-07
Receiving a circular reference exception when using insufficient factory method


sergeyshushlyapin:
Consider the following type with a **private** constructor and a static *factory* method :
```csharp
public class TypeWithPseudoFactoryMethod
{
    private TypeWithPseudoFactoryMethod() { }

    public static TypeWithPseudoFactoryMethod DoSomething(
        TypeWithPseudoFactoryMethod argument)
    {
        return new TypeWithPseudoFactoryMethod();
    }
}
```

Creating an instance of this type leads to a circular dependency because the `DoSomething` method requires a parameter of the type it suppose to create:

```
...
AutoFixture was unable to create an instance of type Ploeh.AutoFixture.Kernel.SeededRequest   
because the traversed object graph contains a circular reference.
...
	Path:
		Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod argument --> 
		Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod --> 
		Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod

```

This method is definitely not suitable for creating specimens so it must not be used as a factory method.  The default object creation exception is much informative and easy to deal with (if there are no other valid factory methods):

```
AutoFixture was unable to create an instance from  
Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod,  
most likely because it has no public constructor, is an abstract or non-public type.
```

moodmosaic:
Thank you for reporting this. – Isn't the current behavior more precise though? :neutral_face:

Factory Methods are supported using a heuristic, and if there's something *wrong* with the requested type, a *specific* exception ought to be thrown – instead of throwing a general one (as proposed here).

sergeyshushlyapin:
The current behaviour is pretty precise in describing *why* AF was unable to create a specimen. My intention was to *avoid* such situations completely when it's clear than the call **will not succeed**. This gives a chance to other *builders* to create the specimen.

I tried to describe the scenario in details in the [PR](https://github.com/AutoFixture/AutoFixture/pull/644#discussion_r66198748).

moodmosaic:
>This gives a chance to other builders to create the specimen.

I'm not 100% sure if that's the case, but it looks like #644 leaves the campground cleaner, indicating that the current behavior is buggy.

ploeh:
Is there any case where the above class design is appropriate? If so, what is it?

[...]

Issue #94, opened 2013-04-15
Exception is thrown when an abstract type with public constructor is requested


mrinaldi:
A MemberAccessException is thrown when an abstract type with a public constructor is requested.

The MethodInvoker does not check if the class is abstract and invokes the constructor if it matches the Query, throwing the MemberAccessException instead of returning NoSpecimen and giving a chance for residue collectors to handle the request.

PR #644, opened 2016-06-07
Change the FactoryMethodQuery behavior to skip pseudo factory methods

sergeyshushlyapin:
Addresses issue #643.

### DIFF
--- a/Src/AutoFixture/Kernel/FactoryMethodQuery.cs
+++ b/Src/AutoFixture/Kernel/FactoryMethodQuery.cs
@@ -28,6 +28,10 @@ namespace Ploeh.AutoFixture.Kernel
         /// In case of two factory methods with an equal number of parameters, the ordering is
         /// unspecified.
         /// </para>
+        /// <para>
+        /// Factory methods that contain parameters of the requested <paramref name="type"/> 
+        /// are skipped in order to avoid circular references.
+        /// </para>
         /// </remarks>
         public IEnumerable<IMethod> SelectMethods(Type type)
         {
@@ -39,6 +43,7 @@ namespace Ploeh.AutoFixture.Kernel
             return from mi in type.GetMethods(BindingFlags.Static | BindingFlags.Public)
                    where mi.ReturnType == type
                    let parameters = mi.GetParameters()
+                   where mi.GetParameters().All(p => p.ParameterType != type)
                    orderby parameters.Length ascending
                    select new StaticMethod(mi) as IMethod;
         }
--- a/Src/AutoFixtureUnitTest/Kernel/FactoryMethodQueryTest.cs
+++ b/Src/AutoFixtureUnitTest/Kernel/FactoryMethodQueryTest.cs
@@ -65,5 +65,18 @@ namespace Ploeh.AutoFixtureUnitTest.Kernel
             Assert.True(expectedMethods.SequenceEqual(result));
             // Teardown
         }
+
+        [Fact]
+        public void SelectMethodsFromTypeWithParameterOfSameTypeReturnsEmptyResult()
+        {
+            // Fixture setup
+            var type = typeof(TypeWithPseudoFactoryMethod);
+            var sut = new FactoryMethodQuery();
+            // Exercise system
+            var result = sut.SelectMethods(type);
+            // Verify outcome
+            Assert.False(result.Any());
+            // Teardown
+        }
     }
 }
--- a/Src/TestTypeFoundation/TestTypeFoundation.csproj
+++ b/Src/TestTypeFoundation/TestTypeFoundation.csproj
@@ -116,6 +116,7 @@
     <Compile Include="TypeWithIndexer.cs" />
     <Compile Include="TypeWithOverloadedMembers.cs" />
     <Compile Include="TypeWithFactoryProperty.cs" />
+    <Compile Include="TypeWithPseudoFactoryMethod.cs" />
     <Compile Include="TypeWithRefMethod.cs" />
     <Compile Include="UnguardedStaticMethodOnStaticTypeHost.cs" />
     <Compile Include="UnguardedStaticPropertyOnStaticTypeHost.cs" />
--- a/Src/TestTypeFoundation/TypeWithPseudoFactoryMethod.cs
+++ b/Src/TestTypeFoundation/TypeWithPseudoFactoryMethod.cs
@@ -1,0 +1,21 @@
+﻿namespace Ploeh.TestTypeFoundation
+{
+    /// <summary>
+    /// This type contains a method that returns an instance of the same type 
+    /// so that it looks like a Factory Method. In fact, it is not because it 
+    /// receives a parameter of the same type, so using it as a factory inevitably 
+    /// leads to circular dependencies.
+    /// </summary>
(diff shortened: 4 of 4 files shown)


Src/AutoFixture/Kernel/FactoryMethodQuery.cs
```diff
@@ -39,6 +43,7 @@ public IEnumerable<IMethod> SelectMethods(Type type)
             return from mi in type.GetMethods(BindingFlags.Static | BindingFlags.Public)
                    where mi.ReturnType == type
                    let parameters = mi.GetParameters()
+                   where mi.GetParameters().All(p => p.ParameterType != type)
```
moodmosaic: This makes the exception message more general though:

```
Ploeh.AutoFixture.ObjectCreationException : AutoFixture was unable to create an
instance from Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod, most likely
because it has no public constructor, is an abstract or non-public type.
```

Compared to the current one:

```
Ploeh.AutoFixture.ObjectCreationException : AutoFixture was unable to create an
instance of type Ploeh.AutoFixture.Kernel.SeededRequest because the traversed
object graph contains a circular reference. Information about the circular path
follows below. This is the correct behavior when a Fixture is equipped with a
ThrowingRecursionBehavior, which is the default. This ensures that you are being
made aware of circular references in your code. Your first reaction should be to
redesign your API in order to get rid of all circular references. However, if
this is not possible (most likely because parts or all of the API is delivered
by a third party), you can replace this default behavior with a different
behavior: on the Fixture instance, remove the ThrowingRecursionBehavior from
Fixture.Behaviors, and instead add an instance of OmitOnRecursionBehavior.

    Path:
        Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod argument --> 
        Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod --> 
        Ploeh.TestTypeFoundation.TypeWithPseudoFactoryMethod
```

moodmosaic:
Thank you for your contribution :+1: – I left a [comment](https://github.com/AutoFixture/AutoFixture/issues/643#issuecomment-224484506) in #643 and another one [here](https://github.com/AutoFixture/AutoFixture/pull/644/files#r66191400).

Src/AutoFixture/Kernel/FactoryMethodQuery.cs
sergeyshushlyapin: That's interesting. The intention was to **simplify** a specimen creation. Please let me describe it in details.

I needed to create an instance of the `System.IO.Stream` class. The class is abstract but I did have AutoNSubstitute on board so that should not be a problem. Instead I received the circular reference exception and started investigating the code.

It took a while to skim across ~900 lines of decompiled code to find out a single 'pseudo' factory method that causes the problem (and blocks AutoNSubstitute BTW):

```csharp
public static Stream Synchronized(Stream stream)
```

So, my intention was to exclude such kind of methods from the specimen creation pipeline at all. If the inability to create an instance is *easily* detectable, why should one need deal with the circular reference errors? ;)

In advance it allows to instantiate the following type:

```csharp
public class TypeWithRealAndPseudoFactoryMethod
{
    private TypeWithRealAndPseudoFactoryMethod() { }

    // Not a real factory method. Causes circular refs but
    // will be taken by default as the most modest one
    public static TypeWithRealAndPseudoFactoryMethod DoSomething(
        TypeWithRealAndPseudoFactoryMethod argument)
    {
        return new TypeWithRealAndPseudoFactoryMethod();
    }

    // valid factory method, but it will be ignored by AF since 
    // it's looking for the most modest method first
    public static TypeWithRealAndPseudoFactoryMethod Create(object arg1, object arg2)
    {
        return new TypeWithRealAndPseudoFactoryMethod();
    }
}
```

Does it make more sense?

adamchester: It's a good point, I think.

I'm not sure if we should consider it to be a _Factory_ method when it requires an instance of itself as the only argument, it seems not.

moodmosaic: Thank you for the detailed explanation! :+1: 

This basically *improves* the existing heuristic for static Factory Methods, then.

moodmosaic: >I'm not sure if we should consider it to be a Factory method when it requires an instance of itself as the only argument, it seems not.

@adamchester, good point.

moodmosaic:
LGTM :+1:

This Pull Request makes sense, as it improves the heuristic we currently have for static Factory Methods.

ploeh:
As far as I understand this pull request, it _skips_ 'impossible' 'factory' methods, effectively ignoring that they exist. Did I misunderstand anything?

sergeyshushlyapin:
That is correct.

ploeh:
What AutoFixture usually does when detecting something that's 'wrong' is to throw an exception with a more detailed error message, explaining what the problem is, and how to address it. See e.g. https://github.com/AutoFixture/AutoFixture/issues/94 or https://github.com/AutoFixture/AutoFixture/issues/19.

Is there any case where the design described in #643 is appropriate?

sergeyshushlyapin:
If I understand you correctly, the suggestion is to stop trying to *bypass* the 'inconvenient' API and give a developer as much details as possible. Is this what you're saying?

The current error message that @moodmosaic has mentioned [earlier](https://github.com/AutoFixture/AutoFixture/pull/644#discussion_r66191400) contains the class info but says nothing about the factory method that cause the problem (`public static Stream Synchronized(Stream stream)` in [my](https://github.com/AutoFixture/AutoFixture/pull/644#discussion_r66198748) case). Would you like to see the method name mentioned in the exception message?

adamchester:
@ploeh FWIW, i think the problem is it gives a lot of irrelevant details about a method that is not a factory method (because it requires an instance of itself as input). Attempting to use it as a factory method then reporting detailed failure is sending the developers in the wrong direction.
AFAICT...

ploeh:
Is there any case where the design described in #643 is appropriate?

sergeyshushlyapin:
Sorry, I didn't get this part completely... Are you asking why would one design a class like I described in #643? Could you please elaborate on this. Thanks.

ploeh:
At the moment, I'm trying to understand if there might be a good reason for designing a method like that.

If there _could_ be a proper reason for such a design, then I agree that the best AutoFixture can do is to skip that method.

When I originally saw this pull request, I was inclined to think it was an issue like #94 - that is: something that compiles, but can never work.

When it comes to [System.IO.Stream.Synchronized](https://msdn.microsoft.com/en-us/library/system.io.stream.synchronized), the design looks a bit odd, but it turns out that it's still possible to use such a method. At first, I didn't realise that, because I'm not used to thinking in inheritance hierarchies :flushed: 

As such, I agree that this isn't a factory, and that the current algorithm needs adjustment.

I've reversed my thinking on this within the last 30 minutes, so I need to let it sit a bit longer, but at the moment, I'm inclined to go forward with this.

ploeh:
Thank you for your contribution; it's now available as [AutoFixture 3.47.7](https://www.nuget.org/packages/AutoFixture/3.47.7).

Issue #612, opened 2016-04-24
Guard clause assertion treating static class as abstract.


mmkal:
I may be doing something wrong here, but it looks like there's a bug somewhere in the GuardClauseAssertion class in AutoFixture.Idioms. I'm trying to use it to test that all arguments passed into a method in a static class are null-checked, but AutoFixture seems to be trying to instantiate the class (even worse, it seems to think it's abstract). Here's a small demo:
```csharp
public interface IMyInterface { }

public static class MyStaticClass
{
    public static void MyMethod(this IMyInterface myInterface)
    {
        if (myInterface == null) throw new ArgumentNullException(nameof(myInterface));
    }
}
public class MyNonStaticClass
{
    public static void MyMethod(IMyInterface myInterface)
    {
        if (myInterface == null) throw new ArgumentNullException(nameof(myInterface));
    }
}
public class TestClass
{
    public static IEnumerable<object[]> TypesToTest => new[] 
    {
        typeof(MyStaticClass),
        typeof(MyNonStaticClass)
    }.Select(t => new[] { t });

    [Theory]
    [MemberData(nameof(TypesToTest))]
    public void GuardClause_UsingAutoMoq(Type type)
    {
        var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
        var guardClauseAssertion = new GuardClauseAssertion(fixture);
        guardClauseAssertion.Verify(type.GetMethods());
    }

    [Theory]
    [MemberData(nameof(TypesToTest))]
    public void GuardClause_ManuallyInjectDependency(Type type)
    {
        var fixture = new Fixture();
        IMyInterface myInterface = Mock.Of<IMyInterface>();
        fixture.Inject(myInterface);
        var guardClauseAssertion = new GuardClauseAssertion(fixture);
        guardClauseAssertion.Verify(type.GetMethods());
    }
}
```
The tests pass when verifying MyNonStatic class, but they fail for MyStatic class with the following exceptions:
For GuardClause_UsingAutoMoq, it throws `System.NotSupportedException: parent does not have a default constructor. The default constructor must be explicitly defined.`

For GuardClause_ManuallyInjectDependency, it throws `Ploeh.AutoFixture.ObjectCreationException: AutoFixture was unable to create an instance from MyStaticClass because it's an abstract class.`

This seems wrong to me because:
1. MyStaticClass is not an abstract class, it's a static class.
[...]

PR #648, opened 2016-06-10
fix issue #612

cocogorilla:
added filters  for tostring and gettype, added no throw test for static
method on static class for guardclauseassertion


ploeh:
Thank you for your contribution - it was nice to get a defect closed :+1: It's now live as [AutoFixture.Idioms 3.47.5](https://www.nuget.org/packages/AutoFixture.Idioms/3.47.5).

Issue #628, opened 2016-05-27
Support FakeItEasy 2


mattiasnordqvist:
FakeItEasy 2 is out. Any plans on supporting it?

ploeh:
No _plans_, but we do take pull requests :wink:

ploeh:
I've added the [_jump-in tag_](https://nikcodes.com/2013/05/10/new-contributor-jump-in), because I think this might be a good task to take on, even of one isn't familiar with the AutoFixture code base.

Not that I think the task is going to be trivial or small, but because, as far as I understand the changes in FakeItEasy 2, a new _AutoFixture.AutoFakeItEasy2_ Glue Library would be a 95% copy of _AutoFixture.AutoFakeItEasy_.

moodmosaic:
>a new _AutoFixture.AutoFakeItEasy2_ Glue Library would be a 95% copy of _AutoFixture.AutoFakeItEasy_.

Agreed. — As @adamralph wrote in #617, _AutoFixture.AutoFakeItEasy_ contains (at least) [two lines of code that are incompatible with FakeItEasy 2.0](https://github.com/AutoFixture/AutoFixture/pull/617#issue-152857988).

adamralph:
Does that mean you intend to create a new package rather than bump the version of the current one?

ploeh:
That's the usual strategy here. This enables people who are not ready to make the change to stay on the old version for a while.

We still have a Glue Library for xUnit 1.x as well.

moodmosaic:
@adamralph Yes, that's also what we've been doing with the xUnit.net and NUnit Glue Libaries.

adamralph:
OK, fair enough. It's a reasonable strategy.

caloggins:
I created a fork, updated to FakeItEasy 2.1.0, and fixed the failing code. What's the next step I should take? Make a pull request?

ploeh:
You could :smile:

ploeh:
With the release of [AutoFixture.AutoFakeItEasy2 3.48.0](https://www.nuget.org/packages/AutoFixture.AutoFakeItEasy2/3.48.0) I think we can close this issue(?)

adamralph:
On behalf of the [FakeItEasy owners](https://github.com/orgs/FakeItEasy/teams/owners), thank you to everyone who made this happen!

/cc @blairconrad @thomaslevesque

TeaDrivenDev:
[...]

PR #659, opened 2016-06-22
Fix for #628 (FakeItEasy 2 support)

caloggins:


ploeh:
Thank you for your interest in contributing to AutoFixture!

Do I understand it correctly that this pull request proposes _changing_ the existing `AutoFakeItEasy` package?

caloggins:
That's what I did. Was that not the intent of the issue?

Not sure why the check failed. Ran the build script locally.

ploeh:
> Was that not the intent of the issue?

No, the intent described in #628 is to create a new library called _AutoFixture.AutoFakeItEasy2_, which can be released side-by-side with _AutoFixture.AutoFakeItEasy_. In this way, we enable users to migrate at their own pace.

caloggins:
Completely misread it.

ploeh:
No worries :smile: If you want to take a shot at that approach, feel free to submit another pull request :smiley_cat:

PR #661, opened 2016-06-22
Add Support for FakeItEasy 2.0+ (#628)

caloggins:
Went with 2.1 plus, since it's the latest as of now.


Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+    <metadata>
+        <id>AutoFixture.AutoFakeItEasy2</id>
+        <version>@build.number@</version>
+        <title>AutoFixture with Auto Mocking using FakeItEasy</title>
+        <authors>Nikos Baxevanis</authors>
+        <owners>Nikos Baxevanis</owners>
```
moodmosaic: I think it's fair to have _your_ name showing up in these two entries :smiley:

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
...
+        <dependencies>
+            <dependency id="AutoFixture" version="@build.number@" />
+            <dependency id="FakeItEasy" version="2.1" />
+        </dependencies>
+        <summary>Turns AutoFixture into an Auto-Mocking Container based on FakeItEasy.</summary>
+        <description>This extension turns AutoFixture into an Auto-Mocking Container. The mock instances are created by FakeItEasy. To use it, add the AutoFakeItEasyCustomization to your Fixture instance. Read more at http://autofixture.codeplex.com/</description>
```
moodmosaic: We could replace the CodePlex URL with the one on GitHub.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
...
+<package>
+    <metadata>
+        <id>AutoFixture.AutoFakeItEasy2</id>
+        <version>@build.number@</version>
+        <title>AutoFixture with Auto Mocking using FakeItEasy</title>
+        <authors>Nikos Baxevanis</authors>
+        <owners>Nikos Baxevanis</owners>
+        <dependencies>
+            <dependency id="AutoFixture" version="@build.number@" />
+            <dependency id="FakeItEasy" version="2.1" />
```
moodmosaic: Shouldn't it be version `2.0` instead? :confused:

Src/AutoFakeItEasy2.UnitTest/DependencyConstraints.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Ploeh.AutoFixture.AutoFakeItEasy.UnitTest
+{
+    public class DependencyConstraints
+    {
+        [Theory]
+        [InlineData("Rhino.Mocks")]
+        [InlineData("Moq")]
+        [InlineData("xunit")]
+        [InlineData("xunit.extensions")]
+        public void AutoFakeItEasyDoesNotReference(string assemblyName)
```
moodmosaic: Could we also include `[InlineData("NSubstitute")]` here?

Src/AutoFakeItEasy2.UnitTest/DependencyConstraints.cs
```diff
@@ -0,0 +1,37 @@
...
+            var references = typeof(AutoFakeItEasyCustomization).Assembly.GetReferencedAssemblies();
+            // Verify outcome
+            Assert.False(references.Any(an => an.Name == assemblyName));
+            // Teardown
+        }
+
+        [Theory]
+        [InlineData("Rhino.Mocks")]
+        [InlineData("Moq")]
+        public void AutoFakeItEasyUnitTestsDoNotReference(string assemblyName)
```
moodmosaic: [Same](https://github.com/AutoFixture/AutoFixture/pull/661/files#r68087815) also here.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
caloggins: Heck, no. I want to blame it on someone else. :)
(I'll update that.)

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
caloggins: Not sure. It's a new assembly. Was thinking v1 of support for FakeItEasy 2.

caloggins:
I don't know why it fails to a missing assembly.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
ploeh: Is there a particular reason for requiring FakeItEasy version 2.1? Or would everything also work with FakeItEasy version 2.0?

Src/AutoFakeItEasy2/FakeItEasyBuilder.cs
```diff
@@ -0,0 +1,95 @@
...
+            var type = request as Type;
+            if (!type.IsFake())
+            {
+#pragma warning disable 618
+                return new NoSpecimen(request);
+#pragma warning restore 618
+            }
+
+            var fake = this.builder.Create(request, context) as FakeItEasy.IHideObjectMembers;
+            if (fake == null)
```
moodmosaic: :+1: Nice. This seems to be compatible with what @adamralph wrote [here](https://github.com/AutoFixture/AutoFixture/pull/617#issue-152857988).

Src/AutoFakeItEasy2/FakeItEasyRelay.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            var type = request as Type;
+            if (type == null)
+#pragma warning disable 618
+                return new NoSpecimen(request);
+#pragma warning restore 618
+
+            var fakeType = typeof(Fake<>).MakeGenericType(type);
+
+            var fake = context.Resolve(fakeType) as IHideObjectMembers;
```
moodmosaic: :+1: Nice. This seems to be compatible with what @adamralph wrote [here](https://github.com/AutoFixture/AutoFixture/pull/617#issue-152857988).

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
adamralph: It should be fine with FIE 2.0.

FIE follows SemVer so hopefully you can keep the dependency at 2.0 for the lifetime of FIE 2.x.

caloggins: It should work. My thinking was that, as of creating this library, the latest available is 2.1. No reason to for the backward compat, since there was no 2.0 support. Easily changed, if you think we should.

ploeh: We usually follow [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) when declaring dependencies. There may be users out there who, for one reason or another, can't readily upgrade from 2.0 to 2.1, and if there's no reason to exclude them, we ought to include them.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package>
+    <metadata>
+        <id>AutoFixture.AutoFakeItEasy2</id>
+        <version>@build.number@</version>
+        <title>AutoFixture with Auto Mocking using FakeItEasy</title>
```
ploeh: Should we have _FakeItEasy 2.x_ here?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
...
+        <authors>Christopher Loggins</authors>
+        <owners>Christopher Loggins</owners>
+        <dependencies>
+            <dependency id="AutoFixture" version="@build.number@" />
+            <dependency id="FakeItEasy" version="2.0" />
+        </dependencies>
+        <summary>Turns AutoFixture into an Auto-Mocking Container based on FakeItEasy.</summary>
```
ploeh: 2.x?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
...
+        <description>This extension turns AutoFixture into an Auto-Mocking Container. The mock instances are created by FakeItEasy. To use it, add the AutoFakeItEasyCustomization to your Fixture instance. Read more at https://github.com/AutoFixture/AutoFixture.</description>
```
ploeh: 2.x?

ploeh: Why are we directing people to [https://github.com/AutoFixture/AutoFixture](https://github.com/AutoFixture/AutoFixture)? There's nothing about `AutoFakeItEasyCustomization` there...

caloggins: Is there a better page that covers this? I only saw the docs on github (Cheat Sheet?), or your post about auto mocking. Which would be better?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
```diff
@@ -0,0 +1,27 @@
...
+<package>
+    <metadata>
+        <id>AutoFixture.AutoFakeItEasy2</id>
+        <version>@build.number@</version>
+        <title>AutoFixture with Auto Mocking using FakeItEasy</title>
+        <authors>Christopher Loggins</authors>
+        <owners>Christopher Loggins</owners>
+        <dependencies>
+            <dependency id="AutoFixture" version="@build.number@" />
+            <dependency id="FakeItEasy" version="2.0" />
```
adamralph: I'd feel more comfortable if this were a version range, with 2.0 being the inclusive minimum and 3.0 being the exclusive maximum.

We won't be pushing out 3.0 any time soon, but we will be doing our best guarantee that the 2.x version range is non-breaking.

Src/AutoFakeItEasy2.UnitTest/AutoFakeItEasyCustomizationTest.cs
```diff
@@ -0,0 +1,103 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using FakeItEasy;
+using Ploeh.AutoFixture.Kernel;
+using Xunit;
+
+namespace Ploeh.AutoFixture.AutoFakeItEasy.UnitTest
```
ploeh: Shouldn't this be `Ploeh.AutoFixture.AutoFakeItEasy2.UnitTest`?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
adamralph: In any case, this is AutoFixture.AutoFakeItEasy2, so the version range makes sense.

Src/AutoFakeItEasy2.UnitTest/DependencyConstraints.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Linq;
+using Xunit;
+using Xunit.Extensions;
+
+namespace Ploeh.AutoFixture.AutoFakeItEasy.UnitTest
```
ploeh: Shouldn't this be `Ploeh.AutoFixture.AutoFakeItEasy2.UnitTest`?

Src/AutoFakeItEasy2.UnitTest/FakeItEasyBuilderTest.cs
```diff
@@ -0,0 +1,150 @@
+﻿using System;
+using FakeItEasy;
+using Ploeh.AutoFixture.Kernel;
+using Ploeh.TestTypeFoundation;
+using Xunit;
+using Xunit.Extensions;
+
+namespace Ploeh.AutoFixture.AutoFakeItEasy.UnitTest
```
ploeh: Shouldn't this be `Ploeh.AutoFixture.AutoFakeItEasy2.UnitTest`?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
ploeh: I'm not sure there's good coverage, but could it be better _not_ having a link?

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
ploeh: Isn't 2.+ implied by the above specification?

(The NuGet specs have changed so much recently that I'm having trouble keeping up...)

Src/AutoFakeItEasy2.UnitTest/AutoFakeItEasyCustomizationTest.cs
caloggins: I was on the fence about that. Yes, it's a version specific thing. But, it's still the AutoFakeItEasy thing. I kept the library's namesapce the same, so it wouldn't affect people when they upgrade. This namespace was kept the same to match the library. Whatever your convention is, I'll follow.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
caloggins: Not sure. The original had a codeplex link, and it was suggested it be changed to GitHub. I think the link is redundant with the project home link. I'll pull it.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
ploeh: Unless we already know that FakeItEasy 3 will have changes that break this library, I see no reason to limit the dependency to less than version 3.

As an example, [AutoFixture.AutoMoq](https://www.nuget.org/packages/AutoFixture.AutoMoq) for years depended on Moq 3.1, but it worked fine with Moq 4.

While there may have been breaking changes between Moq 3.1 and Moq 4, they didn't affect AutoFixture.AutoMoq.

caloggins: I thought adding >=2.0, <3.0 assumes that FIE will never make the mistake of introducing a breaking change into 2.x.

Src/AutoFakeItEasy2.UnitTest/AutoFakeItEasyCustomizationTest.cs
ploeh: We use renamed namespaces in [other places](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/AutoDataAttribute.cs#L9). This makes it explicit what's going on. It may imply a bit more conversion work, but prevents subtle bugs where you think you're using one library, but in reality, it's not what you thought...

caloggins: I'll make the change to the base library, and the test library.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy2.nuspec
ploeh: Wouldn't that prevent a user from upgrading to FakeItEasy 3, because it'd be incompatible with those constraints?

caloggins: It would. Even if, as you said, FIE 3 works w/ this library. That's why I went with the >=2.0 setting, but no upper limit.

adamralph: Fair enough. If, during FIE 3.0 beta/RC we find that this library is compatible, we can restrict the dependency to <3.0 at that time, as we did for FIE 2.0 in https://github.com/AutoFixture/AutoFixture/pull/617

Src/AutoFakeItEasy2.UnitTest/FakeItEasyBuilderTest.cs
```diff
@@ -0,0 +1,150 @@
...
+        public void CreateWithNonFakeRequestReturnsCorrectResult(object request)
+        {
+            // Fixture setup
+            var dummyBuilder = A.Fake<ISpecimenBuilder>();
+            var sut = new FakeItEasyBuilder(dummyBuilder);
+            // Exercise system
+            var dummyContext = A.Fake<ISpecimenContext>();
+            var result = sut.Create(request, dummyContext);
+            // Verify outcome
+#pragma warning disable 618
```
ploeh: These warnings are there in the existing AutoFixture code base because we were concerned that it'd be a breaking change to shift to returning `new NoSpecimen()`. This library, being new, however, doesn't need to make that concession, so ought to use `new NoSpecimen()` directly, so that the `#pragma` lines can be avoided.

Src/AutoFakeItEasy2.UnitTest/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("AutoFakeItEasyUnitTest")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Ploeh")]
+[assembly: AssemblyProduct("AutoFixture")]
+[assembly: AssemblyCopyright("Copyright © Ploeh 2011")]
```
ploeh: I don't think we can claim that the copyright of this particular library stretches back to 2011; it ought to be 2016. Also, I think it'd be fairer to assign the copyright to _AutoFixture_, although that isn't a legal entity...

Src/AutoFakeItEasy2/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("AutoFakeItEasy")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Ploeh")]
+[assembly: AssemblyProduct("AutoFixture")]
+[assembly: AssemblyCopyright("Copyright © Ploeh 2011")]
```
ploeh: Same comment as [above](#r68506263).

ploeh:
Thank you for your interest in contributing to AutoFixture! This looks promising :+1: The few comments I have are superficial, and are easy to address :smile:

Src/AutoFakeItEasy2.UnitTest/FakeItEasyBuilderTest.cs
caloggins: Should they be pulled from the classes being tested (AutoFakeItEasy2.FakeItEasyBuilder, AutoFakeItEasy2.FakeItEasyRelay)?

moodmosaic: Yes, it'd make sense to use `new NoSpecimen()` throughout the whole AutoFixture.AutoFakeItEasy2 source code, [for the reason @ploeh described](https://github.com/AutoFixture/AutoFixture/pull/661#discussion_r68505956).

ploeh:
Thank you for your contribution! It's now live as [AutoFixture.AutoFakeItEasy2 3.48.0](https://www.nuget.org/packages/AutoFixture.AutoFakeItEasy2/3.48.0).

Issue #663, opened 2016-06-29
Examine type of created FakeItEasy Fakes instead of using IHideObjectMembers


blairconrad:
From my https://github.com/AutoFixture/AutoFixture/pull/617#issuecomment-229425417

I think using `IHideObjectMembers` to tell if a Fake were created properly is not the best idea, because
1. `IHideObjectMembers` changed namespaces between FakeItEasy 1.x and 2.0, keeping AutoFakeItEasy from working with FakeItEasy 2.x, and
2. just because something implements `IHideObjectMembers` doesn't mean it's a Fake.

I propose comparing the resolved object's type against `Fake<T>`. It should be a more robust solution, and would even allow for us to remove the upper bound on the FakeItEasy nuget dependency (and as a side effect, obviate the need for AutoFakeItEasy2).

I've spiked the change in AutoFakeItEasy only. It seems to work, and allows me to upgrade to a FakeItEasy 2.2.0 dependency. I'll submit a PR for discussion.

ploeh:
I'm not sure I'm endowed with sufficient insight to evaluate this. @adamralph, what do you think?

blairconrad:
@ploeh, @moodmosaic, I'm sure the change is functionally sound but am always happy to hear what @adamralph thinks. Of course @thomaslevesque may also have some insight.
However, the important question for the AutoFixture owners would be whether or not you'd want to continue with AutoFakeItEasy2. This change makes it redundant, but there's always the tricky question of removing things from a project. Taking away AutoFakeItEasy2 could be considered a breaking change and I was wondering if you'd want to do that. Making the change in both packages is, of course, also an option, but it increases the maintenance work in the future.

adamralph:
The proposal seems sound to me. I agree that checking for an implementation of `IHideObjectMembers` is a very brittle test.

Has AutoFakeItEasy2 already been released?

blairconrad:
@adamralph, thanks for the chime in.

Regarding the status of the AutoFakeItEasy2 release:

* [NuGet package](https://www.nuget.org/packages/AutoFixture.AutoFakeItEasy2)
* [Twitter announcement](https://twitter.com/ploeh/status/747889675042496513) (which you retweeted :wink:)

adamralph:
Thanks for refilling my sieve-like memory :smile: 

[...]

PR #664, opened 2016-06-29
AutoFakeItEasy: Examine created Fake's types instead of using IHideObjectMembers

blairconrad:
Relates to #663.

Note that I've only made the change in AutoFakeItEasy (not AutoFakeItEasy2) just so everyone can see the shape of it. I've attempted to follow what seemed to be the coding conventions, especially with the separate ifs.

### DIFF
--- a/Src/AutoFakeItEasy/FakeItEasyBuilder.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyBuilder.cs
@@ -56,7 +56,7 @@ namespace Ploeh.AutoFixture.AutoFakeItEasy
         /// The Create method checks whether a request is for an interface or abstract class. If so
         /// it delegates the call to the decorated <see cref="Builder"/>. When the specimen is
         /// returned from the decorated <see cref="ISpecimenBuilder"/> the method checks whether
-        /// the returned instance is a FakeItEasy instance.
+        /// the returned instance is a FakeItEasy Fake instance of the correct type.
         /// </para>
         /// <para>
         /// If all pre- and post-conditions are satisfied, the mock instance is returned; otherwise
@@ -73,16 +73,8 @@ namespace Ploeh.AutoFixture.AutoFakeItEasy
 #pragma warning restore 618
             }
 
-            var fake = this.builder.Create(request, context) as FakeItEasy.Configuration.IHideObjectMembers;
-            if (fake == null)
-            {
-#pragma warning disable 618
-                return new NoSpecimen(request);
-#pragma warning restore 618
-            }
-
-            var fakeType = type.GetFakedType();
-            if (fake.GetType().GetFakedType() != fakeType)
+            var fake = this.builder.Create(request, context);
+            if (!type.IsInstanceOfType(fake))
             {
 #pragma warning disable 618
                 return new NoSpecimen(request);
--- a/Src/AutoFakeItEasy/FakeItEasyRelay.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyRelay.cs
@@ -82,11 +82,13 @@ namespace Ploeh.AutoFixture.AutoFakeItEasy
 
             var fakeType = typeof(Fake<>).MakeGenericType(type);
 
-            var fake = context.Resolve(fakeType) as FakeItEasy.Configuration.IHideObjectMembers;
-            if (fake == null)
+            var fake = context.Resolve(fakeType);
+            if (!fakeType.IsInstanceOfType(fake))
+            {
 #pragma warning disable 618
                 return new NoSpecimen(request);
 #pragma warning restore 618
+            }
 
             return fake.GetType().GetProperty("FakedObject").GetValue(fake, null);
         }


Src/AutoFakeItEasy/FakeItEasyRelay.cs
```diff
@@ -82,13 +82,28 @@ public object Create(object request, ISpecimenContext context)
...
+            }
+
+            if (resolvedType.GetFakedType() != type)
+            {
+#pragma warning disable 618
+                return new NoSpecimen(request);
+#pragma warning restore 618
+            }
+
+            return resolvedType.GetProperty("FakedObject").GetValue(fake, null);
```
blairconrad: Left to my own devices, I would've combined the three previous ifs into one, (or at least the last two), but that doesn't seem to be the way the other code was written, so I kept it separate.

I suppose another option would be to see if `resolvedType.GetFakedType()` was assignable to `type`, as I guess returning a `Fake<Derived>` when asked for a `Fake<Base>` could be a valid use case, depending on how AutoFixure's been configured.

blairconrad:
I amended the PR to change both AutoFakeItEasy and AutoFakeItEasy2 and to update the required FakeItEasy version in the AutoFakeItEasy nuspec.

blairconrad:
As an aside, I found that the repo doesn't just build for me on checkout. When I run `build.cmd` or `bash build.sh`, I get the following error:

Src/AutoFoq.sln
Src/Idioms.FsCheck.sln

```
1) Building C:\PortableApps\Documents\Source\GitHub\AutoFixture\Src\AutoFoq.sln failed with exitcode 1.
2) MSB4057: C:\PortableApps\Documents\Source\GitHub\AutoFixture\Src\AutoFoqUnitTest\AutoFoqUnitTest.fsproj(0,0): The target "Clean" does not exist in the project.
3) MSB4057: C:\PortableApps\Documents\Source\GitHub\AutoFixture\Src\AutoFoq\AutoFoq.fsproj(0,0): The target "Clean" does not exist in the project.
```

Is there a dependency I'm missing? I didn't see anything in [CONTRIBUTING.md](https://github.com/AutoFixture/AutoFixture/blob/1d6211908ebb5b436a8c98abc33ebce8dd6f2f2d/CONTRIBUTING.md), but maybe I am just unobservant. I contented myself with excluding the projects from my local build and leaning on your CI system.

Building from the IDE works fine. I did notice that the command-line build uses `Microsoft (R) Build Engine version 14.0.25420.1`. Perhaps that's the difference.

blairconrad:
Ah. I just noticed that the auto-VS-selector opened the AutoFoq sln in VS2012, but I'd been running from a VS2015 command prompt. VS2015 tells me

image

Installing now. That's probably the difference. The `The target "Clean" does not exist in the project.` threw me.

blairconrad:
Ah. That was it. Sorry.

Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
```diff
@@ -8,7 +8,7 @@
         <owners>Nikos Baxevanis</owners>
         <dependencies>
             <dependency id="AutoFixture" version="@build.number@" />
-            <dependency id="FakeItEasy" version="[1.7,2)" />
+            <dependency id="FakeItEasy" version="1.7" />
```
moodmosaic: Isn't it a breaking change if we put back the old, less strict, constraint?

Src/AutoFakeItEasy/FakeItEasyBuilder.cs
```diff
@@ -73,16 +73,24 @@ public object Create(object request, ISpecimenContext context)
...
+            if (!createdType.IsFake())
+            {
+#pragma warning disable 618
+                return new NoSpecimen(request);
+#pragma warning restore 618
+            }
+
             var fakeType = type.GetFakedType();
-            if (fake.GetType().GetFakedType() != fakeType)
+            if (createdType.GetFakedType() != fakeType)
```
moodmosaic: Could lines 76-93 be refactored so that we avoid calling `GetFakedType` twice? :smiley:

Src/AutoFakeItEasy/FakeItEasyRelay.cs
```diff
@@ -82,13 +82,28 @@ public object Create(object request, ISpecimenContext context)
 
             var fakeType = typeof(Fake<>).MakeGenericType(type);
 
-            var fake = context.Resolve(fakeType) as FakeItEasy.Configuration.IHideObjectMembers;
+            var fake = context.Resolve(fakeType);
             if (fake == null)
```
moodmosaic: I would expect the only change being made in this file is to replace `IHideObjectMembers` with a call to `IsFake` extension method (?)

Src/AutoFakeItEasy2/FakeItEasyBuilder.cs
```diff
@@ -56,7 +56,7 @@ public ISpecimenBuilder Builder
         /// The Create method checks whether a request is for an interface or abstract class. If so
         /// it delegates the call to the decorated <see cref="Builder"/>. When the specimen is
         /// returned from the decorated <see cref="ISpecimenBuilder"/> the method checks whether
-        /// the returned instance is a FakeItEasy instance.
```
moodmosaic: Could we tackle *AutoFixture.AutoFakeItEasy2* on a separate Pull Request? :smiley:

Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
blairconrad: I am not sure. I made the change based on the thought that it was an error to have introduced it and your comment https://github.com/AutoFixture/AutoFixture/issues/663#issuecomment-230164610, which seemed to be in favour of the change.

My understanding is that new projects would still depend on 1.7. It's only if someone runs a `nuget update` or equivalent that there'd be a change, yes? You concern is that all of a sudden clients would have an undesirable FakeItEasy, which will be incompatible with their code? I guess that's the case. I can restore the original.

Src/AutoFakeItEasy/FakeItEasyRelay.cs
blairconrad: You know the architecture better than I. The layered builders and relays boggle my mind, to tell the truth. If you're content to just know that `fake` has the quality that it `IsFake`, I'll remove the new stuff.

Src/AutoFakeItEasy2/FakeItEasyBuilder.cs
blairconrad: Of course.

Src/AutoFakeItEasy/FakeItEasyBuilder.cs
blairconrad: Three times, if you count the one in `IsFake`. Sure.

Src/AutoFakeItEasy/FakeItEasyRelay.cs
blairconrad: Being unsure of how the pieces fit together, I was being overly cautious. I'll remove the extra work, in a way that mirrors the change you asked for in `FakeItEasyBuilder`.

moodmosaic:
This Pull Request looks good to me :+1:

Should I know about this FakeItEasy extension method before, we'd probably wouldn't have to release AutoFixture.AutoFakeItEasy2.

Since we *did* release AutoFixture.AutoFakeItEasy2, I think we're doing the right thing in this Pull Request, by keeping it as a pure refactoring.

blairconrad:
Thanks, @moodmosaic. I'll submit a parallel one for AutoFakeItEasy2 later. Do you want an issue for that as well, or will the PR be enough?

blairconrad:
(The AutoFakeItEasy2 PR is #671.)

moodmosaic:
>Do you want an issue for that as well, or will the PR be enough?

Thank you! :smiley: The PR (as you did) should be enough :+1:

blairconrad:
Thanks, @moodmosaic and @ploeh.

Issue #637, opened 2016-06-01
Passive Customization Attributes are ignored when used in combination with [Frozen]


sergeyshushlyapin:
If a test method parameter is decorated with both `[Substitute]` and `[Frozen]` attributes, a concrete instance is created instead of a substitute. The following test fails:

``` csharp
[Theory, AutoNSubstituteData]
public void FreezeConcreteTypeSubstitute([Frozen, Substitute] ConcreteType ct1, ConcreteType ct2)
{
    Assert.Same(ct1, ct2);
    Assert.True(ct1.GetType().BaseType == typeof(ConcreteType)); // fails
    // A real instance is created instead of a mock.
}
```

The issue has been addressed in PR #627 but it was agreed that the solution is sub-optimal. To be implemented in a new PR.

ploeh:
@sergeyshushlyapin, while I haven't added the [jump in](https://nikcodes.com/2013/05/10/new-contributor-jump-in) label to this issue, you're still more than welcome to attempt to address it. You seem to be on top of things :+1:

sergeyshushlyapin:
The issue is not about the `[Substitute]` attribute only. It's about all the [Passive Attributes](http://blog.ploeh.dk/2014/06/13/passive-attributes/) that require corresponding *relays* to create specimens (e.g. [StringLengthAttributeRelay](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DataAnnotations/StringLengthAttributeRelay.cs)):

```csharp
[Theory, AutoData]
public void StringLengthConstraintWorksWithoutFrozen([StringLength(3)]string str)
{
    Assert.True(str.Length == 3); // pass
}

[Theory, AutoData]
public void StringLengthConstraintFailsWithFrozen([Frozen][StringLength(3)]string str)
{
    Assert.True(str.Length == 3); // fails
}
```

The underlying reason is in the wrong assumption that in order to create a specimen only *type* is required. In the Glue Libraries specimens can be defined by types and *Customization Attributes*.

The [FreezeOnMatchCustomization](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/FreezeOnMatchCustomization.cs#L103) resolves specimens using type only which causes the issue:
```csharp
var specimen = context.Resolve(this.TargetType);
```

Zvirja:
[...]

Issue #645, opened 2016-06-07
Fixture creates OmitSpecimens instead of default values


sergeyshushlyapin:
In the `OmitSpecimen` [documentation](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/OmitSpecimen.cs#L9):

> A signal type used to indicate to the Auto-Property feature that a given request should be
> ignored, and no further processing performed.

Despite the explicit mentioning the Auto-Property feature here (which indicates that we're talking about   specimen *'parts'*), I have a strong feeling that the following test should pass:

```csharp
[Test]
public void CreateWithOmittedRequestReturnsNull()
{
    var fixture = new Fixture();
    var overallOmitter = new Omitter(new TrueRequestSpecification());
    fixture.Customizations.Insert(0, overallOmitter);

    var result = fixture.Create<object>();

    Assert.IsNotInstanceOf<OmitSpecimen>(result); // fails
    Assert.Null(result);
}
```

ploeh:
I don't think I even have an opinion on that matter :flushed: 

There are some parts of the design of AutoFixture that, if you start doing weird things to it, will compile, but not necessarily do meaningful things at run-time. That's one disadvantage of an API which is, to be frank, quite weakly typed.

Is there an underlying reason for this issue?

sergeyshushlyapin:
Yep, the question itself sounds a bit strange but there is a reason :)

I think the issue #343 can be solved using an *omitter* that would ignore the test method parameters with the NUnit's `ValuesAttribute`.

I'm experimenting with that [here](https://github.com/sergeyshushlyapin/AutoFixture/commits/nunit2-values-omitter) and now end up with the following exception:
```csharp
[Test, AutoData]
public void FixtureIgnoresParameterWithValuesAttribute([Values(1, -1)] int num)
{
    Assert.True(num == -1 || num == 1);    // fails
    // System.ArgumentException : Unable to cast  "Ploeh.AutoFixture.Kernel.OmitSpecimen" 
    // to "System.Int32".
}
```

//cc: @steph4nc

sergeyshushlyapin:
I was wrong, the *omitter* won't solve the issue #343 (there is [another fix](https://github.com/AutoFixture/AutoFixture/issues/343#issuecomment-224848030) required), so if there is no other scenario when this particular fix could be used, I'm fine with closing the issue.

> an API which is, to be frank, quite weakly typed.

[...]

PR #665, opened 2016-06-29
Fix parameter freezing when used in combination with Passive Attributes (#637, xUnit.net2)

sergeyshushlyapin:
Addresses issue #637 for xUnit.net2 Glue Library.

### DIFF
--- a/Nuget/Ploeh.AutoFixture.nuspec
+++ b/Nuget/Ploeh.AutoFixture.nuspec
@@ -6,6 +6,9 @@
         <title>AutoFixture</title>
         <authors>Mark Seemann</authors>
         <owners>Mark Seemann</owners>
+        <dependencies>
+            <dependency id="Albedo" version="1.0.0" />
+        </dependencies>
         <description>AutoFixture makes it easier for developers to do Test-Driven Development by automating non-relevant Test Fixture Setup, allowing the Test Developer to focus on the essentials of each test case.</description>
         <projectUrl>https://github.com/AutoFixture/AutoFixture</projectUrl>
         <licenseUrl>https://github.com/AutoFixture/AutoFixture/blob/master/LICENCE.txt</licenseUrl>
--- a/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Ploeh.TestTypeFoundation;
 using Xunit;
@@ -415,6 +416,12 @@ namespace Ploeh.AutoFixture.Xunit2.UnitTest
             FieldHolder<string> p2)
         {
             Assert.NotEqual(p1, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeParameterWithStringLengthConstraint([Frozen][StringLength(3)]string p)
+        {
+            Assert.True(p.Length == 3);
         }
     }
 }
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -140,7 +140,8 @@ namespace Ploeh.AutoFixture
         private ISpecimenBuilder FreezeTargetType(IFixture fixture)
         {
             var context = new SpecimenContext(fixture);
-            var specimen = context.Resolve(this.TargetType);
+            var request = (object)this.ParameterInfo ?? this.TargetType;
+            var specimen = context.Resolve(request);
             return new FixedBuilder(specimen);
         }
     }
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -1,6 +1,5 @@
 using System;
-using System.ComponentModel.DataAnnotations;
-using System.Reflection;
+using Ploeh.Albedo;
 using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture
@@ -68,24 +67,23 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
         /// </summary>
-        /// <param name="parameter">
-        /// The <see cref="ParameterInfo"/> used to freeze specimens decorated with advanced
-        /// attributes (e.g. <see cref="StringLengthAttribute"/>).
+        /// <param name="reflectionElement">
+        /// The <see cref="IReflectionElement"/> used to create specimens to freeze.
         /// </param>
         /// <param name="matcher">
         /// The <see cref="IRequestSpecification"/> used to match the requests
         /// that will be satisfied by the frozen specimen.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="parameter"/> or <paramref name="matcher"/> is null.
+        /// <paramref name="reflectionElement"/> or <paramref name="matcher"/> is null.
         /// </exception>
         public FreezeOnMatchCustomization(
-            ParameterInfo parameter,
+            IReflectionElement reflectionElement,
             IRequestSpecification matcher)
         {
-            if (parameter == null)
+            if (reflectionElement == null)
             {
-                throw new ArgumentNullException(nameof(parameter));
+                throw new ArgumentNullException(nameof(reflectionElement));
             }
 
             if (matcher == null)
@@ -93,15 +91,14 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentNullException(nameof(matcher));
             }
 
-            this.ParameterInfo = parameter;
-            this.TargetType = parameter.ParameterType;
+            this.ReflectionElement = reflectionElement;
             this.Matcher = matcher;
         }
 
         /// <summary>
-        /// The <see cref="ParameterInfo"/> describing the specimen to freeze.
+        /// The <see cref="ReflectionElement"/> describing the specimen to freeze.
         /// </summary>
-        public ParameterInfo ParameterInfo { get; }
+        public IReflectionElement ReflectionElement { get; }
 
         /// <summary>
         /// The <see cref="Type"/> of the frozen specimen.
@@ -140,7 +137,7 @@ namespace Ploeh.AutoFixture
         private ISpecimenBuilder FreezeTargetType(IFixture fixture)
         {
             var context = new SpecimenContext(fixture);
-            var request = (object)this.ParameterInfo ?? this.TargetType;
+            var request = (object)this.ReflectionElement ?? this.TargetType;
             var specimen = context.Resolve(request);
             return new FixedBuilder(specimen);
         }
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -49,19 +49,9 @@ namespace Ploeh.AutoFixture
         public FreezeOnMatchCustomization(
             Type targetType,
             IRequestSpecification matcher)
+            : this(new TypeElement(targetType), matcher)
         {
-            if (targetType == null)
-            {
-                throw new ArgumentNullException(nameof(targetType));
-            }
-
-            if (matcher == null)
-            {
-                throw new ArgumentNullException(nameof(matcher));
-            }
-
             this.TargetType = targetType;
-            this.Matcher = matcher;
         }
 
         /// <summary>
@@ -137,8 +127,7 @@ namespace Ploeh.AutoFixture
         private ISpecimenBuilder FreezeTargetType(IFixture fixture)
         {
             var context = new SpecimenContext(fixture);
-            var request = (object)this.ReflectionElement ?? this.TargetType;
-            var specimen = context.Resolve(request);
+            var specimen = context.Resolve(this.ReflectionElement);
             return new FixedBuilder(specimen);
         }
     }
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -1,4 +1,6 @@
 using System;
+using System.ComponentModel.DataAnnotations;
+using System.Reflection;
 using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture
@@ -43,7 +45,7 @@ namespace Ploeh.AutoFixture
         /// that will be satisfied by the frozen specimen.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="targetType"/> or<paramref name="matcher"/> is null.
+        /// <paramref name="targetType"/> or <paramref name="matcher"/> is null.
         /// </exception>
         public FreezeOnMatchCustomization(
             Type targetType,
@@ -62,6 +64,44 @@ namespace Ploeh.AutoFixture
             this.TargetType = targetType;
             this.Matcher = matcher;
         }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
+        /// </summary>
+        /// <param name="parameter">
+        /// The <see cref="ParameterInfo"/> used to freeze specimens decorated with advanced
+        /// attributes (e.g. <see cref="StringLengthAttribute"/>).
+        /// </param>
+        /// <param name="matcher">
+        /// The <see cref="IRequestSpecification"/> used to match the requests
+        /// that will be satisfied by the frozen specimen.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="parameter"/> or <paramref name="matcher"/> is null.
+        /// </exception>
+        public FreezeOnMatchCustomization(
+            ParameterInfo parameter,
+            IRequestSpecification matcher)
+        {
+            if (parameter == null)
+            {
+                throw new ArgumentNullException(nameof(parameter));
+            }
+
+            if (matcher == null)
+            {
+                throw new ArgumentNullException(nameof(matcher));
+            }
+
+            this.ParameterInfo = parameter;
+            this.TargetType = parameter.ParameterType;
+            this.Matcher = matcher;
+        }
+
+        /// <summary>
+        /// The <see cref="ParameterInfo"/> describing the specimen to freeze.
+        /// </summary>
+        public ParameterInfo ParameterInfo { get; }
 
         /// <summary>
         /// The <see cref="Type"/> of the frozen specimen.
--- a/Nuget/Ploeh.AutoFixture.Idioms.nuspec
+++ b/Nuget/Ploeh.AutoFixture.Idioms.nuspec
@@ -8,7 +8,6 @@
         <owners>Mark Seemann</owners>
         <dependencies>
             <dependency id="AutoFixture" version="@build.number@" />
-            <dependency id="Albedo" version="1.0.0" />
         </dependencies>
         <summary>Extension to AutoFixture that contains unit testing idioms.</summary>
         <description>Ubiquitous use of AutoFixture for unit testing has given rise to a number of idiomatic unit tests - unit tests that tend to follow common templates. The AutoFixture Idioms library encapsulates these idioms into reusable classes and methods.</description>
--- a/Src/AutoFixture.NUnit2/AutoFixture.NUnit2.csproj
+++ b/Src/AutoFixture.NUnit2/AutoFixture.NUnit2.csproj
@@ -57,6 +57,10 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Ploeh.Albedo">
+      <HintPath>..\..\Packages\Albedo.1.0.0\lib\net35\Ploeh.Albedo.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="Microsoft.CSharp" />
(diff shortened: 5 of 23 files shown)


Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -64,6 +66,44 @@ public FreezeOnMatchCustomization(Type targetType)
...
+        /// </param>
+        /// <param name="matcher">
+        /// The <see cref="IRequestSpecification"/> used to match the requests
+        /// that will be satisfied by the frozen specimen.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="parameter"/> or <paramref name="matcher"/> is null.
+        /// </exception>
+        public FreezeOnMatchCustomization(
+            ParameterInfo parameter,
```
sergeyshushlyapin: Looking on this change, I can predict a question similar to [this](https://github.com/AutoFixture/AutoFixture/pull/627#discussion_r64986467) one I've been asked earlier. Despite it does solve  the original issue, I cannot come up with any *self contained* use case for the new constructor. I tend to create a new internal version of the `FreezeOnMatchCustomization` in the Glue Library and move there all the changes from this class.

ploeh: Well, yes, sort of. I'm predictable, it seems :neutral_face: 

Sorry to be a bit negative, but I think this needs to go in a slightly different direction. The overall idea, however, seems sound :+1: 

Adding a `ParameterInfo` property pulls in the direction of less cohesion, which I'm not particularly fond of. Now the class has two properties, both of which could be `null`, or both of which, according to the type system, could have a value. This is the sort of feature extension that tends to rot a code base over time.

Some time ago, @ecampidoglio and I had [a discussion about something similar](https://github.com/AutoFixture/AutoFixture/pull/317#discussion_r19058880) that might be worth revisiting. Essentially, the idea is to widen the existing 'greedy' constructor, so that it takes any `object`, instead of only a `Type`. AFAICT, there's nothing in the class's invariants that require it to be a `Type`.

Given that we now seem to have a use for this change, I think it might be worthwhile pursuing.

sergeyshushlyapin: > Sorry to be a bit negative, but I think this needs to go in a slightly different direction.

That's totally fine. I found a *potential* fix, we just need to give it a proper shape ;)

I reviewed the discussion you mentioned, thanks for sharing that with me. That is a possible solution but if I understand correctly, the `TargetType` property and both existing constructors (?) in the `FreezeOnMatchCustomization` class will need to be obsoleted.

I'm still validating other possible solutions. As a POC, I created a new internal freezing [customization](https://github.com/sergeyshushlyapin/AutoFixture/blob/freezing-behavior-internal/Src/AutoFixture.xUnit.net2/FreezeOnParameterMatchCustomization.cs). It allows not to change the existing deep-seated API but it brings some other challenges with [testing internals](http://blog.ploeh.dk/2015/09/22/unit-testing-internals/).

ecampidoglio: Sorry for the late reply but I'm currently on vacation and have only occasional access to the Internet. ☀️🍦

Looking back at the discussion, I noticed that @ploeh did, in fact, [predict this exact scenario](https://github.com/AutoFixture/AutoFixture/pull/317#discussion_r19219817): 

> Based on this, I'm inclined to keep it as is. However, I was wondering if we'd then be missing out on some opportunities, like:

    new FreezeOnMatchCustomization(somePropertyInfo);
    new FreezeOnMatchCustomization(someParameterInfo);
    new FreezeOnMatchCustomization(someFieldInfo);
    // etc.

Unfortunately, I failed to follow up on that comment so the blame is totally on me. 😔

In the face of this new requirement, I agree that relaxing the constructor of `FreezeOnMatchCustomization` sounds lile the right thing to do.

Of course, we would still have to pay complexity tax of handling multiple variants of the `target` parameter. Although it would be nice if we could restrict the subset of types that can be passed as arguments through the type system, I'm afraid that, in the end, it would still have to be `object` since there is no other contravariant type between `PropertyInfo`, `ParameterInfo`, `FieldInfo` and `Type`.

I should mention that I'm saying all this without having researched other possible solutions that don't involve modifying the constructor of `FreezeOnMatchCustomization`.

ploeh: > there is no other contravariant type between `PropertyInfo`, `ParameterInfo`, `FieldInfo` and `Type`.

Well, there's [Albedo](https://github.com/ploeh/Albedo#what-problem-does-it-address), which was created to address that particular problem... :wink:

sergeyshushlyapin: Yep. Albedo does the job!
JFI, AutoFixture Core and all the test framework Glue Libraries have to reference it.

sergeyshushlyapin:
The build error looks intimidating:
```
  1) Building C:\projects\autofixture\Src\AutoFakeItEasy.sln failed with exitcode 1.
  2) CA1506: (-1,0): Microsoft.Maintainability : 'Fixture' is coupled with 81 different types 
from 14 different namespaces. Rewrite or refactor this class's methods to decrease its class 
coupling, or consider moving some of the class's methods to some of the other types it is 
tightly coupled with. A class coupling above 95 indicates poor maintainability, a class 
coupling between 95 and 80 indicates moderate, and a class coupling below 80 indicates 
good maintainability.
```

Unfortunately I cannot reproduce it locally because of another build error with NUnit3 (even in the latest `master` branch!):
```
  1) Fake.UnitTestCommon+FailedTestsException: NUnit test failed. Process finished with 
exit code UnexpectedError (-100).
   at Fake.Testing.NUnit3.NUnit3(FSharpFunc`2 setParams, IEnumerable`1 assemblies) in 
C:\code\fake\src\app\FakeLib\UnitTest\NUnit\NUnit3.fs:line 280
   at FSI_0005.Build.clo@36-6.Invoke(Unit _arg7)
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in 
C:\code\fake\src\app\FakeLib\TargetHelper.fs:line 484
```

sergeyshushlyapin:
> 2) CA1506: (-1,0): Microsoft.Maintainability : 'Fixture' is coupled with 81 different types

Does that mean that somebody *has coupled* Fixture "80 times" and I'm the lucky man who has to decouple this? :fearful:

sergeyshushlyapin:
The issue is caused by the new `PropertyInfoElementRelay` and `TypeElementRelay` registrations in the [Fixture](https://github.com/AutoFixture/AutoFixture/pull/665/files#diff-7c98c8ff3d991f5068150fffcab7611cR116) class. The [graph](https://github.com/sergeyshushlyapin/AutoFixture/blob/648ffd138187e6127b30ac30c04a85df86c3a116/Src/AutoFixture/Fixture.cs#L65) of specimen builders is getting huge.

Any thoughts how to address this issue? Moving the graph initialization to another class(es) may just postpone the problem for a while. Any objections against disabling the rule?

Nuget/Ploeh.AutoFixture.nuspec
```diff
@@ -6,6 +6,9 @@
         <title>AutoFixture</title>
         <authors>Mark Seemann</authors>
         <owners>Mark Seemann</owners>
+        <dependencies>
+            <dependency id="Albedo" version="[1,2)" />
```
ploeh: Is this the correct dependency configuration? The actual dependency added is Albedo 1.0.1, but if I understand this configuration correctly, any 1.x version is supported - including 1.0.0.

Additionally, why exclude a hypothetical version 2 of Albedo?

Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations;
```
ploeh: Does this namespace, and its types, also exist in .NET Core?

Nuget/Ploeh.AutoFixture.nuspec
sergeyshushlyapin: I choose Albedo 1.0.1 simply because it was already referenced in [unit tests](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/packages.config#L3). I overlooked that the dependency is added to the publicly visible AutoFixture package which should reference the _lowest possible version_ (1.0.0).

I'll get back on this!

ploeh: I think it ought to reference the lowest possible version that we know works. If we're using 1.0.1, we should state that as a requirement.

If we wish to be able to state that 1.0.0 is the minimum version supported, we should downgrade our use of Albedo so that we are, in fact, testing against that version.

sergeyshushlyapin: Everything works fine with Albedo 1.0.0.

BTW, there is [AutoFixture.Igioms](https://www.nuget.org/packages/AutoFixture.Idioms) package which is already published with Albedo v1.0.1 dependency. There should be no harm in downgrading this reference to v1.0.0. Please let me know if you have any concerns about this.

ploeh: I think we should have consistent versions for shared dependencies. If AutoFixture.Idioms works with Albedo 1.0.0, I think it'd be more correct to downgrade its dependency as well. I haven't tested, but looking at the change between 1.0.0 and 1.0.1, I think you're correct that it ought to work as well.

ploeh: We should probably start with a separate pull request for that.

Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
sergeyshushlyapin: According to this [MSDN page](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.aspx?f=255&MSPPError=-2147217396) and [source code](https://github.com/dotnet/corefx/tree/master/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations) I assume it does.

ecampidoglio: Yes, the [System.ComponentModel.DataAnnotations](https://www.nuget.org/packages/System.ComponentModel.Annotations/4.1.0) exist in .NET Core, it's just not part of the [NETStandard](https://www.nuget.org/packages/NETStandard.Library/) set of packages so it has to be referenced explicitly – no big deal.

adamchester: Yep, I used it for autofixture [here](https://github.com/AutoFixture/AutoFixture/compare/v4...adamchester:netcore-step2#diff-491c68353068d4c77e3a8a02d1a84ee6R41)

ecampidoglio:
@sergeyshushlyapin When it comes to PRs, we generally prefer to have a [linear history over merge commits](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/). Could you please remove the merge commits f3c74f9 18fb59d and instead rebase your `freezing-behavior` branch on top of `master`?

We – the maintainers – could also do it ourselves once the PR is complete, if you prefer.

sergeyshushlyapin:
No problem, usually I also prefer rebasing over merge commits. I just noticed that rebasing often "collapse" old conversations, so I just tried to preserve them ;)

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -43,27 +44,53 @@ public FreezeOnMatchCustomization(Type targetType)
...
 
         /// <summary>
+        /// The <see cref="ReflectionElement"/> describing the specimen to freeze.
+        /// </summary>
+        public IReflectionElement ReflectionElement { get; }
+
+        /// <summary>
         /// The <see cref="Type"/> of the frozen specimen.
         /// </summary>
         public Type TargetType { get; }
```
ploeh: Should we deprecate the `TargetType` property?

ploeh:
It's looking good :+1: 

The only comment I have is about deprecation of the the `TargetType` property.

Src/AutoFixture/FreezeOnMatchCustomization.cs
sergeyshushlyapin: It's not used in AutoFixture anymore. Yes, I think it can be deprecated.

sergeyshushlyapin: This PR changes some _core_ classes ([FreezeOnMatchCustomization](https://github.com/sergeyshushlyapin/AutoFixture/blob/13513fb9507b3d33f56dc69d2dbd21d6727c48a0/Src/AutoFixture/FreezeOnMatchCustomization.cs)) as well as Glue Library classes ([xUnit2.FrozenAttribute](https://github.com/sergeyshushlyapin/AutoFixture/blob/13513fb9507b3d33f56dc69d2dbd21d6727c48a0/Src/AutoFixture.xUnit.net2/FrozenAttribute.cs)). I started it as xUnit.net2 contribution and intended to fix all the other libraries with follow up requests.

If I deprecate the `TargetType` property and [corresponding](https://github.com/sergeyshushlyapin/AutoFixture/blob/13513fb9507b3d33f56dc69d2dbd21d6727c48a0/Src/AutoFixture/FreezeOnMatchCustomization.cs#L33) [constructors](https://github.com/sergeyshushlyapin/AutoFixture/blob/13513fb9507b3d33f56dc69d2dbd21d6727c48a0/Src/AutoFixture/FreezeOnMatchCustomization.cs#L49) (which are in 'core'), then other libraries start failing:

```
  1) Building C:\sc\autofixture\Src\AutoFixture.NUnit2.sln failed with exitcode 1.
  2) CS0618: FrozenAttribute.cs(113,20): 'FreezeOnMatchCustomization.FreezeOnMatchCustomization(Type, IRequestSpecification)' is obsolete: 'This constructor has been deprecated. Please use constructor that takes IReflectionElement parameter.'
```

What is the best approach to go? Should I suppress the error in other libraries for the time being or postpone obsoleting the property and constructors?

sergeyshushlyapin: The property is still valid for other Glue Libraries, so it's too early to make it obsolete.

ecampidoglio:
I apologize if this feedback comes in late, but I have given this some more thought and I'd like to raise a couple of concerns.

By making `IReflectionElement` part of the public API of `FreezeOnMatchCustomization`, we're effectively coupling the AutoFixture core API to Albedo from a consumer's point of view. Now, that might not be a problem in and of itself – although I believe the AutoFixture core API has relied solely on the BCL up to this point, while Albedo was used internally.

If we're also going to deprecate the constructor that takes `Type` as an argument, consumers who wish use the `FreezeOnMatchCustomization` in their test projects will now have to take a direct dependency on Albedo. Again, this is not a problem per se but it does smell a little bit of _interface pollution_, in the sense that it forces consumers to use a library they shouldn't necessarily have to know about (Albedo) in order to take advantage of one AutoFixture's core features (freezing by criteria). AutoFixture and Albedo address different concerns and, while Albedo is _very useful_ to AutoFixture's implementation, I don't believe their respective APIs belong together.

These are my concerns. What are your thoughts?

sergeyshushlyapin:
From [Albedo](https://github.com/ploeh/Albedo#who-cares) site:
> Albedo was born out of [a need for such abstractions in AutoFixture](https://github.com/AutoFixture/AutoFixture/pull/171#discussion_r6505484). While it's easy to extract a specific interface to address a specific need, you always run the risk of defining an interface that can only be used in one very specific situation; such an interface may not be a good abstraction.

Won't we end up with Albedo2 *inside* AutoFixture if we decide to [extend the existing core API](https://github.com/AutoFixture/AutoFixture/pull/171#discussion_r6504414)?

moodmosaic:
Since some Albedo maintainers are also AutoFixture maintainers, and since it follows Semantic Versioning, it shouldn't be that bad of an idea if we add it as a dependency...

AutoFixture has relied only on BCL, but I can't currently see disadvantages if we'd have to take Albedo as a dependency :confused:

sergeyshushlyapin:
One thing worth mentioning (according to my understanding, correct me if I'm wrong) is that the `FreezeOnMatchCustomization` is [not used by AF consumers directly](http://stackoverflow.com/search?q=%5Bautofixture%5D+FreezeOnMatchCustomization). When one needs to _freeze_ a specimen, he either call the `fixture.Freeze<>()` [method](http://stackoverflow.com/search?q=%5Bautofixture%5D+Freeze) or use `[Frozen]` [attribute](http://stackoverflow.com/search?q=%5Bautofixture%5D+frozen) from Glue Libraries.

ecampidoglio:
We're discussing two different issues here.

@moodmosaic
> AutoFixture has relied only on BCL, but I can't currently see disadvantages if we'd have to take Albedo as a dependency

AutoFixture already depends on Albedo. My point is that, up until now, Albedo has been used _internally_ to simply the implementation of AutoFixture – the consumers don't need to know or care about Albedo in order to use AutoFixture in their tests – but with this PR we're effectively _surfacing_ Albedo as part AutoFixture's public API. The practical effect is that whoever needs to use the `FreezeOnMatchCustomization` is going to have to take a direct dependency on Albedo in order to use its types.

Now, if Albedo was a namespace within AutoFixture that wouldn't be much of a problem. But since Albedo is currently published as a general-purpose package, I see a problem in tying it to AutoFixture's public API.

@sergeyshushlyapin
> When one needs to freeze a specimen, he either calls the fixture.Freeze<>() method or use [Frozen] attribute from Glue Libraries.

Well, the `Freeze` method – in its current implementation – doesn't support matching criteria; if you want to use the imperative API there is currently no other way than using the `FreezeOnMatchCustomization`directly.

On a more philosophical level, I see customizations and specimen builders as AutoFixture's as _building blocks_ that are combined in order to offer functionality which is conveniently exposed through a higher level API (see the `Customize` method or the `AutoDataAttribute`). If a usage scenario isn't directly supported in the API, it should be possible to drop down to the building blocks and combine them in new ways in order to achieve the goal.

That's why I would rather see that the building blocks themselves continue to be free from dependencies on external libraries.

sergeyshushlyapin:
From Albedo we need just three things: `IReflectionElement`, `TypeElement` and `ParameterInfoElement`. Than might sound a bit impertinently, but what if (as a POC) we just "steal" them and remove the reference?... :flushed:

Zvirja:
@sergeyshushlyapin 
I have concerns regarding the approach you decided to follow. You created custom wrappers and later relays which unpacks that wrappers. Such approach leads to the performance overhead (more relays to check, more object are created, extra request is performed during resolve), however the reason behind that is not clear.

Isn't that much simpler to extend the `FreezeOnMatchCustomization` with `object` request which customization should freeze? Technically, that will make customization even more generic, because it will be capable of freezing _any_ type of requests.

It seems that root of your approach is some kind of semantic meaning, but I don't see it. Given that `fixture` accepts `object` as a request, we could also use `object` as a proxy.

P.S. Also, using `object` should reduce the complexity of overall freezing logic and number of changes applied above :)

P.P.S. Excuse me, if I missed something obvious.

sergeyshushlyapin:
> Such approach leads to the performance overhead (more relays to check, more object are created, extra request is performed during resolve)

I don't think that is the case. Relays are [widely used](https://github.com/AutoFixture/AutoFixture/search?l=csharp&q=relay&utf8=%E2%9C%93) in AutoFixture and according to test execution results [there are no performance issues](https://ci.appveyor.com/project/AutoFixture/autofixture/build/1.0.309-yetyxfjr#L6475).

> Isn't that much simpler to extend the FreezeOnMatchCustomization with object request instead of Type request that customization should freeze?

That is one of the three mostly discussed options:
- [Use objects](https://github.com/AutoFixture/AutoFixture/pull/317#discussion_r19058880)
- Introduce [new abstraction](https://github.com/AutoFixture/AutoFixture/pull/171#discussion_r6505484)
- Rely on [Albedo](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r70166045)

> If later we will need support freezing of another request sources (e.g. PropertyInfo), we will need to create one more wrapper and relay.

If we talk about Albedo than yes, we will. But are such scenarios numerous?

> Given that fixture accepts object as a request, we could use object as a request too.

Yep. The `object` parameter _does fit_ the existing design (which is ["quite weakly typed"](https://github.com/AutoFixture/AutoFixture/issues/645#issuecomment-224850261)).

ploeh:
The value proposition offered by Albedo is that it enables a client to model a variety of Reflection types in such a way that they look like they have a common consumption interface: `IReflectionVisitor<T>`. This enables a client to implement the interface. When visited by the various heterogeneous Reflection types, no type tests are required because of the double dispatch design.

This also forces a client (developer) to consider how to deal with all of the concrete `IReflectionElement` subtypes. This adds a degree of type safety to implementations.

None of these features are in play in this pull request. Instead, the two proposed relays perform type tests and use the downcast objects to access the underlying Reflection objects. In this case, I can't see that wrapping `Type` in `TypeElement` and `ParameterInfo` in `ParameterInfoElement` adds any value. We could achieve the same functionality by simply requesting `Type` and `ParameterInfo`.

Thus, I don't see that this particular use of Albedo adds any value, and thus I don't think that the added dependency is warranted.

In order to prevent more thrashing on this particular PR branch, I'm going to close it. @sergeyshushlyapin, as the owner of the fork, you still have the commits, so you should be able to salvage the good parts of this PR from the commits that introduce Albedo, so I'm not saying that everything in this PR is useless. Quite the contrary, and I'd welcome a second attempt, but I think it ought to be in a fresh PR.

---

I do realise that if anyone is responsible for introducing Albedo to this work item, [it's me](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r70166045):

> "Well, there's Albedo, which was created to address that particular problem..."

I wrote this in response to the assertion that there's no polymorphic type that can model both `PropertyInfo`, `ParameterInfo`, `FieldInfo` and `Type`. This is a problem that haunts much of the AutoFixture code base. At a certain time, it became so bad that we created Albedo to address that problem. Albedo is, in fact, a dependency of AutoFixture.Idioms, although not of AutoFixture itself.

The thought of making AutoFixture itself depend on Albedo is not foreign to me. If it's warranted, I'd be happy to entertain the idea (while still listening to arguments _against_ the notion).

I'd consider it warranted if it simplifies the user-aimed API of AutoFixture.

PR #672, opened 2016-07-09
Relaxed version requirements for AutoFakeItEasy

ploeh:
With the new implementation that doesn't rely on IHideObjectMembers, there
seems to be no reason to restrict users to FakeItEasy 1.x.

I've tried to create a local build based off this pull request, and then created a test project that uses these packages:

```
Id                                  Versions
--                                  --------
AutoFixture                         {3.49.0}
AutoFixture.AutoFakeItEasy          {3.49.0}
FakeItEasy                          {2.1.0} 
xunit                               {2.1.0} 
xunit.abstractions                  {2.0.0} 
xunit.assert                        {2.1.0} 
xunit.core                          {2.1.0} 
xunit.extensibility.core            {2.1.0} 
xunit.extensibility.execution       {2.1.0} 
```

(Do not that the two AutoFixture packages are my special builds, so not the official 3.49.0 packages.)

Then I wrote this test:

```C#
public class Class1
{
    [Fact]
    public void Test()
    {
        var fixture = new Fixture().Customize(new AutoFakeItEasyCustomization());
        var foo = fixture.Create<IFoo>();
        Assert.NotNull(foo);
    }
}
```

With this binding redirect, the test passes:

```XML
<configuration>
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="FakeItEasy" publicKeyToken="eff28e2146d5fd2c" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-2.1.0.0" newVersion="2.1.0.0" />
      </dependentAssembly>
    </assemblyBinding>
  </runtime>
</configuration>
```

If I remove the binding redirect, the test fails as expected, because it can't locate FakeItEasy version 1.7.4257.42.

ping, @blairconrad, @adamralph

### DIFF
--- a/Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
+++ b/Nuget/Ploeh.AutoFixture.AutoFakeItEasy.nuspec
@@ -8,7 +8,7 @@
         <owners>Nikos Baxevanis</owners>
         <dependencies>
             <dependency id="AutoFixture" version="@build.number@" />
-            <dependency id="FakeItEasy" version="[1.7,2)" />
+            <dependency id="FakeItEasy" version="1.7" />
         </dependencies>
         <summary>Turns AutoFixture into an Auto-Mocking Container based on FakeItEasy.</summary>
         <description>This extension turns AutoFixture into an Auto-Mocking Container. The mock instances are created by FakeItEasy. To use it, add the AutoFakeItEasyCustomization to your Fixture instance. Read more at http://autofixture.codeplex.com/</description>


adamralph:
LGTM. Does this mean the source for AutoFakeItEasy and AutoFakeItEasy2 is identical?

ploeh:
I haven't done a diff of the sources, but I'm considering discontinuing AutoFakeItEasy2.

adamralph:
Indeed, once this version of AutoFakeItEasy is released, AutoFakeItEasy2 is redundant.

I guess you could do something like this https://github.com/AutoFixture/AutoFixture/issues/663#issuecomment-229646017

Alternatively, you could link the source files between the two projects and push out both of them each time.

moodmosaic:
>I haven't done a diff of the sources, but I'm considering discontinuing AutoFakeItEasy2.

Is there an easy way of doing this without causing confusion?

What @adamralph [suggests](https://github.com/AutoFixture/AutoFixture/issues/663#issuecomment-229646017) makes sense, though isn't it a bit confusing for the existing AutoFakeItEasy2 users?

ploeh:
If you look at the [package statistics](https://www.nuget.org/packages/AutoFixture.AutoFakeItEasy2), it looks like there are some 50-odd users. I'm willing to take that hit.

moodmosaic:
And then we'll re-release/continue AutoFakeItEasy2 if there are breaking changes in FakeItEasy2 in the future?

blairconrad:
I'm late to the party, but this was exactly my thought when I proposed not relying on `IHideObjectMembers`: that AutoFakeItEasy2 could be made redundant.
I did worry, as @moodmosaic does, a little about stranding the loyal users of the package, but figured you _might_ want to take the hit since the package was so new.

The only differences are in how the `NoSpecimen` objects were created: including the original request or not.

> And then we'll re-release/continue AutoFakeItEasy2 if there are breaking changes in FakeItEasy2 in the future?

That sounds reasonable, although we do not intend to release breaking changes in FakeItEasy 2.x. SemVer dictates that we'd release FakeItEasy 3.x. Which we're talking about actually, but the breaking change we're considering so far would not impact AutoFakeItEasy.

moodmosaic:
Thanks, @blairconrad and @adamralph, for the great feedback!

PR #673, opened 2016-07-09
Support netstandard1.X (step 1)

adamchester:
This begins support for `netstandard` (and coreclr / dotnet core / whatever it's called now) by doing the following:

 * Use the [evolved reflection APIs](https://blogs.msdn.microsoft.com/dotnet/2012/08/28/evolving-the-reflection-api/) (mostly `GetTypeInfo`)
 * Introduces `SYSTEM_NET_MAIL` and `SYSTEM_RUNTIME_SERIALIZATION` defines. `System.Net.Mail` APIs (used by AutoFixture in the `MailAddressGenerator`) are currently only available on the full .net framework (e.g. net45), so we wrap them entirely in an `#if SYSTEM_NET_MAIL` directive. Similarly, serialization (including `SerializableAttribute` and binary remoting serialization) is only available on the full .net framework, so we use `#if SYSTEM_RUNTIME_SERIALIZATION` to allow us to compile on targets that don't have those features.

After these changes, I was able to create a file named `Src\AutoFixture\project.json`:
```json
{
  "version": "1.0.0-*",
  "buildOptions": {
    "debugType": "portable"
  },
  "dependencies": {},
  "frameworks": {
    "net45": {
      "buildOptions": { "define": [ "SYSTEM_NET_MAIL", "SYSTEM_RUNTIME_SERIALIZATION" ] },
      "frameworkAssemblies": {
        "System.ComponentModel.DataAnnotations": ""
      }
    },
    "netstandard1.5": {
      "dependencies": {
        "NETStandard.Library": "1.6.0",
        "System.Threading.Thread": "4.0.0",
        "System.ComponentModel.Annotations": "4.1.0"
      }
    }
  }
}
```

Which allowed me to compile the core AutoFixture library:
image

### DIFF
--- a/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
+++ b/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
@@ -36,7 +36,7 @@ namespace Ploeh.AutoFixture.Kernel
                     GetFriendlyName(methodInfo.ReturnType),
                     methodInfo.DeclaringType.FullName,
                     methodInfo.Name,
-                    string.Join(", ", methodInfo.GetGenericArguments().Select(a => a.ToString())),
+                    string.Join(", ", methodInfo.GetTypeInfo().GetGenericArguments().Select(a => a.ToString())),
                     string.Join(", ", methodInfo.GetParameters().Select(p => GetFriendlyName(p.ParameterType)))
                     ))
         {
@@ -90,7 +90,7 @@ namespace Ploeh.AutoFixture.Kernel
                 return string.Format(CultureInfo.CurrentCulture, 
                     "{0}<{1}>", 
                     type.Name.Split('`')[0], 
-                    string.Join(", ", type.GetGenericArguments().Select(GetFriendlyName)));
+                    string.Join(", ", type.GetTypeInfo().GetGenericArguments().Select(GetFriendlyName)));
 
             return type.Name;
         }
--- a/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
+++ b/Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
@@ -72,7 +72,7 @@ namespace Ploeh.AutoFixture.Kernel
         /// </param>
         public TypeArgumentsCannotBeInferredException(string message, Exception innerException)
             : base(message, innerException)
-         {
+        {
         }
 
 #if SYSTEM_RUNTIME_SERIALIZATION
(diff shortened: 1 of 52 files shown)


Src/AutoFixture/Kernel/TypeArgumentsCannotBeInferredException.cs
```diff
@@ -68,9 +72,10 @@ public TypeArgumentsCannotBeInferredException(string message)
         /// </param>
         public TypeArgumentsCannotBeInferredException(string message, Exception innerException)
             : base(message, innerException)
-        {
+         {
         }
```
moodmosaic: Oops! There's an extra space added here (?) 😃 😉

moodmosaic:
Nice work 👍

I don't think there's a better way, other than introducing these compiler directives.

I've gone through the changes and they look good to me.

adamchester:
Warning: this PR is currently set to merge into the `v4` branch. I'm not so sure about that anymore. When I started the branch, I think I was assuming incorrectly that I would be introducing breaking changes, but that hasn't happened yet.

I now have a proposed way to integrate this into the current build system, however it would require us to use some preview tooling, so I haven't included that in this PR.

The plan is to compile against `netstandard1.3` via `dotnet build` (RTM/preview2) which gives us compiler errors if we accidentally introduced new breaking changes (for example, we might accidentally start using the _old_ reflection APIs again). It would use the preview2 `xproj` style project (aka. the msbuild-integrated way of compiling project.json-based projects).

We know that the current `project.json` is being deprecated in favour of a more msbuild-integrated approach before the tooling is released. However, we don't know when that will happen, and it could be many months away. The `project.json` style has been widely used by many projects now, to target the newer dotnet core platforms.

moodmosaic:
Interesting analysis.

If they're deprecating `project.json`, probably the best thing we could do is wait and see what's going to happens (?)

hoetz:
FWIW this is from the .NET blog

> Q4 2016 / Q1 2017
This will be the first minor update, mainly focused on replacing .xproj/`project.json` with .csproj/MSBuild. Project format update should be automatic. Just opening a 1.0 project will update it to the new project format. There will also be new functionality and improvements in the runtime and libraries.

Judging from that, there will be an easy migration path when leaving `project.json` behind.

ploeh:
This looks good, but I agree that it should go into _master_, since there are no breaking changes in this PR.

Could you re-issue the PR to be based on _master_?

adamchester:
My mistake; this will require `net45` (and this `v4` branch) because `GetTypeInfo` and related [evolved reflection APIs](https://blogs.msdn.microsoft.com/dotnet/2012/08/28/evolving-the-reflection-api/) are only available in `net45`+

hoetz:
Is there anything that could help to do "step 2" ?

adamchester:
Hi @hoetz, I made a start at enabling a .net core build verification step you can see here https://github.com/AutoFixture/AutoFixture/compare/v4...adamchester:netcore-step2

I simply haven't gotten back to it yet :cry:

hoetz:
@adamchester ok if there is anything I can do to help, let me know

PR #675, opened 2016-07-11
 Added example of using OmitOnRecursionBehavior in exception message

### LINKED ISSUE #667: Add more details to "traversed object graph contains a circular reference" error
When using AutoFixture, I've got the following exception :

> Ploeh.AutoFixture.ObjectCreationException: AutoFixture was unable to create an instance of type System.Reflection.RuntimePropertyInfo
> because the traversed object graph contains a circular reference. Information about the circular path follows below. This is the correct behavior when a Fixture is equipped with a ThrowingRecursionBehavior, which is the default. This ensures that you are being made aware of circular references in your code. Your first reaction should be to redesign your API in order to get rid of all circular references. However, if this is not possible (most likely because parts or all of the API is delivered by a third party), you can replace this default behavior with a different behavior: on the Fixture instance, remove the ThrowingRecursionBehavior from Fixture.Behaviors, and instead add an instance of OmitOnRecursionBehavior.
> 

It will be good to provide URL to the help page with example, e.g. to Wiki  https://github.com/AutoFixture/AutoFixture/wiki/Examples-of-using-behaviors

rahulpnath:
This pull request fixes #667. It adds a sample usage of OmitOnRecursionBehavior to the exception message. 

Attaching the console output snapshots for reference
**Before fix**
before fix

**After fix**
after fix


ploeh:
Thank you for your contribution! It's now live as [AutoFixture 3.49.1](https://www.nuget.org/packages/AutoFixture/3.49.1).

rahulpnath:
@ploeh  aah so that happened when Damian was talking ☺. Glad to meet you in person today and great talk!

PR #676, opened 2016-07-12
Install NUnit3TestAdapter in order to run NUnit3 tests in VS Test Explorer

sergeyshushlyapin:
This PR allows VS [Test Explorer](https://msdn.microsoft.com/en-us/library/hh270865.aspx) to run NUnit3 tests.

Before the [NUnit3TestAdapter](https://www.nuget.org/packages/NUnit3TestAdapter/3.0.8-ctp-8) is installed (2375 tests):
image

After (2456 tests):
image


ploeh:
Shouldn't this commit also contain the binaries as well?

sergeyshushlyapin:
Hm.. I did add the package content to the source control but you're right, the [binaries](https://github.com/AutoFixture/AutoFixture/pull/676/files#diff-4c306cfc5f9443f16916d1d57f966cf2R68)  appeared to be missing. Will add them. Thanks.

ploeh:
When I try this PR out, I get 30 tests not run because of an unknown error:

image

Is this expected?

sergeyshushlyapin:
No, I don't think it's expected. It looks like a bug in the Test Adapter and I found it described [here](https://github.com/nunit/nunit3-vs-adapter/issues/188).

It seems that the Test Adapter (in it's current state) is unable to report properly if all the tests pass. One still have to run `build.bat` or use alternative runner ([NCrunch](http://www.ncrunch.net/) works perfectly fine). Does it make this PR redundant?

ploeh:
> Does it make this PR redundant?

I don't know; you tell me. I don't use the Visual Studio test runner, so I have no use for this PR at all.

It's fine that people have other workflows than mine, and I think that it's important to support these as well. If this PR addresses a concern you have, then I'm happy to accept it (it doesn't, AFAICT, incur much of a maintenance overhead), but I can't evaluate that. That's the reason I'm asking.

moodmosaic:
>One still have to run `build.bat`

It'd be nice if we could get rid, or automate, this step.

>or use alternative runner

Some other test-runners work fine OOTB, such is the [TestDriven.Net](http://testdriven.net/). This PR is for discovering and running unit tests with Microsoft Visual Studio's Test Explorer, so it'd be nice if it 'just works' there.

sergeyshushlyapin:
> It'd be nice if we could get rid, or automate, this step.

I don't see any problems with running `build.bat` manually *before* pushing changes to the remote. That's totally fine.

> This PR is for discovering and running unit tests with Microsoft Visual Studio's Test Explorer, so it'd be nice if it 'just works' there.

@moodmosaic, agree with you, it should 'just work'. If the NUnitTestAdapter bug I [mentioned before](https://github.com/AutoFixture/AutoFixture/pull/676#issuecomment-238068206) solves the test running issue (when/if fixed), I want this PR to be updated and merged. Otherwise no NUnit adapters should appear in AF code base ;)

sergeyshushlyapin:
Have no value till https://github.com/nunit/nunit3-vs-adapter/issues/188 is fixed. Might be reopen later on.

Issue #678, opened 2016-07-13
Feature: Create a string specimen with Min/MaxLength constraint


sergeyshushlyapin:
The following two tests fail:

``` csharp
[Theory, AutoData]
public void CreateParameterWithMinLengthConstraint([MinLength(5)]string p)
{
    Assert.True(p.Length == 5); // fails; actual is 37
}

[Theory, AutoData]
public void CreateParameterWithMaxLengthConstraint([MaxLength(5)]string p)
{
    Assert.True(p.Length == 5); // fails; actual is 37
}
```

In order to make it pass two new _relays_ similar to [StringLengthAttributeRelay](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DataAnnotations/StringLengthAttributeRelay.cs) need to be implemented.

**Important:** the [MaxLengthAttribute](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.maxlengthattribute%28v=vs.110%29.aspx) and [MinLengthAttribute](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.minlengthattribute%28v=vs.110%29.aspx) classes are available since .Net 4.5 so those AutoFixture projects that are still on .Net 4.0 have to be upgraded (that should cause no problems since 4.5 is [backward compatible](https://blogs.msdn.microsoft.com/dotnet/2011/09/26/compatibility-of-net-framework-4-5/) with 4.0).

micheleissa:
@zvirja Could you point me to the right direction on this one. I feel I can do it and will be a good start on the real stuff :)

zvirja:
@micheleissa I'd look at #711 and fix all the concerns we had there. The approach looked fine :wink:

micheleissa:
Gotcha.I'll look into #711

damian-krychowski:
@ztsmith  shouldn't the tests be rather: 

```c#
[Theory, AutoData]
public void CreateParameterWithMinLengthConstraint([MinLength(5)]string p)
{
    Assert.True(p.Length >= 5); //fails; actual is 37 - if actual is 37 than it is ok, as it is longer than 5
}

[Theory, AutoData]
public void CreateParameterWithMaxLengthConstraint([MaxLength(5)]string p)
{
    Assert.True(p.Length <= 5); // fails; actual is 37 - here something is wrong
}
```

Don't these make more sense? Asserting that length should be always equal to 5 is quite strange for me.

[...]

PR #679, opened 2016-07-13
Upgrade AutoFixture to .NET 4.5

sergeyshushlyapin:
This PR is an enabler for #678.


ploeh:
The _v4_ branch is already on .NET 4.5, for the reason that I consider it a breaking change to upgrade.

If people are still using AutoFixture on .NET 4.0, then upgrading to 4.5 will prevent users from moving forward. Therefore, it's a breaking change.

Sorry that maintaining an open source project is so difficult sometimes, but such is the environment.

PR #688, opened 2016-08-07
Downgrade the required Albedo package version to 1.0.0

sergeyshushlyapin:
which satisfies the current AutoFixture needs and in case of referencing by AutoFixture packages will not force consumers to upgrade their dependencies.

This PR is a result of [this discussion](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r73807476).


ploeh:
Looks good, but didn't you forget to change the dependency configuration in `Nuget/Ploeh.AutoFixture.Idioms.nuspec`?

sergeyshushlyapin:
Nope. It's [1.0.0](https://github.com/AutoFixture/AutoFixture/pull/688/commits/8eff189a4233c516d20c67bd80bbae44a742ea7d#diff-3657e6426db4722bdd50ea7e8af8dd2fR11):

`           <dependency id="Albedo" version="1.0.0" />`

ploeh:
I don't know how I managed to miss that, but it could be the jetlag :flushed: Sorry about that!

Issue #689, opened 2016-08-11
Minimum string length constraint does not work


sergeyshushlyapin:
The following test fails:
```csharp
[Theory, AutoData]
public void CreateLongStringSpecimen(
    [StringLength(50, MinimumLength = 50)] string p)
{
    Assert.Equal(50, p.Length); // fails. Actual is always 36
}
```

PR #690, opened 2016-08-11
Minimum string length constraint does not work (#689)

sergeyshushlyapin:
Addresses issue #689.

### DIFF
--- a/Src/AutoFixtureUnitTest/DataAnnotations/StringLengthArgumentSupportTests.cs
+++ b/Src/AutoFixtureUnitTest/DataAnnotations/StringLengthArgumentSupportTests.cs
@@ -16,10 +16,10 @@ namespace Ploeh.AutoFixtureUnitTest.DataAnnotations
             // Fixture setup
             var fixture = new Fixture();
             // Exercise system
-            var actual = fixture.Create<ClassWithLengthConstrainedConstructorArgument>();
+            var actual = fixture.Create<ClassWithShortStringLengthConstrainedConstructorArgument>();
             // Verify outcome
             Assert.True(
-                actual.ShortText.Length <= ClassWithLengthConstrainedConstructorArgument.ShortTextMaximumLength,
+                actual.ShortText.Length <= ClassWithShortStringLengthConstrainedConstructorArgument.ShortTextMaximumLength,
                 "AutoFixture should respect [StringLength] attribute on constructor arguments.");
             // Teardown
         }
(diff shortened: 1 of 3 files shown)


Src/AutoFixtureUnitTest/DataAnnotations/StringLengthArgumentSupportTests.cs
```diff
@@ -24,15 +24,33 @@ public void FixtureCorrectlyCreatesShortText()
...
         private class ClassWithLengthConstrainedConstructorArgument
         {
             public const int ShortTextMaximumLength = 3;
+            public const int LongTextLength = 50;
             public readonly string ShortText;
+            public readonly string LongText;
 
             public ClassWithLengthConstrainedConstructorArgument(
-                [StringLength(ShortTextMaximumLength)]string shortText)
+                [StringLength(ShortTextMaximumLength)]string shortText,
```
ploeh: How do the changes in this class impact existing tests? Could it cause existing tests to become false negatives?

sergeyshushlyapin: In this particular case I don't think so. The class is used [only once](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/DataAnnotations/StringLengthArgumentSupportTests.cs#L19), there are no any customizations applied. But I understand what you're talking about. Sometimes it might be hard to track if a test stopped working.

I'm fine with splitting it.

Src/AutoFixtureUnitTest/DataAnnotations/StringLengthAttributeRelayTest.cs
```diff
@@ -112,5 +112,31 @@ public void CreateWithConstrainedStringRequestReturnsCorrectResult(int maximum)
...
+        {
+            // Fixture setup
+            var stringLengthAttribute = new StringLengthAttribute(maximum) { MinimumLength = 1 };
+            var providedAttribute = new ProvidedAttribute(stringLengthAttribute, true);
+            ICustomAttributeProvider request = new FakeCustomAttributeProvider(providedAttribute);
+            var expectedRequest = new ConstrainedStringRequest(stringLengthAttribute.MinimumLength, stringLengthAttribute.MaximumLength);
+            var expectedResult = new object();
+            var context = new DelegatingSpecimenContext
+            {
+#pragma warning disable 618
```
ploeh: Instead of disabling the warning, why not return `new NoSpecimen()`?

Src/AutoFixtureUnitTest/DataAnnotations/StringLengthAttributeRelayTest.cs
```diff
@@ -112,5 +112,29 @@ public void CreateWithConstrainedStringRequestReturnsCorrectResult(int maximum)
             Assert.Equal(expectedResult, result);
             // Teardown
         }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(3)]
+        public void CreateWithStringRequestConstrainedbyMinimunLengthReturnsCorrectResult(int maximum)
```
sergeyshushlyapin: This test is a *slightly* extended version of the [previous one](https://github.com/sergeyshushlyapin/AutoFixture/blob/4d526237d012bbfea42fe9e4e9d75db92d8dabc9/Src/AutoFixtureUnitTest/DataAnnotations/StringLengthAttributeRelayTest.cs#L94) which has got `MinimumLength` configured explicitly ...

```csharp
 var stringLengthAttribute = new StringLengthAttribute(maximum) { MinimumLength = 1 };
```

... instead of the default value used in previous test:

```
 var stringLengthAttribute = new StringLengthAttribute(maximum);
```

It's impossible to break the old test without breaking the new one. In the same time, the new one checks the scenario that is not covered in the old test. This makes the old test redundant, and if we all agree on that, I'd like to remove it.

Any objections?

moodmosaic: That [other](https://github.com/sergeyshushlyapin/AutoFixture/blob/4d526237d012bbfea42fe9e4e9d75db92d8dabc9/Src/AutoFixtureUnitTest/DataAnnotations/StringLengthAttributeRelayTest.cs#L94) one exercises just the `MaximumLength`, while this one does both. I think it wouldn't hurt to keep both. I'd be nice, IMO, if we can keep [test suites append-only](http://blog.ploeh.dk/2013/04/02/why-trust-tests/).

ploeh: I'm not sure that I understand. The GitHub diff view seems to indicate that this test is a new test - not a modification of an existing test...

sergeyshushlyapin: I copy-pasted if from the previous one and extended slightly.

ploeh:
Thank you for your contribution :+1: It's now live as [AutoFixture 3.50.1](https://www.nuget.org/packages/AutoFixture/3.50.1).

PR #695, opened 2016-08-16
Passive Customization Attributes are ignored when used in combination with [Frozen] (#637, xUnit.net2)

sergeyshushlyapin:
Addresses #637 introducing _Request Factory_ in the `FreezeOnMatchCustomization` which allows to use `ParameterInfo` instead of `TargetType` to resolve specimens in glue libraries.

### DIFF
--- a/Src/AutoFixture.xUnit.net2/FrozenAttribute.cs
+++ b/Src/AutoFixture.xUnit.net2/FrozenAttribute.cs
@@ -110,7 +110,7 @@ namespace Ploeh.AutoFixture.Xunit2
                 .Or(ByParameterName(type, name))
                 .Or(ByFieldName(type, name));
 
-            return new FreezeOnMatchCustomization(type, filter);
+            return new FreezeOnMatchCustomization(parameter, filter);
         }
 
         private static IRequestSpecification ByEqual(object target)
--- a/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
@@ -36,6 +36,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.ComponentModel.DataAnnotations" />
     <Reference Include="System.Core" />
     <Reference Include="xunit.abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c, processorArchitecture=MSIL">
       <HintPath>..\..\Packages\xunit.abstractions.2.0.0\lib\net35\xunit.abstractions.dll</HintPath>
--- a/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/Scenario.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using Ploeh.TestTypeFoundation;
 using Xunit;
@@ -415,6 +416,13 @@ namespace Ploeh.AutoFixture.Xunit2.UnitTest
             FieldHolder<string> p2)
         {
             Assert.NotEqual(p1, p2.Field);
+        }
+
+        [Theory, AutoData]
+        public void FreezeParameterWithStringLengthConstraintShouldCreateConstrainedSpecimen(
+            [Frozen, StringLength(3)]string p)
+        {
+            Assert.True(p.Length == 3);
         }
     }
 }


Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -75,6 +80,40 @@ public FreezeOnMatchCustomization(Type targetType)
...
+        /// <para>
+        /// A typical use case is to create a request which includes significant 
+        /// information about a specimen. For instance, <see cref="ParameterInfo"/> 
+        /// which is widely used in Glue Libraries can contain custom attributes
+        /// such as <see cref="StringLengthAttribute"/> or <see cref="RegularExpressionAttribute" />.
+        /// These attributes are essential part of the request since they influence 
+        /// the final specimen.
+        /// </para>
+        /// </remarks>
+        public Func<object> RequestFactory
```
ploeh: Why is this a factory, instead of a value?

sergeyshushlyapin: That's a good question. I started with a virtual `CreateRequest` method that could create a _proper_ request in Glue Libraries (overriding the default implementation in new classes like `xUnit2.FreezeOnPatameterMatchCustomization`). Then I realized that the method isn't really necessary and we can easily avoid creating such classes using delegates...

I haven't really considered using a property (maybe because we've [discussed it before](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r69581627))... It seems to me that the factory method here is a bit more eloquent. Please compare how it might be used in the `FrozenAttribute`:

```csharp
return new FreezeOnMatchCustomization(type, filter) {
    RequestFactory = () => parameter
};
```

```csharp
return new FreezeOnMatchCustomization(type, filter) {
    Request = parameter
};
```

ploeh: What value does using a factory provide?

sergeyshushlyapin: It appeared to me that this factory is a method which _emphasizes_ its goal to crate requests. The difference with a property is minimal and I don't have a strong opinion that it must be a delegate.
No other reasons. Would you prefer to convert it into `Request` property?

ploeh: Clearly, a factory's goal is to create something, but I don't understand why `FreezeOnMatchCustomization` needs to _create_ a new request every time it does what it has to do. Particularly not if the implementation always closes over a single object anyway...

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -75,6 +80,41 @@ public FreezeOnMatchCustomization(Type targetType)
...
+        /// should be set into the Request property.
+        /// </para>
+        /// </remarks>
+        public object Request
+        {
+            get
+            {
+                return this.request;
+            }
+            set
```
ecampidoglio: It seems to me that [resolving the request as `object`](https://github.com/AutoFixture/AutoFixture/pull/695/files#diff-ff3511dac1948f3316dc6bf6343595eaL103) makes `TargetType` obsolete, so there is no need to keep it around. Instead, we could add a constructor that accepts the `request` object and set it that way instead of using a property setter.

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -17,6 +19,8 @@ namespace Ploeh.AutoFixture
     /// </remarks>
     public class FreezeOnMatchCustomization : ICustomization
     {
+        private object request;
```
ecampidoglio: Why not assign this field through [a new constructor](https://github.com/AutoFixture/AutoFixture/pull/695/files#r75438371)? That way we could also make it _immutable_.

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -61,6 +65,7 @@ public FreezeOnMatchCustomization(Type targetType)
 
             this.TargetType = targetType;
             this.Matcher = matcher;
+            this.Request = this.TargetType;
```
ecampidoglio: Instead of assigning the `request` through a property, this constructor could simply invoke the [new one that accepts the `request` as `object`](https://github.com/AutoFixture/AutoFixture/pull/695/files#r75438371).

Src/AutoFixture.xUnit.net2/FrozenAttribute.cs
```diff
@@ -110,7 +110,10 @@ private ICustomization FreezeByCriteria(ParameterInfo parameter)
                 .Or(ByParameterName(type, name))
                 .Or(ByFieldName(type, name));
 
-            return new FreezeOnMatchCustomization(type, filter);
+            return new FreezeOnMatchCustomization(type, filter)
```
ecampidoglio: By adding a [new constructor that accepts the `request` as `object`](https://github.com/AutoFixture/AutoFixture/pull/695/files#r75438371), it should be enough to say `return new FreezeOnMatchCustomization(request, filter)`, shouldn't it?

Src/AutoFixture/FreezeOnMatchCustomization.cs
sergeyshushlyapin: @ecampidoglio you know, there was a long discussion in #317, we discussed it [recently in #665](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r70165230). Those discussions revealed so many pros and cons that finding an acceptable solution looks almost unreachable.

In order to fix #637, the `TargetType` needs to be changes in a _single place_ hidden from the consumer eyes (deep in `[FrozenAttribute]`). Now I don't think it worth changing `TargetType` to `object` publicly in the `FreezeOnMatchCustomization` (opening [undesired use cases](https://github.com/AutoFixture/AutoFixture/pull/317#discussion_r19219817)) because of this tiny change. Does it may sense?

Src/AutoFixture/FreezeOnMatchCustomization.cs
sergeyshushlyapin: That is not one but two new constructors have to be defined to replace the [existing](https://github.com/sergeyshushlyapin/AutoFixture/blob/e3d79a8b583b3fe7783471b0b9897ce6ac2b641b/Src/AutoFixture/FreezeOnMatchCustomization.cs#L36) [two](https://github.com/sergeyshushlyapin/AutoFixture/blob/e3d79a8b583b3fe7783471b0b9897ce6ac2b641b/Src/AutoFixture/FreezeOnMatchCustomization.cs#L52).

Making the `Request` property protected and further configuring it in a derived _internal_ customization class in a Glue Library will solve the mutation issues (and will prevent the public API from changes).

sergeyshushlyapin:
Please take a look on the latest commits that do not change the public API and do not introduce controversial constructor overloads and obsoletes. Please let me know what do you think about such approach.

sergeyshushlyapin:
The `Request` property is still not immutable but now it's protected. Is this really a _flaw_ comparing to the fact that well-know API stays untouched and we won't need to deal with obsolete constructors and `TargetType` property?

sergeyshushlyapin:
@ecampidoglio, finally I got your idea with obsoletes! (I hope :))
In the long term, the new `Request` property should _completely substitute_ the existing `TargetType`. It adds some overhead now as [you've mentioned before](https://github.com/AutoFixture/AutoFixture/pull/665#discussion_r70165230) due to numerous usages (in tests) but should it should be simplified once the obsoletes are removed in v4.

Since obsoleting `TargetType` impacts _all the glue libraries_, I'll come up with a new PR for that (and get back to this one later on).

ploeh:

ecampidoglio:
Based on what @sergeyshushlyapin wrote in [this comment](https://github.com/AutoFixture/AutoFixture/pull/695#issuecomment-241165973):

> Since obsoleting `TargetType` impacts all the glue libraries, I'll come up with a new PR for that (and get back to this one later on).

I think it's safe to close this PR for now.

ecampidoglio:
Closing based on the latest feedback from the author. Please, feel free to reopen it or submit a new PR entirely if you wish to continue working on this issue.

sergeyshushlyapin:
@ecampidoglio, sorry if I wasn't really clear here... This PR is broken since it is _waiting_ for #697.... It's still actual and will became green as soon as #697 is merged.

ecampidoglio:
@sergeyshushlyapin Oh, I didn't realize that. Reopened.

sergeyshushlyapin:
@zvirja, why do you call it _enhancement_? It's a [bugfix](https://github.com/AutoFixture/AutoFixture/issues/637#event-679359649).

zvirja:
@sergeyshushlyapin Yep, you are right, fixed.

zvirja:
#802 has been merged. That means that we should re-target this one on top of `v4` and, probably, do it for all the test frameworks together (to not miss them later).

@sergeyshushlyapin Could you please follow up? Do not create a new PR - simply modify this one.

zvirja:
Closing this one in favor of #816.

PR #697, opened 2016-08-20
Obsolete TargetType property in FreezeOnMatchCustomization

sergeyshushlyapin:
Enabler for #695 and list of further fixes for all the glue libraries.

In the `FreezeOnMatchCustomization` class, obsoletes `TargetType` property and introduces new `Request` property.

### DIFF
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -70,6 +70,23 @@ namespace Ploeh.AutoFixture
         /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
         /// </summary>
         /// <param name="request">The request used to create a specimen to freeze.</param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="request"/> is null.
+        /// </exception>
+        public FreezeOnMatchCustomization(object request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            this.Request = request;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
+        /// </summary>
+        /// <param name="request">The request used to create a specimen to freeze.</param>
         /// <param name="matcher">
         /// The <see cref="IRequestSpecification"/> used to match the requests
         /// that will be satisfied by the frozen specimen.
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -32,16 +32,7 @@ namespace Ploeh.AutoFixture
             }
 
             this.Request = request;
-
-            var targetType = request as Type;
-            if (targetType != null)
-            {
-                this.Matcher = new ExactTypeSpecification(targetType);
-            }
-            else
-            {
-                this.Matcher = new EqualRequestSpecification(request);
-            }
+            this.Matcher = new EqualRequestSpecification(request);
         }
 
         /// <summary>
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -29,6 +29,7 @@ namespace Ploeh.AutoFixture
         /// <exception cref="ArgumentNullException">
         /// <paramref name="targetType"/> is <see langword="null"/>.
         /// </exception>
+        [Obsolete("Please use FreezeOnMatchCustomization(object, IRequestSpecification) instead.")]
         public FreezeOnMatchCustomization(Type targetType)
             : this(targetType, new ExactTypeSpecification(targetType))
         {
@@ -45,6 +46,7 @@ namespace Ploeh.AutoFixture
         /// <exception cref="ArgumentNullException">
         /// <paramref name="targetType"/> or <paramref name="matcher"/> is null.
         /// </exception>
+        [Obsolete("Please use FreezeOnMatchCustomization(object, IRequestSpecification) instead.")]
         public FreezeOnMatchCustomization(
             Type targetType,
             IRequestSpecification matcher)
@@ -96,6 +98,7 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// The <see cref="Type"/> of the frozen specimen.
         /// </summary>
+        [Obsolete("Please use the Request property instead.")]
         public Type TargetType { get; }
 
         /// <summary>
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -37,9 +37,6 @@ namespace Ploeh.AutoFixture
             if (targetType != null)
             {
                 this.Matcher = new ExactTypeSpecification(targetType);
-#pragma warning disable 618
-                this.TargetType = targetType;
-#pragma warning restore 618
             }
             else
             {
@@ -80,7 +77,7 @@ namespace Ploeh.AutoFixture
         /// The <see cref="Type"/> of the frozen specimen.
         /// </summary>
         [Obsolete("Please use the Request property instead.")]
-        public Type TargetType { get; }
+        public Type TargetType => this.Request as Type;
 
         /// <summary>
         /// The request used to resolve specimens. By default that is TargetType.
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -104,19 +104,19 @@ namespace Ploeh.AutoFixture
                 throw new ArgumentNullException(nameof(fixture));
             }
 
-            FreezeTypeForMatchingRequests(fixture);
+            FreezeSpecimenForMatchingRequests(fixture);
         }
 
-        private void FreezeTypeForMatchingRequests(IFixture fixture)
+        private void FreezeSpecimenForMatchingRequests(IFixture fixture)
         {
             fixture.Customizations.Insert(
                 0,
                 new FilteringSpecimenBuilder(
-                    FreezeTargetType(fixture),
+                    FreezeSpecimen(fixture),
                     this.Matcher));
         }
 
-        private ISpecimenBuilder FreezeTargetType(IFixture fixture)
+        private ISpecimenBuilder FreezeSpecimen(IFixture fixture)
         {
             var context = new SpecimenContext(fixture);
             var specimen = context.Resolve(this.Request);
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -20,46 +20,9 @@ namespace Ploeh.AutoFixture
         /// <summary>
         /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
         /// </summary>
-        /// <remarks>
-        /// The constructed customization will use the frozen specimen
-        /// to satisfy requests for the exact same type.
-        /// This mimics the behavior of the <see cref="FreezingCustomization"/>.
-        /// </remarks>
-        /// <param name="targetType">The <see cref="Type"/> to freeze.</param>
-        /// <exception cref="ArgumentNullException">
-        /// <paramref name="targetType"/> is <see langword="null"/>.
-        /// </exception>
-        [Obsolete("Please use FreezeOnMatchCustomization(object) instead.")]
-        public FreezeOnMatchCustomization(Type targetType)
-            : this((object)targetType)
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
-        /// </summary>
-        /// <param name="targetType">The <see cref="Type"/> to freeze.</param>
-        /// <param name="matcher">
-        /// The <see cref="IRequestSpecification"/> used to match the requests
-        /// that will be satisfied by the frozen specimen.
-        /// </param>
-        /// <exception cref="ArgumentNullException">
-        /// <paramref name="targetType"/> or <paramref name="matcher"/> is null.
-        /// </exception>
-        [Obsolete("Please use FreezeOnMatchCustomization(object, IRequestSpecification) instead.")]
-        public FreezeOnMatchCustomization(
-            Type targetType,
-            IRequestSpecification matcher)
-            : this((object)targetType, matcher)
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
-        /// </summary>
         /// <param name="request">The request used to create a specimen to freeze.</param>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="request"/> is null.
+        /// <paramref name="request"/> is <see langword="null"/>.
         /// </exception>
         public FreezeOnMatchCustomization(object request)
         {
@@ -69,6 +32,7 @@ namespace Ploeh.AutoFixture
             }
 
             this.Request = request;
+
             var targetType = request as Type;
             if (targetType != null)
             {
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -81,6 +81,7 @@ namespace Ploeh.AutoFixture
             }
 
             this.Request = request;
+            this.Matcher = new TrueRequestSpecification();
         }
 
         /// <summary>
--- a/Src/AutoFixture/FreezeOnMatchCustomization.cs
+++ b/Src/AutoFixture/FreezeOnMatchCustomization.cs
@@ -43,7 +43,7 @@ namespace Ploeh.AutoFixture
         /// that will be satisfied by the frozen specimen.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="targetType"/> or<paramref name="matcher"/> is null.
+        /// <paramref name="targetType"/> or <paramref name="matcher"/> is null.
         /// </exception>
         public FreezeOnMatchCustomization(
             Type targetType,
@@ -60,6 +60,36 @@ namespace Ploeh.AutoFixture
             }
 
             this.TargetType = targetType;
+            this.Request = targetType;
+            this.Matcher = matcher;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FreezeOnMatchCustomization"/> class.
+        /// </summary>
+        /// <param name="request">The request used to create a specimen to freeze.</param>
+        /// <param name="matcher">
+        /// The <see cref="IRequestSpecification"/> used to match the requests
+        /// that will be satisfied by the frozen specimen.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="request"/> or <paramref name="matcher"/> is null.
+        /// </exception>
+        public FreezeOnMatchCustomization(
+            object request,
+            IRequestSpecification matcher)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            if (matcher == null)
+            {
+                throw new ArgumentNullException(nameof(matcher));
+            }
+
+            this.Request = request;
             this.Matcher = matcher;
         }
 
@@ -67,6 +97,11 @@ namespace Ploeh.AutoFixture
         /// The <see cref="Type"/> of the frozen specimen.
         /// </summary>
         public Type TargetType { get; }
+
+        /// <summary>
+        /// The request used to resolve specimens. By default that is TargetType.
+        /// </summary>
+        public object Request { get; }
 
         /// <summary>
         /// The <see cref="IRequestSpecification"/> used to match the requests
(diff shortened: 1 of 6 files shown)


Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -29,6 +29,7 @@ public class FreezeOnMatchCustomization : ICustomization
         /// <exception cref="ArgumentNullException">
         /// <paramref name="targetType"/> is <see langword="null"/>.
         /// </exception>
+        [Obsolete("Please use FreezeOnMatchCustomization(object, IRequestSpecification) instead.")]
```
ploeh: For a user so far calling `FreezeOnMatchCustomization(t)`, we're now telling him or her to use an overload where a second `IRequestSpecification` argument is required, where before, none was required. What should that user supply as that second argument?

ploeh:
While I understand that this change is a precursor to #695, does it add any value viewed in isolation?

sergeyshushlyapin:
That's correct. No other value. If this one is merged, there will be three other PRs similar to #695 (xUnit.net, NUnit2 and NUnit3).

Src/AutoFixture/FreezeOnMatchCustomization.cs
sergeyshushlyapin: The code from the [removed constructor](new ExactTypeSpecification(targetType)):  `new ExactTypeSpecification(targetType)`.
That might be a bit unclear. I'll extend the obsolete message.

ploeh: Wouldn't it be a better user experience if we could add a constructor overload that doesn't demand of the user that he or she supplies an explicit `IRequestSpecification`?

Src/AutoFixture/FreezeOnMatchCustomization.cs
```diff
@@ -43,8 +44,9 @@ public FreezeOnMatchCustomization(Type targetType)
         /// that will be satisfied by the frozen specimen.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        /// <paramref name="targetType"/> or<paramref name="matcher"/> is null.
+        /// <paramref name="targetType"/> or <paramref name="matcher"/> is null.
         /// </exception>
+        [Obsolete("Please use FreezeOnMatchCustomization(object, IRequestSpecification) instead.")]
```
ploeh: What would happen if, instead of deprecating this overload, we change the type of the first argument from `Type` to `object`?

sergeyshushlyapin: That may lead to the `InvalidCastException` if one passes not a `Type` instance:
```csharp
var sut = new FreezeOnMatchCustomization(new object());
// System.InvalidCastException: System.InvalidCastException : Unable to cast object
// of type 'System.Object' to type 'System.Type'.
```
Where the constructor is the following:
```csharp
    public FreezeOnMatchCustomization(object request)
            : this(request, new ExactTypeSpecification((Type)request))
```

Src/AutoFixture/FreezeOnMatchCustomization.cs
sergeyshushlyapin: I was thinking about such a scenario, but if we use the `object request` parameter, the constructor overload that does not require the explicit `IRequestSpecification` [may fail in some cases](https://github.com/AutoFixture/AutoFixture/pull/697#discussion_r75891426).

Src/AutoFixture/FreezeOnMatchCustomization.cs
ploeh: Wouldn't that also be the case if we deprecate this constructor and add `FreezeOnMatchCustomization(object, IRequestSpecification)`?

sergeyshushlyapin: Yes, it will be possible to get such exception with the [new constructor](https://github.com/sergeyshushlyapin/AutoFixture/blob/02a793050a14bc052c8320145ad9bc75280cac8f/Src/AutoFixture/FreezeOnMatchCustomization.cs#L80). That is the cost of allowing the constructor receive objects.

That is also the reason why I tried to avoid changing the public API of the `FreezeOnMatchCustomization` class introducing the [extended internal version of the customization](https://github.com/sergeyshushlyapin/AutoFixture/blob/4280168a380867bb288f5660849b1b22b491afaf/Src/AutoFixture.xUnit.net2/FreezeOnParameterMatchCustomization.cs). Unfortunately in this case we have to deal with both `TargetType` and `Request` properties [simultaneously](https://github.com/sergeyshushlyapin/AutoFixture/blob/4280168a380867bb288f5660849b1b22b491afaf/Src/AutoFixture/FreezeOnMatchCustomization.cs#L99). None of them will be obsoleted.

sergeyshushlyapin:
@ploeh, @ecampidoglio, you're still having some concerns regarding obsoleting the `TargetType`, aren't you? If that is the case, may I suggest you left it alone and introduce brand new `FreezeOnParameterMatchCustomization` alongside the existing one? Please share your thoughts. Thanks.

ploeh:
I'm not convinced that it's necessary. Isn't it enough to [widen the the current constructors?](https://github.com/AutoFixture/AutoFixture/pull/697#discussion_r75881830)

sergeyshushlyapin:
Changing the `public FreezeOnMatchCustomization(Type targetType)` constructor to `public FreezeOnMatchCustomization(object request)` won't change the fact that *only* `Type` argument is allowed. Passing any other argument will lead to the [type cast exception](https://github.com/AutoFixture/AutoFixture/pull/697#discussion_r75891426). Shouldn't we avoid such kind of implicit behavior?

ploeh:
This pull requests suggest to _add_ a constructor overload with the signature `public FreezeOnMatchCustomization(object request, IRequestSpecification matcher)`. How is that different?

ecampidoglio:
@sergeyshushlyapin
> Changing the public FreezeOnMatchCustomization(Type targetType) constructor to public FreezeOnMatchCustomization(object request) won't change the fact that only Type argument is allowed. Passing any other argument will lead to the type cast exception. Shouldn't we avoid such kind of implicit behavior?

Maybe I'm missing something here, but where is this cast to `Type` happening? It seems to me that the `request` is being [passed straight](https://github.com/AutoFixture/AutoFixture/pull/697/files#diff-ff3511dac1948f3316dc6bf6343595eaR141) to `context.Resolve(object)`.

sergeyshushlyapin:
The suggested constructor is the *only* constructor which *highlights* the necessity to think about both the parameters being passed. For instance, if we introduce the the constructor with a single `object request` parameter, the following code will look 'legal' but won't work:
```
new FreezeOnMatchCustomization(myParameterInfo);
```

sergeyshushlyapin:
@ecampidoglio the cast will have to be added if we introduce the `FreezeOnMatchCustomization(object request)` [constructor](https://github.com/sergeyshushlyapin/AutoFixture/blob/02a793050a14bc052c8320145ad9bc75280cac8f/Src/AutoFixture/FreezeOnMatchCustomization.cs#L34).

sergeyshushlyapin:
Actually, nothing prevents us from being a bit more "careful" when dealing with a single request argument and use `ExactTypeSpecification` only if possible (otherwise `TrueRequestSpecification`). Sorry, it seems I missed this part.

Zvirja:
In this case it would be better to always use the `EqualRequestSpecification` instead of switching between `ExactTypeSpecification` and `TrueRequestSpecification`.

In this case, we have two overloads:
- `FreezeOnMatchCustomization(object request)` - freeze only if request matches. No issues here. Supports any type of request, not `Type` requests only.
- `FreezeOnMatchCustomization(object request, IRequestSpecification specification)` - freeze if specification is matched. Potential `InvalidCastException` if specification result type is not compatible with request type. But the latest should be OK - it's up to caller to ensure that valid data is provided.

In this case we can just change signature for existing ctors - they are binary compatible. The only place that should be obsoleted is the `TargetType` property.

sergeyshushlyapin:
> The only place that should be obsoleted is the TargetType property.

Isn't renaming constructor parameters a [breaking change](http://stackoverflow.com/questions/1456785/a-definitive-guide-to-api-breaking-changes-in-net)? For now I preserve the old constructors.

Zvirja:
Indeed, the binary compatibility will be broken. You will need to re-compile the code. I believe that it's not a _critical_ issue, while that indeed could violate semver.

No other issues:

- Source level compatibility - no issues.
- Semantic meaning - no changes.

Note, if we decide to keep both constructor overloads (one with `Type`, other with `object` arg) , it will be a nightmare for consumers:

```csharp
Type type = typeof(string);
var specification = new TrueRequestSpecification();

new FreezeOnMatchCustomization(type); // Obsolete usage
new FreezeOnMatchCustomization((object)type); // OK

new FreezeOnMatchCustomization(type, specification); // Obsolete usage
new FreezeOnMatchCustomization((object)type, specification); // OK
```

They will need to case type to object to use correct overload - looks very annoying.

sergeyshushlyapin:
@ploeh, @ecampidoglio, from your point of view, should we decorate [old](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/FreezeOnMatchCustomization.cs#L32) [constructors](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/FreezeOnMatchCustomization.cs#L48) with the `[Obsolete]` attribute or should we just replace the `Type targetType` argument with `object request`?

ecampidoglio:
In spite of having [argued for the opposite](https://github.com/AutoFixture/AutoFixture/pull/317#discussion_r19199307) in the past, I vote for making the constructor _more liberal_ by accepting the request as `object`. Yes, I changed my mind :wink:

sergeyshushlyapin:
Clear, thanks @ecampidoglio ;)
But my question is slightly different. Should we **add** new *liberal* constructors (obsoleting the old ones) or just **change** existing (which might be considered as a breaking change)?

ecampidoglio:
Widening the constructor by accepting `object` instead of `Type` shouldn't break any existing calls so I think we should just do that.

sergeyshushlyapin:
If there are no other comment's I'd like to squash last 6 or 7 commits since they bring no value for the commit history.

Zvirja:
I'd like to add my five cents. Some of them are copied from my [previous comment](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-242701966).

1.  It's incorrect to use this kind of specification for `object` freezing requests.
    ```csharp
    if (targetType != null)
    {
        this.Matcher = new ExactTypeSpecification(targetType);
    #pragma warning disable 618
        this.TargetType = targetType;
    #pragma warning restore 618
    }
    else
    {
        this.Matcher = new TrueRequestSpecification();
    }
    ```

    This way we break `Fixture` because it will return passed to constructor object for any request. The `EqualRequestSpecification` should be used instead.

1. It's redundant to switch between `ExactTypeSpecification` and `EqualRequestSpecification` in constructor. Rather, we could always delegate work to `EqualRequestSpecification`.

1. `TargetType` property could be calculated as `Request as Type` dynamically - no need to assign it statically in constructor.

1. The following private methods should be renamed: 
    - `FreezeTypeForMatchingRequests`
    - `FreezeTargetType`

    Now they freeze any type of requests, not `Type` requests only. If we keep old name's they might be confusing for code readers in future.

sergeyshushlyapin:
> It's redundant to switch between ExactTypeSpecification and EqualRequestSpecification in constructor. Rather, we could always delegate work to EqualRequestSpecification.

Get stuck how to write a test which would show the necessity of using the `EqualRequestSpecification` type. Wouldn't like to write a [structural test](http://enterprisecraftsmanship.com/2016/07/21/unit-testing-anti-patterns-structural-inspection/) expecting `EqualRequestSpecification` (similar to [this one](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs#L65) which will fail after the suggested change BTW) since it *does not really prove* the chosen type is the one we need.

Zvirja:
Unless I'm missing something, we should just verify that specimen is frozen for different type of requests. 

Example:

1. Create `PropertyInfo` for string property. 
2. Customize fixture using value from point 1 in constructor. 
3. Resolve `PropertyInfo` value for 2 times. 
3. Assert strings are the same

In the same time, you should assert that if you resolve `PropertyInfo.Type`, you get different string for each time (because `Type` requests are not frozen). 

You can use any other request type instead of `PropertyInfo` - ensure that result for that exact request is frozen, but not for other requests. 

The mentioned structural test should be rewritten to verify that `matcher` returns `true` for value, passed to ctor, but false for any other object.

sergeyshushlyapin:
@Zvirja, I covered all the scenario as you described except the second one. Make sense to me. Thanks!
Still have some doubts if scenario 2 should be implemented since we'll have to change the *[existing test](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs#L65)* and violate the [append-only tests](http://blog.ploeh.dk/2013/04/02/why-trust-tests/) principle.

@ploeh, @ecampidoglio, any other comments regarding this pr?

ploeh:

Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
```diff
@@ -77,6 +90,63 @@ public void InitializeWithMatcherShouldSetCorrespondingProperty()
...
+        [Fact]
+        public void InitializeWithSingleRequestShouldSetCorrespondingProperty()
+        {
+            var request = new object();
+            var sut = new FreezeOnMatchCustomization(request);
+            Assert.Same(request, sut.Request);
+        }
+
+        [Fact]
+        public void InitializeWithObjectRequestShouldSetTrueRequestSpecificationMatcher()
```
zvirja: Fix test name

zvirja: The name of the test should be updated.
BTW, aren't we introducing [yet another](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-243843127) structural test which might be broken in future. Ins't it enough to test the behavior only?

Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
```diff
@@ -114,6 +184,69 @@ public void CustomizeWithCompetingSpecimenBuilderForTheSameTypeShouldReturnTheFr
...
+        }
+
+        [Fact]
+        public void CustomizeWithNotEqualRequestsShouldNotFreezeSpecimen()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var context = new SpecimenContext(fixture);
+
+            var freezingRequest = typeof(ConcreteType).GetProperty("Property1");
```
zvirja: Ins't it better to use `nameof(ConcreteType.Property1)` for such cases?

zvirja:
Let's move this one from the dead point. 

@sergeyshushlyapin I've added a couple of comments.

@moodmosaic, @ecampidoglio Do you have other concerns regarding the current implementation?

moodmosaic:
I'm sorry, I don't know enough about `FreezeOnMatchCustomization` in order to review this PR.

I've actually tried to review the PR several times the past few days, but I don't feel enough confident with the `FreezeOnMatchCustomization` and its surrounding parts on the codebase.

Maybe @ecampidoglio or @adamchester can help (?)

Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
sergeyshushlyapin: Are you talking about [Structural Inspection anti-pattern](http://enterprisecraftsmanship.com/2016/07/21/unit-testing-anti-patterns-structural-inspection/)? Yep, that is a structural test. I suppose it was added mechanically following the test-first practice.

Despite the code change _is_ covered by the other tests in [the same](https://github.com/AutoFixture/AutoFixture/pull/697/commits/4cd5ee87615aa62e698808671cee6748a863caa7) commit, I do not see any issues in keeping this 'structural' test. It's tiny and it may serve the code documentation function much better than the [`CustomizeWithNotEqualRequestsShouldNotFreezeSpecimen`](https://github.com/AutoFixture/AutoFixture/pull/697/files#diff-3b20e30bf8f7c8e5020d5f1e40f2907bR234) test which checks that the behavior is correct.

zvirja: Well, I'm fine with structural inspection as soon as we agree that it's fine to not follow the [append-only](http://blog.ploeh.dk/2013/04/02/why-trust-tests/) approach with such tests. Recently I had another PR where I [had to alter](https://github.com/AutoFixture/AutoFixture/pull/781/files#diff-68dade75f99c180341e9e21ce0f0d3be) all such tests because they are pretty volatile.

Currently we have the following block of code:
```csharp
var targetType = request as Type;
if (targetType != null)
{
    this.Matcher = new ExactTypeSpecification(targetType);
}
else
{
    this.Matcher = new EqualRequestSpecification(request);
}
```
This code is redundant and can be simplified to the `EqualRequestSpecification` only, but you [don't want](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-243843127) to simplify it because of [this](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs#L65) test which is a pure structural inspection. 

In such light I'm pretty reluctant to add this test in which you pin that specification should be necessarily of the `EqualRequestSpecification` type. If later we'll need to tune behavior even more (e.g. add some conditions), we will be limited by this test.

If we agree that tests should be append only (even structural inspection), then I'd vote to remove this test. If we agree that structural inspection tests could be altered, let's simplify the code block above and adjust [this](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs#L65) test.

@sergeyshushlyapin What do you think about this? 😉
@moodmosaic Do you have any opinion about the [structural inspection](http://enterprisecraftsmanship.com/2016/07/21/unit-testing-anti-patterns-structural-inspection/) tests? Do you agree that it's fine to change them if necessary like I did [here](https://github.com/AutoFixture/AutoFixture/pull/781/files#diff-68dade75f99c180341e9e21ce0f0d3be)?

sergeyshushlyapin: > This code is redundant and can be simplified to the EqualRequestSpecification only, but you don't want to simplify it because of this test which is a pure structural inspection.

I wouldn't say so. This PR solves the original issue and is well covered with tests which enables further refactoring. Since it's one year old PR, at this particular moment I have no idea what's going on inside because I simply do not remember ;)

I (probably) haven't dig into details so much to notice that the mentioned `if-else` statement can be simplified. If that's true and the only 'blocker' is those structural test, I'm fine with killing it ;)

zvirja: @sergeyshushlyapin Yes, if no other test fails (and it should be so), I'd say that it's better to make the code even more simple. I wouldn't consider that as a breaking change, because we expose `IRequestSpecification` rather than a concrete type, so it's fine to change the underlying implementation.

zvirja: @sergeyshushlyapin I've fixed that myself via commit to your PR branch 😉

zvirja:
@ecampidoglio We polished this PR and it looks fine to me. Also it seems that all your concerns are addressed as well. Could you please check whether you have some other concerns and if not - approve this one? Thanks.

zvirja:
I've just tested this code more and found that we cannot merge it to the `master` in the current form. The reason are the removed constructors that [we agreed](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-243363884) to delete. If we push this code to `master`, it will break all the already compiled binaries. If there is a binary compiled against e.g. `v3.25.0` that uses `FreezeOnMatchCustomization` and user updates AutoFixture to the latest version, it will get the following error during execution:
```
Unhandled Exception: System.MissingMethodException: Method not found: 'Void Ploeh.AutoFixture.FreezeOnMatchCustomization..ctor(System.Type)'.
  at FreezeCustomizationUsage.Program.Main(String[] args)
```

It could happen that there are third-party assemblies that depends on us and are published to the NuGet. All them will stop to work after user updates to the latest AutoFixture version. That is prohibited behavior according to the SemVer.

Given that keeping all 4 constructor overload will make the usage [ugly](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-242710636), I vote to merge this change to the `v4` branch instead. That change is fine for the major update.

@sergeyshushlyapin @ecampidoglio Are you fine with merging the code to `v4`? If so, I'll rebase the branch.

sergeyshushlyapin:
> Are you fine with merging the code to v4?

Nope. There are no breaking changes (if there are, they must be addressed). It should be in v3. And, BTW, that is not an _enhancement_, that is an enabler for another [bugfix](https://github.com/AutoFixture/AutoFixture/pull/697#issue-172243551).

zvirja:
> Nope. There are no breaking changes (if there are, they must be addressed).
Got it 😉 

In this case we need to revert the constructors removal, so that finally we have 4 constructors:
```csharp
public FreezeOnMatchCustomization(object request);
public FreezeOnMatchCustomization(object request, IRequestSpecification matcher)
[Obsolete]
public FreezeOnMatchCustomization(Type targetType);
[Obsolete]
public FreezeOnMatchCustomization(Type targetType, IRequestSpecification matcher)
```

This will cause [inconveniences](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-242710636), however change will not break the existing code.

Do you have time do apply that change? If not - let me know and I'll do that myself.

sergeyshushlyapin:
Oops. it seems I overlooked this part. I'm fine merging it into v4 since I do not want to start yet another round of constructor params discussion (all the conversation above is about that).

zvirja:
@sergeyshushlyapin That also means that #695 should also go to v4, as this one is a precursor for it. If you are fully fine with that - could you please rebase on top of the `v4` branch and change the merge target?

zvirja:
Closing this one in favor of #802.

PR #698, opened 2016-08-23
Improve NSubstitute

Zvirja:
Apply a few code changes to improve NSubstitute integration codebase.
Each time I have issues with NSubstitute integration (e.g. because of concurrency) I find that it's hard to understand what is going on in the integration. Therefore, I decided to simplify a few places and add comment.

Also, I resolved a couple of small issues.

If you wish, I could split into multiple PRs. I found that changes are rather tiny, so no real sense for that.

---
BTW: The current code analysis rules don't allow to use string interpolation from C# 6 . Each time it's used, code analysis fails and asks to use culture specific overload of `string.Format()`. Do we need to disable that rule? Obviously, AF doesn't worry about that - it's a dev tool. See example in warning 3 [here](https://ci.appveyor.com/project/AutoFixture/autofixture/build/1.0.317-kggqwrqe).


ploeh:
@mrinaldi, I wouldn't mind your eyes on this, if you have the time and inclination :smile:

mrinaldi:
LGTM :+1:
Thanks @Zvirja

Zvirja:
@ploeh
Could you please reply to the part, related to a string interpolation? :)

ploeh:
> Obviously, AF doesn't worry about that - it's a dev tool.

Doesn't worry about what?

Zvirja:
@ploeh Sorry, I wrote that in context of the error message from CI

>Microsoft.Globalization : Because the behavior of 'string.Format(string, object)' could vary based on the current user's locale settings, replace this call in 'SubstituteRequest.ToString()' with a call to 'string.Format(IFormatProvider, string, params object[])'. If the result of 'string.Format(IFormatProvider, string, params object[])' will be displayed to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider' parameter. Otherwise, if the result will be stored and accessed by software, such as when it is persisted to disk or to a database, specify 'CultureInfo.InvariantCulture'.

I meant that AF shouldn't worry about formatting in different cultures - it's a pure dev tool. It doesn't matter how data (e.g. numbers) are formatted - in context or independent culture.

ploeh:
> It doesn't matter how data (e.g. numbers) are formatted - in context or independent culture.

What's the default, implicit culture if none is provided?

Zvirja:
The Current culture is default. I believe that should come from regional settings of your OS.

ploeh:
True. I interpret that CA rule as making the implicit explicit. This makes the code clearer, because it's explicit which type of culture is being used.

There's [a least one example in the AutoFixture code base where the invariant culture is used](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/MailAddressGenerator.cs#L63), so I think that the ability to distinguish could be valuable.

As the [Zen of Python](https://www.python.org/dev/peps/pep-0020) states:

> Explicit is better than implicit

Zvirja:
OK. The only concern from my side was that this rule prevents us from usage of a modern syntax, available in C#. However, if you believe that this way we could prevent some potential issues and inconveniences worth it - let's keep the rule :)

If you don't have any objections regarding the code changes - let's finally merge the PR.

ploeh:
I don't think I have much opinion either way, to be honest. I can see value in both.

I wonder what @moodmosaic, @adamchester, or @ecampidoglio thinks..?

ploeh:
BTW, that you for your contribution :+1:

adamchester:
I agree that explicit is better than implicit in this case, because otherwise the rules are not obvious.

We can use string interpolation with invariant (or other) cultures like this:
http://stackoverflow.com/a/29338846/672329

Having said that, I'm ok with not specifying the culture (using `current` implicitly) if the vast majority of string interpolations are for display to the user and the current culture is nearly always the most appropriate, it feels like a developer in that code base should know the rules. There could be a lot of code noise when always specifying invariant everywhere.

moodmosaic:
I usually keep these 2 rules in mind:

* Use [CultureInfo.CurrentCulture](https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.currentculture.aspx) when displaying output to the user
*  Use [CultureInfo.InvariantCulture](https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.invariantculture.aspx) when persisting linguistically meaningful but culturally agnostic data

The above are merely taken from [here](https://msdn.microsoft.com/en-us/library/dd465121.aspx#recommendations_for_string_usage), as I can't currently remember if I read those in either [CLR via C#](https://www.amazon.com/dp/0735627045/) or [Framework Design Guidelines](https://www.amazon.com/dp/0321545613).

Zvirja:
@moodmosaic Thanks for the input. Could you please also answer to the following question? Do you thing it would be fine to disable code analysis rule that enforces to always explicitly specify culture or you believe that we need it?

moodmosaic:
@Zvirja It's up to the user to specify the appropriate `CultureInfo` instance, so there isn't much benefit from that CA rule really...

Zvirja:
@ploeh Could you please trigger a new release code change this PR introduces? I need the bug fix I did here.

Also, it seems that guys above voted to disable that CA rule :)

Thank you.

ploeh:
Is there a bug fix in this PR? What is it?

ploeh:
> it seems that guys above voted to disable that CA rule

I count one vote for and one vote neither for nor against. Hardly a resolved issue, I think.

ploeh:
No need to create an issue, but I could use some help with what to write in the release commit :flushed:

Zvirja:
I'd say something like "Fix NSubstitute integration could fail if test is executed in context of custom TaskScheduler".

However, I don't know whether it looks OK for release message.

ploeh:
Thank you for your contribution. It's now live as [AutoFixture.AutoNSubstitute 3.50.2](https://www.nuget.org/packages/AutoFixture.AutoNSubstitute/3.50.2).

PR #702, opened 2016-09-04
Support fake classes (not just interfaces) when using FakeItEasy 2.0

### LINKED ISSUE #693: Dependency against old FakeItEasy version?
I'm using:
- AutoFixture 3.50.1
- AutoFixture.AutoFakeItEasy 3.50.1
- AutoFixture.Xunit2 3.50.1
- FakeItEasy 2.2.0
- Xunit 2.0.0

I'm writing tests that extend Xunit's AutoData attribute to add the AutoFakeItEasy customization, and then use abstract class types in the test method parameters. The expected behaviour is that FakeItEasy will handle the creation of the abstract type test doubles, however instead I'm getting the following exception:

> Could not load type 'FakeItEasy.Creation.IFakeOptionsBuilder`1' from assembly 'FakeItEasy, Version=2.2.0.0, Culture=neutral, PublicKeyToken=eff28e2146d5fd2c'.

Stack trace:

>    at Ploeh.AutoFixture.AutoFakeItEasy.FakeItEasyMethodQuery.FakeMethod`1.Invoke(IEnumerable`1 parameters)
>    at Ploeh.AutoFixture.Kernel.MethodInvoker.Create(Object request, ISpecimenContext context)

Here's some test code to replicate this exception:

``` c#
using Ploeh.AutoFixture;
using Ploeh.AutoFixture.AutoFakeItEasy;
using Ploeh.AutoFixture.Xunit2;
using Xunit;

namespace Tests
{
    public abstract class AbstractClass
    {
    }

    public class AutoFixtureBugTest
    {
        public class ArrangeAttribute : AutoDataAttribute
        {
            public ArrangeAttribute() : base(new Fixture().Customize(new AutoFakeItEasyCustomization())) { }
        }

        [Theory, Arrange]
        public void TestMethod(AbstractClass test)
        {
        }
    }
}
```

Looking through FakeItEasy's github repo, I noticed that there hasn't been a FakeItEasy.Creation.IFakeOptionsBuilder since version 1.25.3; the interface appears to have been renamed to IFakeOptions since version 2.0.0. 

So this leads me to wonder, was the AutoFixture 3.50.1 nuget package built against a pre-2.0.0 version of FakeItEasy?

blairconrad:
Fixes #693.

Resorting to finding the `Fake<T>` constructor by reflection and invoking that one, so we don't rely on it having the same interface as in FakeItEasy 1.0.

The first step was to add a test project for AutoFakeItEasy that targeted FakeItEasy 2.0. I opted to use FakeItEasy 2.2.0 with the tests because that's the version that #693 was reported against, and in this case it seemed that testing for a newer version was a good choice, in contrast to AutoFixture's policy of having the framework depend on the lowest possible version of a library.

Of course if you hate this whole approach to testing FakeItEasy 2.0 integration, or the library version I chose, or anything, I'm happy to make any changes.

The actual fix seems to be functional, but is not optimized. If you like, I could attempt to do so, by caching found constructors, creating  delegates to invoke, or something similar.
Of course, such optimizations could be introduced in later issues/PRs as well…

### DIFF
--- a/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
@@ -74,8 +74,46 @@ namespace Ploeh.AutoFixture.AutoFakeItEasy
 
             public object Invoke(IEnumerable<object> parameters)
             {
-                return new Fake<T>(
-                    b => b.WithArgumentsForConstructor(parameters));
+                var genericFakeType = typeof (Fake<>).MakeGenericType(typeof (T));
+
+                foreach (var constructor in genericFakeType.GetConstructors())
+                {
+                    var constructorParameterInfos = constructor.GetParameters();
+                    if (constructorParameterInfos.Length != 1)
+                    {
+                        continue;
+                    }
+
+                    var parameterType = constructorParameterInfos[0].ParameterType;
+                    if (!parameterType.IsGenericType ||
+                        parameterType.GetGenericTypeDefinition() != typeof (Action<>))
+                    {
+                        continue;
+                    }
+
+                    // The parameter is an action of type
+                    // Action<IFakeOptionsBuilder<T>> (FakeItEasy 1.x) or
+                    // Action<IFakeOptions<T>> (FakeItEasy 2.0+).
+                    // Each of the options-type interfaces contains a WithArgumentsForConstructor method
+                    // that we'll use to pass arguments to the fake object's constructor.
+                    var fakeOptionsType = parameterType.GetGenericArguments()[0];
+
+                    var withArgumentsForConstructorMethod = fakeOptionsType.GetMethod(
+                        "WithArgumentsForConstructor",
+                        new[] {typeof (object[])});
+
+                    if (withArgumentsForConstructorMethod == null)
+                    {
+                        continue;
+                    }
+
+                    Action<object> addConstructorArgumentsToOptionsAction =
+                        options => withArgumentsForConstructorMethod.Invoke(options, new object[] {parameters});
+
+                    return constructor.Invoke(new object[] {addConstructorArgumentsToOptionsAction});
+                }
+
+                return null;
             }
         }
     }
--- a/Src/All.sln
+++ b/Src/All.sln
@@ -67,6 +67,8 @@ Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "Idioms.FsCheckUnitTest", "I
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "IdiomsUnitTest", "IdiomsUnitTest\IdiomsUnitTest.csproj", "{730E1D38-BA80-48C7-B05C-E2BD29F38851}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFakeItEasy.FakeItEasy2UnitTest", "AutoFakeItEasy.FakeItEasy2UnitTest\AutoFakeItEasy.FakeItEasy2UnitTest.csproj", "{7F957CC0-79C3-48A1-9A3E-47BED7539248}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -266,6 +268,12 @@ Global
 		{730E1D38-BA80-48C7-B05C-E2BD29F38851}.Release|Any CPU.Build.0 = Release|Any CPU
 		{730E1D38-BA80-48C7-B05C-E2BD29F38851}.Verify|Any CPU.ActiveCfg = Release|Any CPU
 		{730E1D38-BA80-48C7-B05C-E2BD29F38851}.Verify|Any CPU.Build.0 = Release|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Release|Any CPU.Build.0 = Release|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Verify|Any CPU.ActiveCfg = Release|Any CPU
+		{7F957CC0-79C3-48A1-9A3E-47BED7539248}.Verify|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
@@ -1,0 +1,103 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProductVersion>8.0.30703</ProductVersion>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>{7F957CC0-79C3-48A1-9A3E-47BED7539248}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Ploeh.AutoFixture.AutoFakeItEasy.UnitTest</RootNamespace>
+    <AssemblyName>Ploeh.AutoFixture.AutoFakeItEasy.FakeItEasy2UnitTest</AssemblyName>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <SignAssembly>true</SignAssembly>
+    <TargetFrameworkProfile />
+    <CodeAnalysisCulture>en-US</CodeAnalysisCulture>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Castle.Core">
+      <HintPath>..\..\Packages\Castle.Core.2.5.2\lib\NET35\Castle.Core.dll</HintPath>
+    </Reference>
+    <Reference Include="FakeItEasy, Version=2.0.0.0, Culture=neutral, PublicKeyToken=eff28e2146d5fd2c, processorArchitecture=MSIL">
+      <HintPath>..\..\Packages\FakeItEasy.2.0.0\lib\net40\FakeItEasy.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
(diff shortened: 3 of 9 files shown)


Src/AutoFakeItEasy.FakeItEasy2.UnitTest/AutoFakeItEasy.FakeItEasy2.UnitTest.csproj
```diff
@@ -0,0 +1,103 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
moodmosaic: Could we do without having to add (and maintain) and additional test project?

Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
```diff
@@ -74,8 +74,46 @@ public IEnumerable<ParameterInfo> Parameters
 
             public object Invoke(IEnumerable<object> parameters)
             {
-                return new Fake<T>(
-                    b => b.WithArgumentsForConstructor(parameters));
+                var genericFakeType = typeof (Fake<>).MakeGenericType(typeof (T));
```
moodmosaic: Would it be possible to cover these changes (lines 77 to 116) with a regression test?

moodmosaic:
Thank you for taking the time to investigate on #693 and opening this PR :+1: 

I had a quick look at this, and while it looks promising, I think it comes with some maintenance overhead. 

I'm a little bit concerned with the fact that we now have to take *two* versions of FakeItEasy 2 in mind, and deal with a new Unit Test project.

Wouldn't reinstating AutoFixture.AutoFakeItEasy2 and constraining AutoFixture.AutoFakeItEasy be a more straightforward solution?

Src/AutoFakeItEasy.FakeItEasy2.UnitTest/AutoFakeItEasy.FakeItEasy2.UnitTest.csproj
blairconrad: If we want to ensure that AutoFakeItEasy works with FakeItEasy 1.x and 2.y, then I'm not sure how.
I suppose there could be some test runner trickery where the same project is run twice, by swapping in FakeItEasy 2.x for the second test run (and adding the redirect). That was not something I'd felt up to proposing, as I'm unsure of the mechanism and am not confident that I could introduce such a thing into the build system. However, if it appeals, I'd try.

I doubt there'd be much maintenance for this test project. We could replace the `Compile Include`s with a more sophisticated mechanism [as used in xBehave](https://github.com/xbehave/xbehave.net/blob/ed11fcdda746edb9a111ac1b109d4d37f13a7c2a/tests/Xbehave.Test.Acceptance.Net45/Xbehave.Test.Acceptance.Net45.csproj#L80-L84), if you think it's likely that new files will be added to the original test project and you worry about them being forgotten here.

Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
blairconrad: I'm afraid I don't quite follow.
The existing AutoFakeItEasyUnitTest project covers this code (as does the new AutoFakeItEasy.FakeItEasy2.UnitTest project that you were hoping to avoid). This code was written specifically to cause those two test suites to pass.

Granted the tests verify what happens when there's no constructor with length other than 1, or where the parameter isn't an action, or where the action doesn't act on a type that has a `WithArgumentsForConstructor`, but all FakeItEasys after 1.7.4257.42 meet the requirements, and the tests serve to ensure that any FakeItEasy dropped into the tests continues to do so. The same statements could be made about the original tests and code as well.

blairconrad:
Thanks for your comments, @moodmosaic. I hear your concerns and hope I've done something to allay them.

> I'm a little bit concerned with the fact that we now have to take two versions of FakeItEasy 2

In some sense, we already (without considering my PR) have to take 3 versions of FakeItEasy 2 in mind, because 2.0, 2.1, and 2.2 have been released. And who knows when 2.3 will be coming out? 
Just because the AutoFakeItEasy tests import one version of FakeItEasy doesn't mean the project is only concerned with one version of FakeItEasy.

If the two imported versions of FakeItEasy make you uneasy, the new test project could easily depend on FakeItEasy 2.0. Or the AutoFakeItEasy2.UnitTest project could switch to 2.2.
(An aside: I understand @ploeh's attitute toward having AutoFixture rely on the lowest possible version of FakeItEasy (and other libraries), but I've been a little thinking that it makes sense for the _tests_ to rely on the newest (or at least newer) versions, to ensure that the libraries stay compatible with shiny new releases. But I wasn't really trying to start down that road today. Ideally, we'd test against all possible releases, but that's just not practical.)

> and deal with a new Unit Test project.

I understand that, but one that should be very low maintenance, and we do discuss in the [line-by-line comments](https://github.com/AutoFixture/AutoFixture/pull/702#discussion_r77485551) a way to potentially reduce the maintenance further, or even remove the new test project.
For my money (and it's not my project, so my money isn't worth much here), to reduce the complexity of the solution, I'd drop AutoFakeItEasy2 and its test project. Then you end up with one _fewer_ test project than today. And much less duplicated code, even with the new AutoFakeItEasy.FakeItEasy2.UnitTest project I'm trying to introduce.

> Wouldn't reinstating AutoFixture.AutoFakeItEasy2 and constraining AutoFixture.AutoFakeItEasy be a more straightforward solution?

Straightforward for the maintainers? Probably. With the overhead that there are two code bases, one dealing with FakeItEasy 1.x and one with FakeItEasy 2.x (although by maintaining the dormant AutoFakeItEasy2, you sort of have that already).

For the clients? I don't think so. I was working toward the solution I propose because there were a few users who were exposed to AutoFakeItEasy2 to begin with and then had it ripped out from under them (a tough decision for the AutoFixture maintainers, and one I supported), and now reinstating it would in effect rip the replacement (AutoFakeItEasy) out from under them as well. As well as making all those users who are already using AutoFakeItEasy with FakeItEasy 2.x switch packages.

Still, this isn't my project. It's all well and good for me to surmise about what would be the best experience for your users, and to comment on maintenance issues and how many versions of libraries should be imported, but ultimately it's not my responsibility and you've no guarantee that I'd be around to help out with any problems that arise (although I would be). If there's anything I can do to make this PR (which I believe in) more palatable, I will happily do so. If you decide to go another way, such as reinstating FakeItEasy2, I understand. And will submit a PR, if you don't want to handle it internally.

moodmosaic:
@blairconrad, thank you for your comment :+1: 

>In some sense, we already (without considering my PR) have to take 3 versions of FakeItEasy 2 in mind, because 2.0, 2.1, and 2.2 have been released. And who knows when 2.3 will be coming out?

>[..] thinking that it makes sense for the *tests* to rely on the newest (or at least newer) versions

AFAICT, if FakeItEasy 2.x uses Semantic Versioning, then we shouldn't have to worry about taking each and every release in mind (?)

blairconrad:
> AFAICT, if FakeItEasy 2.x uses Semantic Versioning, then we shouldn't have to worry about taking each and every release in mind (?)

It's true. If that's the main barrier to taking the PR, I'm happy to change AutoFakeItEasy.FakeItEasy2.UnitTest to rely on FakeItEasy 2.0.

Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
moodmosaic: I see... So, AFAICT, #693 is caused because [a public interface is now renamed in FakeItEasy 2.0.x](https://github.com/FakeItEasy/FakeItEasy/issues/520).

If that's the case, then it looks like AutoFixture.AutoFakeItEasy is not truly compatible with FakeItEasy 2.0.x. 

Instead of trying to 'hack' AutoFixture.AutoFakeItEasy, wouldn't it be easier if we apply the change in AutoFixture.AutoFakeItEasy2?

blairconrad: No change would be required to AutoFixture.AutoFakeItEasy2. It works out of the box.
I was suggesting the change to AutoFixture.AutoFakeItEasy so as to spare the users from "nuget package whiplash".

> wouldn't it be easier

I suppose it depends on your focus. You have a working PR in your hands, so the initial development is done, without the need to explain to users that they should go to AutoFakeItEasy2. That being said, the reflection-based approach I've committed is reflection-based, and therefore somewhat less desirable. I understand your concerns about it. Personally, I'd take the PR (easy for me to say; I wrote it) in order to provide the users with a seamless experience.
I don't think there will ever be a need to revisit the method in question during FakeItEasy 2.x's run (and probably not for a while beyond that), so maintenance should be nil.

Of course that same argument applies to the duplicated code if you decide to go with reinstating AutoFakeItEasy2. Me, I think the internal costs and benefits are so close that I'd let the decision come down to what's going to work for the users. But I know that it's you that has to live with whatever decision you make.

ploeh:

moodmosaic:
I've pulled the latest from this branch, and now the newly added test project references FakeItEasy 2.0.x :+1: 

image

@blairconrad, one small nitpick – we could probably rename the newly added test project to something like *AutoFakeItEasyUnitTest.FakeItEasy2*, to be consistent with the existing *AutoFakeItEasyUnitTest* project. @ploeh, after that, we should be good to go!

blairconrad:
@moodmosaic Sorry, yes. I updated the referenced version right after https://github.com/AutoFixture/AutoFixture/pull/702#issuecomment-244905562. I should've said.

I'm happy to rename the test project to AutoFakeItEasyUnitTest.FakeItEasy2.
I'd just chosen the one I did because It looked to me like the build.fsx was finding test projects by virtue of them ending in `Test`, and the name I came up with was the best I could think of to conform.
I'll be explicit when I've pushed the new project name this time!

moodmosaic:
>I'd just chosen the one I did because It looked to me like the build.fsx was finding test projects by virtue of them ending in `Test`

Fair enough :+1: :smiley:

blairconrad:
> It looked to me like the build.fsx was finding test projects by virtue of them ending in `Test`

Confirmed. If I rename, the tests are no longer found.

I can rename differently (although I'll lean on you for the new name, as I've already used my best version), or attempt to update the `build.fsx`, if that's your preference.

If you want a change to the `build.fsx`, would you like an explicit inclusion for the one project, or an alteration of the include pattern?
I warn you that I've not yet found a good general purpose pattern, as some of the currently included tests are not `*UnitTest`, and there are DLLs that we don't want to include that contain `Test`.

blairconrad:
> I can rename differently (although I'll lean on you for the new name, as I've already used my best version), or attempt to update the `build.fsx`, if that's your preference.

Should've been

1. rename differently or
2. rename as you suggest and update `build.fsx`

My original comment made it sound like I was advocating updating `build.fsx` _without_ renaming the project. Which is silly.

moodmosaic:
I'd rather leave `build.fsx` as-is :smiley:

The more precise, and as the same time *consistent* name I can think of, is

*AutoFakeItEasy.FakeItEasy2UnitTest*

which basically is the same name with the one you have just with one `.` less :sweat_smile: – But for sure, it won't hurt much if we also leave it as-is.

blairconrad:
@moodmosaic If you want one less '.', then one less '.' you shall have. Probably within 12 hours.

blairconrad:
@moodmosaic, pushed!

blairconrad:
@moodmosaic, if everything's in a good state or you're just busy, please disregard, but there's been no movement here for about 4 weeks and I was wondering if I'd missed a cue and you were waiting on some action from me.

moodmosaic:
@blairconrad, this Pull Request looks good to me. See #703 on why this might not being merged just yet.

blairconrad:
Great, @moodmosaic. Sorry to be a bother. I just worried that I was the holdup.
If there's anything I can do, short of joining the governing council, don't hesitate to ask.

zvirja:
@blairconrad We have established the governance model, so we can proceed. I've reviewed your PR and I'm fine with too. However, I'd like to ask you about a minor change before we merge it.

Recently [we created](https://github.com/AutoFixture/AutoFixture/pull/756) the `All.sln` file that contains all the projects. Could you please rebase your branch on top of the fresh `master` and add the `AutoFakeItEasy.FakeItEasy2UnitTest.csproj` to the `All.sln` solution in addition to the `AutoFakeItEasy.sln`? This way we'll ensure that test is always build and invoked via our `build.fsx`.

Let me know if you don't have time for that, so I'll create a PR for your PR 😉

blairconrad:
@zvirja, I'm happy to make the changes you describe, but am unavailable for
about a week. If you can wait that long, I'll amend the PR when I return
home. If not, I won't be offended if you take what I have here and build on
it.

zvirja:
@blairconrad This PR waited for more than half of a year, so I believe it's fine if it waits for one more week :wink:

blairconrad:
@zvirja, I added the change to `All.sln`, rebased and squashed (because I figured the change was small enough that you could handle the difference). Thanks!

zvirja:
@blairconrad Thanks a lot! 👍

@moodmosaic Given that you and me are fine with this PR, is that enough to merge this PR or we need either @ecampidoglio or @adamchester approval as well?

moodmosaic:
@zvirja, given that I originally wrote AutoFixture.AutoFakeItEasy, and since this pull request has been discussed extensively, and looks good, I think that yes, we could just merge it.

zvirja:
Great!

One last question to @blairconrad and @moodmosaic - I noticed that previously we had a simple method invocation, while currently we have a set of plays with reflection. This reflection handles stuff which is bound to the assembly version and doesn't change in runtime. Do we need to introduce a caching here to not resolve constructor over and over again (e.g. cache a `Func<IEnumerable<object>, object>` factory in a static field)? My concern is that new approach is slower than the previous one and whether that will not cause significant degradation. On other hand, AF is full of reflection and this change might be not even noticeable... :)

moodmosaic:
Perhaps this is a good point @zvirja, but I think [we may not worry about this yet](http://wiki.c2.com/?PrematureOptimization).

zvirja:
@moodmosaic Thanks 😉 Wish one day I'll reach out Zen like you and will not think about optimizing everything before we had real profiling results 😆

I'm going to trigger a new release with the minor increased - it looks like a bug fix. Do you agree with minor?

moodmosaic:
I think it's a minor one 😉

blairconrad:
Thanks for the merge, folks. And FWIW, I agree with @moodmosaic—the optimization may very well not be worth it. That being said, the offer in [my initial comment](https://github.com/AutoFixture/AutoFixture/pull/702#issue-174949675) still stands. If ever you decide you want caching, I'm happy to add it. (Also happy to consult on any FakeItEasy matters in the future, if you need it.)

Issue #703, opened 2016-09-05
New governance model


ploeh:
AutoFixture needs a new governance model. From it's humble [beginnings in 2009](http://blog.ploeh.dk/2009/03/22/AnnouncingAutoFixture), I've been using the [benevolent dictator governance model](http://www.outercurve.org/resources/governance/benevolent_dictator.html), and for many years, this has worked well (in my opinion).

It's always been my responsibility to steer AutoFixture in the direction of my vision. For the last couple of years, however, I've lost the vision for AutoFixture. In my opinion, it's _done_. I don't need it to do more than it already does.

This is exacerbated by the fact that I rarely write C# code anymore. While I could use AutoFixture in F#, I prefer [FsCheck](https://fscheck.github.io/FsCheck) instead.

Running the AutoFixture project has always been a labour of love for me. Increasingly, however, I find that I've lost the passion. Recent contributors have probably felt this, and for that, I apologise.

It's time to try something else in order to enable the project to move forward. I can think of at least three options, but I'm open to more:
- Continue with the _benevolent dictator_ governance model, but with a new leader. This person must be someone willing to take on the responsibility associated with that role, as well as be someone the AutoFixture community trusts with that role. This could be one of the current core contributors to AutoFixture (@adamchester, @ecampidoglio, @moodmosaic), but could also be someone coming in from the outside.
- Establish a new governance model. It'll be up to the people taking over to decide which one.
- Establish a sponsorship. While I've lost the passion for AutoFixture, I still _care_ about it, and would be willing to continue if compensated. Managing AutoFixture has never been a full-time job, so I don't expect a big sum; frankly, I don't expect this option to be realistic, but if I don't ask, the answer is automatically _no_ :wink: 

Obviously, I'll be happy to assist handing over the responsibilities of the project to someone else. You may, for example, wonder how to build, sign, and release new NuGet packages. It's easy, but you'll need the signing key.

It's been great leading AutoFixture for all those years. It's a project that I originally created and nurtured. It's difficult to let go :cry: 

My hope is that this is like a child growing up, now becoming mature enough to move on without his or her parents.

moodmosaic:
I would lean towards the third option; *Establish a sponsorship*, hoping to keep AutoFixture at the same *fine* quality it currently is.

Mykezero:
@ploeh Thank you for all the hard work and dedication you've put into this project! Of course, I would always be willing to sponsor you by buying you those cups of coffee ^^

[...]

PR #709, opened 2016-10-04
NUnit3: change the way test name is created so that it doesn't change between discovery and execution

frblondin:
There are some problems when using the test explorer for running tests that use the AutoData / InlineAutoData attributes. When the parameter values are converted as string for generating the test name, the name contains the value that will change each time AutoFixture is requested to produce a value. In Visual Studio there are two distinct steps: discovery and execution. Because of the test name that changes constantly the test framework will fail at running/debuggin the test. Here are the steps to reproduce:
1. Create an empty C# project and add the Autofixture.NUnit3 and NUnit3TestAdapter packages
2. Declare one simple method: `[Test, AutoData]
   public void SomeMethod(string unfixedArgumentProducedByTheGreatAutofixture) { }`
3. Go to the Test Explorer, it displays one test SomeMethod("<some value that constantly change from one discovery / execution to another>")
4. Try to Run/Debug -> doesn't work

This can be reproduced with any primitive type, string, or types/structs whose ToString() method shows the property values that have been created by AutoFixture (Version, Guid, StringBuider...).

Even if there was no such error, I don't think there is real value to display values that constantly change. The only benefit would be _if_ a test randomly fails because of a particular value but that's very unlikely in general. This PR make it so that:
- AutoData: the test name and full test name will display respectively **SomeMethod(auto<string>)** and **NameSpace.SomeMethod(auto<string>)**
- InlineAutoData: the test name and full test name will display respectively **SomeMethod("fixed value 1", "fixed value 2"..., auto<string>, auto<MyClass>)** and **NameSpace.SomeMethod("fixed value 1", "fixed value 2"..., auto<string>, auto<MyClass>)**

### DIFF
--- a/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesFixture.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesFixture.cs
@@ -1,0 +1,68 @@
+﻿namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    public class TestNameStrategiesFixture
+    {
+        private static IFixture CreateFixtureWithInjectedValues()
+        {
+            var result = new Fixture();
+            // Make so that fixed values will be returned
+            result.Inject(42);
+            result.Inject("foo");
+            return result;
+        }
+
+        public class AutoDataFixedNameAttribute : AutoDataAttribute
+        {
+            public AutoDataFixedNameAttribute()
+            {
+                TestMethodBuilder = new FixedNameTestMethodBuilder();
+            }
+        }
+
+        public class AutoDataVolatileNameAttribute : AutoDataAttribute
+        {
+            public AutoDataVolatileNameAttribute() : base(CreateFixtureWithInjectedValues())
+            {
+                TestMethodBuilder = new VolatileNameTestMethodBuilder();
+            }
+        }
+
+        public class InlineAutoDataFixedNameAttribute : InlineAutoDataAttribute
+        {
+            public InlineAutoDataFixedNameAttribute(params object[] arguments) 
+                : base(arguments)
+            {
+                TestMethodBuilder = new FixedNameTestMethodBuilder();
+            }
+        }
+
+        public class InlineAutoDataVolatileNameAttribute : InlineAutoDataAttribute
+        {
+            public InlineAutoDataVolatileNameAttribute(params object[] arguments) 
+                : base(CreateFixtureWithInjectedValues(), arguments)
+            {
+                TestMethodBuilder = new VolatileNameTestMethodBuilder();
+            }
+        }
+
+        [AutoDataFixedName]
+        public void FixedNameDecoratedMethod(int expectedNumber, MyClass sut)
+        {
+        }
+
+        [AutoDataVolatileName]
+        public void VolatileNameDecoratedMethod(int expectedNumber, string value)
+        {
+        }
+
+        [InlineAutoDataFixedName("alpha", "beta")]
+        public void InlineFixedNameDecoratedMethod(string p1, string p2, string p3)
+        {
+        }
+
+        [InlineAutoDataVolatileNameAttribute("alpha", "beta")]
+        public void InlineVolatileNameDecoratedMethod(string p1, string p2, string p3)
+        {
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
@@ -1,0 +1,139 @@
+﻿using NUnit.Framework;
+using NUnit.Framework.Internal;
+using System;
+using System.Linq;
+using static Ploeh.AutoFixture.NUnit3.UnitTest.TestNameStrategiesFixture;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    public class TestNameStrategiesTest
+    {
+        [Test]
+        public void AutoDataAttributeUsesRealValueByDefault()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new AutoDataAttribute();
+            // Verify outcome
+            Assert.That(sut.TestMethodBuilder,
+                Is.TypeOf<VolatileNameTestMethodBuilder>());
+            // Teardown
+        }
+
+        [Test]
+        public void InlineAutoDataAttributeUsesRealValueByDefault()
+        {
+            // Fixture setup
+            // Exercise system
+            var sut = new InlineAutoDataAttribute();
+            // Verify outcome
+            Assert.That(sut.TestMethodBuilder,
+                Is.TypeOf<VolatileNameTestMethodBuilder>());
+            // Teardown
+        }
+
+        [Test]
+        public void AutoDataUsesFixedValuesForTestName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<AutoDataFixedNameAttribute>(nameof(TestNameStrategiesFixture.FixedNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(@"FixedNameDecoratedMethod(auto<Int32>,auto<MyClass>)"));
+            // Teardown
+        }
+
+        [Test]
+        public void AutoDataFixedNameUsesFixedValuesForTestFullName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<AutoDataFixedNameAttribute>(nameof(TestNameStrategiesFixture.FixedNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.FullName,
+                Is.EqualTo(@"Ploeh.AutoFixture.NUnit3.UnitTest.TestNameStrategiesFixture.FixedNameDecoratedMethod(auto<Int32>,auto<MyClass>)"));
+            // Teardown
+        }
+
+        [Test]
+        public void AutoDataVolatileNameUsesRealValuesForTestName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<AutoDataVolatileNameAttribute>(nameof(TestNameStrategiesFixture.VolatileNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(@"VolatileNameDecoratedMethod(42,""foo"")"));
+            // Teardown
+        }
+
+        [Test]
+        public void AutoDataVolatileNameUsesRealValuesForTestFullName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<AutoDataVolatileNameAttribute>(nameof(TestNameStrategiesFixture.VolatileNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.FullName,
+                Is.EqualTo(@"Ploeh.AutoFixture.NUnit3.UnitTest.TestNameStrategiesFixture.VolatileNameDecoratedMethod(42,""foo"")"));
+            // Teardown
+        }
+
+        [Test]
+        public void InlineAutoDataUsesFixedValuesForTestName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataFixedNameAttribute>(nameof(TestNameStrategiesFixture.InlineFixedNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(@"InlineFixedNameDecoratedMethod(""alpha"",""beta"",auto<String>)"));
+            // Teardown
+        }
+
+        [Test]
+        public void InlineAutoDataFixedNameUsesFixedValuesForTestFullName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataFixedNameAttribute>(nameof(TestNameStrategiesFixture.InlineFixedNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.FullName,
+                Is.EqualTo(@"Ploeh.AutoFixture.NUnit3.UnitTest.TestNameStrategiesFixture.InlineFixedNameDecoratedMethod(""alpha"",""beta"",auto<String>)"));
+            // Teardown
+        }
+
+        [Test]
+        public void InlineAutoDataVolatileNameUsesRealValuesForTestName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataVolatileNameAttribute>(nameof(TestNameStrategiesFixture.InlineVolatileNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(@"InlineVolatileNameDecoratedMethod(""alpha"",""beta"",""foo"")"));
+            // Teardown
+        }
+
+        [Test]
+        public void InlineAutoDataVolatileNameUsesRealValuesForFullName()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataVolatileNameAttribute>(nameof(TestNameStrategiesFixture.InlineVolatileNameDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.FullName,
+                Is.EqualTo(@"Ploeh.AutoFixture.NUnit3.UnitTest.TestNameStrategiesFixture.InlineVolatileNameDecoratedMethod(""alpha"",""beta"",""foo"")"));
+            // Teardown
+        }
+
+        private static TestMethod GetTestMethod<TAttribute>(string testName) where TAttribute : Attribute, NUnit.Framework.Interfaces.ITestBuilder
+        {
+            var method = new MethodWrapper(typeof(TestNameStrategiesFixture), testName);
+            var inlineAttribute = (TAttribute)Attribute.GetCustomAttribute(method.MethodInfo, typeof(TAttribute));
+            var testMethod = inlineAttribute.BuildFrom(method, null).Single();
+            return testMethod;
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
+++ b/Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Linq;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    /// <summary>
+    /// Builder that generates <see cref="TestMethod"/> with fixed names.
+    /// This might be needed by some test runners to correctly identify tests between the discovery and execution
+    /// (e.g. Nunit test adaptor for Visual Studio, NCrunch).
+    /// </summary>
+    public class FixedNameTestMethodBuilder : ITestMethodBuilder
+    {
+        /// <inheritdoc />
+        public virtual TestMethod Build(IMethodInfo method, Test suite, Func<object[]> argsFactory, int autoDataStartIndex)
+        {
+            if (method == null)
+            {
+                throw new ArgumentNullException(nameof(method));
+            }
+
+            if (argsFactory == null)
+            {
+                throw new ArgumentNullException(nameof(argsFactory));
+            }
+
+            return new NUnitTestCaseBuilder().BuildTestMethod(method, suite, GetParametersForMethod(method, argsFactory, autoDataStartIndex));
+        }
+
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes",
+            Justification = "This method is always expected to return an instance of the TestCaseParameters class.")]
+        private static TestCaseParameters GetParametersForMethod(IMethodInfo method, Func<object[]> argFactory, int autoDataStartIndex)
+        {
+            try
+            {
+                var parameterValues = argFactory();
+                return GetParametersForMethod(method, parameterValues.ToArray(), autoDataStartIndex);
+            }
+            catch (Exception ex)
+            {
+                return new TestCaseParameters(ex);
+            }
+        }
+
+        private static TestCaseParameters GetParametersForMethod(IMethodInfo method, object[] args, int autoDataStartIndex)
+        {
+            var result = new TestCaseParameters(args);
+
+            EnsureOriginalArgumentsArrayIsNotShared(result);
+
+            var methodParameters = method.GetParameters();
+            for (int i = autoDataStartIndex; i < result.OriginalArguments.Length; i++)
+            {
+                result.OriginalArguments[i] = new TypeNameRenderer(methodParameters[i].ParameterType);
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// Before NUnit 3.5 the Arguments and OriginalArguments properties are referencing the same array, so
+        /// we cannot safely update the OriginalArguments without touching the Arguments value.
+        /// This method fixes that by making the OriginalArguments array a standalone copy.
+        /// <para>
+        /// When running in NUnit3.5 and later the method is supposed to do nothing.
+        /// </para>
+        /// </summary>
+        private static void EnsureOriginalArgumentsArrayIsNotShared(TestCaseParameters parameters)
+        {
+            if (ReferenceEquals(parameters.Arguments, parameters.OriginalArguments))
+            {
+                var clonedArguments = new object[parameters.OriginalArguments.Length];
+                Array.Copy(parameters.OriginalArguments, clonedArguments, parameters.OriginalArguments.Length);
+                
+                // Unfortunately the property has a private setter, so can be updated via reflection only.
+                // Should use the type where the property is declared as otherwise the private setter is not available.
+                var property = typeof(TestParameters).GetProperty(nameof(TestCaseParameters.OriginalArguments));
+                property.SetValue(parameters, clonedArguments, null);
+            }
+        }
+
+        private class TypeNameRenderer
+        {
+            public Type Type { get; }
+
+            public TypeNameRenderer(Type type)
+            {
+                this.Type = type ?? throw new ArgumentNullException(nameof(type));
+            }
+
+            public override string ToString()
+            {
+                return "auto<" + this.Type.Name + ">";
+            }
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3.UnitTest/AutoFixture.NUnit3.UnitTest.csproj
+++ b/Src/AutoFixture.NUnit3.UnitTest/AutoFixture.NUnit3.UnitTest.csproj
@@ -62,8 +62,12 @@
     <Compile Include="NoAutoPropertiesAttributeTest.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Scenario.cs" />
+    <Compile Include="TestNameStrategiesFixture.cs" />
+    <Compile Include="TestNameStrategiesTest.cs" />
     <Compile Include="ThrowingStubFixture.cs" />
     <Compile Include="TypeWithCustomizationAttributes.cs" />
+    <Compile Include="FixedNameTestMethodBuilderTest.cs" />
+    <Compile Include="VolatileNameTestMethodBuilderTest.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\AutoFixture.NUnit3\AutoFixture.NUnit3.csproj">
--- a/Src/AutoFixture.NUnit3.UnitTest/FixedNameTestMethodBuilderTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/FixedNameTestMethodBuilderTest.cs
@@ -1,0 +1,22 @@
+﻿using NUnit.Framework;
+using NUnit.Framework.Internal;
+using System;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    public class FixedNameTestMethodBuilderTest
+    {
+        [Test]
+        public void FixedNameTestMethodBuilderIsResilientToFactoryException()
+        {
+            // Fixture setup
+            var dummyMethod = new MethodWrapper(typeof(TestNameStrategiesFixture), nameof(TestNameStrategiesFixture.FixedNameDecoratedMethod));
+            var sut = new FixedNameTestMethodBuilder();
+            // Exercise system
+            var testMethod = sut.Build(dummyMethod, null, () => throw new Exception(), 0);
+            // Verify outcome
+            Assert.That(testMethod.Name, Is.EqualTo(nameof(TestNameStrategiesFixture.FixedNameDecoratedMethod)));
+            // Teardown
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
@@ -378,7 +378,7 @@ namespace Ploeh.AutoFixture.NUnit3.UnitTest
 
         [Theory, AutoData]
         public void NoAutoPropertiesAttributeLeavesPropertiesUnset(
-            [NoAutoProperties]PropertyHolder<object> ph1, 
+            [NoAutoProperties]PropertyHolder<object> ph1,
             [NoAutoProperties]PropertyHolder<string> ph2,
             [NoAutoProperties]PropertyHolder<int> ph3
             )
--- a/Src/AutoFixture.NUnit3.UnitTest/VolatileNameTestMethodBuilderTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/VolatileNameTestMethodBuilderTest.cs
@@ -1,0 +1,22 @@
+﻿using NUnit.Framework;
+using NUnit.Framework.Internal;
+using System;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    public class VolatileNameTestMethodBuilderTest
+    {
+        [Test]
+        public void VolatileNameTestMethodBuilderIsResilientToFactoryException()
+        {
+            // Fixture setup
+            var anyMethod = new MethodWrapper(typeof(TestNameStrategiesFixture), nameof(TestNameStrategiesFixture.VolatileNameDecoratedMethod));
+            var sut = new VolatileNameTestMethodBuilder();
+            // Exercise system
+            var testMethod = sut.Build(anyMethod, null, () => throw new Exception(), 0);
+            // Verify outcome
+            Assert.That(testMethod.Name, Is.EqualTo(nameof(TestNameStrategiesFixture.VolatileNameDecoratedMethod)));
+            // Teardown
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3/AutoDataAttribute.cs
+++ b/Src/AutoFixture.NUnit3/AutoDataAttribute.cs
@@ -5,6 +5,7 @@ using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Builders;
 using Ploeh.AutoFixture.Kernel;
+using System.Diagnostics.CodeAnalysis;
 
 namespace Ploeh.AutoFixture.NUnit3
 {
@@ -12,11 +13,22 @@ namespace Ploeh.AutoFixture.NUnit3
     /// This attribute uses AutoFixture to generate values for unit test parameters. 
     /// This implementation is based on TestCaseAttribute of NUnit3
     /// </summary>
+    [SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", 
+        Justification = "This attribute is the root of a potential attribute hierarchy.")]
     [AttributeUsage(AttributeTargets.Method)]
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", Justification = "This attribute is the root of a potential attribute hierarchy.")]
     public class AutoDataAttribute : Attribute, ITestBuilder
     {
         private readonly IFixture _fixture;
+
+        private ITestMethodBuilder _testMethodBuilder = new VolatileNameTestMethodBuilder();
+        /// <summary>
+        /// Gets or sets the current <see cref="ITestMethodBuilder"/> strategy.
+        /// </summary>
+        public ITestMethodBuilder TestMethodBuilder
+        {
+            get => _testMethodBuilder;
+            set => _testMethodBuilder = value ?? throw new ArgumentNullException(nameof(value));
+        }
 
         /// <summary>
         /// Construct a <see cref="AutoDataAttribute"/>
@@ -49,31 +61,14 @@ namespace Ploeh.AutoFixture.NUnit3
         /// <returns>One or more TestMethods</returns>
         public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
         {
-            var test = new NUnitTestCaseBuilder().BuildTestMethod(method, suite, this.GetParametersForMethod(method));
+            var test = this.TestMethodBuilder.Build(method, suite, () => GetParameterValues(method).ToArray(), 0);
 
             yield return test;
         }
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "This method is always expected to return an instance of the TestCaseParameters class.")]
-        private TestCaseParameters GetParametersForMethod(IMethodInfo method)
+        private IEnumerable<object> GetParameterValues(IMethodInfo method)
         {
-            try
-            {
-                var parameters = method.GetParameters();
-
-                var parameterValues = this.GetParameterValues(parameters);
-
-                return new TestCaseParameters(parameterValues.ToArray());
-            }
-            catch (Exception ex)
-            {
-                return new TestCaseParameters(ex);
-            }
-        }
-
-        private IEnumerable<object> GetParameterValues(IEnumerable<IParameterInfo> parameters)
-        {
-            return parameters.Select(Resolve);
+            return method.GetParameters().Select(Resolve);
         }
 
         private object Resolve(IParameterInfo parameterInfo)
--- a/Src/AutoFixture.NUnit3/AutoFixture.NUnit3.csproj
+++ b/Src/AutoFixture.NUnit3/AutoFixture.NUnit3.csproj
@@ -71,6 +71,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="FixedNameTestMethodBuilder.cs" />
     <Compile Include="InlineAutoDataAttribute.cs" />
     <Compile Include="AutoDataAttribute.cs" />
     <Compile Include="CustomizeAttribute.cs" />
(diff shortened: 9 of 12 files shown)


frblondin:
FYI I created a temporary [NuGet package](https://www.nuget.org/packages/AutoFixture.NUnit3.PatchedName/3.50.2.2) because I needed to share this.

Any chance you can have a look at this PR?

moodmosaic:
As [added to the README](https://github.com/AutoFixture/AutoFixture/commit/ab829640ed8e02776e4f4730d0e72ab3cc382339)

>[AutoFixture is getting a new governance model](https://github.com/AutoFixture/AutoFixture/issues/703). This most likely means that you can expect delays if you have questions, issues, or pull requests.

zvirja:
@frblondin Could you clarify whether you are still interested and have time to work on this PR? 😉 That is to understand whether the review makes sense..

frblondin:
@zvirja Sure I am, please let me know what I need to do so that this PR gets merged 😄

zvirja:
@frblondin Finally I had a chance to review the suggested changes and think a bit about that. As far as I understand this issue is specific to VS runner only and other runners (like console, R#) works well. We also have a [similar situation for xUnit](https://github.com/xunit/xunit/issues/1386) and the priority seems to be the same as here.

I've checked your PR and have a few concerns regarding the implementation:
- It's not very good idea to enable this by default for all the consumers. There might be people who are fine how it works now (e.g. they don't use VS runner) and they don't need this feature. Also this feature makes it hard to later troubleshoot the issue as you don't see the values that failed test. On other hand, I don't see great ways how to make it configurable - each approaches has some drawbacks.
- The code which manually concatenates the test name scares me. It's very low-level and very specific, it's hard to be sure that we covered all the cases. If later NUnit3 will fix some bug or change the rendering approach, we will produce names that will differ from the standard ones. Ideally, we should be able to use some NUnit API to format test name, however change volatile argument values to static ones. If current NUnit doesn't expose any of such APIs, we should discuss that at NUnit side and implement such an API.

What are your thoughts regarding that?

frblondin:
* About not displaying values and preventing from quickly know the values causing the error it's a very valid point. As you pointed out, there is no good alternative I can think of. In any case from my experience this is a rare case and the logging tools or good assertion messages seems acceptable to me.
* Test names are normally created by NUnit in [NUnitTestCaseBuilder](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L54) using the [TestNameGenerator](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/TestNameGenerator.cs). As you said there are raw concatenations but I followed what used to be done in the existing implementation of AutoFixture.Nunit. I agree with you, we should change the implementation so that it uses it. I will do the changes and push a new commit.

What do you think?

zvirja:
> As you pointed out, there is no good alternative I can think of. In any case from my experience this is a rare case and the logging tools or good assertion messages seems acceptable to me.

My point is that we _shouldn't_ change the existing behavior. The new behavior should be _disabled_ by default and only those ones who wants it might enable it. I'm thinking about [property injection](https://jeremybytes.blogspot.com/2014/01/dependency-injection-property-injection.html) to `AutoDataAttribute` and `InlineAutoDataAttribute` with default [strategy](https://sourcemaking.com/design_patterns/strategy) that a doesn't modify the display name.

Later, if you wish to have a static test names, you could create your own `AutoData` attributes and enable the appropriate strategy.

Only after some time if we see that most of people with this strategy we could probably make it default.

> I agree with you, we should change the implementation so that it uses it. I will do the changes and push a new commit.

I don't know NUnit code base, so, in fact, I don't imagine how it will look like 😟 If you believe that it will address the concerns I added - feel free to push those changes so we can see them 😉

frblondin:
For the activation of the "static" test names, I cannot think of an easy way to switch to this mode:
- Additional attribute to be applied on methods: really heavy solution. If you want to switch to this mode I don't see why you'd want it to be applied for some tests but not for others...
- Additional entry in the assembly app.config file: same as above, not really convenient

**Since these tests can only work if static names are used in VS** I think we could change them without having to use an option, only in the context of VS. It may sound like hacky but we a simple way to detect which names should be generated can be checking the process name hosting the running code. Something like `System.Diagnostics.Process.GetCurrentProcess().MainModule.FileName.StartsWith("vstest", StringComparison.OrdinalIgnoreCase)`.

Crazy?

zvirja:
> Crazy?

Yep! 😉 I think that it will be very tricky approach and it might be non-stable. It's better to have consistency, so it works the same in all the runners.

> Additional attribute to be applied on methods: really heavy solution.

It seems that you misunderstood my idea with property injection and strategy. Let me show how it looks:
```csharp
public class AutoDataAttribute
{
    public ITestNameStrategy NameStrategy { get; set; } = new DefaultNameStrategy();
    ....
}
```

later you can create your own data attribute in solution:
```charp
public class AutoDataStableNameAttribute : AutoDataAttribute
{
    public AutoDataStableNameAttribute()
    {
        this.NameStrategy = new StableNameStrategy();
    }
}
```

and later you simply use your own `AutoDataStableNameAttribute` everywhere instead of default `AutoData`:
```charp
[Test, AutoDataStableName]
public void SomeTest(...) { ... }
```

This approach is not so bad, as:
- It's explicit and you have total control over the look.
- Usually you use AF together with `AutoMoq` or `AutoNSubstitute` libraries, so you already have your own attribute.

This way allows users to opt-in only if they want and out of the box the name will not be changed.
Later we can review the approach further and probably use some other default strategy (e.g. controlled by [`AppContext`](https://msdn.microsoft.com/en-us/library/system.appcontext(v=vs.110).aspx) switch).

zvirja:
I've recently learned this question deeply for **xUnit** and found that actually it supports auto-data out of the box. It expects that theory data might be dynamic and adds a special support for that.

What we are doing here for NUnit looks more like a workaround, rather than a real solution. We are not the guy who is doing something wrong. Rather, the issue is with NUnit that doesn't have support for that.

Probably, we should go one step ahead and raise the discussion on their side, so they support this feature and we simply consume it.

frblondin:
I think I explorer this path already but I'll to it once more: NUnit provides the ability to generate values that change over time - see [Random](https://github.com/nunit/docs/wiki/Random-Attribute) as an example. They implemented a way so that the value can be passed between the vstest discovery process to the vstest runner.

There is one big difference though: supported value types for RandomAttribute are simple serializable values (see the [source file](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Attributes/RandomAttribute.cs), only integers, doubles, longs...). Other than for simple types (scalar types, strings, int...) it is **impossible** to create test values during discovery and pass them to the runner. These are two different processes.

That could mean that for simple types we could try to reuse the same behavior than for the NUnit RandomAttribute but for more complex type we should display autodata... but you are reluctant to do so. For most cases the test name for a complex type parameter will always produces a non-meaningfull value - NameSpace.Type<SomeOtherNameSpace.OtherType> as an example. This is why I'm still not getting why changing the behavior is not acceptable.

Before considering how to implement it, what do we want to do?

zvirja:
Notice, as I mentioned before, I'm with xUnit background and personally I don't use NUnit. Therefore, I apologize if there are some areas in which I have wrong vision :worried:

I've just played more and got some vision how NUnit decorate names. I used the following code for testing:
```csharp
public class Data
{
    public int IntValue { get; set; }
}

public class DataWithToString
{
    public int IntValue { get; set; }
    public override string ToString() => $"MyData: {IntValue}";
}

public interface ISomeData
{
    int IntProp { get; }
}

public class AutoNData : AutoDataAttribute
{
    public AutoNData():base(new Fixture().Customize(new AutoConfiguredNSubstituteCustomization()))
    {
    }
}

public class TestNameTester
{
    [Test, AutoNData]
    public void Sample(int a, Data d1, DataWithToString d2, ISomeData d3)
    {
        Assert.IsTrue(true);
    }
}
```
and this is the result I got:
```
NUnit3RunnerTest709.TestNameTester.Sample(105,NUnit3RunnerTest709.Data,MyData: 236,Castle.Proxies.ISomeDataProxy)
```

As we can see from this (and [here](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/TestNameGenerator.cs)) potentially any argument (primitive or complex) could have dynamic name.

Our goal (if I see that clearly) is to ensure that we have a stable test name. From this perspective it doesn't make any sense to distinguish between primitive and non-primitive types as both kinds could lead to variable test name (see a sample above).

> That could mean that for simple types we could try to reuse the same behavior as for the NUnit RandomAttribute but for more complex type we should display autodata... but you are reluctant to do so. For most cases the test name for a complex type parameter will always produces a non-meaningfull value - NameSpace.Type<SomeOtherNameSpace.OtherType> as an example.

I think that we are a bit unaligned, so let me explain that for one more time. Our library is quite popular and we have a lot of people using it to test their enterprise or opensource code. We don't know about all the scenarios, therefore it's unsafe to make decision that could affect them. Most of those people might use different runners (e.g. R#, Console) and are not affected by this issue. If you change test name, this will make them harder to troubleshoot if they sporadically failed on CI. Consider the sample I used above:
**Currently:**
```
NUnit3RunnerTest709.TestNameTester.Sample(105,NUnit3RunnerTest709.Data,MyData: 236,Castle.Proxies.ISomeDataProxy)
```

**After:**
```
NUnit3RunnerTest709.TestNameTester.Sample(auto<System.Int32>,auto<NUnit3RunnerTest709.Data>,auto<NUnit3RunnerTest709.Data>,auto<Castle.Proxies.ISomeDataProxy>)
```

If test sporadically failed on CI, currently you can troubleshoot that likely either `105` or `236` caused the issue. After we accept your fix and enable it, users will not have that information.

I _do understand_ that we need this change to fix VS runner, but that should be something you opt-in, something that you must _explicitly enable_ understanding the trade-off and benefits. That's why I suggest to implement this feature, however keep it in the _disabled_ state by default. It will be very easy to enable it in your project if you need it, however you will not affect people that prefer diagnostics information over working VS runner they don't use.

------
As further steps I suggest the following:
1. Rework current approach, so that we don't decorate string manually, but rather use NUnit API for that (e.g. `new TestNameGenerator("{m}{a:40}").GetDisplayName(test, ....)`).
1. Rework implementation so that feature is disabled by default and you need to enable it manually. I'd prefer the Strategy pattern I already described [here](709#issuecomment-324591069). Don't use some static switchers for now (like static variable, AppConfig, etc), we could introduce them later if strongly needed.
1. Don't distinguish between primitive and non-primitive types as both them could produce dynamic names. It will be complex and benefit is little.
1. Register an issue (or find an existing) for NUnit to support tests with dynamic data out of the box, so that we don't need to invent these [kludges](https://en.wikipedia.org/wiki/Kludge) to simply work.

Hope that makes sense. If not - describe your concerns.

zvirja:
It appeared that not only VS runner becomes crazy - NCrunch is affected as well. Therefore, we definitely should provide a way for users to support those runnres.

@frblondin Looking forward to your follow up so we can proceed! 😉

frblondin:
Thanks for this deep analysis. I did some more tests and things are getting more and more confusing (for xUnit)... but I think I have an acceptable plan in the end for NUnit.

I did some more tests in the attached solution:
* I wanted to see how it works with xUnit... VS 2017 15.3.2 + xUnit + AutoFixture seems to be also failing. Use this simple procedure: 1/ Open VS, 2/ Add breakpoint for each unit test in XUnit.cs, 3/ Go to VS Test Explorer, 4/ Click on debug -> the tests decorated with `AutoDataAttribute` or `InlineAutoDataAttribute` fails at being debugged. May be I missed something and this is not the primary topic for this PR - but I thought you might be interested. Also I'm a bit confused why this would not be worrying anyone since VS is used by 99% of .Net developers and xUnit is a mainstream unit test framework.
* I created a simple `CustomDataAttribute` attribute that mimics the `AutoDataAttribute` attribute from AutoFixture.NUnit. It turns out that what is the identifier for a test case between the test discovery and the test execution is the `FullName` property, not the `Name` property. `FullName` gets usually automatically computed from the Name. In my `CustomDataAttribute` I only modified the `FullName` and things are working.
Of course I don't know of all test execution platforms (dev platforms, CI tools, coverage...) but it seems acceptable to only change the fullname, given that the `Name` would remain unchanged and would contain the actual values (`105` or `236` caused the issue...).
Please see the attached solution: you should be able to debug `NUnit_HomeMade(...)`, not `NUnit_AutoData(...)`.

[AutoFixtureCases.zip](https://github.com/AutoFixture/AutoFixture/files/1254160/AutoFixtureCases.zip)
***
As next steps I propose:
1. Do not override the test `Name` property any longer. From what I can see there is no need for AutoFixture.NUnit to override the way the test name is created. NUnit will do that for you and will use the customized naming formatting if any.
2. Override the `FullName` property. Note that for the proper implementation we will need to only provide inlined argument values, not auto data argument values. They reason if that if the same test method has multiple test cases (with different inline values) the full name must be able to identify them uniquely.

Hope this sounds good. If that's not the case we will need to reach out to NUnit folks.

zvirja:
> I wanted to see how it works with xUnit... VS 2017 15.3.2 + xUnit + AutoFixture seems to be also failing. 

That is because you have installed `AutoFixture.xunit` glue library instead of `AutoFixture.xunit2`. If you install the valid library - everything will be fine 😉

> Do not override the test Name property any longer.
> Override the FullName property.

I don't like this idea and vote to override both the properties. [Internally, nUnit uses](https://github.com/nunit/nunit/blob/4d31aea6175dc338c0ce66dcd9145d62c52fa781/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L54) the `Name` property to generate the `FullName` one. I've checked source code and found about 75 usages of the `Name` property. I'm not sure that we will not break something if we keep those 2 ones inconsistent.

Therefore, let's override both the properties - it will be safer.

frblondin:
> Therefore, let's override both the properties - it will be safer.

Are you saying that you are ok with generating a name that doesn't the values anymore (auto(<xxx> instead)? I'd be suprised that you changed your mind about this. I probably misunderstood...

zvirja:
> Are you saying that you are ok with generating a name that doesn't the values anymore (auto( instead)?

Yes, let's override stable name for both `Name` and later `FullName` properties as otherwise it could lead to unpredictable results. I don't know nUnit enough to assert that it will work fine in all the cases.

> I'd be suprised that you changed your mind about this. I probably misunderstood...

I'm not sure why you see that I changed my mind.. I still expect that this feature will be _disabled_ by default. What we are currently discussing - is just a detail of implementation of the strategy that changes the test name.

frblondin:
Thanks for the clarification.

Ok, I'll implement the plan you suggested [here](https://github.com/AutoFixture/AutoFixture/pull/709#issuecomment-324911836) and will keep you posted.

frblondin:
During the last days I had limited availability to complete the PR. I have one last remark/question before I start.

The purpose of this PR is to circumvent a limit that affects the runtime (NUnit3 combined with VS test runner or other runners). I believe the source code should not be affected when we need one naming convention or another. We shouldn't have to change the attribute because we want/need another naming generation.

There are two choices: we can add an option using some test file (ie. if there is a file named XXX.config then read it) or we could also decide to use one NuGet package or another.

Today there is the [AutoFixture.NUnit3](https://www.nuget.org/packages/AutoFixture.NUnit3) and [AutoFxiture.Nunit3.PatchedName](https://www.nuget.org/packages/AutoFixture.NUnit3.PatchedName). Finally I believe that this is an acceptable solution.

I propose to add one more nuget package in this github repository so that any update made to AutoFixture.NUnit3 would benefit to the fixed name nuget package. We would need to find a nicer name that the nuget package I created - something like AutoFixture.NUnit3.FixedName or AutoFixture.NUnit3.ImmutableName.

What do you think?

zvirja:
@frblondin Thanks for sharing some new details and your vision!

> I propose to add one more nuget package in this github repository so that any update made to AutoFixture.NUnit3 would benefit to the fixed name nuget package. 

The approach with a standalone NuGet package looks attractive at the first glance, however I don't like it too much. That is simply a copy of the `AutoFixture.NUnit3` project with a slightly tuned `AutoData` and `InlineAutoData` attributes. We already have a plenty of [almost identical glue libraries](https://github.com/AutoFixture/AutoFixture/tree/master/Src) in our code base (like NUnit2, xUnit, xUnit2) and it would cost a lot to maintain one more without any strong reason. Another issue is that it would be hard for people to depend on us and create their own public libs based on our `AutoFixture.NUnit3` glue library - they will need to explicitly support both `AutoFixture.NUnit3` and `AutoFixture.NUnit3.PatchedName` as classes from these two assemblies will have different identities.

> we can add an option using some test file (ie. if there is a file named XXX.config then read it)

This option looks probably reasonable, however we should do that using iterations. We could finally come up with that solution in the future if that's indeed strongly demanded. What I'm actually suggesting now is to do steps _one by one_ - not all at once! The first step would be to implement that as a configurable strategy. Later we could improve that further so that default strategy will be a switcher and will look at config first.

> We shouldn't have to change the attribute because we want/need another naming generation.

I don't see actually too much issues with that, but it's probably my experience. I never used `AutoFixture.xUnit2` without `AutoFixture.AutoNSubstitute` (or `AutoFixture.AutoMoq`). Also I always needed to tune the default behavior of AutoFixture and add project-specific customizations. As result, I always ended up in creation of my own `AutoData` and `InlineAutoData` attributes and using them all over the project instead of the default ones.

With such experience behind I think that a lot of people are doing the same. That's why I expect that it will not be a big issue to change the default naming strategy as you already have your own data attributes. However, if that appears to not be so - we are free to improve that in future by adding the configuration support.

---

The bottom line is that I'd suggest to proceed with the strategy approach described above and improve that later if needed. It will simplify both review and testing the changes.

frblondin:
I finally found a much better way to modify the test name. The TestParameters has two properties: [OriginalArguments](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/TestParameters.cs#L150) and [Arguments](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/TestParameters.cs#L110). The latter is used for the execution whereas the first is only used for the display. Instead of overriding the test name generation the code is now simply changing the name used by the test generation. Note that I had to set values on the existing array using indexed setters as these properties are readonly. Both arrays are two distinct instances, see [here](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/TestParameters.cs#L87).

Also, instead of setting a string value in the OriginalArguments array, I had to find an object which is not wrapped in double quotes ("auto<int>" vs. auto<int>) by the TestNameGenerator (see [here](https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/framework/Internal/TestNameGenerator.cs#L305)). This is why I'm storing StringBuilder instances in the OriginalArguments as the [TestNameGenerator simply calls ToString()](https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/framework/Internal/TestNameGenerator.cs#L232).

I decided to implement the logic in AutoDataAttribute instead of AutoDataFixedNameAttribute: as discussed if we later find a better way to decide whether fixed test names should be generated, the internal boolean (false by default) will just need to be set differently.

zvirja:
Thank you for the follow up! 👍

> I finally found a much better way to modify the test name. 

I don't like this approach. While the documentation for the `OriginalArguments` property says that it's being used for the display, the name of the property doesn't imply that. Therefore, such approach might be non-stable and looks a bit cumbersome. In future versions NUnit could add some additional responsibility to this property and our integration will fail or will break something inside the NUnit as it will not expect a trash there.

The previous approach when we format the name manually looks much better and I'd vote to stick to  that.

> This is why I'm storing StringBuilder instances

For me usage of StringBuilder looks like an overhead as this class is a bit complicated under the hood. I'd suggest to create a custom type instead with overridden `ToString()` method. It will be much quicker and safer as NUnit could start to support StringBuilder in the future versions.

Also ensure to use _full type name_ in the placeholder to avoid collisions when multiple types have the same name, but reside in different namespaces.

> I decided to implement the logic in AutoDataAttribute instead of AutoDataFixedNameAttribute

Previously we agreed to implement the logic via the strategy, however you changed the design. Could you please stick to that decision and follow it? The current approach has drawbacks:
- It adds [additional responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) to the AutoData attribute and from design perspective it's better to keep that logic in a separate class. It will look much simpler and cleaner.
- It adds unnecessary dependency between `InlineAutoData` and `AutoData` attributes as now you call the `AutoDataAttribute.UpdatedFixedTestArguments` static method. It's again a good indicator that strategy will work better.
- You have introduced `AutoDataFixedNameAttribute` attribute while we haven't agreed on that before. Please don't introduce that attribute for now as it could be easily created later in users solution. We could add it later via a separate PR if needed, however that is a whole new story.
- The `GeneratesFixedUnitTestName` property has protected setter - any reason for that? I believe that setter should be public (of course, in our case that will be a strategy property rather than a bool flag). Ensure you also have guards, so null value cannot be set.

> Lastly I had to update NUnit because of a bug in the TestNameGenerator - see last commit.

You haven't provided any information in the commit message what is the bug. Could you share this info? Notice, we cannot accept these updated NUnit dependency in v3 as this is a _HUGE_ breaking change. We are following the [SemVer](http://semver.org/) in this project and we provide a guarantee to customers that no breaking changes will appear withing the same major version. If you update the dependency of the package, it will be impossible for part of users to longer use this library.

Unless there is very strong reason, I suggest to keep the dependency version the same as it is. Otherwise, the chance we will be fine to accept these changes is lower 😟

> Merge remote-tracking branch 'upstream/master'

Please don't use back-merge from master into your feature branch and use the _rebase_ instead. The history will look much simpler and it will be easier to later track what is going on. Please rebase your branch on top of the latest master and ensure that merge commit is not present. If that is present, use the [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) to get rid of it.

Also don't forget to squash the obsoleted commits to make the commits history look cleaner. Of course, the important commits should be preserved, but the obsolete or non-significant ones should be squashed.

-----

Thank you for the time you spend on this - highly appreciated. Sorry that we cannot the changes in the current form - we should think about the project maintability and the suggested approach has some undesired implications.

Notice, previously we [have already agreed](https://github.com/AutoFixture/AutoFixture/pull/709#issuecomment-325312438) on approach, but you implemented it in a completely different way 😕 It's a bit hard to interact in such a way as it confuses and requires a lot of efforts from both my and your sides. I'd suggest to try to stick to the discussed approach as much as possible so we can proceed quicker.

Looking forward to your follow-up! 😉

frblondin:
You must think I'm crazy after reading your comment lol.

I don't think I really changed the [plan](https://github.com/AutoFixture/AutoFixture/pull/709#issuecomment-324911836):

> 1. Rework current approach, so that we don't decorate string manually, but rather use NUnit API for that (e.g. new TestNameGenerator("{m}{a:40}").GetDisplayName(test, ....)).

This is exactly what I intended to do. When you use xUnit the test names are including the namespace but test name under NUnit do not. This is something that could evolve and NUnit could provide a general option to amend the way test names are constructed. So I wanted to rely on NUnit API to generate the names just as you suggested, and not directly provide to the TestNameGenerator the pattern of test names that NUnit may change. Moreover if you change the test name you also need to change the FullTestName.

My primary implementation was relying on a direct use of TestNameGenerator but had the limitation I just mentioned... and I ended up seeing that the OriginalArguments is used to generate test names. I agree that the property name is crappy but the description says it's used for this purpose only.
Finally there is no perfect solution but I vote for the use of OriginalArguments.

After hearing these arguments let me know if you agree now.

>>This is why I'm storing StringBuilder instances

>I'd suggest to create a custom type instead with overridden ToString() method.

Agreed

> Also ensure to use full type name in the placeholder to avoid collisions when multiple types have the same name, but reside in different namespaces.

You mean that we should display auto<System.String> instead of auto\<String>, right? This will make the test names longer, reduce the readability, and could lead to other issues. I remember that Visual Studio Test Explorer has limitations with loooong test names (couldn't find any reference on the web but remember this was causing problems with xUnit 2). Moreoever the risk of having in the same test method two test cases that are having a reference to two different types with same name but different namespaces seems very unlikely to me.

>> I decided to implement the logic in AutoDataAttribute instead of AutoDataFixedNameAttribute

>Previously we agreed to implement the logic via the strategy, however you changed the design. Could you please stick to that decision and follow it?

Ok

>> Lastly I had to update NUnit because of a bug in the TestNameGenerator - see last commit.

> You haven't provided any information in the commit message what is the bug. Could you share this info?

Because I tried different attempts and because TestNameGenerator is declaring additional constructors in later versions of NUnit I did my developments locally on a more recent version of NUnit. Since I finally didn't use TestNameGenerator I primarily didn't push the NUnit update. The tests were then failing because appveyor used NUnit 3.0.0. I then pushed the NUnit 3.5.0 update without clearly identifying which commit resolves the issue.
I was not aware of the SemVer so I'll see if I can find a workaround in 3.0.0.

> Please use the rebase instead

Sure I'll do so for the next changes.

zvirja:
> So I wanted to rely on NUnit API to generate the names just as you suggested, and not directly provide to the TestNameGenerator the pattern of test names that NUnit may change.
> I agree that the property name is crappy but the description says it's used for this purpose only.

I've checked the implementation more carefully and must admit that the suggested approach with `OriginalArguments` looks better. I was about to reject this approach, but just noticed [this](https://github.com/nunit/nunit/blob/acdc5477eae0aa489331199ec07053c9e43b69f2/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L97) line for `FullName` formatting. It would be too much to keep this logic in our integration. While I don't like usage of `OriginalArguments` property as it isn't descriptive, it looks like less pain. Also I noticed that NUnit allows to set default test name pattern via console, so it's indeed better to support this feature by not hardcoding the pattern on our side.

So please go ahead with this approach, but please extract it to a standalone strategy. Probably, you should extract the [whole TestMethod creation](https://github.com/zvirja/AutoFixture/blob/3665511115c6eb69d6860874abb51e394672bb46/Src/AutoFixture.NUnit3/AutoDataAttribute.cs#L52-L52) to the strategy, so later strategy can decide whether it wants to patch `OriginalArguments` or the `Name & FullName` properties. That's in case we find some issues in future and cannot bring the breaking changes.

> I remember that Visual Studio Test Explorer has limitations with loooong test names
> Moreoever the risk of having in the same test method two test cases that are having a reference to two different types with same name but different namespaces seems very unlikely to me.

It might be right as I noticed that previous version of NUnit had argument values limit set to [40 chars](https://github.com/nunit/nunit/commit/4690d6024c140417ccb5c78ff9e0c7ee09e54147#diff-7842b7101011893b22318d35b285ed47), while later the limit was removed..

OK, let's proceed with the suggested approach and put the type name only. We could change that later if users report about name collision issues.

---

So please go ahead and
- rebase your branch on top of fresh `master`;
- clean-up the commit history (e.g. revert the NUnit upgrade and remove that ugly merge commit);
- extract name management to the strategy;
- remove the introduced `AutoDataFixedNameAttribute` like attributes.

Will be looking forward to your updates and thanks for your efforts 😉

P.S. Notice, I'll be in vacation for one week starting the upcoming Saturday, so you might observe the delay in that period.

frblondin:
> remove the introduced `AutoDataFixedNameAttribute` like attributes

I don't get this one... we need a dedicated attribute to use the new strategy instead of the default strategy that displays the argument value.

zvirja:
> we need a dedicated attribute to use the new strategy instead of the default strategy that displays the argument value.

Sure we need! But that dedicated attribute will be created by our clients in their solutions, rather than by us 😉 I don't see any reason to add it and support in the AutoFixture code base. Later we could change the default strategy, so users might not even need to create a custom attribute.

frblondin:
New & hopefully the last update :-)

Attention point [here](https://github.com/frblondin/AutoFixture/blob/master/Src/AutoFixture.NUnit3/Builders/FixedNameTestCaseParameterBuilder.cs#L83): that's the only I found to overcome the issue the version of NUnit being used. Note that with newer version the function won't do anything.

Src/AutoFixture.NUnit3/DefaultTestCaseParameterBuilder.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    /// <summary>
+    /// Default builder used by <see cref="AutoDataAttribute"/> and <see cref="InlineAutoDataAttribute"/> to create
```
zvirja: Please rewrite doc to not mention who uses this type and how. Instead simply document the behavior of this strategy as if it would be one among many other similar strategies.

Src/AutoFixture.NUnit3/FixedNameTestCaseParameterBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Linq;
+using System.Text;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using System.Globalization;
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    /// <summary>
+    /// Builder that generates fixed values for the <see cref="TestCaseParameters"/>.OriginalArguments
```
zvirja: Rephrase doc to emphasize that this strategy generate static test name and omit the implementation details.

Src/AutoFixture.NUnit3/FixedNameTestCaseParameterBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// In NUnit 3.5+ the Arguments and OriginalArguments properties are containing two different instances,
+        /// not in earlier versions.<BR/>
+        /// Unfortunately Arguments has a private setter and there is not way to change the instance other
+        /// than by using reflection...<BR/>
+        /// When running in NUnit3.5+ the test below will be false.
```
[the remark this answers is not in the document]
zvirja: If you meant the condition at the beginning of the method - rephrase to "When... method is supposed to do nothing".

frblondin:
Thanks a lot for your review. I'm really impressed with the quality of your comments.

I just pushed the changes and I think I took into account all your comments. Note that I'm a but confused with the github interface...

frblondin:
May be I should have pushed new commits instead of squashing them...

zvirja:
@frblondin Thank you a lot for those changes! Now code looks significantly cleaner! 😎

I've noticed that some of the comments were not addresses. Notice, it's always better to add some marks to the comments when you addressed them (e.g. 👍) and comments when you won't fix.

I'll take a look again and will keep the leftover comments.

Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Globalization;
+
+namespace Ploeh.AutoFixture.NUnit3
+{
+    /// <summary>
+    /// Builder that generates <see cref="TestMethod"/> with fixed names.
+    /// This is needed for some test runners such as the Nunit test adaptor for Visual Studio.
+    /// </summary>
+    /// <seealso cref="MutableNameTestMethodBuilder" />
+    public class FixedNameTestMethodBuilder : MutableNameTestMethodBuilder
```
zvirja: Please do not use the inheritance here and implement the interface directly. I see both strategies as independent ones, rather than "FixedName" strategy "is" one "VolatileName" strategy as that is implied by inheritance. Inheritance should be only used when child _is_ parent (like Cat is Animal), but not when two classes look similar and you want to reuse some code.

Another point is that code reuse is really tiny, so inheritance doesn't worse it as maintenance overhead is larger than benefits.

When you refactor this, alter all the protected members in the "VolatileName" strategy as they are no longer required. That will allow to remove most of the guard clauses, as they are not needed for the private methods.

Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// not in earlier versions.<BR/>
+        /// Unfortunately Arguments has a private setter and there is not way to change the instance other
+        /// than by using reflection...<BR/>
+        /// When running in NUnit3.5+ the method is supposed to do nothing.
+        /// </summary>
+        private static void EnsureOriginalArgumentsArrayIsNotShared(TestCaseParameters parameters)
+        {
+            if (ReferenceEquals(parameters.Arguments, parameters.OriginalArguments))
+            {
+                var clonedArguments = new object[parameters.Arguments.Length];
```
zvirja: Please make a copy of the `OriginalArguments` array and update that property. It will be more correct, as we update the `OriginalArguments` property, not the `Arguments` one.

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesFixture.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Ploeh.AutoFixture.NUnit3.UnitTest
+{
+    public class TestNameStrategiesFixture
+    {
+        #region Custom attribute definitions
+        public const int InjectedIntValue = 1;
```
zvirja: Always use 42 for such values 😆

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        [Test]
+        public void AutoDataUsesRealValuesForTestNameByDefault()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<AutoDataDefaultNameAttribute>(nameof(TestNameStrategiesFixture.DefaultAttributeStrategyDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(nameof(TestNameStrategiesFixture.DefaultAttributeStrategyDecoratedMethod) + "(" + TestNameStrategiesFixture.InjectedIntValue + @",""" + TestNameStrategiesFixture.InjectedStringValue + @""")"));
```
zvirja: I'd vote to hard-code int and string values and don't use the concatenation. Currently test are read bad because of this concatenation..

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        [Test]
+        public void InlineAutoDataUsesRealValuesForTestNameByDefault()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataDefaultNameAttribute>(nameof(TestNameStrategiesFixture.DefaultInlineAttributeStrategyDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.Name,
+                Is.EqualTo(nameof(TestNameStrategiesFixture.DefaultInlineAttributeStrategyDecoratedMethod) + @"(""alpha"",""beta"",""" + TestNameStrategiesFixture.InjectedStringValue + @""")"));
```
zvirja: Embed injected value and avoid concatenation - readability is poor :(

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        [Test]
+        public void InlineAutoDataFixedNameUsesRealValuesForTestFullNameByDefault()
+        {
+            // Fixture setup
+            // Exercise system
+            var testMethod = GetTestMethod<InlineAutoDataDefaultNameAttribute>(nameof(TestNameStrategiesFixture.DefaultInlineAttributeStrategyDecoratedMethod));
+            // Verify outcome
+            Assert.That(testMethod.FullName,
+                Is.EqualTo(typeof(TestNameStrategiesFixture).FullName + "." + nameof(TestNameStrategiesFixture.DefaultInlineAttributeStrategyDecoratedMethod) + @"(""alpha"",""beta"",""" + TestNameStrategiesFixture.InjectedStringValue + @""")"));
```
zvirja: Embed injected value and avoid concatenation - readability is poor :(

Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
```diff
@@ -0,0 +1,84 @@
...
+                result.OriginalArguments[i] = new TypeNameRenderer(methodParameters[i].ParameterType);
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// In NUnit 3.5+ the Arguments and OriginalArguments properties are containing two different instances,
+        /// not in earlier versions.<BR/>
+        /// Unfortunately Arguments has a private setter and there is not way to change the instance other
```
zvirja: Fix to "OriginalArguments has a..", as you will change this property.

Src/AutoFixture.NUnit3/MutableNameTestMethodBuilder.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "This method is always expected to return an instance of the TestCaseParameters class.")]
+        private TestCaseParameters GetParametersForMethod(IMethodInfo method, Func<object[]> argFactory, int autoDataStartIndex)
+        {
+            try
+            {
+                var parameterValues = argFactory();
+                return GetParametersForMethod(method, parameterValues.ToArray(), autoDataStartIndex);
+            }
+            catch (Exception ex)
```
zvirja: Please add a test for exception case. Ensure to add tests for both the strategies.

zvirja:
@frblondin Do you know whether you will have time in the nearby future to fix my comments? We are on the final way and I wish to publish it soon 😅 Also I'm actively working on v4 release and it would be good to finish this one, so I could apply safely apply the breaking changes to v4 and don't worry about merge conflicts.

frblondin:
> Specify the strategy explicitly here. Don't rely on a fact which one is default - it might be changed in future. Don't forget to update the name and related test names 😉

The purpose of this test was to test what it the default strategy for non-regression testing - eg. make sure that the name won't be altered after people update their NuGet package.

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesFixture.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            public AutoDataFixedNameAttribute()
+            {
+                TestMethodBuilder = new FixedNameTestMethodBuilder();
+            }
+        }
+
+        public class AutoDataDefaultNameAttribute : AutoDataAttribute
+        {
+            public AutoDataDefaultNameAttribute() : base(CreateFixtureWithInjectedValues())
```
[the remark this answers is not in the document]
zvirja: > The purpose of this test was to test what it the default strategy for non-regression testing - eg. make sure that the name won't be altered after people update their NuGet package.

Currently you have two kind of attributes - `AutoDataFixedNameAttribute` and `AutoDataDefaultNameAttribute`. In this scenario context it's better to have `AutoDataFixedNameAttribute` and `AutoDataVolatileNameAttribute` - to cover all the cases individually.

If you wish add tests for regression, we should:
- Assert the default strategy - already done (`AutoDataAttributeUsesRealValueByDefault()` and `InlineAutoDataAttributeUsesRealValueByDefault()`).
- Add yet another set of tests where you test the `AutoDataAttribute` directly and assert that name is mutating by default.

I'd say that the last point is not obligatory, but if you wish - feel free to add it.

P.S. Please add reply comments in place, rather than in global scope - it's easier to track them 😉

zvirja:
@frblondin Replied in appropriate thread. Please add replies in the related threads for better readability.

Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesFixture.cs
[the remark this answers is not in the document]
zvirja: The main issue with current approach is that when we change the default strategy, all those tests will become failing and no tests will cover the `VolatileTestBuilder` anymore.

frblondin:
> Fix to "OriginalArguments has a..", as you will change this property.
> Please make a copy of the OriginalArguments array and update that property. It will be more correct, as we update the OriginalArguments property, not the Arguments one.

We are changing the Arguments property, not OriginalArguments.

frblondin:
Please let me know for the two remaining comments and I'll push my updates.

zvirja:
@frblondin Could you please reply in threads as I asked you about [here](https://github.com/AutoFixture/AutoFixture/pull/709#issuecomment-333571560)?

Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
zvirja: > We are changing the Arguments property, not OriginalArguments.

That's the main point concern - we should modify the `OriginalArguments` array instead. It's a bit strange that we patch the `OriginalArguments` array, but make a copy of the `Arguments` one. See my comment below - it's exactly about that.

zvirja:
I've replied to all your questions - please take a look.

Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs
zvirja: I've just checked the source code and found that actually the setter is present, but it's private. You can use the following code snippet to set the value:
```c#
// We should use the declaring type, otherwise private setter is not available.
var property = typeof(TestParameters)
    .GetProperty(nameof(TestCaseParameters.OriginalArguments));
property.SetValue(parametersObject, newValue);
```

Currently it doesn't matter whether you access `internal` or `private` setter - code is ugly in any way 😅 However, it will be more correct to modify the `OriginalArguments` array.

zvirja:
@frblondin Pinging just in case you missed my reply :wink:

frblondin:
@zvirja was in a long flight... I just pushed the round of updates - let me know!

zvirja:
@moodmosaic I've worked closely with Thomas on this one and finally we made a shape when PR is fine to me.

A short story - NUnit doesn't support tests with volatile names as it's unable to identify them between discovery and execution (opposed to xUnit which support this feature). Therefore, some runners like VS and NCrunch simply cannot run tests with AutoFixture. In this PR we add strategies to produce different kind of test names. One of the strategy is to use the real argument values for test name (the default NUnit behavior), while another one is to replace real argument values with placeholders (`auto<ArgTypeName>` in our case). That allows user to create their own data attributes using the latter strategy to correctly work with NCrunch and xUnit. Of course, it's a bit clumsy behavior and this feature should be supported by NUnit rather than by us, however currently NUnit doesn't provide such API and our users suffer.

In this PR we don't change the default behavior and users should enable it manually. Therefore, this PR is a perfect candidate for v3. I have a plan to make this `fixed name` strategy the default one in v4, so we support all the test runners out of the box. Later if some people with to disable this feature, they could override the strategy.

Nikos, do you want to review the PR by yourself? If not - let me know and I'll proceed with a merge - it looks OK to me.

@frblondin Thank you a lot for all the time you spent on this. One last consideration - we should probably create a wiki article and describe all this stuff - so we can provide users with a link.. Or simply update ReadMe - I'm unsure..

moodmosaic:
@zvirja, go ahead 👍

zvirja:
@frblondin Merged! One more time thank you for the contribution. 🥇

I'll try to allocate some time later and create a page on wiki describing how to use this feature. Later I'll release a new minor version of `v3`, so you can consume the contribution.

frblondin:
Thanks a lot @zvirja for your time & support!

zvirja:
@frblondin The feature is released as v3.51.0. It's the first minor update since the ownership change 🎉
Mentioned you on twitter: https://twitter.com/AutoFixture/status/915953470578024449.

frblondin:
Awesome... and thanks for the credit :-)

BTW hope you'll have a chance to [share your opinion](https://github.com/AutoFixture/AutoFixture/issues/858) about this potential (much smaller) pull request that I could quickly create.

zvirja:
@frblondin I'm currently looking into that issue and will update you soon.

Do you have any plans to unlist the `AutoFixture.NUnit3.PatchedName` package from NuGet? It doesn't seem to be required anymore, while might lead to people using it because they are unaware of this feature being supported by AutoFixture. The existing clients will be still able to download it, while new clients will use the AutoFixture only.

What do you think?

frblondin:
Sure I was planning to hide this NuGet package. The remaining bit is the lazy IFixture load which we are relying on for speeding up test discovery - purpose of this issue.

PR #710, opened 2016-10-12
Add tests and impl for a random DateTimeOffset generator

steve-codemunkies:
This generator is based on the existing RandomDateTimeSequenceGenerator.
By randomly setting the offset you are able to ensure you are dealing
correctly with data from different time zones.

We (On The Beach) needed to generate DateTimeOffsets to test some of our
code. We have been using this generator for a while now, and in cleaning
up our code-base decided it was time to contribute this back.

Thanks for the great tool!

### DIFF
--- a/Src/AutoFixture/RandomDateTimeOffsetSequenceGenerator.cs
+++ b/Src/AutoFixture/RandomDateTimeOffsetSequenceGenerator.cs
@@ -1,0 +1,94 @@
+﻿using System;
+using Ploeh.AutoFixture.Kernel;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Creates random <see cref="DateTimeOffset"/> specimens.
+    /// </summary>
+    /// <remarks>
+    /// The generated <see cref="DateTimeOffset"/> UTC values will be within
+    /// a range of ± two years from today's date,
+    /// unless a different range has been specified in the constructor.
+    /// </remarks>
+    public class RandomDateTimeOffsetSequenceGenerator : ISpecimenBuilder
+    {
+        private readonly RandomNumericSequenceGenerator _dateRandomizer;
+        private readonly RandomNumericSequenceGenerator _offsetRandomizer;
+
+        // Offsets have a maximum ± of 14hrs...
+        private static long MinutesInFourteenHours = 840L;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RandomDateTimeOffsetSequenceGenerator"/> class.
+        /// </summary>
+        public RandomDateTimeOffsetSequenceGenerator()
+            : this(DateTimeOffset.Now.AddYears(-2), DateTimeOffset.Now.AddYears(2))
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RandomDateTimeOffsetSequenceGenerator"/> class
+        /// for a specific range of dates.
+        /// </summary>
+        /// <param name="minDate">The lower bound of the date range.</param>
+        /// <param name="maxDate">The upper bound of the date range.</param>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="minDate"/> is greater than <paramref name="maxDate"/>.
+        /// </exception>
+        public RandomDateTimeOffsetSequenceGenerator(DateTimeOffset minDate, DateTimeOffset maxDate)
+        {
+            if (minDate >= maxDate)
+            {
+                throw new ArgumentException("The 'minDate' argument must be less than the 'maxDate'.");
+            }
+
+            _dateRandomizer = new RandomNumericSequenceGenerator(minDate.ToUniversalTime().Ticks, maxDate.ToUniversalTime().Ticks);
+            _offsetRandomizer = new RandomNumericSequenceGenerator(MinutesInFourteenHours * -1, MinutesInFourteenHours);
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="DateTimeOffset"/> specimen based on a request.
+        /// </summary>
+        /// <param name="request">The request that describes what to create.</param>
+        /// <param name="context">Not used.</param>
+        /// <returns>
+        /// A new <see cref="DateTimeOffset"/> specimen, if <paramref name="request"/> is a request for a
+        /// <see cref="DateTimeOffset"/> value; otherwise, a <see cref="NoSpecimen"/> instance.
+        /// </returns>
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            return IsNotDateTimeOffsetRequest(request)
+                ? new NoSpecimen()
+                : CreateRandomDate(context);
+        }
+
+        private static bool IsNotDateTimeOffsetRequest(object request)
+        {
+            return !typeof(DateTimeOffset).IsAssignableFrom(request as Type);
+        }
+
+        private object CreateRandomDate(ISpecimenContext context)
+        {
+            // This ensures that the date/time falls within the specified range
+            var utcDto = new DateTimeOffset(GetRandomNumberOfTicks(context), TimeSpan.Zero);
+            // This then applies a randomly generated offset
+            return utcDto.ToOffset(GetRandomOffset(context));
+        }
+
+        private TimeSpan GetRandomOffset(ISpecimenContext context)
+        {
+            return new TimeSpan((long)_offsetRandomizer.Create(typeof(long), context) * TimeSpan.TicksPerMinute);
+        }
+
+        private long GetRandomNumberOfTicks(ISpecimenContext context)
+        {
+            return (long)this._dateRandomizer.Create(typeof(long), context);
+        }
+    }
+}
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -193,6 +193,7 @@
     <Compile Include="Kernel\TypeRelay.cs" />
     <Compile Include="OmitOnRecursionBehavior.cs" />
     <Compile Include="RandomCharSequenceGenerator.cs" />
+    <Compile Include="RandomDateTimeOffsetSequenceGenerator.cs" />
     <Compile Include="RandomDateTimeSequenceGenerator.cs" />
     <Compile Include="RandomBooleanSequenceCustomization.cs" />
     <Compile Include="RandomNumericSequenceCustomization.cs" />
--- a/Src/AutoFixture/DefaultPrimitiveBuilders.cs
+++ b/Src/AutoFixture/DefaultPrimitiveBuilders.cs
@@ -40,6 +40,7 @@ namespace Ploeh.AutoFixture
             yield return new MailAddressGenerator();
             yield return new EmailAddressLocalPartGenerator();
             yield return new DomainNameGenerator();
+            yield return new RandomDateTimeOffsetSequenceGenerator();
         }
 
         /// <summary>
--- a/Src/AutoFixtureUnitTest/AutoFixtureUnitTest.csproj
+++ b/Src/AutoFixtureUnitTest/AutoFixtureUnitTest.csproj
@@ -186,6 +186,7 @@
     <Compile Include="MarkerNode.cs" />
     <Compile Include="ProvidedAttribute.cs" />
     <Compile Include="RandomCharSequenceGeneratorTest.cs" />
+    <Compile Include="RandomDateTimeOffsetSequenceGeneratorTest.cs" />
     <Compile Include="RandomDateTimeSequenceGeneratorTest.cs" />
     <Compile Include="RandomBooleanSequenceCustomizationTest.cs" />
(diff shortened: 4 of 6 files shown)


moodmosaic:
@steve-codemunkies, thank you for opening a pull request 😃 I've started a review and left a comment.

steve-codemunkies:
Hello, is there a problem with the pull request, the only comment is the one above?

Thanks, Steve

moodmosaic:
>Hello, is there a problem with the pull request, the only comment is the one above?

[This](https://github.com/AutoFixture/AutoFixture/pull/710#discussion_r83076449) was the comment. I've no idea why it didn't show up (?) 😕

zvirja:
@steve-codemunkies Will you be able to find some and and follow up on this one so we can finally close it? ☺️

Src/AutoFixture/RandomDateTimeOffsetSequenceGenerator.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// Creates random <see cref="DateTimeOffset"/> specimens.
+    /// </summary>
+    /// <remarks>
+    /// The generated <see cref="DateTimeOffset"/> UTC values will be within
+    /// a range of ± two years from today's date,
+    /// unless a different range has been specified in the constructor.
+    /// </remarks>
+    public class RandomDateTimeOffsetSequenceGenerator : ISpecimenBuilder
+    {
+        private readonly RandomNumericSequenceGenerator _dateRandomizer;
```
steve-codemunkies: I've tried the change locally, and to be honest it add anything to the generator, but neither does it take away.

Do you want me to make the change to use the `RandomDateTimeSequenceGenerator`?

steve-codemunkies:
Sorry, got a bit busy... :) Anyway, it doesn't really add or detract to making the change, so please advise

Src/AutoFixture/RandomDateTimeOffsetSequenceGenerator.cs
zvirja: From my understanding I see 2 ways to re-use the existing generator:
1. Simply use the suggested `RandomDateTimeSequenceGenerator` instead of `RandomNumericSequenceGenerator ` and manual work with offsets when you calculate the number of ticks.

2. Don't manage the `DateTime` creation at all and use the `specimenContext.Resolve(typeof(DateTime))` instead. This way you will not care about `DateTime` creation so it could be configured/customized separately. Later you could create `DateTimeOffset` based on that `DateTime` value.

I'd prefer the second way as it's more natural for this project and is more customizable.
Do you see any issues with such approach?

Kralizek:
I never went too deep here, so i am just shooting in the dark. I hope nobody gets hurt :)

Would it be too big of a change to make DT as a special case of DTO?

Basically move the real logic in the Sequence Generator for DateTimeOffset and change the DateTime one to use the former and just return the value of the DateTime property of the created object.

steve-codemunkies:
@zvirja and @Kralizek so to answer your various points:

* Using `specimenContext.Resolve(typeof(DateTime))` is a non-starter (especially in the scenario that I created this generator for) as you have no control of the range of dates created. The code that I am testing requires different actions depending on the date/time, and the system I am integrating supplies the date/times as `DateTimeOffset` (making use of the offset part).

* It is possible to swap to use `RandomDateTimeSequenceGenerator`, but it does not add to the clarity of the code, further -

* `System.DateTime` and `System.DateTimeOffset` are **not** super/sub classes of each other. `DateTimeOffset` does make use of `DateTime` internally, but it also makes use of `TimeSpan` and a number of other types as well. They represent similar and closely related concepts, but they do that very differently.

Therefore I don't think that the changes requested are suitable or would aid or improve the clarity of either `RandomDateTimeOffsetSequenceGenerator` or `RandomDateTimeSequenceGenerator`.

zvirja:
Thanks for the replies. Let's proceed to have the complete clarity because I'm likely still missing something 😉

> System.DateTime and System.DateTimeOffset are not super/sub classes of each other. DateTimeOffset does make use of DateTime internally, but it also makes use of TimeSpan and a number of other types as well. They represent similar and closely related concepts, but they do that very differently.

I do also agree to not alter the `DateTime` creation logic and follow the append-only approach.

> Using specimenContext.Resolve(typeof(DateTime)) is a non-starter (especially in the scenario that I created this generator for) as you have no control of the range of dates created. 

I don't see how the existing way allows you to control that. Currently you internally create `DateTime` within (-2;+2) range, exactly how it will be done by `context.Resolve(typeof(DateTime))`. Later you [register builder](https://github.com/AutoFixture/AutoFixture/pull/710/files#diff-0b8f9434f4a792a034e92f352ecd2f54R43) with this predefined range in the `DefaultPrimitiveBuilders` instance. User doesn't have any control on the ranges this builder uses later. With such approach the only way for user to customize ranges will be to add _a custom `RandomDateTimeOffsetSequenceGenerator` instance_ to the `fixture.Customizations` with customized range. 

If we follow the way I suggested, we will need to customize the `DateTime` creation and register a `DateTime` builder with custom range instead. Later `DateTimeOffset` will use the customized range indirectly.

For me `DateTimeOffset` looks like `DateTime` + time zone offset. Therefore, it looks naturally that it delegates `DateTime` creation outside to the fixture (and depends on `DateTime` range customizations if there are such ones).

Have I overlooked something in my judgement or missed some usage scenario that is not covered? Do we indeed need isolation between `DateTime` and `DateTimeOffset`? I haven't worked with `DateTimeOffset` too much, so I might miss something. My intention is to use multi-dispatch polymorphism as much as possible and do not control sub-components creation if that is not necessarily required for correct work or handling the particular scenarios.

>It is possible to swap to use RandomDateTimeSequenceGenerator, but it does not add to the clarity of the code, further -

Could you please attach a sample of code how `RandomDateTimeOffsetSequenceGenerator` will look if you use `RandomDateTimeSequenceGenerator` instead, so we can compare?

Thank you :)

zvirja:
@steve-codemunkies I hope our discussion haven't caused you to lost the passion to this PR.. 😕

zvirja:
@steve-codemunkies Just a friendly reminder 😅 Hope, you have free time these days, so we can proceed further and finally merge the PR.

zvirja:
@steve-codemunkies It's being a while since your last comment. Could you please reply? If we don't merge this one in the nearby future (in a week or two) I'll be forced to close this PR, as I'm going to apply the final breaking changes to the `v4` branch. Otherwise, it will be a hell to merge the code.

Thanks!

steve-codemunkies:
Hi,

Sorry, I've been very busy. I've put in a new pull request (https://github.com/AutoFixture/AutoFixture/pull/860) that now uses the `RandomDateTimeSequenceGenerator`. As I have previously said though I do not feel that this adds anything to this generator.

Also I'm afraid @zvirja you're still conflating `DateTime` and `DateTimeOffset`. You are absolutely correct that you need to remove this generator and then re-add to get `DateTimeOffset`s in the range that you require. And this is what I do in my code that uses this. What I don't do is remove the `RandomDateTimeSequenceGenerator` and set this up as I am _not_ generating `DateTime` objects. To me your suggestion is completely counter-intuitive. It is like asking a user to remove and add a configured version of `Int32SequenceGenerator` to control how `string`s are generated.

Thanks, Steve

zvirja:
> I've put in a new pull request (#860) that now uses the RandomDateTimeSequenceGenerator

Thanks for that! 👍 Personally, it looks much better. While the number of lines of code is almost the same, idiomatically this code is more correct as `RandomDateTimeOffsetSequenceGenerator` now have _less_ responsibilities. However, let's resort to that approach only if we don't agree with `context.Resolve(typeof(DateTime))` - see below.

> It is like asking a user to remove and add a configured version of Int32SequenceGenerator to control how strings are generated.

Well, you are cheating a bit 😉 The `DateTimeOffset` type is composite and consists of `DateTime` and `Offset`, while string is not composite. Let's compare it better with `MailAddressGenerator` or `UriGenerator` - for me they look similar to `DateTimeOffset`.

I've read more about `DateTimeOffset` on MSDN and Microsoft seem to often mention that it's a composition of `DateTime` and `Offset`.
[Here](https://docs.microsoft.com/en-us/dotnet/standard/datetime/choosing-between-datetime):
> The DateTimeOffset type includes all of the functionality of the DateTime type along with time zone awareness.

[Here](https://msdn.microsoft.com/en-us/library/system.datetimeoffset(v=vs.110).aspx):
> The DateTimeOffset structure includes a DateTime value, together with an Offset property that defines the difference between the current DateTimeOffset instance's date and time and Coordinated Universal Time (UTC). 

It makes me to consider `DateTimeOffset` as a composition, so type should be classified in the same way as `MailAddress` or `Uri`.

----

I still don't see benefits from the approach when you _own_ DateTime creation, rather than delegate it to the `context.Resolve(typeof(DateTime))`. It doesn't seem that you need to constrain `DateTime` value in any special manner - you might construct the `DateTimeOffset` instance using any arbitrary `DateTime` object. Another story is with `Offset` - any arbitrary value will not work here, so we are forced to own the creation.

The reason why I'm persisting on this approach is that we follow it all over our code base. When we create composite types, we ask for the required parts, rather than own their generators. For instance:
- `MailAddressGenerator` [delegates](https://github.com/AutoFixture/AutoFixture/blob/cc12c22e47b94d232a6a3b540d791de2f6b0639e/Src/AutoFixture/MailAddressGenerator.cs#L51-L52) `EmailAddressLocalPart` and `DomainName` creation, rather than owns the generators locally.
- `UriGenerator` [delegates](https://github.com/AutoFixture/AutoFixture/blob/cc12c22e47b94d232a6a3b540d791de2f6b0639e/Src/AutoFixture/UriGenerator.cs#L31) `UriScheme` and `authority` creation, rather than owns generators.

In this case we need to customize e.g. `DomainName` generator to affect the `MailAddress` creation and it's perfectly fine here.

Therefore, I'd like to ask you again to clearly describe _all the reasons_ in form of a _bullet list_ why we need to own the `DateTime` creation and cannot delegate it to the kernel using `context. Resolve(typeof(DateTime))` instead. That will allow us to evaluate all the points and decide more correct approach.

P.S. Sorry for being strict and taking your extra-time on this question - I want to align our visions and understand why we might decide to follow the approach that differs from what we are doing in the similar classes now. That is a way to make everybody happy - you as a contributor and us as maintainers, who will support that code later 😉

steve-codemunkies:
Internally yes, DateTimeOffset is a composite (I've looked in the code too), however the two are still separate, though related concepts.

Thanks anyway, but I'll continue with the generator in my own codebase.

zvirja:
@steve-codemunkies Honestly - I'm a bit frustrated by your actions. I'm not charged to work on this project and spend my free time on it instead. I've already invested some amount of my resources thinking about your case and analyzing how it's better to perform the actions. Also I spend a few more hours to compose the answers to you and review how you implemented the feature and compare how similar features are implemented in our codebase. By doing this I supposed that we finally reach out to some agreement and the PR would be merged in some form. However instead I'm left with a feeling that I simply wasted my time and obviously such things don't inspire.

How I see it now, you simply closed the PR after you see that we have concerns regarding the current PR implementation and want to communicate things. If that is so - this is _disrespectful_ to me and other members who participated in review of this PR. I could understand if man closes a PR because of some important reason - e.g. he doesn't have time more to work on that, or it has been agreed to not include the feature. However even in such cases it's being agreed before, so there is a chance to pick up code by somebody else before PR is closed. Such things are usually implicitly meant by contribution etiquette and by contributing to the project you should realize that there might be some ongoing discussions and you might need to apply additional changes before things are merged. If you are not ready to invest to that - it's probably better to fire an issue, rather than a PR - so less resources are invested into this.

The reason you provided is still not strong enough to explain the current approach.
> Internally yes, DateTimeOffset is a composite (I've looked in the code too), however the two are still separate, though related concepts.

Composite entity consists of other members, but _is not_ those members. The `MailAddress` consists of `DomainName`, but you would agree that `MailAddress` and `DomainName` "are still separate, though related concepts". Therefore, if `MailAddressGenerator` delegates `DomainName` creation, I don't see why `DateTimeOffsetGenerator` should own the `DateTime` creation.

Wish you luck with your subsequent contributions and hope you'll not repeat the same pattern again.

Kralizek:
I would add that whilst it is true that DateTimeOffset is composing DateTime and TimeSpan, DTO uses only a specific Kind of DateTime.

Long story short (and forgive me for not providing sources), an instance of DateTime can be of Kind Local, Utc or Unspecified (casing is because those are the names used in the BCL). DateTimeOffset only uses DateTime of Kind Utc.

That being said, I am not sure whether or not the DTO constructor that accepts a DT purifies its inputs. If that is not the case, using any instance of DateTime would not be optimal.

As for the TimeSpan, that could be reduced to an Int32 generator that picks signed values between -11 and +12, at the price of losing the half time zones.

zvirja:
> If that is not the case, using any instance of DateTime would not be optimal.

You could easily normalize the value using `DateTime` API - that is still not a reason to own the `DateTime` activation :blush:

> As for the TimeSpan, that could be reduced to an Int32 generator

The existing implementation looked also fine for me and I'd kept it as is :wink:

For now, I don't see a reason to discuss this further and spend resources on this as the PR was closed.

PR #713, opened 2016-10-26
Refactored collections to support derived collection types

### LINKED ISSUE #700: Please support collections derived from ObservableCollection<T>
In version 3.50.2 of AutoFixture, ObservableCollection<T> is supported when used directly.

However it doesn't work for collection types that derive from the BCL type.

Could you please add support so that any type derived from this collection type is treated the same as the base type or List<T>?

tylerohlsen:
Closes #700 

I implemented this change as a replacement for the multiple implementations of collection types to a single implementation that is more generic as it works for all derived collection types. Alternatively, I may have been able to add some generic relays, but I felt this was a better solution. I'm happy to take critical feedback.

### DIFF
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -97,6 +97,7 @@
   <ItemGroup>
     <Compile Include="AutoPropertiesTarget.cs" />
     <Compile Include="BehaviorRoot.cs" />
+    <Compile Include="CollectionFillerCommand.cs" />
     <Compile Include="DomainName.cs" />
     <Compile Include="DomainNameGenerator.cs" />
     <Compile Include="ElementsBuilder.cs" />
--- a/Src/AutoFixture/CollectionFillerCommand.cs
+++ b/Src/AutoFixture/CollectionFillerCommand.cs
@@ -1,0 +1,101 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using Ploeh.AutoFixture.Kernel;
+using System.Linq;
+
+namespace Ploeh.AutoFixture
+{
+    /// <summary>
+    /// Contains methods for populating collections with specimens.
+    /// </summary>
+    public class CollectionFillerCommand : ISpecimenCommand
+    {
+        /// <summary>
+        /// Adds many items to a collection.
+        /// </summary>
+        /// <param name="specimen">The collection to which items should be added.</param>
+        /// <param name="context">The context which can be used to resolve other specimens.</param>
+        /// <remarks>
+        /// <para>
+        /// This method mainly exists to support AutoFixture's infrastructure code (particularly
+        /// <see cref="MultipleCustomization" /> and is not intended for use in user code.
+        /// </para>
+        /// </remarks>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="specimen"/> is not an instance of <see cref="ICollection{T}" />.
+        /// </exception>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use the instance method Execute instead.")]
+        public static void AddMany(object specimen, ISpecimenContext context)
+        {
+            new CollectionFillerCommand().Execute(specimen, context);
+        }
+
+        /// <summary>
+        /// Adds many items to a collection.
+        /// </summary>
+        /// <param name="specimen">The collection to which items should be added.</param>
+        /// <param name="context">The context which can be used to resolve other specimens.</param>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="specimen"/> is not an instance of <see cref="ICollection{T}" />.
+        /// </exception>
+        public void Execute(object specimen, ISpecimenContext context)
+        {
+            if (specimen == null)
+                throw new ArgumentNullException(nameof(specimen));
+            if (context == null)
+                throw new ArgumentNullException(nameof(context));
+
+            var typeArguments = GetInheritedCollectionGenericArguments(specimen.GetType());
+            if (typeArguments.Length != 1)
+                throw new ArgumentException("The specimen must be an instance of ICollection<T>.", nameof(specimen));
+
+            if (!typeof(ICollection<>).MakeGenericType(typeArguments).IsAssignableFrom(specimen.GetType()))
+                throw new ArgumentException("The specimen must be an instance of ICollection<T>.", nameof(specimen));
+
+            var filler = (ISpecimenCommand)Activator.CreateInstance(
+                typeof(Filler<>).MakeGenericType(typeArguments));
+            filler.Execute(specimen, context);
+        }
+
+        private static Type[] GetInheritedCollectionGenericArguments(Type type)
+        {
+            var collectionInterfaces =
+                from i in type.GetInterfaces()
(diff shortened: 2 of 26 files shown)


tylerohlsen:
I implemented this change as a replacement for the multiple implementations of collection types to a single implementation that is more generic as it works for all derived collection types.  Alternatively, I may have been able to add some generic relays, but I felt this was a better solution.  I'm happy to take critical feedback.

zvirja:
@tylerohlsen Thanks for your contribution and sorry for the belated reply caused by the [Governance model change](703).

Could you please clarify whether you still have interest and time to work on this PR? That will help to understand whether it makes sense to review it 😉

tylerohlsen:
I do have interest and time to work on the PR.  Please let me know if you have any questions.

zvirja:
I lived couple of days thinking about the suggested changes and found a few aspects which I'd like to change.

- Please don't change the _existing_ behavior and do not fill collection-like types _by default_. That could be a breaking change for the existing user tests. Users could have their own types which implement `ICollection<>`, however doesn't accept any auto-generated specimen (e.g. they should meet a specific conditions). It isn't expected that after you perform a [minor update](http://semver.org/) half of your existing tests become red without a way to quickly fix the situation :)

- Please don't modify any of the existing types (like `CollectionSpecification`, `DictionarySpecification` or `DictionaryFiller`). Such changes are not allowed according to the [SemVer](http://semver.org/) and could break the existing user projects that use them directly. Instead, please create new types with appropriate logic.

- By default AutoFixture uses [a couple of construction ways](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/DefaultRelays.cs#L32), while with your implementation any type implementing `ICollection` will be created via the `ModestConstructorQuery` only. We did that for the known collections like `HashSet<>`, but there it was fine as we knew the types. I'm not sure it's OK to disable other construction methods for unknown types. Notice, we cannot simply copy construction methods from engine as user could have his own engine.

- If user decides to customize its type via `fixture.Customize()` or any other way, the auto-fill behavior will be automatically disabled. It should be considered whether this is an expected behavior or not.

- I'm concerned a bit about performance. Currently we check whether type is generic and if so, we take open type and compare it to a known one (e.g. `List<>`). In the suggested code you take all the implemented interfaces and iterate over them. Given that built-in types implement a lot of interfaces and types could be a part of a huge hierarchy, it might lead to the performance issues. However, we shouldn't apply [premature optimizations](http://wiki.c2.com/?PrematureOptimization) - just think whether we can improve something.

Given the aforementioned points, I basically see a few ways how we could proceed with this. All points assume that we don't touch how we handle the existing collections and keep the existing code as is. We could improve it later via a separate PR. It will be too much for a single PR.

## Way 1 - Create a new _generic_ customization
We introduce a new customization that could be applied to the fixture. That consists of the following steps:

- We have new specifications that check that request type is a writable collection (e.g. `WritableCollectionSpecification`, `WritableDictionarySpecification`).
- We have new fillers that fill collections/dictionaries using the interface methods (like existing `CollectionFiller`, but ensure that naming is consistent for both collections and dictionaries).
- We have a new customization that could be applied to a fixture to enable behavior when _any_ `ICollection/IDictionary` is automatically filled with some items.

When user wants such behavior - he will simply apply the customization. We should think how we can solve the concerns above about construction methods. Also we should ensure that user has a way to opt-out the particular type to not fill it.

## Way 2 - Create a new _generic_ behavior
Unlike cusomization, [behavior](https://github.com/AutoFixture/AutoFixture/wiki/Internal-Architecture#behaviors) could fill a newly created specimen even if it's customized. It might fit better for this feature, however that should be checked.

In this case, you create a behavior that intercepts any request and if this is a writable collection/dictionary, it fills it with values. This should also include some ways to disable this behavior for some particular type.

I'm not sure too much about this option and whether it's better than customization. 


## Way 3 - Create a new _type specific_ customization

It's like the point 1, but we enable this behavior per type only. User should specify which particular type should be filled and we fill only requests of that type. Later, we could extend our glue libraries to have helper attributes for this customization (e.g. like the [Frozen](https://github.com/AutoFixture/AutoFixture/blob/03e9ee8368a0fe7adce3349238b2641ca1567e4c/Src/AutoFixture.xUnit.net2/FrozenAttribute.cs) xUnit/NUnit attribute).

Here we can also consider whether internally we register it as a [customization](https://github.com/AutoFixture/AutoFixture/wiki/Internal-Architecture#customizations) or as a [behavior](https://github.com/AutoFixture/AutoFixture/wiki/Internal-Architecture#behaviors).

----

Personally, I'd see way 3 as the best one. It might be too much to customize any ICollection and here you will have more precise control. Also this way offers the extensibility. When implemented properly, there will be an [IRequestSpecification](https://github.com/AutoFixture/AutoFixture/blob/03e9ee8368a0fe7adce3349238b2641ca1567e4c/Src/AutoFixture/Kernel/IRequestSpecification.cs) parameter somewhere to control which requests should be filled. If you need to fill some particular type only - you can use the [ExactTypeSpecification](https://github.com/AutoFixture/AutoFixture/blob/03e9ee8368a0fe7adce3349238b2641ca1567e4c/Src/AutoFixture/Kernel/ExactTypeSpecification.cs), if more - any other one with wider range.

----

I know that it might be a step-back, however the design of this feature was never discussed. It would be cool to find the most appropriate shape for this feature so it meets all the needs.
If you have any questions or concerns regarding my points - let's discuss that. 

It might be better to describe the design you choose before you start to implement it, so we can agree on it Alternatively, you can try re-implementing the feature and show a new shape (if that's simpler to you), but there is a risk that you will need to change it later.

Thank you for your efforts and hope my reply didn't look too scary 😄

zvirja:
And, as usual, thoughts from other @AutoFixture/core members are welcome 😅

zvirja:
@tylerohlsen Did you have any chance to review my reply? 😉 Let us know if you are busy and currently don't have resources to longer work on this one, so we could decide what to do further.

zvirja:
@tylerohlsen Do you know whether you will any time in the nearby future to review my reply?

tylerohlsen:
I keep on telling myself that I will make time to come back to this, but realistically it will take a while.  If you have someone else that would like to look at it, please do not wait for me.

zvirja:
@tylerohlsen Thanks for the reply! For now I don't have enough capacity to work on this.

Let's keep the PR and see how it goes. If you don't find time to work on this in a month or two, we'll close this one to re-open it later when time is present.

zvirja:
@tylerohlsen It doesn't seem that I have enough capacity for this issue. That would be a cool feature for v4, but I don't have enough resources to work on it for now.

In a week I'm going to [change the default namespace](745) and that would be a HUGE breaking change for this repo. Since that point the `master` branch will be frozen and nothing more will be pushed to the v3. The `master` branch will be un-frozen after the v4 release only.

To make the situation more clear I would like to close this PR, so we clearly see that we don't have any ongoing merges to the `master`. 

I'm again thanking you for all the efforts you made and I'm sorry that I followed up too late 😟 However, that was out of our control.

Could you please _don't remove_ the `issue-700-supportAllDerivedCollections` branch in your fork? Later when we (or somebody else) have resources to work on your issue, it will be possible to use your code as a base. Also would be cool if you could find some resources and follow up on the issue later after we release v4 (so all this release mess is behind) 😉

PR #716, opened 2016-10-30
Ignore abstract methods in GuardClauseAssertion

### LINKED ISSUE #715: GuardClauseAssertion returns false positive when applied to abstract methods
Given the following type:

``` csharp
public abstract class AbstractTypeWithAbstractMethod
{
    public abstract void Method(object arg);
}
```

OOTB, AutoFixture does not create instances of abstract types, so the following test expectedly fails:

``` csharp
[Theory, AutoData]
public void GuardClauseAssertionWithMockedType(GuardClauseAssertion assertion)
{
    // 'AutoFixture was unable to create an instance of type AbstractTypeWithAbstractMethod.'
    assertion.Verify(typeof(AbstractTypeWithAbstractMethod));
}
```

When this type is instantiated using a mocking Glue Library (e.g. NSubstitute), the `GuardClauseAssertion` tries to validate its abstract method and fails:

``` csharp
[Theory, AutoNSubstituteData]
public void GuardClauseAssertionWithMockedType(GuardClauseAssertion assertion)
{
    // Ploeh.AutoFixture.Idioms.GuardClauseException: An attempt was made to assign the 
    // value null to the parameter "arg" of the method "Method", and no Guard Clause 
    // prevented this. Are you missing a Guard Clause?
    // Method Signature: Void Method(System.Object)
    // Parameter Type: System.Object
    // Declaring Type: AbstractTypeWithAbstractMethod
    // ...
    assertion.Verify(typeof(AbstractTypeWithAbstractMethod));
}
```

This fail is not expected since the method is abstract and hasn't got any implementation. The guard should just ignore it.

sergeyshushlyapin:
Fixes #715.

### DIFF
--- a/Src/Idioms/GuardClauseAssertion.cs
+++ b/Src/Idioms/GuardClauseAssertion.cs
@@ -120,7 +120,8 @@ namespace Ploeh.AutoFixture.Idioms
             if (methodInfo.IsEqualsMethod() ||
                 methodInfo.IsGetHashCodeMethod() ||
                 methodInfo.IsToString() ||
-                methodInfo.IsGetType())
+                methodInfo.IsGetType() ||
+                methodInfo.IsAbstract)
                 return;
 
             methodInfo = this.ResolveUnclosedGenericType(methodInfo);
--- a/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
+++ b/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
@@ -1773,5 +1773,19 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
                     throw new ArgumentNullException("arg1");
             }
         }
+
+        [Fact]
+        public void VerifyOnAbstractMethodDoesNotThrow()
+        {
+            var method = typeof(AbstractTypeWithAbstractMethod)
+                .GetMethod("Method");
+            var sut = new GuardClauseAssertion(new Fixture());
+            sut.Verify(method);
+        }
+
+        private abstract class AbstractTypeWithAbstractMethod
+        {
+            public abstract void Method(object arg);
+        }
     }
 }


moodmosaic:
Given the thorough description in #715, the changes proposed in this pull request look reasonable to me.

Src/Idioms/GuardClauseAssertion.cs
```diff
@@ -120,7 +120,8 @@ public override void Verify(MethodInfo methodInfo)
             if (methodInfo.IsEqualsMethod() ||
                 methodInfo.IsGetHashCodeMethod() ||
                 methodInfo.IsToString() ||
-                methodInfo.IsGetType())
+                methodInfo.IsGetType() ||
+                methodInfo.IsAbstract)
```
moodmosaic: 👍

Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
```diff
@@ -1773,5 +1773,19 @@ public NestedGenericParameterTestType(Func<T1, IEnumerable<IEnumerable<T2>>, T1[
...
             }
         }
+
+        [Fact]
+        public void VerifyOnAbstractMethodDoesNotThrow()
+        {
+            var method = typeof(AbstractTypeWithAbstractMethod)
+                .GetMethod("Method");
+            var sut = new GuardClauseAssertion(new Fixture());
+            sut.Verify(method);
```
moodmosaic: 👍

zvirja:
@moodmosaic Are you fine with this PR so it could be merged? Unfortunately, I don't know Idioms at all 😢

zvirja:
@moodmosaic Given that majority agreed to follow the rule of a single approval, do you think it's fine to merge this PR with your approval only now?

moodmosaic:
>Given that majority agreed to follow the rule of a single approval

Well, in reality, only @adamchester responded. It'd be great to have @ecampidoglio's and @klimisa's opinion regarding [that](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-319909069) comment 😃

zvirja:
@adamchester Thanks! 🍺 👍

PR #718, opened 2016-11-01
Idioms: Verify methods with deferred execution

### LINKED ISSUE #717: Idioms: Unable to validate methods with deferred execution
When AutoFixture Idioms meet method like the following:
```csharp
public virtual IEnumerable<object> Method(object arg)
{
    if (arg == null)
        throw new ArgumentNullException(nameof(arg));

    yield return new object();
}
```

... it throws extremely detailed exception message explaining what has happened:
`
Ploeh.AutoFixture.Idioms.GuardClauseException: A Guard Clause test was performed on a method that 
may contain a deferred iterator block, but the test failed. See the inner exception for more details. 
However, because of the deferred nature of the iterator block, this test failure may look like a false 
positive. Perhaps you already have a Guard Clause in place, but in conjunction with the 'yield' keyword (if 
you're using C#); if this is the case, the Guard Clause is dormant, and will first be triggered when a client 
starts looping over the iterator. This doesn't adhere to the Fail Fast principle, so should be addressed.
See e.g. http://codeblog.jonskeet.uk/2008/03/02/c-4-idea-iterator-blocks-and-parameter-checking/ for 
more details.
`

That's good that AF has described the problem. But what is not clear is how to bypass this since I'm trying to validate _all_ the methods on an assembly level and I _do not want_ to perform any changes in the production code as described in the linked post.

IMO, forcing the deferred method execution in AF would be much more _valuable_ solution: 

```csharp
// MethodInvokeCommand.cs
public void Execute(object value)
{
    // Instead if calling 
    // this.method.Invoke(this.expansion.Expand(value))

    // force the method execution casting to IEnumerable and calling .ToList()
    var r = this.method.Invoke(this.expansion.Expand(value))
        as IEnumerable<object>;
    r?.ToList<object>();
}
```

sergeyshushlyapin:
Closes #717.

There are three test cases in the [GuardClassAssertionTest](https://github.com/AutoFixture/AutoFixture/blob/ab829640ed8e02776e4f4730d0e72ab3cc382339/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs#L450) checking that AF throws an exception with message saying that it was unable to verify the guard:
```csharp
[InlineData(typeof(ClassWithDeferredNullGuard))]
[InlineData(typeof(ClassWithDeferredGuidGuard))]
[InlineData(typeof(ClassWithDeferredGuidGuardReturningEnumerator))]
public void VerifyMethodWithDeferredGuardThrowsExceptionWithExtraHelpfulMessage(...)
```

This PR _enables_ guard clause verification for the first two cases.

### DIFF
--- a/Src/Idioms/MethodInvokeCommand.cs
+++ b/Src/Idioms/MethodInvokeCommand.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Globalization;
 using System.Reflection;
 using Ploeh.AutoFixture.Kernel;
@@ -80,7 +81,8 @@ namespace Ploeh.AutoFixture.Idioms
         /// </remarks>
         public void Execute(object value)
         {
-            this.method.Invoke(this.expansion.Expand(value));
+            var result = this.method.Invoke(this.expansion.Expand(value)) as IEnumerable;
+            result?.GetEnumerator().MoveNext();
         }
 
         /// <summary>
--- a/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
+++ b/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
@@ -485,12 +485,11 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
             }
         }
 
-        [Theory]
-        [InlineData(typeof(ClassWithDeferredGuidGuardReturningEnumerator))]
-        public void VerifyMethodWithDeferredGuardThrowsExceptionWithExtraHelpfulMessage(
-            Type type)
+        [Fact]
+        public void VerifyMethodWithDeferredGuardThrowsExceptionWithExtraHelpfulMessage()
         {
             // Fixture setup
+            var type = typeof (ClassWithDeferredGuidGuardReturningEnumerator);
             var sut = new GuardClauseAssertion(new Fixture());
             var method = type.GetMethod("GetValues");
             // Exercise system and verify outcome
--- a/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
+++ b/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
@@ -446,7 +446,6 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
         [Theory]
         [InlineData(typeof(ClassWithDeferredNullGuard))]
         [InlineData(typeof(ClassWithDeferredGuidGuard))]
-        [InlineData(typeof(ClassWithDeferredGuidGuardReturningEnumerator))]
         public void VerifyMethodWithDeferredGuardDoesNotThrow(
             Type type)
         {
@@ -484,6 +483,21 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
                 yield return someGuid;
                 yield return someGuid;
             }
+        }
+
+        [Theory]
+        [InlineData(typeof(ClassWithDeferredGuidGuardReturningEnumerator))]
+        public void VerifyMethodWithDeferredGuardThrowsExceptionWithExtraHelpfulMessage(
+            Type type)
+        {
+            // Fixture setup
+            var sut = new GuardClauseAssertion(new Fixture());
+            var method = type.GetMethod("GetValues");
+            // Exercise system and verify outcome
+            var e =
+                Assert.Throws<GuardClauseException>(() => sut.Verify(method));
+            Assert.Contains("deferred", e.Message);
+            // Teardown
         }
 
         private class ClassWithDeferredGuidGuardReturningEnumerator
@@ -654,7 +668,7 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
         {
             public Dictionary<string, string> GetValues(string someString)
             {
-                return new Dictionary<string, string> 
+                return new Dictionary<string, string>
                 {
                     { "uniqueKey1", someString },
                     { "uniqueKey2", someString }
@@ -990,7 +1004,7 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
                     mockVerification = true;
                 }
             };
-            
+
             var sut = new GuardClauseAssertion(fixture, behaviorExpectation);
             var methodInfo = typeof(DynamicInstanceTestConstraint<>).GetMethod("Method");
             // Exercise system
@@ -1096,7 +1110,7 @@ namespace Ploeh.AutoFixture.IdiomsUnitTest
 
                 return System.Threading.Tasks.Task.Factory.StartNew(() => obj.ToString());
             }
-            
+
             public System.Threading.Tasks.Task TaskWithCorrectGuardClause(object obj)
             {
                 if (obj == null)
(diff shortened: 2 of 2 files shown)


zvirja:
I don't know a lot about this guard, however the exception message looks reasonable to me. Indeed, if your body consists of `yield return`, the guard statements will not play immediately and the Fail Fast principle will be violated.

If we would like to fix that, I'd expect to have something like:
- check whether invocation throws with null. If throws - it's OK.
- if doesn't throw and returns `IEnumerable` - try to iterate.
- if was thrown during the iteration - fail with exception that guard clause is deferred.

This looks like more valid behavior to me. However, I might be missing something, so let's see what @AutoFixture/core think.

zvirja:
@moodmosaic Could you please also reply to my concerns? I want to know what I'm missing to be more aware in future 😉

zvirja:
@sergeyshushlyapin Could you please follow up regarding my and @moodmosaic concerns (about v4)?

sergeyshushlyapin:
I'm a bit curious what is the reason to delay it till v4? This PR adds one more crenario which instead of throwing an error can be successfully verified by Idioms. Is this a breaking change?

zvirja:
@sergeyshushlyapin Kindly asking you for the 3rd time - could you please reply to my concerns?

sergeyshushlyapin:
Sorry, but I don't really understand your concerns. I do not really remeber what is the PR about and have no idea which piece of code you'd like to improve.

The only thing I know is that there was smth not working properly; I added a test to reproduce it and the fix. Coult you please point me to the exact line?

zvirja:
Sure! Read your [original message](https://github.com/AutoFixture/AutoFixture/issues/717#issue-186484602) in the bug report.

This code:

```csharp
public virtual IEnumerable<object> Method(object arg)
{
    if (arg == null)
        throw new ArgumentNullException(nameof(arg));

    yield return new object();
}
```

causes this exception:
`
Ploeh.AutoFixture.Idioms.GuardClauseException: A Guard Clause test was performed on a method that may contain a deferred iterator block, but the test failed. See the inner exception for more details. However, because of the deferred nature of the iterator block, this test failure may look like a false positive. Perhaps you already have a Guard Clause in place, but in conjunction with the 'yield' keyword (if you're using C#); if this is the case, the Guard Clause is dormant, and will first be triggered when a client starts looping over the iterator. This doesn't adhere to the Fail Fast principle, so should be addressed. See e.g. http://codeblog.jonskeet.uk/2008/03/02/c-4-idea-iterator-blocks-and-parameter-checking/ for more details.
`

This exception is meaningful and it makes sense. I do agree with what the message says and the code sample you provided _indeed suffers_ from the issue. 

In this PR you changed the behavior so that code no longer fails. You _changed the existing test_ (`VerifyMethodWithDeferredGuardThrowsExceptionWithExtraHelpfulMessage`), to indicate that you changed the behavior.

I have concerns about this change as I don't agree that behavior of guard should be modified. Instead, you should rewrite your code so something like this:

```csharp
public virtual IEnumerable<object> Method(object arg)
{
    if (arg == null)
        throw new ArgumentNullException(nameof(arg));

    return Do();

    IEnumerable<object> Do()
    {
        yield return new object();
    }
}
```

Such code with work properly and the fail-fast will be followed - it will fail immediately, rather than during the iteration.

Am I missing something?

sergeyshushlyapin:
> Instead, you should rewrite your code so something like this

@zvirja, you suggest not to change AF and rewrite _my production_ code? Unfortunately that won't play nice to me since next time I meet this scenario I will remember nothing about it and will have to spend time figuring out what is wrong with my code or AF.

If you consider this a behavior change, let it be v4 as @moodmosaic [suggested](https://github.com/AutoFixture/AutoFixture/pull/718#pullrequestreview-54788901).

zvirja:
>  you suggest not to change AF and rewrite my production code?

Yes and there is nothing wrong with that. If you missed to verify argument for null and AutoFixture fails, you go and modify the method - not the AutoFixture. The same situation is here - your code doesn't fail fast for null argument - AutoFixture verified that correctly and put that to your desk.

If you are fine with that and don't see that as an issue, you should simply suppress the verification for particular methods as you accepted the issuen. It looks awkward to me that if you have code with issues you tune the verification library rather than your code.

> next time I meet this scenario I will remember nothing about it and will have to spend time figuring out what is wrong with my code or AF.

If that is indeed so, it means that the exception message should be improved. Currently it's verbose enough and describes the potential reason clearly. You can improve it by adding a sample how to fix such issues, so next time you see that exception, you will have a hint what to do.

> If you consider this a behavior change, let it be v4

This is definitely a breaking change as you by fact _remove_ verification for deferred assertions which is useful and it's a _very good catch_  of non-obvious behavior. ATM I don't agree with this change at all as I don't see a reason why issue is with AutoFixture, but not with the code sample you demonstrated.

sergeyshushlyapin:
> The same situation is here - your code doesn't fail fast for null argument - AutoFixture verified that correctly and put that to your desk.

Yep. And, that is not enough for succesfull testing. My code uses defered execution so it does not fail fast. That is for purpose. But, I do want it to be _verified_ no matter what and I DO want AutoFixture to force the validation if necessary.

The reason for that is the following. As soon as a method with deferred execution appears in my codebase, AF starts complaining about that (even if the method has got `null` argumet validation). So the test that checks dozen of cases (e.g. all the methods per assembly) just start failing. There is no easy way to exclude this single failing (potentially false-positive) test case. In order to do that I have to rewrite the following nice-looking test

```
var fixture = new Fixture();
var assemblyToVerity = typeof(MyType).Assembly;
var assertion = new GuardClauseAssertion(fixture);
assertion.Verify(assemblyToVerify);
```

into something dirty like this:

```
var fixture = new Fixture();
var typesToVerify = typeof(MyType)
    .Assembly
    .GetTypes()
    .Where(t => t != typeof(SomeTypeWithDefferedExecution))
    .Where(t => t != typeof(SomeOtherTypeWithDefferedExecution))
    .Where(t => t != typeof(YetAnotherTypeWithDefferedExecution));
var assertion = new GuardClauseAssertion(fixture);
assertion.Verify(typesToVerify);
```

That is why I want AutoFixture to be one step forward performing real validation instead of just warning me with false positives. Does it make sense?

zvirja:
> But, I do want it to be verified no matter what and I DO want AutoFixture to force the validation if necessary.

In this case we should follow a bit different way. Instead of disabling the validation in the Idioms library, we should refactor the `GuardClauseAssertion` class and make the `MethodInvokeCommand` commands creation configurable (e.g. via property injection of factory).

In this case you will be able to adjust the execution behavior in your solution as you wish (e.g. tomorrow you will need to await tasks) without affecting other solutions which want to follow the fail-fast approach.

> That is why I want AutoFixture to be one step forward performing real validation

The existing validation _is not_ a false-positive. The demonstrated code sample _do suffer_ from the issue and the fail-fast principle _is violated_. Therefore, the current validation is "real". If you don't need this guard it doesn't mean that it's useless 😉

---

Given that this AutoFixture is used by a variety of different projects, I feel it ins't OK to remove the functionality that might be required by other people. Rather, we should add handles to customize behavior per your needs.

Does it make sense? 😉

Issue #720, opened 2016-11-10
Generic methods in AutoNSubstitute


pipedreambomb:
Long time user, first time etc. I've cloned this project to see if I can do something about the lack of automocking when it comes to generic methods. For instance, the following code would return null, even when using AutoConfiguredNSubstituteCustomization:

`MyEntity mappedEntityFromDto = new Fixture().Freeze<IMapper>().Map<MyEntity>(myDto)`

If it were a non-generic method, it would return a randomly generated MyEntity, which would negate null reference exceptions and thus make my tests nice and short and able to focus on other things. Instead, all my tests end up with a line like 

`fixture.Freeze<IMapper>().Map<MyEntity>(Arg.Any<MyDto>()).Returns(fixture.Create<MyDto>());`

However, now I've looked into the source, I find a test in `AutoConfiguredFixtureIntegrationTest` called `GenericMethodsAreIgnored` that seems to suggest this isn't a bug, but a feature. I just wondered why that is; is it a technical limitation, a design decision or just left undone and an opportunity for improvement?

It was worked on by @mrinaldi.

mrinaldi:
The way the customization works, it has to setup all the methods before they are used. Since we cannot know the type parameter beforehand, we cannot setup generic methods.
I sent a PR to `NSubstitute` to better support `AutoFixture` and support this scenario, but it was never merged.
However, @Zvirja is trying to get from where I left it: https://github.com/nsubstitute/NSubstitute/pull/259
I didn't review his PR, but I'm hoping it extends `NSubstitute` in such a way that `AutoFixture` can support this scenario.

Zvirja:
Yes, I'm continuing the @mrinaldi's idea and am adding a way to register custom auto-value provider to NSubstitute. This will give us generics for free, because NSubsitute will ask us for values ("passive" approach instead of "active" - how it's now). Therefore a bit of patience and time should solve this issue :)

pipedreambomb:
@Zvirja Exciting! Keep us posted :)

zvirja:
Is finally fixed in v4. We reworked the integration approach and now generic methods are fully supported. Will be able to consume that when we release v4.

@pipedreambomb Would appreciate if you early test that using our [private feed](https://github.com/AutoFixture/AutoFixture#vnext-feed).

Issue #707, opened 2016-09-15
Cannot configure return value for out parameter for NSubstitute mock


zvirja:
Consider the following code sample:

``` csharp
public interface ITestInterface
{
  bool GetResult(string key, out string result);
}

[Fact]
public void ReturnFixedResult()
{
  //arrange
  var fixture = new Fixture().Customize(new AutoConfiguredNSubstituteCustomization());
  var key = fixture.Create("key_");
  var result = fixture.Create("result_");

  var mockFromAF = fixture.Create<ITestInterface>();
  string dummy;
  mockFromAF.GetResult(key, out dummy).Returns(c => { c[1] = result; return true; });

  var mockFromSub = Substitute.For<ITestInterface>();
  string dummy2;
  mockFromSub.GetResult(key, out dummy2).Returns(c => { c[1] = result; return true; });

  //act
  string actualFromAF;
  mockFromAF.GetResult(key, out actualFromAF);

  string actualFromSub;
  mockFromSub.GetResult(key, out actualFromSub);

  //assert
  Assert.Equal(result, actualFromAF); //FAILS. Actual - guid string
  Assert.Equal(result, actualFromSub);
}
```

I configure value for `out` parameter as described in [NSubstitute documentation](http://nsubstitute.github.io/help/setting-out-and-ref-arguments/).
When mock is created by AutoFixture, it doesn't seem to trigger my `Return` callback and rather just substitutes automatically resolved value.
For mocks created by `NSubstitute` directly everything works fine.

Also, everything works fine if I use `AutoNSubstituteCustomization` instead of `AutoConfiguredNSubstituteCustomization`.

Likely, issue happens because NSubstitute integration changes the `out` value before it tries to invoke my custom return callback, so it doesn't match the criteria more (`result` arg is not `null` more).

The following workaround works:

``` csharp
var mockFromAF = fixture.Create<ITestInterface>();
string dummy = Arg.Any<string>();
mockFromAF.GetResult(key, out dummy).Returns(c => { c[1] = result; return true; });
```

@mrinaldi Will it be fixed after you finish work on [this PR](https://github.com/nsubstitute/NSubstitute/pull/234)?

zvirja:
Fixed in v4. If argument is modified during call dispatch (e.g. in `When...Do` block, AutoFixture no longer updates the argument).

Issue #653, opened 2016-06-16
NSubstitute - AutoConfigured properties are resolved via Type requests


zvirja:
**Bug:** When NSubstitute integration resolves value for mock's property, it performs `Type` request rather than `PropertyInfo` request. Because of that some features (like `[Frozen(Matching.PropertyName)]` xUnit attribute) doesn't work properly for substitutes.

Consider the following example (pretty verbose though):

``` csharp
  public class NSubstituteIntegrationTest
  {
    [Fact]
    public void InterfacePropertyShouldBeResolvedCorrectly()
    {
      var fixture = new Fixture();
      fixture.Customize(new AutoConfiguredNSubstituteCustomization());

      //For each Data Property request return etalon
      var expectedData = fixture.Create<string>();
      fixture.Customizations.Insert(0,
        new FilteringSpecimenBuilder(
          new FixedBuilder(expectedData),
          new PropertySpecification(typeof(string), "Data")));


      var classValue = fixture.Create<TestClass>();
      var ifValue = fixture.Create<ITestInterface>();

      //assert
      Assert.NotSame(expectedData, classValue.OtherData);
      Assert.NotSame(expectedData, ifValue.OtherData);

      Assert.Same(expectedData, classValue.Data);
      Assert.Same(expectedData, ifValue.Data); //___this assertion fails___
    }

    public interface ITestInterface
    {
      string Data { get; }
      string OtherData { get; }
    }

    public class TestClass
    {
      public string Data { get; set; }
      public string OtherData { get; set; }
    }
  }
```

A few things to clarify before fixing:  
1. Please confirm that `PropertyInfo` request should be fired to resolve `ITestInterface.Data` property value. I'm not 100% confident that `PropertyInfo` are purposed for this kind of situations (while this looks obvious, though).
2. @mrinaldi I've noticed that you fired [another PR](https://github.com/nsubstitute/NSubstitute/pull/234) which supposes that current integration could be heavily refactored. How do you feel, does it make sense to fix current integration or it's better to wait till "large refactoring"? If you feel that that direction could take a lot of time - I could fix issue for current integration.

mrinaldi:
Indeed, the current integration will be heavily refactored.
However, since we're having a discussion on the changes yet, I'm not sure how long it'll take to have everything in place to refactor the integration.

So, I wouldn't invest time to change the current integration, but since I'm not the one facing the bug, it's up to you. If you feel you need this badly, go for it.

Let me know if I can help you with anything.

[...]

Issue #592, opened 2016-04-04
AutoConfiguredNSubstituteCustomization not working correctly in XUnit theories with parallel tasks


MichaelLogutov:
We've been facting strange issues when running tests that invokes multiple tasks (to simulate multithreaded environment). I found that the issue is related to AutoConfiguredNSubstituteCustomization. Here is the test:

```
using System;
using System.Linq;
using System.Threading;
using System.Threading.Tasks;
using NSubstitute;
using Ploeh.AutoFixture;
using Ploeh.AutoFixture.AutoNSubstitute;
using Xunit;

namespace Bug
{
    public interface IServiceA
    {
        IServiceB GetService();
    }

    public interface IServiceB
    {
    }


    public class Tests
    {
        [Theory]
        [InlineData(32), InlineData(64), InlineData(128), InlineData(256)]
        public Task UsingAutoConfiguredNSubstituteCustomization(int dop)
        {
            // arrange
            var fixture = new Fixture();
            fixture.Customize(new AutoConfiguredNSubstituteCustomization());

            var service = fixture.Freeze<IServiceA>();

            // act, assert
            return TestAsync(dop, service);
        }


        [Theory]
        [InlineData(32), InlineData(64), InlineData(128), InlineData(256)]
        public Task AutoNSubstituteCustomization(int dop)
        {
            // arrange
            var fixture = new Fixture();
            fixture.Customize(new AutoNSubstituteCustomization());

            var service = fixture.Freeze<IServiceA>();

            // act, assert
            return TestAsync(dop, service);
        }


        [Theory]
        [InlineData(32), InlineData(64), InlineData(128), InlineData(256)]
        public Task UsingNSubstitute(int dop)
        {
            // arrange
            var service = Substitute.For<IServiceA>(); // no error

            // act, assert
            return TestAsync(dop, service);
        }


        private static async Task TestAsync(int dop, IServiceA service)
        {
            // arrange
            var start = new SemaphoreSlim(0, dop);
            var cts = new CancellationTokenSource(TimeSpan.FromSeconds(1));

            var tasks = Enumerable
                .Range(0, dop)
                .Select(_ => Task.Run
                                 (async () =>
                                        {
                                            await start.WaitAsync(cts.Token).ConfigureAwait(false);
                                            service.GetService();
                                        },
                                  cts.Token));


            // act
            start.Release(dop);
            await Task.WhenAll(tasks).ConfigureAwait(false);


            // assert
[...]

PR #726, opened 2016-12-22
Rework nsubstitute integration approach

zvirja:
The NSubstitute library [was recently extended](https://github.com/nsubstitute/NSubstitute/pull/259) to provide a graceful API for auto-value providers. This was [quite](https://github.com/nsubstitute/NSubstitute/pull/156) [a long](https://github.com/nsubstitute/NSubstitute/pull/234) way, but I'm glad that it finished. Usage of this API allows to extend NSubstitute in very optimized way and avoid huge limitations which we have now.

This rework introduces the breaking changes to the AutoFixture, therefore I did them in the `v4` branch, where such kind of changes is allowed. These breaking changes basically extend the level of integration rather that are intended to limit the existing logic. Consider the following set of changes:  

1. Generic methods are now fully supported (#720).
1. `ref` and `out` argument are now configured for the void methods as well.
1. The `PropertyInfo` request is used to resolve property values (#653).
1. It is possible now to configure methods with `ref` and `out` parameters: #707 (that is achieved by both NSubsitute and integration changes).
1. Concurrency support is improved and substitution methods could be safely invoked from the multiple threads, even for the first time. There are chances that the issue #592 will be resolved now.

So far no tests failed except ones where I changed behavior intentionally (generics and void methods).

I'm currently consuming Release Candidate because I want to start the discussion as early as possible :) I've already put some efforts to push changes on NSubstitute side, so hope everything will go smoothly here as well :) After we agree on approach, we'll wait till NSubsitute v2.0 release and push changes after that.

I'd like to invite @mrinaldi, @dtchepak and @alexandrnikitin to review my implementation and advice whether they see some concerns regarding my approach.

P.S. In case you already started to review this PR, I've improved the initial implementation a bit mostly to reduce the `AutoFixtureValuesHandler` responsibilities.


dtchepak:
Sorry @Zvirja, I'm not familiar enough with AutoFixture code to be of help here.

Zvirja:
@mrinaldi I know that you were the guy who wrote the initial implementation for the integration. Would you be able to find some time and review the reworked integration to check whether the potential pitfalls exist? 😊

Issue #722, opened 2016-11-24
Range attribute on Enum throws InvalidCastException


Jogge:
I just encountered this issue:

```
public enum MyEnum
{
    First = 1,
    Second = 2,
    Third = 3,
}
```

```
public class MyEntity
{
    [Range(1, 3)]
    public MyEnum MyEnum { get; set; }
}

```
```
public class AutoFixtureVsEnumRangeAttribute
{
    public AutoFixtureVsEnumRangeAttribute()
    {
        var fixture = new Fixture();

        fixture.Create<MyEntity>();
    }
}
```

When creating MyEntity with AutoFixture it throws InvalidCastException and says "Invalid cast from 'Int32' to 'MyEnum'".

cvbarros:
This is because AutoFixture's support for DataAnnotations (in order to make it easier to create valid model instances in tests) is trying to assign a created `Int32` to your property of type `MyEnum`.
So, with this behavior, AutoFixture tries to create the value *based on annotation first*, then on the target property type. The `Range` attribute expects integersm so that's what you get.

In order to disable this behavior, you can customize your fixture by adding a built-in customization such:

```C#
public class AutoFixtureVsEnumRangeAttribute
{
    public AutoFixtureVsEnumRangeAttribute()
    {
        var fixture = new Fixture();
        fixture.Customize(new NoDataAnnotationsCustomization());
        fixture.Create<MyEntity>();
    }
}
```
However, in this case I'd create a custom annotation for allowing only certain values from an enum instead of using a `Range`.

moodmosaic:
You may read [this](https://github.com/AutoFixture/AutoFixture/issues/453#issuecomment-149463476) comment on why you're getting an error for ranges other than `Int` and `Double`.

PR #729, opened 2017-01-03
Woodburysoft/722 range attribute on enum throws exception

woodburysoft:
This is intended to fix the bug raised here https://github.com/AutoFixture/AutoFixture/issues/722. 

My original intention was to create a new Range attribute that took 2 Enum values but when I tried this I had a warning that the project was no longer CLI compliant because Enum was exposed and so I changed course. The solution implemented here uses the standard Range attribute and simply detects if the underlying type is an enum and behaves accordingly. I have added new unit tests where I felt it appropiate.

This is my first pull request to an open source project and while I've read the guidelines and hopefully followed them all I expect there might be some things I've missed.

### DIFF
--- a/Src/AutoFixture.sln
+++ b/Src/AutoFixture.sln
@@ -1,6 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
+MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFixture", "AutoFixture\AutoFixture.csproj", "{400AC174-9A4A-4C7D-815B-F2A21130A0D1}"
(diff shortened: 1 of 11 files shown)


zvirja:
Thanks for your contribution @woodburysoft! Sorry for the belated reply caused by the [Governance Model change](703).

I've briefly reviewed the PR and likely the approach should be re-designed a bit. This PR was created more than half of a year, therefore are you still interested on working on this and do you have time to work on this now? I'm asking to understand whether it makes sense to perform the code review and provide the feedback 😉

zvirja:
@woodburysoft It seems that you don't have free time for now as you kept the silence for about 1.5 months. I do understand that previously we ignored this PR for even longer period of time, however that was out of our control, unfortunately.

I decided to close this PR for now as there is no activity here. Another reason is that the approach should be redesigned, so significant investments are required to proceed with this one. It's unlikely that you have time for that now. However, if later you find time to follow up on this PR, please let us know and re-open the PR.

Thank you for all your efforts and sorry for how it went 😞

PR #740, opened 2017-03-12
Support for .NET Standard

Kralizek:
The target of this PR is to enable the basic usage of AutoFixture in .NET Standard projects, as discussed in the issue #404.
To allow this, I converted the AutoFixture.csproj file to the new .NET Standard class library template.
This introduces few breaking changes:
* Visual Studio 2015 and below are not supported.
* MSBuild 15 is required
* .NET Core SDK 1.0.0 is required

These breaking changes are likely to break the current CI setup.
Few suggested changes:
* Update FAKE to latest version to support MSBuild 15 https://github.com/fsharp/FAKE/issues/1442
* Change AppVeyor build settings to use Visual Studio 2017 image https://github.com/appveyor/ci/issues/1179

Additional notes:
All tests in the Src/AutoFixture.sln solution are correctly ran to completion but are executed only in .NET Framework mode.
I was thinking about converting the test projects as well, but then we would have lost the possibility to test with SYSTEM_NET_EMAIL and SYSTEM_RUNTIME_SERIALIZATION variables set.

### DIFF
--- a/Src/AutoFixture/AutoFixture.csproj
+++ b/Src/AutoFixture/AutoFixture.csproj
@@ -1,80 +1,23 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProductVersion>9.0.30729</ProductVersion>
-    <SchemaVersion>2.0</SchemaVersion>
-    <ProjectGuid>{400AC174-9A4A-4C7D-815B-F2A21130A0D1}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>Ploeh.AutoFixture</RootNamespace>
-    <AssemblyName>Ploeh.AutoFixture</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <FileAlignment>512</FileAlignment>
-    <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>
-    </AssemblyOriginatorKeyFile>
-    <SccProjectName>
-    </SccProjectName>
-    <SccLocalPath>
-    </SccLocalPath>
-    <SccAuxPath>
-    </SccAuxPath>
-    <SccProvider>
-    </SccProvider>
-    <FileUpgradeFlags>
-    </FileUpgradeFlags>
-    <OldToolsVersion>3.5</OldToolsVersion>
-    <UpgradeBackupLocation />
-    <IsWebBootstrapper>false</IsWebBootstrapper>
-    <PublishUrl>publish\</PublishUrl>
-    <Install>true</Install>
-    <InstallFrom>Disk</InstallFrom>
-    <UpdateEnabled>false</UpdateEnabled>
-    <UpdateMode>Foreground</UpdateMode>
-    <UpdateInterval>7</UpdateInterval>
-    <UpdateIntervalUnits>Days</UpdateIntervalUnits>
-    <UpdatePeriodically>false</UpdatePeriodically>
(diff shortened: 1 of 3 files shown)


Kralizek:
knock knock. anybody here?

adamchester:
Hey @Kralizek, thanks for looking into this. It's likely that #703 needs to be resolved first.

maggiepint:
Not wanting to create noise, but I have a very strong interest in this pull request. I'm an OSS maintainer myself, and I have no wish to rush the core team - it's hard out there. That said, if there is any help I can offer, someone feel free to delegate.

zvirja:
@maggiepint We are [on our way](https://github.com/AutoFixture/AutoFixture/issues/703) to transfer the ownership, so let's wait for a bit. After the ownership transfer is finished and we agree on all the details, we will continue our work on v4. We do everything to make it happen as quick as possible.

Src/AutoFixture/AutoFixture.csproj
```diff
@@ -1,80 +1,23 @@
...
-    <UpdateIntervalUnits>Days</UpdateIntervalUnits>
-    <UpdatePeriodically>false</UpdatePeriodically>
-    <UpdateRequired>false</UpdateRequired>
-    <MapFileExtensions>true</MapFileExtensions>
-    <ApplicationRevision>0</ApplicationRevision>
-    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
-    <UseApplicationTrust>false</UseApplicationTrust>
-    <BootstrapperEnabled>true</BootstrapperEnabled>
-    <TargetFrameworkProfile />
+    <TargetFrameworks>netstandard1.5;net45</TargetFrameworks>
```
Kralizek: This could be changed to `<TargetFramework>netstandard1.5</TargetFramework>` but it requires a bit of testing because I've experienced problems with netstandard-only packages.

zvirja: We cannot, as we can NET 45 specific code. If we decide to have 2 projects with different targets it will be even worse to maintain. I believe that NET 45 support is temporal and with NETSTANDARD 2.0 we will not need it. 

I'm currently checking what would be the best way to support all this stuff. Likely, will start a new discussion somewhere. It isn't clear much what to do projects which cannot be targeted to standard because of dependencies.

Kralizek: @zvirja I didn't think about it :)

Kralizek:
Hi everybody. Is there anything I can do to help?

zvirja:
@Kralizek Sorry for the delay, it took some time before we established the release process.

I've played more with .NET Core and found that migration should be tightly coupled with the whole build and deploy process. When we migrate projects to .NET Core, we should ensure that:

- build and CI are configured properly and don't fail
- validation build is taken into account
- tests are run for all the target frameworks
- we produce NuGet packages with all the artifacts

While current PR switches the AutoFixture project to .NET Core, it doesn't makes all the required changes. Another blocker is that I'm in progress of modifying the build scripts, so we will be able to migrate project only after I finish.

Currently I'd push this PR on hold and suggest the following plan:
- wait until I finish all the work on CI. Currently I'm working on enabling the private feed for the `v4` branch on MyGet, so we could consume packages early;
- agree on our [dependency strategy](https://github.com/AutoFixture/AutoFixture/issues/760);
- configure CI to be ready for .NET Core (e.g. change VS version, add target for NuGet restore, etc);
- I'll migrate this project myself after everything is ready using your PR as a base. I'll need to change our build scripts to make it work and I want the vision to be consistent.

I've [already started](https://github.com/AutoFixture/AutoFixture/tree/zvirja/update-projects-to-net-core) the migration, however currently I'm blocked by the issues above. (Notice, [this](https://github.com/AutoFixture/AutoFixture/tree/zvirja/update-projects-to-net-core) branch is just a draft.)

It definitely takes longer than expected, however from my side I'm doing everything to apply changes as quick as possible. It's unlikely we could skip any steps 😞

zvirja:
Sorry, accidentally closed this PR. Let's keep it open for now.

bklooste:
This has been waiting a while now is it possible to publish a beta nuget ? You will probably want a version bump or different package name for standard anyway .

zvirja:
@bklooste I'm working on the private feed (see #762) so we'll get it soon. The other issue is that .NET Core support is not ready yet. I'm currently working on porting, so support should appear soon (at least for the basic `AutoFixture` project).

zvirja:
@Kralizek Thanks for the contribution - I used your changes as a base and it definitely helped me to better understand what to do 👍 
I'm closing this one in favor of #773.

PR #749, opened 2017-04-07
Rework buildscripts for easier management

zvirja:
I've reworked build scripts and CI configuration to simplify release management. Now current package version is calculated automatically based on the last tag and number of commits since tag. Also I've configured both CI and our `build.fsx` to automatically push packages to appropriate NuGet feeds.

The matrix is following (`0` means any digit):

|     Trigger                                    | Result Version  | Action                               |
| ----------------------------------             | --------------- | ------------------------------------ |
| Add tag `v0.0.0` or `v.0.0.0-RC0`              | `0.0.0` or `0.0.0-RC0`     | Build, push to NuGet and MyGet feeds |
| Add tag `v0.0.0-aplha0` (any pre suffix)               | `0.0.0-aplha0`  | Build, push to MyGet feed            |
| Push 2 commits after tag `v0.0.0` to master/v4 | `0.0.0.2`       | Build, push to MyGet feed            |
| Create a PR                                    | -               | Build                                |

Other changes:
- Update FAKE to latest version. Switch to `FAKE.Core` package usage as it takes less space.
- Add tasks to Fake script to calculate Git version, patch assembly version attributes and push to NuGet.
- Add `AssemblyInformationalVersion` to each assembly.
- Move CI configuration to `appveyor.yml`
- Add production keys to `appveyor.yml`
- Recreated project in AppVeyor, because old project didn't work with `appveyor.yml`. Likely, that is because of [this](https://github.com/appveyor/ci/issues/545).

After I merge this PR to `master`, I'll merge it to `v4` and will tune it a bit for .NET Core projects (e.g. change VS image version).

P.S. Large number of changes is caused by FAKE update. The only interesting changes might be `build.fsx`, `appveyor.yml` and `AssemblyInfo.cs` files.


moodmosaic:
>Now current package version is calculated automatically based on the last tag and number of commits since tag.

What does this mean regarding [Semantic Versioning](http://semver.org/)? ([AutoFixture follows Semantic Versioning 2.0.0](https://github.com/AutoFixture/AutoFixture/blob/master/README.md#versioning))

zvirja:
@moodmosaic A good question! :) 

For the Production feed (NuGet) the SemVer is not violated by any means - we push on tags only. Release manager will manage tags according to SemVer.

For the Dev feed (MyGet) SemVer will not be violated for the `v4` branch, as I'm going to push a tag `v4.0.0-alpha` at the beginning of our work (or just hardcode version on CI). 
However it might be violated for the `master` branch after we push commits since last tag. Example: we create a tag `v3.7.1` and merge a PR to master after that. The next artifact on Dev feed will have version e.g. `3.7.1.2`- this violates SemVer as bugfix or a new feature might appear. However I'd say this is fine as the Dev feed is not designed for the public usage. Rather, it's a feed where you can find the latest artifacts and it will be used in rare cases only. In this case you are aware that you are using something internal. 

It would be very redundant to track SemVer for the Dev feed, so we 2 options only - disable Dev feed for the Master branch or have it knowning that this is a private feed with latest artifacts. I'd vote for the last option as such a feed might be very useful sometimes

zvirja:
I've just rebased my code to remove doc for FAKE from repo - it takes about 10MB :)

Build.fsx
```diff
@@ -173,6 +173,7 @@ Target "PublishNuGetRelease" (fun _ -> publishPackagesToNuGet
                                             (getBuildParam "NuGetReleaseKey"))
 
 Target "CompleteBuild" (fun _ -> ())
+Target "PublishNuGetAll" (fun _ -> ())
```
ecampidoglio: How are the changes in this file related to the commit message? Shouldn't this be part of another commit or even be committed separately?

Build.fsx
```diff
@@ -19,7 +19,9 @@ let build target configuration =
     let properties = [ "Configuration", configuration ] @ keyFile
 
     solutionsToBuild
-    |> MSBuild "" target properties
+    |> Seq.iter (fun s -> build (fun p -> { p with Verbosity = Some(Minimal)
```
ecampidoglio: Why set the verbosity to minimal? Don't we want to have as much information as possible when troubleshooting?

Build.fsx
```diff
@@ -1,15 +1,43 @@
...
+open System.Text.RegularExpressions
 
 let releaseFolder = "Release"
 let nunitToolsFolder = "Packages/NUnit.Runners.2.6.2/tools"
 let nuGetOutputFolder = "NuGetPackages"
 let solutionsToBuild = !! "Src/*.sln"
 let signKeyPath = FullName "Src/AutoFixture.snk"
 
+type GitVersion = { apiVersion:string; nugetVersion:string; }
+let getGitVersion =
```
ecampidoglio: I like the idea of letting the version number be calculated based on the history. 👍
However, I strongly suggest we use an established tool like [GitVersion](https://github.com/GitTools/GitVersion) for that instead of rolling our own.

Build.fsx
```diff
@@ -13,7 +13,7 @@ let nuGetOutputFolder = "NuGetPackages"
 let solutionsToBuild = !! "Src/*.sln"
 let signKeyPath = FullName "Src/AutoFixture.snk"
 
-type GitVersion = { apiVersion:string; nugetVersion:string; }
+type GitVersion = { apiVersion:string; nugetVersion:string }
```
ecampidoglio: I think this whole commit should be squashed into 6e6dad9 as it's unclear what it does on its own. What's the "_optimization_" here?

Build.fsx
```diff
@@ -17,7 +17,7 @@ type GitVersion = { apiVersion:string; nugetVersion:string }
 let getGitVersion = 
     let desc = Git.CommandHelper.runSimpleGitCommand "" "describe --tags --long --match=v*"
     // Example for regular: v3.50.2-288-g64fd5c5b, for prerelease: v3.50.2-alpha1-288-g64fd5c5b
-    let result = Regex.Match(desc, @"^v(?<maj>\d+)\.(?<min>\d+)\.(?<rev>\d+)(?<pre>-(alpha|beta|rc)\d*)?-(?<num>\d+)-g(?<sha>[a-z0-9]+)$", RegexOptions.IgnoreCase).Groups
+    let result = Regex.Match(desc, @"^v(?<maj>\d+)\.(?<min>\d+)\.(?<rev>\d+)(?<pre>-\w+\d*)?-(?<num>\d+)-g(?<sha>[a-z0-9]+)$", RegexOptions.IgnoreCase).Groups
```
ecampidoglio: Again, it would be better if we didn't have to support all these use cases on our own. See my [previous comment](https://github.com/AutoFixture/AutoFixture/pull/749/commits/6e6dad941de349d9c538379d9d09b79cddbdb38d#r110601832).

Build.fsx
```diff
@@ -1,4 +1,4 @@
-﻿#r @"packages/FAKE.4.19.0/tools/FakeLib.dll"
+﻿#r @"packages/FAKE.Core/tools/FakeLib.dll"
```
ecampidoglio: It's hard to get an overview of the changes related to updating FAKE when there are another 150K+ deleted lines. Could you please commit the deletion of the old package separately?

ecampidoglio:
@zvirja Thank you for your contribution! I left a few comments in my review.

Also, I noticed that the [build on AppVeyor](https://ci.appveyor.com/project/AutoFixture/autofixture) is currently broken on `master`. 😢 I understand that you needed to modify the configuration in order to test these changes, but I really think we should get the build up and running as quickly as possible while this PR is being reviewed.

Could you please make a new PR with a new `appveyor.yml` file that reproduces the old process? Of course, if you don't have time, I can also do it myself.

Build.fsx
zvirja: Do you want me to perform the rebase and just separate add/remove as separate commits, or want me to rebase and remove this from PR at all, so it will be fired via separate PR later?

ecampidoglio: I think a separate commit for deleting the old package (and its related documentation) is enough. No need for a new PR. 😉

Build.fsx
zvirja: I did that intentionally, as it was impossible to track review the build log on CI - browser hung after the pressure of number of build entries.
However, later (in migration to .NET Core branch I'm currently working on) I found that we perform a lot of redundant rebuilds, so probably we could revert this setting back. Will update soon.

zvirja:
I do understand your concern regarding the build failing for the master, however there was no other way. I needed to remove the previous AppVeyor project, so currently it relies on `appveyor.yml`. This file never existed before, therefore now master is broken.

Do you believe that it makes sense to create a temp `appveyor.yml`? I expect this PR will be merged soon, so everything will become fixed. We don't have too much activities on AF project now, so it might be acceptable to wait for a couple of days. Let me know if you disagree with that and still persist on creation of a temp `appveyor.yml` via separate PR... :)

ecampidoglio:
@zvirja 
> I expect this PR will be merged soon

In OSS, "_soon_" could literally mean anything 😉

I still think we should get the build running on `master` in the meantime. We might have to fix something in the released version, you never know. Plus, it's [unhealthy](https://pragprog.com/the-pragmatic-programmer/extracts/software-entropy) for the project to have a broken build for any extended period of time.

Build.fsx
zvirja: Done.

Build.fsx
zvirja: Done - squashed with previous commit.

Build.fsx
zvirja: Done - keep normal verbosity

Build.fsx
zvirja: Done - squashed with previous commit

Build.fsx
zvirja: See reply below. Let's close this thread ;)

Build.fsx
zvirja: I do like idea of switching to GitVersion tool, FAKE even have helpers for it, however I don't have experience with that tool. It would take some time for me to read the documentation.

Currently I'm focused on migration to .NET Core, so would like to have this PR merged as soon as possible.
Would it be OK for you if merge this PR as is and switch to `GitVersion` in `v4` branch later?

zvirja:
@ecampidoglio I've followed up all the points above and fixed them (except the GitVersion one). Thank you for such a brilliant code review.

Let's follow this way. If PR is not approved by you and {@moodmosaic or @adamchester} till evening, I'll create this file and add it directly to master - it's pretty simple.

zvirja:
@ecampidoglio I noticed that you liked the reply, but haven't replied [here](https://github.com/AutoFixture/AutoFixture/pull/749#discussion_r110618072) and haven't approved the PR. Does that mean that you don't agree? 😏

zvirja:
@ecampidoglio I've [pushed](https://github.com/AutoFixture/AutoFixture/blob/master/appveyor.yml) legacy configuration, so now our CI is [green](https://ci.appveyor.com/project/AutoFixture/autofixture/branch/master).

I've also spent few hours playing with GitVersion tool and I hate it 😠 While I find it awesome in some aspects, there is one thing which blows my head up. Tool simply doesn't work property with pre-release tags ending with numbers. Consider the following flow:
1. You release `RC1` and add a tag `v4.0.0-rc1`. Tool version: `v4.0.0-rc0001`. I understand that padding, because NuGet doesn't support SemVer 2.0 properly.
2. You add 2 more commits. Tool version: `v4.0.0-rc0002`. WHAT?
3. You add 2 more commits and create a tag `v4.0.0-rc2`. What do you guess is the tool version? It's `4.0.0-rc0002`!

The issue is that SemVer tool simply trims numbers at the end of your pre-release tag. Meaning, that you cannot release `rc1` and `rc2` builds, as you might end up with same version for absolutely different releases.

If you tweak configuration a bit and play with `ContinuousDelivery`, `ContinuousDeployment` and branch specific configuration, you might observe different set of behavior (e.g. it simply ignores commits after the tag assigning them the same NuGet version). But none of them is what I expect to see.

I'm too silly to make it work fine 😢 Therefore, if you could find free time, please rework approach to use this tool instead of the regular expressions. In the meanwhile, I'd like to ask to approve as is.

moodmosaic:
I've been viewing this item on-and-off the past few days and I have to admit that I'm having a hard time to understand what it does exactly 😕 Though, as long as @ecampidoglio is reviewing this, I feel safe 😃

But maybe we should consider splitting this in smaller pull requests?

zvirja:
@moodmosaic Thank you for the time you spent on this. Likely, the most confusion is caused by the FAKE update which brought a huge number of diffs.

Could you clarify which exact parts caused the confusion to you? We could split it to a separate PRs if that is needed, but I need to understand which exact commits you want to group.

moodmosaic:
@zvirja, I just can't understand what happens next if/when this pull request gets merged 😕 (I'm not saying that the pull request is bad, just that I can't understand exactly what it does.)

ecampidoglio:
From what I gather, this PR does a few different things:

- Updates to a new version of FAKE
- Switches to using the `Fake.Core` package
- Adds automatic versioning based on the state of the history
- Adds deployment to NuGet for stable releases
- Adds deployment to MyGet for _prerelease_ releases

They all go under the umbrella of automating the release process, but I have to say that multiple PRs for each of those items would be easier to manage.

moodmosaic:
@ecampidoglio, @zvirja, if you agree, based on [this](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293185149) and [this](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293185842), it will be faster then to get this done if it gets split into smaller, separate pull requests.

zvirja:
@ecampidoglio appeared to be quicker than me 😉 
Agreed! I'll split that to 2 PRs - Update of the FAKE and everything else. Are you fine with that? I don't see a reason to make them smaller..

moodmosaic:
@zvirja, probably it's just me, but I see each item on [this](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293185149) list by @ecampidoglio as a separate pull request.

ecampidoglio:
Actually, I would prefer to have separate PRs for each of the items that I listed [here](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293185149), assuming I'm not missing anything.

The reason is that we might not want to have _everything_. For example, while the automatic versioning might be good, we might want to hold off on the MyGet feed.

zvirja:
Thanks for the feedback. Likely, that is how review system works in OSS projects - you need to make small granularity changes.

I could rework that to the small sub-sets, the are couple of issues with that:
1. All these changes are iterative, meaning that I will need to create PRs one by one. That will take a lot of time, before all them are finally merged. It doesn't play well with our plan to release `v4` in the nearest future :)
2. Do you really want to throw something out? If yes - it would make sense to do that. If not - it would be just a process for a process. I do understand that it was my mistake to mix all these changes together and I promise to not do that in future. But this will require me to rework everything just to make granular changes. I've spent almost 1.5 weeks trying to find the working solution for that it it's sad for me to rework everything just for a good history, while the current history is not too bad.

However, by submitting that I'm ready to do that. If you really need that, feel that it would be better for everybody and we cannot accept 2 PRs with FAKE update and everything else - I'll do that! I appreciate the time you spend on this, I want to collaborate and to not make a chaos in our project 😉

ecampidoglio:
@zvirja Thank you for understanding. Your efforts are _very appreciated_ 😃

In the future, it's better to discuss your proposed changes with the team first, _before_ you do any actual work (with the exception of a small POC, if necessary). That way, you'll avoid investing a huge chunk of your time doing something that might not align with the vision the rest of the team has for the project.

The preferred way to do that is to create an issue and have the initial discussion there. Once we agree on the course of action, you can submit one or more PRs in reference to that issue.

Again, I want to empathize that I really appreciate your enthusiasm and your will to do good! We just need to dial it back a little so that we don't rock the boat too hard 😉

zvirja:
@ecampidoglio Thanks for the appreciation 😌 

Ok, let's follow this way. I've already created the [first PR](https://github.com/AutoFixture/AutoFixture/pull/751) with FAKE update and will create more after it will be approved.

I close this one as it will not be merged.

PR #751, opened 2017-04-11
Update the FAKE package

zvirja:
PART 1 of [Rework buildscripts for easier management](https://github.com/AutoFixture/AutoFixture/pull/749).

Update the `FAKE` package as current version doesn't work with MSBuild 15 we'll need in future.
Also I switched to the `FAKE.Core` package instead as it's smaller and contains everything we need.


update-fake.bat
```diff
@@ -0,0 +1,3 @@
+cls
+Packages\NuGet.CommandLine.2.8.6\tools\NuGet.exe install "FAKE" -OutputDirectory "Packages" -ExcludeVersion
```
ecampidoglio: Where is this file used? Do we need it?

zvirja: Next time you need to update FAKE - you will just run this file instead of Googling parameters for invocation.
Other projects (e.g. [NSubstitute](https://github.com/nsubstitute/NSubstitute)) have this file as well.

zvirja:
@ecampidoglio and @moodmosaic Do you have any more concerns regarding the FAKE update? If no - can I just merge this change? I have 3 more already developed PR waiting for you :)

moodmosaic:
@zvirja, great work 👍 💯

zvirja:
@ecampidoglio Thanks for your approval and time! 

May I also ask you to put comments to the latest version of code if possible? Sometimes you leave comments to the lines that were changed in the subsequent commits and that might be a bit confusing:
1. Lines might be obsolete and not actual more. 
2. It's easier to miss comments and replies to them, as they are hidden under the "Outdated" spoiler. 
2. It's inconvenient a bit to expand spoilers for each time to read comment and replies. 
3. Usually outdated comments might be an indicator that you changed code since the last review. Here this is not the case as they are initially outdated. 

It's just a recommendation, so, of course, do whatever is much convenient for you during the reviews 😉 I decided to left this feedback just in case it doesn't too much matter for you and is not an intentional decision ;-) Thank you.

PR #753, opened 2017-04-17
Calculate build version based on repo status

zvirja:
This PR is intended to simplify the release management of the project. I've implemented version calculation based on the state of the repository. Last tag and number of commits since last tag is used to determine the current version.

## Main changes:

- [x] Add code to calculate current version using the [git describe](https://git-scm.com/docs/git-describe) command output as a base.
- [x] Add build target to automatically patch all the `AssemblyInfo.[cs,fs]` files before the build.
- [x] Use the [AssemblyInformationalVersion](https://msdn.microsoft.com/en-us/library/system.reflection.assemblyinformationalversionattribute(v=vs.110).aspx) attribute to store the NuGet version of the particular DLL file. As a part of this change I've modified all the `AssemblyInfo` files to reset default version to `1.0.0.0` and to add the missing attribute (to be able to update it's value during the build).

## Version resolve logic
As it was stated above, the `git describe` command output is used to calculate current version. The logic depends on the last/current tag name, whether it's a release or a pre-release tag and number of commits since last tag. The result is following:

| Repo state            | Calculated version | Explanation |
--------------------- | --------------------| ---------------- |
| on tag `v3.52.0` | `3.52.0` | If we are on a tag, we use it's version without any changes. |
| on tag `v3.52.0-beta2` | `3.52.0-beta2` | If we are on a tag, we use it's version without any changes. |
| 15 commits since tag `v3.52.2` | `3.52.3-pre0015` |  We patch version to have a new version that is larger than `3.52.2`. We know nothing about changes, so we speculatively increase the `revision` number. We also add a `-preXXXX` pre-release suffix to both indicate that this version is a pre-release and to distinguish between different commits on the same branch. Number padding is required because nuget doesn't support SemVer 2.0 and uses lexical comparison for numbers. |
| 15 commits since tag `v3.52.2-beta2` | `3.52.2-beta2-build0015` | We simply append the `-buildXXXX` suffix to produce a version that is larger than `3.52.2-beta2`. NuGet uses the lexical comparison for the pre-release part, so the produced version is larger. |

Pay attention that I intentionally use the different suffixes in last 2 examples. That is because versions have completely different semantics:
- the version `3.52.3` was never released before, so we use the `-pre` suffix.
- the version `3.52.0-beta2` was already released and current version it's just something that contains more changes. Therefore, I use the `-build` suffix.

If you feel this logic too much confusing, let's find the better approach.

__Note!__ Versions from the last 2 table rows are never supposed to be published to the release (nuget.org) feed. They could be potentially published to the pre-release feed.

Also I know about the [GitVersion](https://github.com/GitTools/GitVersion) tool, however didn't use it intentionally. That is because it [doesn't work well](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293077239) with pre-release tags ending with numbers (e.g. `-rc1`, `-beta2`, etc) while I want to have such a possibility for such versions.

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -76,7 +76,10 @@ let build target configuration =
         | "" -> []
         | x  -> [ "AssemblyOriginatorKeyFile", FullName x ]
 
-    let properties = [ "Configuration", configuration ] @ keyFile
+    let properties = keyFile @ [ "Configuration", configuration
+                                 "AssemblyVersion", buildVersion.assemblyVersion
+                                 "FileVersion", buildVersion.fileVersion
+                                 "InformationalVersion", buildVersion.infoVersion ]
 
     solutionsToBuild
     |> MSBuild "" target properties
--- a/Build.fsx
+++ b/Build.fsx
@@ -55,6 +55,20 @@ let buildVersion = match getBuildParamOrDefault "BuildVersion" "git" with
 
 
 
+Target "PatchAssemblyVersions" (fun _ ->
+    printfn 
+        "Patching assembly versions. Assembly version: %s, File version: %s, Informational version: %s" 
+        buildVersion.assemblyVersion
+        buildVersion.fileVersion
+        buildVersion.infoVersion
+
+    let filesToPatch = !! "Src/*/Properties/AssemblyInfo.*"
+    ReplaceAssemblyInfoVersionsBulk filesToPatch 
+                                    (fun f -> { f with AssemblyVersion              = buildVersion.assemblyVersion
+                                                       AssemblyFileVersion          = buildVersion.fileVersion
+                                                       AssemblyInformationalVersion = buildVersion.infoVersion })
+)
+
 let build target configuration =
     let keyFile =
         match getBuildParam "signkey" with
@@ -191,8 +205,9 @@ Target "CompleteBuild" (fun _ -> ())
 "CleanReleaseFolder" ==> "Verify"
 "CleanAll"           ==> "Verify"
 
-"Verify"    ==> "Build"
-"BuildOnly" ==> "Build"
+"Verify"                ==> "Build"
+"PatchAssemblyVersions" ==> "Build"
+"BuildOnly"             ==> "Build"
 
 "Build"    ==> "Test"
 "TestOnly" ==> "Test"
--- a/Build.fsx
+++ b/Build.fsx
@@ -3,6 +3,7 @@
 open Fake
 open Fake.Testing
 open System
+open System.Diagnostics;
 open System.Text.RegularExpressions
 
 let releaseFolder = "Release"
@@ -184,10 +185,7 @@ Target "CleanNuGetPackages" (fun _ ->
 )
 
 Target "NuGetPack" (fun _ ->
-    let version = "Src/AutoFixture/bin/Release/Ploeh.AutoFixture.dll"
-                  |> GetAssemblyVersion
-                  |> (fun v -> sprintf "%i.%i.%i" v.Major v.Minor v.Build)
-
+    let version = buildVersion.nugetVersion
     let nuSpecFiles = !! "NuGet/*.nuspec"
 
     nuSpecFiles
--- a/Build.fsx
+++ b/Build.fsx
@@ -3,12 +3,57 @@
 open Fake
 open Fake.Testing
 open System
+open System.Text.RegularExpressions
 
 let releaseFolder = "Release"
 let nunitToolsFolder = "Packages/NUnit.Runners.2.6.2/tools"
 let nuGetOutputFolder = "NuGetPackages"
 let solutionsToBuild = !! "Src/*.sln"
 let processorArchitecture = environVar "PROCESSOR_ARCHITECTURE"
+
+type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string }
+let calculateVersionFromGit buildNumber =
+    // Example of output for a release tag: v3.50.2-288-g64fd5c5b, for a prerelease tag: v3.50.2-alpha1-288-g64fd5c5b
+    let desc = Git.CommandHelper.runSimpleGitCommand "" "describe --tags --long --match=v*"
+
+    let result = Regex.Match(desc,
+                             @"^v(?<maj>\d+)\.(?<min>\d+)\.(?<rev>\d+)(?<pre>-\w+\d*)?-(?<num>\d+)-g(?<sha>[a-z0-9]+)$",
+                             RegexOptions.IgnoreCase)
+                      .Groups
+    let getMatch (name:string) = result.[name].Value
+
+    let major, minor, revision, preReleaseSuffix, commitsNum, sha =
+        getMatch "maj" |> int, getMatch "min" |> int, getMatch "rev" |> int, getMatch "pre", getMatch "num" |> int, getMatch "sha"
+
+    
+    let assemblyVersion = sprintf "%d.%d.%d.0" major minor revision
+    let fileVersion = sprintf "%d.%d.%d.%d" major minor revision buildNumber
+    
+    // If number of commits since last tag is greater than zero, we append another identifier with number of commits.
+    // The produced version is larger than the last tag version.
+    // If we are on a tag, we use version specified modification.
+    // Examples of output: 3.50.2.1, 3.50.2.215, 3.50.1-rc1.3, 3.50.1-rc3.35
+    let nugetVersion = match commitsNum with
+                       | 0 -> sprintf "%d.%d.%d%s" major minor revision preReleaseSuffix
+                       | _ -> sprintf "%d.%d.%d%s.%d" major minor revision preReleaseSuffix commitsNum
+
+    let infoVersion = match commitsNum with
+                      | 0 -> nugetVersion
+                      | _ -> sprintf "%s-%s" nugetVersion sha
+
+    { assemblyVersion=assemblyVersion; fileVersion=fileVersion; infoVersion=infoVersion; nugetVersion=nugetVersion }
+
+// Calculate version that should be used for the build. Define globally as data might be required by multiple targets.
+// Please never name the build parameter with version as "Version" - it might be consumed by the MSBuild, override 
+// the defined properties and break some tasks (e.g. NuGet restore).
+let buildVersion = match getBuildParamOrDefault "BuildVersion" "git" with
+                   | "git"       -> calculateVersionFromGit (getBuildParamOrDefault "BuildNumber" "0" |> int)
+                   | assemblyVer -> { assemblyVersion = assemblyVer
+                                      fileVersion = getBuildParamOrDefault "BuildFileVersion" assemblyVer
+                                      infoVersion = getBuildParamOrDefault "BuildInfoVersion" assemblyVer
+                                      nugetVersion = getBuildParamOrDefault "BuildNugetVersion" assemblyVer }
+
+
 
 let build target configuration =
     let keyFile =
--- a/Src/AutoFakeItEasy/Properties/AssemblyInfo.cs
+++ b/Src/AutoFakeItEasy/Properties/AssemblyInfo.cs
@@ -33,8 +33,9 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
 
 [assembly: CLSCompliant(true)]
 [assembly: NeutralResourcesLanguage("en")]
--- a/README.md
+++ b/README.md
@@ -95,6 +95,17 @@ AutoFixture is available via NuGet:
 
 AutoFixture follows [Semantic Versioning 2.0.0](http://semver.org/spec/v2.0.0.html).
 
+## Build
+
+AutoFixture uses [FAKE](http://fsharp.github.io/FAKE/) as a build engine. If you would like to build the AutoFixture locally, run the `Build.cmd` file and wait for the result.
+
+The repository state (the last tag name and number of commits since the last tag) is used to determine the build version. If you would like to override the auto-generated AutoFixture version, pass the `BuildVersion` parameter to the `Build.cmd` file. For example:
+```
+Build.cmd BuildVersion=3.52.0
+```
+
+Refer to the [Build.fsx](Build.fsx) file to get information about all the supported build keys.
+
 ## Documentation ##
 
 * [CheatSheet](https://github.com/AutoFixture/AutoFixture/wiki/Cheat-Sheet)
--- a/Src/AutoFakeItEasy2.UnitTest/Properties/AssemblyInfo.cs
+++ b/Src/AutoFakeItEasy2.UnitTest/Properties/AssemblyInfo.cs
@@ -31,5 +31,6 @@ using System.Runtime.InteropServices;
 //
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
--- a/Src/AutoFakeItEasy2/Properties/AssemblyInfo.cs
+++ b/Src/AutoFakeItEasy2/Properties/AssemblyInfo.cs
@@ -33,8 +33,9 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
 
 [assembly: CLSCompliant(true)]
 [assembly: NeutralResourcesLanguage("en")]
--- a/Src/AutoFakeItEasyUnitTest/Properties/AssemblyInfo.cs
+++ b/Src/AutoFakeItEasyUnitTest/Properties/AssemblyInfo.cs
@@ -31,5 +31,6 @@ using System.Runtime.InteropServices;
 //
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
--- a/Src/AutoFixture.NUnit2.Addins.UnitTest/Properties/AssemblyInfo.cs
+++ b/Src/AutoFixture.NUnit2.Addins.UnitTest/Properties/AssemblyInfo.cs
@@ -32,5 +32,6 @@ using System.Runtime.InteropServices;
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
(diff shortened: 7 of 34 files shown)


Build.fsx
```diff
@@ -3,20 +3,75 @@
...
+    let result = Regex.Match(desc, @"^v(?<maj>\d+)\.(?<min>\d+)\.(?<rev>\d+)(?<pre>-\w+\d*)?-(?<num>\d+)-g(?<sha>[a-z0-9]+)$", RegexOptions.IgnoreCase).Groups
+    let getMatch (name:string) = result.[name].Value
+
+    let major, minor, revision, preReleaseSuffix, commitsNum, sha =
+        getMatch "maj" |> int, getMatch "min" |> int, getMatch "rev" |> int, getMatch "pre", getMatch "num" |> int, getMatch "sha"
+
+    
+    let assemblyVersion = sprintf "%d.%d.%d.0" major minor revision
+    let fileVersion = sprintf "%d.%d.%d.%d" major minor revision buildNumber
+
```
moodmosaic: It'd be nice to keep these lines under 120 characters.

From [CONTRIBUTING.md](https://github.com/AutoFixture/AutoFixture/blob/master/CONTRIBUTING.md)
>Please keep line lengths under 120 characters. Line lengths over 120 characters don't fit into the standard GitHub code listing window, so it requires vertical scrolling to review.

Src/AutoFakeItEasy/Properties/AssemblyInfo.cs
```diff
@@ -33,8 +33,9 @@
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("3.50.2.0")]
-[assembly: AssemblyFileVersion("3.50.2.0")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyInformationalVersion("1.0.0.0")]
```
moodmosaic: Hmm... Wouldn't this be a little confusing to those trying to compile AutoFixture directly from source code?

zvirja: I believe we don't have too much options here as we will not update the committed `AssemblyInfo` files on each update more. Rather, we temporarily update files immediately before the build and don't expect these changes to be committed. Therefore, there is no way to overcome the issue when you compile your project directly - version will be non-actual in any case.

If we don't reset committed version to `1.0.0.0`, it will lead to even larger confusion. Instead of having the `1.0.0.0` that could indicate that you just have a default version, you will always get `3.50.2` - the latest version which was released using the old release approach.

I'd say that it should not be an issue. We just need to put a notice somewhere in the wiki that you should use FAKE to get assemblies of the desired version.

Build.fsx
zvirja: I'll try to trim lines to 120 characters whenever that is possible. However there are places when that would be hard (e.g. line with regular expression will look less readable if splitted into multiple lines).

zvirja:
> Overall, this looks great. AFAICT, we keep doing Semantic Versioning, but also offer 'nightly' builds as well.

Exactly! :)

However I'd say that nightly builds is not the only goal I pursue. I also want to simplify the release procedure to a simple tag creation. At the end you should be able to simply draft a release from the GitHub Releases UI, publish it and a new version will be released to feeds.

Build.fsx
zvirja: Fixed by rebasing all the code and changing indents in places.

Src/AutoFakeItEasy/Properties/AssemblyInfo.cs
moodmosaic: >Rather, we temporarily update files immediately before the build and don't expect these changes to be committed.

Okay, I think that's fine (by me), since the 'normal' way to build AutoFixture is by running its build script.

moodmosaic: >I'd say that confusion should not be an issue. We just need to put a notice somewhere in the wiki

Or maybe on the README.md. My only concern with the wiki is that it's not part of the Git repository, it's a _GitHub_ thing (I could be wrong though).

zvirja: I've updated the README file (see [here](https://github.com/AutoFixture/AutoFixture/blob/zvirja/calculate-build-version-based-on-repo-status/README.md#build)) and added information about build. Please review and let me know if you see how to improve the message :)

moodmosaic: Yes, it looks fine. 👍

Build.fsx
```diff
@@ -183,9 +184,7 @@ Target "CleanNuGetPackages" (fun _ ->
 )
 
 Target "NuGetPack" (fun _ ->
-    let version = "Src/AutoFixture/bin/Release/Ploeh.AutoFixture.dll"
-                  |> GetAssemblyVersion
-                  |> (fun v -> sprintf "%i.%i.%i" v.Major v.Minor v.Build)
+    let version = FileVersionInfo.GetVersionInfo("Src/AutoFixture/bin/Release/Ploeh.AutoFixture.dll").ProductVersion
```
zvirja: My initial thoughts were to decouple build from packaging, so that for the NuGet pack task knows nothing about how version is resolved - we simply take version from the DLL.

However, given that NuGet pack target has dependency on the build, it could be never triggered separately, it might make sense to use global variable instead.

zvirja:
@ecampidoglio Thanks for the reply and description of your schema! :sparkles:

## SemVer 2.0

I was under impression that NuGet doesn't work with SemVer 2.0 properly - I've read a set of conflicting articles. However, it seems that it works. After the more detailed testing I found the following:

- NuGet 3.5+ - works
- VS 2015 - _doesn't_ work
- VS 2017 - works
- NuGet.org - doesn't allow to publish SemVer 2.0 packages
- MyGet.org - allows to publish SemVer 2.0 packages

That means if we decide to use SemVer 2.0 schema, we'll get the following situation:
- We are not able to use SemVer 2.0 for public builds (NuGet.org feed) and 1.0 is the only option (e.g. `3.50.2`, `4.0.0-rc2`). This is fine as we don't need pre-release builds there.
- If we decide to use SemVer 2.0 for the 'nightly' builds, only users with VS 2017 (or e.g. [Rider](https://www.jetbrains.com/rider/)) will be able to consume these builds.

It might be fine, given that we don't expect our 'nightly' builds to be used widely.. But should be explicitly agreed whether we have a good reason for such a limitation.

## Build in the version
The build section (`+build.XXX`) isn't an option as it does't participate in the package precedence. It isn't intended to sort assemblies, which is required here. If you read the "Build Metadata" section [here](http://blog.nuget.org/20140924/supporting-semver-2.0.0.html) you will see that this is definitely not the behavior we would like to get.

Rather, I'd like to the the following precedence of the packages:
- Last Tag
- Last Tag + 10 commits
- Last Tag + 20 commits
- Next Tag

If I add a pre-release feed to my project and a new package is pushed to the feed, I want to get a nuget notification about that. I __don't__ want NuGet to randomly pick any of the builds.

## Pre-release feed
In addition to the release feed, I would like to have a pre-release feed. That feed would include the following packages:
- All the release packages (mirror them).
- Packages for each commit to the `master`
- Packages for each commit to the `vNext` (v4 in our case).

The reasons why I want to have `master` on the pre-release feed in addition to the `vNext` are following:
- there is a delay between PR merge and the release itself. Feed could enable early adoption for urgent cases;
- it might be required to perform some testing before tagging a new release (e.g. you spotted something after PR merge);
- it costs nothing to have it, so we could cover other potential scenarios.

This will be discussed in the different PR, added here just for a reference :)

## Answering to your questions
> I don't see the need to further distinguish between pre and build versions; if we want to generate a new version number for every commit, the number of commits since the last tag should be more than enough.

That's exactly what I want to suggest - a mechanism based on number of commits since the last tag. However, there are requirements that this version should be somewhere here: 

`[previous release] < VERSION <[next release]`. 

Unfortunately, we cannot always simply append the `-buildXXXX` suffix to `v3.50.0`, because that will violate our needs (`v3.50.0` > `v.3.50.0-buildXXXX`). 
If you see the better way to achieve the requirements given all the restrictions we have - please suggest 🙏 

> For example:  4.0.0-beta1  4.0.0-beta1+1

As was stated above, it will not work, because for SemVer (and NuGet) `4.0.0-beta1` and `4.0.0-beta1+1` are the same versions.

## Version generation
Given all the information above (build metadata doesn't cover our needs; VS 2015 doesn't work with SemVer 2.0) do you have some other ideas how to alter the approach? 😉

Build.fsx
zvirja: @ecampidoglio Thank you for this suggestion. I've improved code a bit, so now:
1. Use calculated NuGet version directly from the variable, rather than from assembly attribute.
2. AssemblyInformationalVersion attribute was changed to have the following format: `NuGetVersion-CommitSHA`. That would be very useful for the "nightly" builds as you will be able to quickly find the corresponding base commit. If we are on the tag now, SHA code isn't appended (because you can find commit by tag).

Let me know whether you agree with this change.

ecampidoglio:
I should probably clarify the premises of my suggestion:

1. We release _deliberately_ (i.e. not automatically on every commit)
2. We release by creating a tag with the version number

For the past few years of AutoFixture development, we've been following the principles of [Continuous Delivery](https://martinfowler.com/bliki/ContinuousDelivery.html); instead of bundling a bunch of improvements into a single release, we've released a feature or a bug fix as soon as it met our standards. Each release was tagged with a version number following the rules of [Semantic Versioning](http://semver.org).

I was under the impression that this PR was about automating our release process; in reality, it seems to be moving towards a higher goal: rework the way we do releases.

Don't get me wrong, I don't think it's a bad thing. However, we should all be clear about what we're doing here.

Now, if the majority of @AutoFixture/core team thinks we should be publishing CI builds on [MyGet](https://www.myget.org) or do nightly builds, that's fine by me. In that case, I think we should come up with a better versioning scheme than simply appending `-preXYZ` or `-buildXYZ` because, frankly, they don't tell much about what will be in the package. Maybe including the branch name would be more descriptive.

If, on the other hand, we continue to do releases the way we've been doing them so far, appending the commit count since the last tag as build metadata is enough to generate a unique version number for the _build_; packages get their version number from the release tags.

zvirja:
It seems that we have a bit of misunderstanding here. Sorry if I'm being too chatty and confuse you by that.
> I was under the impression that this PR was about automating our release process; in reality, it seems to be moving towards a higher goal: rework the way we do releases.

Simple automation is exactly what I want to do - I want us to continue following the same way as before. I don't to change the approach because I like it. I want the `master` branch to be publishable at any time and I want us to have quick release cycles. I still expect that ideally we have release after each PR merge. I __don't expect__ to have larger release cycles when we roll out a bulk of changes.
The changes of this PR are intended, as you mentioned above, to automate the process - simple tag creation should be enough to trigger the release process.

For the people you use the public feed only nothing will change and will be the same as it was before - all these auto-generated suffixes will be invisible (if we are on a tag, we simply use tag version).

### Pre-release feed
I want us to have a pre-release feed and the main goal is enable the early adoption of the `vNext` branch. We could also want the `master` branch to be published to the pre-release feed. The reason is that there _could be_ potential delays between the PR merge and subsequent release. Delays might be caused _not_ because of the policy change, but simply because of the human factor. As was stated in the Governance model discussion, nobody from the Core team wants to care about releases, while you still are ready to review and approve the PRs. As result, it could happen that you accept PR, merge it to master and I'll release only a few days after. That could create a gap between PR merges and public releases. If we have pre-release feed, everybody (if that is urgently required or e.g. just for testing) could go to our MyGet and consume packages from there in the meanwhile. However, as it was stated above, nightly builds for the `master` branch are optional and up to discuss later.

### Auto-generated versions
If we want to have pre-release feed (and we want at least for the vNext), we need a mechanism to automatically generate NuGet versions which meet the criteria I described [above](https://github.com/AutoFixture/AutoFixture/pull/753#issuecomment-294850359): `[lastTag] < [lastTag + 1 commit] < [lastTag + 15 commits] < [nextTag]`. This should work well if `lastTag` is both release and pre-release (e.g. `v3.50.1` and `v3.50.1-beta2`). The only purpose of these auto-generated versions is to indicate that we are "N" commits after the last tag. They cannot carry any semantics as it's impossible.

If you don't like the schema I suggested above, let's try another one:

| Repo state| Result version | Comment |
| ------------| --------------- | ----------- |
| on tag `v3.50.1` | 3.50.1 | don't change |
| on tag `v3.50.1-beta2` | 3.50.1-beta2 | don't change |
| 2 commits since tag `v3.50.1` | 3.50.1.2 | append the patch block. While this doesn't meet SemVer (it doesn't contain the 4th number), NuGet works fine with such versions. |
| 2 commits since tag `v3.50.1-beta2` | v3.50.1-beta2.2 | append another pre-release section. This is SemVer 2.0, so will not work in VS 2015 |

This schema doesn't contain any confusing suffixes and states clearly that it's just the previous version with appended commits. The drawback is that it might not work in VS 2015.

> Maybe including the branch name would be more descriptive.

Honestly, I don't see a reason to include it - that would be always `master` or `vNext` only. It's clearly visible from the version on which branch we are (e.g. `3.x` - `master`, `4.x` - `vNext`).

> If, on the other hand, we continue to do releases the way we've been doing them so far, appending the commit count since the last tag as build metadata is enough to generate a unique version number for the build; packages get their version number from the release tags.

I'm not sure I understood you. Do you mean release or the pre-release packages? For the release packages we don't need to include number of commits at all - version corresponds to the particular tag. For the pre-release packages (the MyGet feed) we should have unique package versions.

-----

If you feel that we still speak about different things, please write small areas and clarify which exact questions/points you want to clarify.

ecampidoglio:
> I want us to have a pre-release feed and the main goal is enable an early adoption of the `vNext` branch. We could also want the `master` branch to be published to the pre-release feed.

I don't know about you, but I see this as a change in the release process: 😉

- **Before:** tag = release
- **After:** tag = release _and also_ potentially release any commit from `master` or `vNext` to a pre-release feed

What I'm saying is that if we _only_ ever release with tags—to which we _manually_ assign a version number—we don't need to generate NuGet-compatible version numbers for every commit. Hence, appending the commit count as build metadata is enough.

I'm also skeptical to the idea of maintaining a separate _pre-release_ feed. What does that buy us? If we want to have pre-release versions from `vNext` couldn't we just tag them appropriately (e.g. `4.0.0-beta3`) and release them to the regular NuGet feed?

> The reason is that there could be potential delays between the PR merge and subsequent release. Delays might be caused not because of the policy change, but simply because of the human factor. As was stated in the Governance model discussion, nobody from the Core team wants to care about releases, while you still are ready to review and approve the PRs. As result, it could happen that you accept PR, merge it to master and I'll release only a few days after.

Frankly, I don't see this happening. If the release process is automated and can be triggered by simply pushing a tag, I don't think it will take days between a merged PR and a release. Do you?

Also, I personally never said that I _don't care_ about our releases; by now it should be obvious that I do, in fact, care a great deal 😉 That's why I welcome the idea of automating the release process. However, I do question the need for the added complexity that comes from having multiple feeds and long package versions.

moodmosaic:
>*and also* potentially release any commit from `master` or `vNext` to a pre-release feed

That seems like overkill, actually.

Why would anyone want to upgrade to some latest 'unofficial' version just to try commit(s) that might not even compile, or not being even approved? 😕

zvirja:
> Why would anyone want to upgrade to some latest 'unofficial' version just to try commit(s) that might not even compile, or not being even approved? 😕

Well, `master` and `vNext` branches contain merged (and therefore approved) PRs only. The pre-release feed will contain something that has been already approved and will be for sure a part of the product, but hasn't been officially released yet.

moodmosaic:
So, essentially it means that every night, everything that has been approved and merged into `master` or `v4` branch will end up in a "nightly build".

zvirja:
@moodmosaic I expected that they will me built not every night, but just immediately after the PR was merged. Therefore, that are not the "nightly" builds per se - just the pre-release packages.

zvirja:
@ecampidoglio Here is my reply :)

> I don't know about you, but I see this as a change in the release process...

Thanks, now I see what exactly you consider as a change. I didn't considered that as a change, because from perspective of a regular NuGet user nothing changed. But yes, internally, we are going do add one more feed.

> I'm also skeptical to the idea of maintaining a separate pre-release feed. What does that buy us? If we want to have pre-release versions from vNext couldn't we just tag them appropriately (e.g. 4.0.0-beta3) and release them to the regular NuGet feed?

The difference is that release cycle for the pre-release feed is usually much larger. That might be caused e.g. by incompleteness of some features. For instance, we could migrate to .NET Core the AutoFixture project only, while the rest projects doesn't support the .NET Core yet. I would never release that as a public `beta` - product isn't finished enough to be published.
Also, usually I saw not too much pre-release versions (e.g. I never saw something like `v4.0.0-alpha95`) on public feeds, meaning that they are be released rarely enough. Given that I'd vote to have pre-release feed for `vNext` at least. If anybody wants to migrate particular project to .NET Core (e.g. xunit), they could do that via PR, wait till it's approved and consume that immediately in their project. Only after we find that it's stable enough, we could publish a public RC.

My point is that I see value from it, while it costs almost nothing.

> Frankly, I don't see this happening. If the release process is automated and can be triggered by simply pushing a tag, I don't think it will take days between a merged PR and a release. Do you?

It seems that we should agree on that. I thought that the Release Manager person is responsible for releases and he is the only one who will trigger the release procedure. It could lead to confusion if everybody will trigger releases. And, therefore, if we have single person (or group of persons) who is responsible for releases, delays could potentially happen.

> Also, I personally never said that I don't care about our releases; by now it should be obvious that I do, in fact, care a great deal 

Sorry, that was definitely not what I wanted to say 😟 Of course I believe that everybody cares 😉  

-----

Next things to agree - are we going to support pre-release feed for `vNext` at least? I vote to have that - it will enable really early adoption, which is what for need, especially for .NET Core. I don't like idea to release 200 beta versions to official NuGet feed or wait for mounts till you could consume your change :worried:

zvirja:
Ok, let's agree to __not have__ the pre-release feed for the `master` branch for now as it's release cycle is pretty short.  Let's __have__ it for the `vNext` branch only. The reason is described above - `vNext` release cycle is much larger (we accumulate a set of changes before we release the next RC). It will be really handy here and it costs nothing for us to have this feed.

@ecampidoglio, @moodmosaic Do you have any objections regarding this?

ecampidoglio:
I think having a separate feed for the `vNext` branch would make sense, since it lets us publish _very early_ builds without polluting the main feed on NuGet.

As long as we release these pre-release packages _deliberately_ with _tags_ (as opposed to pushing a package on every build), I'm up for it.

zvirja:
@ecampidoglio Hard times for common agreements :tired_face: You discard the idea of the pre-release feed and push on each commit for the `vNext`, but I don't understand the reason.

Usually, the schema is following:
1. NuGet - release builds.
2. NuGet - public `alpha`, `beta`, `rc`, whatever.
3. Private feed - latest builds, in our case for the `vNext` branch only, as release cycle __is not too frequent__.

We cannot agree on the 3rd point. I'd say that this is the exact purpose of the __private__ feed - to have latest builds _early_. This feed isn't expected to be used by everybody - it's intended to be used only by people who know what this is for. This is very convenient mechanism to distribute the latest releasable version of product when release cycle is large (what is exactly the case for vNext).

Could you please provide exact reasons why you don't agree to have this feed? I do understand that this is the change and something we didn't have before, however this should not be a reason per se. All this situation is really weird as:
1. Feed is free, we pay _nothing_ to have it.
2. It doesn't affect the public feed and is transparent.
3. This is a private feed. If people don't need these packages - they will simple don't use the feed.
4. This feed is useful and convenient in particular scenarios, even if they happen not on the everyday basis.

ecampidoglio:
I think we have another misunderstanding. I thought your suggestion was to not publish _any_ pre-release packages to NuGet and instead _only_ make them available through the private feed.

If we're going to have _deliberate_ pre-release packages in the NuGet feed (tagged as _beta_, _rc_ and what not), then, sure, let's have a private feed with the CI builds from the `vNext` branch. 👍

I hope this clarifies things.

zvirja:
> If we're going to have deliberate pre-release packages in the NuGet feed (tagged as beta, rc and what not), then, sure, let's have a private feed with the CI builds from the vNext branch. 👍

Agreed 😌 Exactly - push tagged builds (pre- and release) to the NuGet, push latest vNext builds to the private feed.

-------

Now, let's get back to this PR and agree on the auto-generated versions schema for the private feed.
It seems that you don't like the original `-pre` and `-build` suffixes for the private feed as they are a bit confusing. How about this one (I suggested it [a few answers before](https://github.com/AutoFixture/AutoFixture/pull/753#issuecomment-295174386)):

| Repo state| Result version | Comment |
| ------------| --------------- | ----------- |
| on tag `v3.50.1` | 3.50.1 | don't change |
| on tag `v3.50.1-beta2` | 3.50.1-beta2 | don't change |
| 2 commits since tag `v3.50.1` | 3.50.1.2 | append the Revision part with commits number. While this doesn't meet SemVer (it doesn't contain the 4th number), NuGet works fine with such versions. |
| 2 commits since tag `v3.50.1-beta2` | v3.50.1-beta2.2 | append another pre-release section. This is SemVer 2.0, so will not work in VS 2015 |

This schema doesn't contain any confusing suffixes and states clearly that it's just the previous version with appended commits. The drawback is that it might not work in VS 2015.

Notice, versions from the last 2 rows will be published to the _private_ feed only (as they are not tagged).

@ecampidoglio Do you agree with this one? If so - I'll tune the build.fsx file appropriately.

moodmosaic:
>@ecampidoglio, @moodmosaic Do you have any objections regarding this?

I agree with what @ecampidoglio commented [here](https://github.com/AutoFixture/AutoFixture/pull/753#issuecomment-295891287).

zvirja:
@moodmosaic Thanks for answer, so we agreed to a private feed with automatic builds :sunglasses:

@ecampidoglio I've updated my code to use the versioning schema described in the [previous reply](https://github.com/AutoFixture/AutoFixture/pull/753#issuecomment-295906592). Please review the code and let me know  if you have more comments.

Thanks, will be looking to your follow up 😉

zvirja:
@ecampidoglio Did you have any luck to review the latest changes? 

If you don't have time now and given that I fixed your initial concerns regarding the cumbersome suffixes, could we proceed further and merge the PR? We need a couple more steps to re-enable the releases and I would like to finish that as soon as possible..

Issue #746, opened 2017-04-02
Should we keep our sign key secret?


zvirja:
Previously @ploeh kept the signing key for all the assemblies in secret. I'd like to discuss whether it indeed makes sense.

If we check other popular testing helpers, we'll find that all them has opened public key:
1. xunit - [key is open](https://github.com/xunit/xunit/blob/master/src/signing.snk).
1. nunit - [key is open](https://github.com/nunit/nunit/blob/master/src/nunit.snk).
1. NSubstitute - [key is open](https://github.com/nsubstitute/NSubstitute/blob/master/Source/nsubstitute.snk).
1. Moq - [key is open](https://github.com/moq/moq4/blob/master/Moq.snk).
1. FakeItEasy - [key is open](https://github.com/FakeItEasy/FakeItEasy/blob/master/FakeItEasy.snk).
1. FluentAssertions - [key is open](https://github.com/fluentassertions/fluentassertions/blob/develop/Src/FluentAssertions.snk).

I don't see too much sense to keep AF key closed and believe we can also add it to the repository. Indeed, signing allows to prove the assembly's identity, but I don't believe we strictly require that. Our days NuGet is used to distribute packages, rather than third-party sites where the identity indeed does matter.

If we publish key to the repository, that will have a lot of benefits:
1. It will be easier to generate NuGet with signed assemblies by CI.
1. It will be easier for people to make own AutoFixture assemblies and test whether it solves particular issue. Consider the following scenario:

    1. You found some bug in AF. You download source code locally and fix it.
    2. Now you have AF assemblies with fixed bug and you want to test whether it helped.

    Here is where the issues start. If you assemblies are unsigned, you cannot simply put the modified assemblies to the local NuGet cache - project will not compile. Rather, you need to update all your projects to target to a new assembly - just for a simple test. The things become even worse when that is the core `AutoFixture` assembly, as other depending libraries (like `AutoFixture.xunit` or `AutoFixture.NSubstitute`) will need to be updated as well.

    If we ship signing key issue will be solved.
1. You don't need to store one more secret :)

As for the @ploeh, it seems he [didn't have](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-280425693) strong reason to keep it secret.

Personally, I'd vote using both hands to open the key and simplify everybody's life.
Guys (@moodmosaic, @adamchester, @ecampidoglio, @klimisa) - what do you think about this?

adamralph:
👍 to open the key.

[...]

PR #755, opened 2017-04-26
Add automatic deployment support

zvirja:
This PR is intended to simplify the release procedure by triggering the automatic NuGet push on tag creation event. It will allow to trigger release by simple drafting a release in GitHub.

PR brings the following changes:
- add NuGet push tasks to the FAKE build script. Tasks enable publish to both public (nuget.org) and private (myget.org) feeds;
- extend the `appveyor.yml` config to trigger build and push based on different events;
- publish the sign key to the repository as it was agreed in #746.

There was added a rule to not publish the `alpha` versions to the _public_ feed. It was done intentionally and reasons are following:
- alpha versions are not stable [by definition](https://en.wikipedia.org/wiki/Software_release_life_cycle), so don't suite well for public consumption;
- we need a way to start vNext versioning in the vNext branch without making any public releases, so `alpha` will be reserved for such a purpose.

All other tagged versions (both release and pre-release) are published to the public feed.
Push to the private feed is currently commented out and will be enabled later for the `v4` branch.

P.S. Some lines in the `appveyor.yml` file are longer than 120 chars - I was unable to wrap them to not break the PS syntax.


appveyor.yml
```diff
@@ -1,11 +1,65 @@
 image: Visual Studio 2015
 
+environment:
+  NUGET_API_KEY:
+    secure: Rh/UHYn7HGj8a7LFWo9C1rSA0sHSQNV3wIqj9bvwRR8yLmCXu9JmbNkmkctk2cBn
```
moodmosaic: I'm very confused looking at this 😕

zvirja: What exactly confused you? :) 

This is the way how appveyor allows you to store sensitive information without worrying that somebody will steal it. These lines are decrypted by the build system when script runs (except for PRs). You can read more about them [here](https://www.appveyor.com/docs/build-configuration/#secure-variables).

We need to store NuGet key here to support auto-push to feeds.

moodmosaic: >These lines are decrypted by the build system when script runs

Oh, OK. I didn't know about this 👍

zvirja: No worries 😉 Let me know if you have another concerns before approval.

moodmosaic: For pull requests similar to this one, related to build automation, I think @ecampidoglio and @adamchester have a much stronger opinion than me (?)

zvirja: Yes, let's wait for the @ecampidoglio reply. However, it seems that @adamchester is totally absent, so I'm not sure how much it could take to wait for his review... 

@adamchester Are you here? :)

appveyor.yml
```diff
@@ -1,11 +1,65 @@
...
+  MYGET_API_KEY:
+    secure: hA4Ut1N2lrrdEtAN24Bty/FNiU0d/Ur/dLYSqpr8jKHOvoO7MU4jD+KwzUvATh+E
+
 pull_requests:
   do_not_increment_build_number: true
 
 build_script:
 - ps: |
-    & .\build.cmd CompleteBuild
+    # A trick to test if variable is defined. Number could never be null.
```
ecampidoglio: There's _a lot_ going on here. 😟 Couldn't we move this logic into the build script itself? I'm not fond of the idea of having the build process split across multiple files (much less using different languages).

appveyor.yml
```diff
@@ -1,11 +1,65 @@
...
+    $branchName = $Env:APPVEYOR_REPO_BRANCH
+
+    Write-Host "[BUILD CONTEXT INFO] Is PR: $isPullRequest, is tag: $isTagBuild, branch name: '$branchName'" -ForegroundColor Green
+
+    if($isTagBuild)
+    {
+      $tagName = $Env:APPVEYOR_REPO_TAG_NAME
+      $isSemVerTag = $tagName -match '^v.*'
+      # We consider version as alpha if it contains the "-alpha" pre-release tag.
+      $isAlphaVerTag = $tagName -match '^v[\d\.]+-alpha[\d\.]*$'
```
ecampidoglio: Do we really have to create a tag in order to publish to the private feed? I'm afraid that is going to leave us with _a lot_ of tags over time. 😟 Can't we simply publish all builds from the `vNext` branch to the private feed, instead? The branch name could be included as _pre-release_ tag in the version number to differentiate the packages.

appveyor.yml
```diff
@@ -1,11 +1,65 @@
...
+      }
+    }
+    # Should be checked _before_ branch name, because branch name is set to "base" branch for Pull Requests
+    elseif($isPullRequest -eq $true)
+    {
+      Write-Host "[PR] Trigger verification build only" -ForegroundColor Green
+      & .\build.cmd CompleteBuild BuildVersion=git BuildNumber=$($Env:APPVEYOR_BUILD_NUMBER)
+    }
+    # Push to private feed on commit - so package could be accessed before it's tagged.
+    #elseif($branchName -eq "master")
```
ecampidoglio: What's this commented code doing here? Should it be removed?

appveyor.yml
```diff
@@ -1,11 +1,65 @@
...
+      $isSemVerTag = $tagName -match '^v.*'
+      # We consider version as alpha if it contains the "-alpha" pre-release tag.
+      $isAlphaVerTag = $tagName -match '^v[\d\.]+-alpha[\d\.]*$'
+
+      if(-not $isSemVerTag)
+      {
+        Write-Host "[Tag] Trigger build only - this is not a SemVer tag. Tag name: $tagName" -ForegroundColor Green
+        & .\build.cmd CompleteBuild BuildVersion=git BuildNumber=$($Env:APPVEYOR_BUILD_NUMBER)
+      }
+      # Alpha version is published to the private feed only as tag's purpose is to start the vNext versions count.
```
ecampidoglio: I have a feeling that publishing all CI builds from the `vNext` branch to the private feed would simplify things quite a bit.

appveyor.yml
zvirja: I don't think that this is a really good idea to move this code to the FAKE build script and reasons are following:
- This code is very AppVeyor specific. It knows a lot about environment variables and specifics of how they are filled according to different events. I don't want our FAKE script be bound to the particular CI.

- Current approach allows to split the responsibilities. The FAKE is responsible for doing the job which you ask it to do. That is the simple command (like `CompleteBuild` or `PublishNuGetPublic`) and required parameters. On other hand, AppVeyor knows about the different triggers for the build, so it's logic is to pick up the most appropriate task for that.

- It might require FAKE file redesign, as it doesn't expect that one target will _dynamically_ invoke other ones. Rather, it's supposed that you define dependencies between targets statically. If we have "CI" target we don't know its dependencies statically. Rather, it will match the current state and will need to invoke another target dynamically. I'm not sure that FAKE was even designed to handle such usage scenario and it will be possible to accomplish such a task gracefully.

- If we move all the logic to FAKE, it would be another target like "CI" with 10 parameters as input. There is no chance you are able to invoke it correctly locally, because number of parameters will be large. Also you will spend some efforts capturing the context and parsing it back in FAKE (I'm more about coding, not about performance).

- If you check the code, it's just a HUGE number of if/else statements. It will be a bit cumbersome to repeat that in FSharp and it might look less readable.

Given that I'd vote to keep it as is - CI knows about different trigger and decides which action should be applied, while FAKE is responsible for doing the actual work. Also code in AppVeyor doesn't look too complex - just a set of different if/else statments and logic to run the appropriate target.

appveyor.yml
zvirja: I could redesign it and alter the version generation logic to use branch name as a pre-release tag, but only if we have a really good reason for that.

[Previously](https://github.com/AutoFixture/AutoFixture/pull/753) we agreed that we will push the `vNext` branch to the private feed only. I don't think that single `alpha` tag for each `vNext` is actually too much (e.g. `v4.0.0-alpha`, `v5.0.0-alpha`, etc). Even if we have 5 more major versions which will take a couple of years, we will have 5 more tags. Given that we generate a tag per each release, you will not even notice these tags among 164 tags (!!) we currently have for `v3`.

Also I would like to keep version generation logic as simple as possible, so it's easily to understand what is going on. If I start to mix branch name, it will complicate the logic too much.

Do you still see that single `alpha` tag for each `vNext` is too much?

appveyor.yml
zvirja: It could happen that we decide to have Release Candidates, which are still on `vNext` branch, however should be published to the public feed. 
I want our configuration to allow such things happen. If don't need it - we will simply not create tags on vNext branch (except the `alpha` ones), however if we need that - we have that for free.

appveyor.yml
zvirja: I left this place intentionally, as this is the place where you will enable private builds for the vNext branch.
The reason I kept it is the following:
- it's much easier to enable early builds for `vNext` - just uncomment out a few lines, instead of writing the boilerplate code again and again;
- logic should be located in this particular place and it will lead to mistakes if we place it in another place. If we keep it here, there are less chances that you will make mistake in future;
- this boilerplate code already contains all the necessary parameters you should pass to the FAKE to run trigger the task. I don't remove it, later you will need to create these lines again.

Given that I'm going to consume these lines in `v4` branch after PR is merged (if it even will be...), I believe it's OK to keep them.

zvirja:
@ecampidoglio Thanks for the code review. I've replied you in the comments.

While I agree that it became more complex than it was before (when we didn't have any logic at all), I don't agree that it's unnecessarily over-complicated and should be definitely simplified :anguished: I spent a lot of time testing the different approaches and looking for the ways to make changes as compact as possible. The changes I applied are required to achieve the level of automation we need and to handle gracefully both release and vNext activities.

I don't see what else could be stripped from the code without removing the feature we need. However, will be happy to see your advises - probably I'm just blindly missing something :confused:

appveyor.yml
zvirja: I've updated my reply above for a few times, so ensure to pick that changes :point_up:

zvirja: I've tried to refactor the build script for better readability. You can find result [here](https://github.com/AutoFixture/AutoFixture/blob/zvirja/alternative-ci-configuration/appveyor.yml). If you like it - I'll put this implementation to current PR. It looks less complex, while it's still verbose.

In any case, this logic should be a part of CI and `appveyor.yml` file, as I stated above.

zvirja: I've updated my reply above for a few times, so ensure to pick that changes :point_up:

zvirja:
@ecampidoglio I've reworked the AppVeyor file since your last review, to please check it for one more time. Now it should be much simpler to understand it's logic and figure out what is going on.

I've updated my replies to all our comments.

appveyor.yml
adamchester: I'm still watching, but I haven't been able to completely keep up with issues here in the last weeks 😢

zvirja: @adamchester I could create an issue with specs and describe the whole feature I'm trying to implement. Will you be interested in code review and traking things up then? Do you have currently time for that? :)

adamchester: I haven't been able to keep up with the issues here due to lack of time. I'll still be watching, and I'll try to "chime in" where I can, but I'm not going to have much time available in the coming weeks 😢

ecampidoglio:
@zvirja Thank you for taking the time to refactor your script, it's definitely more readable now.

I still don't like the idea of having to create a tag in order to publish something to the private feed. I think tags are reserved for _releases_ while the private feed should contain the packages created by the _CI builds_ for the `vNext` branch; in other words, every build on the `vNext` branch gets published to the private feed and official releases go the public feed.

Also, I am still of the opinion that the `appveyor.yml` file should only contain a call to the FAKE script with the appropriate target; the script itself should then handle both the build and the deployment of the packages. I don't see a problem with having the script reference some AppVeyor-specific environment variables; after all, the script represents the build and deployment pipeline so it's OK if it's aware of the CI environment, if that's necessary.

On a separate note, I completely understand if you find this back and forth frustrating. However, _this_ is the design and planning conversation we should have had _before_ you started implementing anything. As I [wrote you before](https://github.com/AutoFixture/AutoFixture/pull/749#issuecomment-293193114), it's important to get buy-in from the @AutoFixture/core team before doing anything and this topic in particular—the release process—requires special attention being so central to the process.

All that is to say that this is going to take some time to get right and shouldn't be rushed. Another alternative could be to just automate our _current_ process (that is publish tagged commits to the public feed) so we get something out the door and _then_ we can see how we can improve it.

What do you think?

zvirja:
@ecampidoglio The reason I'm starting discussion in place rather than beforehand is that for me it looked like we agreed on everything in the previous PR. It seems that we just have misunderstanding on implementation, because I see that you describe things a bit differently to how they are implemented.

> I still don't like the idea of having to create a tag in order to publish something to the private feed.

We __don't__ need to create a tag to push to private feed. The design is that we will _automatically_ publish on _each commit_ to the `vNext` branch. The tag creation is required __not to push the package__, but to start the `4.0.0` versioning. It's a really small trick and it _single_ alpha tag per each major doesn't look like something really horrible. It allows simplify other things a lot.

> I don't see a problem with having the script reference some AppVeyor-specific environment variables;

I still don't agree with this idea and don't like it very much. I like how it's now, when AppVeyor specific stuff is located on the AppVeyor side and it's the only place which knows about AppVeyor and triggers. I don't see any valuable reason to mix the responsibilities and make things unnecessary complicated just because we want to keep everything in single place. `appveyor.yml` file will not be empty in any case, so you still should be aware of it's content and depend on it.
Also, I as I mentioned above, FAKE is designed for _static_ target dependencies, rather than dynamic ones. If you move implementation to FAKE, it might look really ugly just because of how the FAKE is designed.

>  Another alternative could be to just automate our current process (that is publish tagged commits to the public feed) so we get something out the door and then we can see how we can improve it.

Let's follow this way. I'll strip all the pre-release specific logic and implement it later on the `v4` branch. It will be much easier to see how exactly it's handled.

zvirja:
@ecampidoglio I've stripped all the private feed related logic and currently it's contains publish to the public feed only.
Do you have more concerns regarding the implementation?

ecampidoglio:
> It's a really small trick and it single alpha tag per each major doesn't look like something really horrible. It allows simplify other things a lot.

I understand. But since the version number is already generated in the build script, wouldn't having the deployment process be part of the same script simplify things even further? That way, we wouldn't have to rely on tags to share that information between AppVeyor and the build script.

> I don't see any valuable reason to mix the responsibilities and make things unnecessary complicated just because we want to keep everything in single place.

The way I see it, the responsibilities we're talking about here are:

1. Implement the build and deployment pipeline.
2. Start the pipeline from the CI environment.

The FAKE script does 1) while the `appveyor.yml` file does 2).

> Also, I as I mentioned above, FAKE is designed for static target dependencies, rather than dynamic ones. If you move implementation to FAKE, it might look really ugly just because of how the FAKE is designed.

Wouldn't something like this work?

```fsharp
"CI"
  ==> "CompleteBuild"
  =?> ("PublishNuGetPrivate", isAlphaVersion)
  =?> ("PublishNuGetPublic", isReleaseVersion)
```

> Let's follow this way. I'll strip all the pre-release specific logic and implement it later on the v4 branch. 

Sounds good 👍

zvirja:
@ecampidoglio Thank you for the reply and patience! I really appreciate that.

> I understand. But since the version number is already generated in the build script, wouldn't having the deployment process be part of the same script simplify things even further? That way, we wouldn't have to rely on tags to share that information between AppVeyor and the build script.

Likely, we still have misunderstanding regarding the tags purpose, because I don't use it for any information sharing between CI and FAKE... :pensive: 

Let's look at the example, it could happen that I'm missing something. Consider the following commits graph:
```
  v3.10.1      
 /
A --- B --- A' --- B' --- C'
      ^                   ^
   master                 v4
```
You push packages to the private feed on each commit for `v4` _automatically_. As result, you have packages in private feed for the `A'`, `B'` and `C'` commits. However, the _NuGet_ package versions for these commits will be:

- A' - v3.10.1.2
- B' - v3.10.1.3
- C' - v3.10.1.4

That is a really confusing and I expect that for `v4` branch you have `4` as a major number. Therefore, I suggest the following trick:
```
  v3.10.1     v4.0.0-alpha 
 /           /
A --- B --- A' --- B' --- C'
      ^                   ^
   master                 v4
```
This way, you will have the following package versions:

- A' - v4.0.0-alpha
- B' - v4.0.0-alpha.1
- C' - v4.0.0-alpha.2

Now it's clear that `v4.0.0-alpha.X` corresponds to the `v4` branch. As you can see, the `v4.0.0-alpha` tag is _not a push trigger_, rather it's just a way to increase major version to `4` for the `v4` branch.

I find that it's pretty elegant way to solve the versioning issue as it costs nothing. Do you see a better way to solve this problem? 

---

> The way I see it, the responsibilities we're talking about here are:
> 1. Implement the build and deployment pipeline.
> 1. Start the pipeline from the CI environment.

It seems that we have different vision about the point 2. I also see that CI is responsible for starting the pipeline, however I expect that CI starts the _particular_ pipeline. Usually you configure triggers on CI, not in your build scripts, so CI is the only instance who knows about different triggers. If you have TeamCity, you usually don't pass all the TeamCity state to the build script to decide what to do later. Rather, you create a several build configurations with different triggers to start the different tasks based on events.

The same situation I see here. I want to handle __triggers__ on CI part and start the particular pipeline, while FAKE is responsible for the pipeline implementation only. I don't want FAKE knows about triggers we could have in CI and I believe that triggers is not FAKE's responsibility.

For TeamCity we configure triggers in UI, while for the AppVeyor we have a configuration file. However that should not change the conception.

Currently, all the code present in the `appveyor.yml` file is only about triggers handling. We just have different triggers and all them trigger the different build pipelines. We don't handle repository state, rather we works with build triggers.

> Wouldn't something like this work?

Likely it would work, I also started to imagine that implementation after some time thinking about that :flushed: However, I still stick the idea that __triggers handle__ is CI responsibility, not the build script one. Don't you agree with that particular point? 😉

ecampidoglio:
> Do you see a better way to solve this problem?

How about we use the `vNext` branch name to derive the version number? For example, a branch named `v4` would result in version number `v4.0.0-alpha.N` where `N` is the number of commits from `master`.  That way, we can keep using tags for releases only.

> I want to handle triggers on CI part and start the particular pipeline, while FAKE is responsible for the pipeline implementation only. I don't want FAKE knows about triggers we could have in CI and I believe that triggers is not FAKE's responsibility.

I think you're mixing different concepts here. In our project, there's only one kind of event that can trigger a CI build: someone pushes a commit on a certain branch. At that point the CI server simply needs to say "start the build" by running _one and only one target_ in the build script.

Right now, `appveyor.yml` is doing more than just starting the pipeline; it's _parsing the version number_ out of the tag (and before also _checking the branch name_) in order to decide _which target_ to run in the build script using a bunch of `if` statements. This logic of _what to deploy_ might as well be in the build script itself, where the version number is already available as a local variable.

> triggers handle is CI responsibility, not the build script one.

Yes, handling the trigger is the responsibility of the CI server, but there's really only one trigger: someone pushed a commit. Everything else (deciding where to push what packages) is part of the _deployment process_ which, IMO, belongs to the build script.

zvirja:
> How about we use the vNext branch name to derive the version number?

We can use, but it's not a very flexible solution. What should happen after you release a `v4.0.0-RC1` on the `v4` branch? I'd expect that next builds will be `v4.0.0-RC1.NUM`, rather than still `v4.0.0-alpha.NUM`. If we start to handle such cases as well, it will make the version resolve logic much more complicated.

With the "alpha tag" approach logic will remain simple and transparent, while you will get support for `beta`, `rc`, whatever. That's why I think it looks like a better much optimal and suitable.

---

> In our project, there's only one kind of event that can trigger a CI build: someone pushes a commit on a certain branch
> Yes, handling the trigger is the responsibility of the CI server, but there's really only one trigger: someone pushed a commit.

I don't agree that we really have a single trigger and everything else doesn't matter. There are several absolutely different triggers:
- A pull request
- Tag pushed
- Branch pushed

And there is a HUGE difference for us which trigger it is. If you omit all the trigger details and just say FAKE that "commit was pushed somewhere", you will get bugs:
- What happens if you create a branch on tagged commit and push it? Should we simply trigger the build or start the release procedure as this commit is tagged? We need to check which particular trigger we have - branch push or tag push.
- What if somebody create PR with `vNext` branch name? Formally, current branch name is `vNext`, so should be deploy this build to the private feed? Likely, no.
- etc

Therefore, it would be not a full truth to say that we have single trigger only - a commit push. If trigger information does matter to correctly handle different scenarios, it means that we have set of standalone and independent triggers, rather than a single one.

We will need to pass all the information we have about _trigger_ to the FAKE, so it will need to clarify which exact trigger it was and what to do later. Clarification of a _trigger_ (not repository state, but the trigger details) and picking up the appropriate task is the CI responsibility, IMO :) 

> it's parsing the version number out of the tag (and before also checking the branch name) in order to decide which target to run in the build script using a bunch of if statements.

It clarifies the tag format only to recognize the particular trigger. After that it just triggers the appropriate pipeline. Currently we have a few independent triggers:
- Not SemVer tag was pushed
- Alpha tag was pushed
- SemVer and non-alpha tag was pushed
- PR was created
- vNext branch pushed
- Other

All the logic the `appveyor.yml` script does is to clarify which exact trigger we have and start the appropriate FAKE task. It doesn't analyze the repository state, only trigger details. Everything else (version calculation based on repository state, how to build and how to deploy) is up to FAKE.

Do you still see that we have a single trigger only?

ecampidoglio:
> What should happen after you release a v4.0.0-RC1 on the v4 branch?

Who said anything about release candidates? AFAIK, we're still talking about the process _as is_. Besides, do we really need to have betas and release candidates to describe _different degrees of instability_ or is it enough to differentiate between _stable_ (`master`) and _unstable_ (`vNext`)? 

> What happens if you create a branch on tagged commit and push it? Should we simply trigger the build or start the release procedure as this commit is tagged? We need to check which particular trigger we have - branch push or tag push.

Why? If the commit is tagged, there should be a release, shouldn't there?

> What if somebody create PR with vNext branch name?

AppVeyor will tell you if a commit is on a PR branch; the build script can get that information from an environment variable.

> etc

What would that be?

> Currently we have a few independent triggers:
> 
> - Not SemVer tag was pushed
> - Alpha tag was pushed
> - SemVer and non-alpha tag was pushed
> - PR was created
> - vNext branch pushed
> - Other
>
> All the logic the appveyor.yml script does is to clarify which exact trigger we have and start the appropriate FAKE task. It doesn't analyze the repository state, only trigger details.

A part from _Other_ (and I'm not really sure what you mean by that) all the other "triggers" are about tags and branch names. Aren't those part of the repository state? Couldn't the build script get that information by reading the Git history or by accessing an environment variable?

zvirja:
> Who said anything about release candidates? AFAIK, we're still talking about the process as is. Besides, do we really need to have betas and release candidates to describe different degrees of instability or is it enough to differentiate between stable (master) and unstable (vNext)?

Nobody says that we _should_ support them, however there was also no agreement and guarantee that "we will not need them". My actions doesn't change the approach - previously RC were supported by calculation logic and they continue to be supported if we need them. You suggest to change the schema to one that definitely doesn't support them. If we intentionally implement feature that limits our opportunities, we should agree that we will not need it 100%. I'd vote to enable such an opportunity, rather than make it limited and rework later if we need.

>> What happens if you create a branch on tagged commit and push it? Should we simply trigger the build or start the release procedure as this commit is tagged? We need to check which particular trigger we have - branch push or tag push.

> Why? If the commit is tagged, there should be a release, shouldn't there?

Because you don't know whether you have already pushed the release. NuGet will fail if you try to push package with same version again. You might push it on the "tag" event, so later on the "branch" event you shouldn't push release again, while you still stay on branch. You need to know whether current event is "branch" or "tag"

>> What if somebody create PR with vNext branch name?

> AppVeyor will tell you if a commit is on a PR branch; the build script can get that information from an environment variable.

This means that you don't have a simple "code pushed" trigger, but rather you still need to distinguish the particular trigger. 

>> etc

> What would that be?

The samples above were shown to demonstrate that you don't have a simple "code pushed" trigger (as you mentioned [here](https://github.com/AutoFixture/AutoFixture/pull/755#issuecomment-297982879)), but rather you should take care about trigger details. That means that you have a set of triggers, rather than a single one. And it's CI responsibility to identify the exact trigger and run the appropriate action.

> Aren't those part of the repository state? Couldn't the build script get that information by reading the Git history or by accessing an environment variable?

It seems that we are talking about different things. There are 2 kind of information:
1. Repository state
2. CI Trigger information.

My approach is that FAKE should work with (1) only, while everything which belongs to CI trigger information (2) should be handled by CI only. Environment variables you mention belongs to (2).

You tried to simplify things saying that "there is only single event - commit pushed" like everything else doesn't matter to FAKE. I showed you that this is not a truth, because trigger detail does matter for action selection. That means that we have _a few different_ triggers, not the simple _something was pushed_.

Technically, we can achieve everything we want with both current approach and approach when logic resides in FAKE. The question is about the way we decide to follow and I vote to keep it as it is now.

-----

It seems that we stuck. You have one vision, I have different one. Likely, my arguments will not change your strong opinion, while I also don't agree that it should be a part of FAKE given all the arguments above :weary: 
We should find a way to solve this somehow. Probably, involve other guys? What do you think?

ecampidoglio:
I agree that this discussion needs more input. 👍

But, just for the sake of the argument, let me ask you a few questions.

> You might push it on the "tag" event, so later on the "branch" event you shouldn't push release again, while you still stay on branch.

This is Git we're talking about, right? What do you mean by "tag event" and "branch event"?

> There are 2 kind of information:
>
> - Repository state
> - CI Trigger information

What do you mean by "CI trigger information"? How is that different from the state of the repository?

Also, you didn't answer my question: if part of the deployment logic is in the CI server, how would you be able to run the same deployment process locally?

zvirja:
> This is Git we're talking about, right? What do you mean by "tag event" and "branch event"?

I refer to "events" as reason why this build was triggered. AppVeyor allows to distinguish why exactly this build was triggered - e.g. because tag was pushed (or because event was pushed). And this information is vitally important when we decide what we should do.

> What do you mean by "CI trigger information"? How is that different from the state of the repository?

In fact, I refer to all the AppVeyor environment variables we have. That variables allow to determine reason why this build was triggered. I refer to the AppVeyor context as "CI trigger information".

On other state, repository state is everything which is stored on the git.

In the `appveyor.yml` file I never work with Repository state and work with environment variables only.

> if part of the deployment logic is in the CI server, how would you be able to run the same deployment process locally?

Currently CI is _not_ a part of deployment logic. You can easily run the process you need locally - just specify FAKE what you want to do.

- to run verification build - `build.cmd CompleteBuild`
- to publish to public feed - `build.cmd PublishNuGetPublic NuGetPublicKey="KEY"`
- to publish to private feed - `build.cmd PublishNuGetPrivate NuGetPrivateKey="KEY"`

zvirja:
@ecampidoglio Do you have a plan how we could proceed on this one? :)

ecampidoglio:
@zvirja Well, I still think the CI server should just invoke a target in the build script and let that handle the deployment process. And, yes, I think that deciding _where_ to push the packages (if at all) is part of the deployment process.

Look, we said that our process looks like this:

- Commit on `vNext` → `PublishNuGetPrivate`
- Tag → `PublishNuGetPublic`

We're always going to build a specific commit: the one `HEAD` points to when the build script runs. That commit will be in a particular branch and may or may not have a tag pointing to it. That information is part of the _repository state_ (not the CI server) and can just as well be read by the build script; in fact, we're already doing it to generate a version number.

As I mentioned [before](https://github.com/AutoFixture/AutoFixture/pull/755#issuecomment-297968329), creating a target like this is totally feasible:

```fsharp
"CI"
  ==> "CompleteBuild"
  =?> ("PublishNuGetPrivate", isAlpha)
  =?> ("PublishNuGetPublic", isRelease)
```

The `isAlpha` function can use the name of the current branch to determine the outcome while `isRelease` could read the tag on the current commit, if present.

That would leave the `appveyor.yml` file with just:

```yml
build_script:
  - ps: & .\build.cmd CI
```

Which would allow us to run the _entire_ deployment process locally without relying on the CI server—nor our own judgment—to decide what to do.

zvirja:
I have a feeling that we have different vision on the basic thing, which causes the confusion. It seems that you have an assumption that FAKE will be able to decide what to do by analyzing the repository state only. However, that is not true and, likely, that is the point which creates difference in the vision.

If you create a "CI" task in FAKE, you will be _required_ to read AppVeyor environment variables to understand the current trigger information. And that is exactly the place I disagree - my point is that FAKE should not be aware of CI and triggers.

> That commit will be in a particular branch and may or may not have a tag pointing to it. That information is part of the repository state (not the CI server) and can just as well be read by the build script;

That information is a part of repository, however __it's not enough__ to make the decision what to do.

Could you please confirm that we have same vision on this point and you agree that in implementation you suggest FAKE __will be aware__ of particular AppVeyor variables and triggers? Do you agree that repository state __is not enough__ to make a decision what action should be applied (e.g. build vs push), so AppVeyor trigger information (environment variables) should be examined?

ecampidoglio:
It's true, we need the `APPVEYOR_REPO_TAG` variable to be able to tell whether we should do a public release or not—I had forgotten about that. 😳

Personally, I don't have a problem with the FAKE script reading CI environment variables but I'll understand if you want to keep the AppVeyor pieces separate. Tell you what: what if we have a target that looks like this:

```fsharp
"CI"
  ==> "CompleteBuild"
  =?> ("PublishNuGetPrivate", isAlpha)
```

Then, in the `appveyor.yml` file we do something like:

```yml
build_script:
  - ps: >-
      & .\build.cmd CI
      if ($Env:APPVEYOR_REPO_TAG -eq $true) {
        & .\build.cmd PublishNuGetPublic
      }
```

Would that work?

zvirja:
OK, so now you also see that AppVeyour trigger state is required :) However, that not the only variable we need and logic is not that simple as you suggested above. Consider the following facts:

- We shouldn't perform a release if tag is not a semver. This repository already have such tags, so we should have a guard against that if somebody creates such a tag. Also I don't want our repo has limitation of semver tags only. 
- We need the `$Env:APPVEYOR_REPO_BRANCH` variable as well, as it might happen that somebody creates and pushes a custom branch on top of `vNext`. In this case you might have multiple branches (e.g. `vNext` and `custom`) pointing to the same commit. You should not push to the private feed when custom branch is pushed. 
- We need the `$Env:APPVEYOR_PULL_REQUEST_NUMBER` variable when we examine the `$Env:APPVEYOR_REPO_BRANCH` variable. That is because for PRs to  `vNext` the latter variable will be set to `vNext`, so we could mistakenly decide to push to feed when we shouldn't. 

> Then, in the appveyor.yml file we do something like: ... 

Not all the cases mentioned above are covered here. However even if they would be, in this case you will have multiple builds as `PublishNuGetPublic` task has dependency on the `Build` task as well. 

-------

The existing implementation covers all the cases. If you check the `appveyor` file, you will see that I inspect the Environment Variables only, so it's unlikely that we can move this logic to FAKE without removal of some logic.

ecampidoglio:
> I don't want our repo has limitation of semver tags only.

Why not?

> You should not push to the private feed when custom branch is pushed.

Can't this be solved by looking at the generated version number? If it contains `alpha`, push to private feed, otherwise no.

> That is because for PRs to `vNext` the latter variable will be set to `vNext`.

Again, couldn't we look at the version number?

In any case, I feel we're going around in circles here. You are OK with splitting the deployment process between `appveyor.yml` and the build script while I want everything in the build script, even if it has to read AppVeyor-specific environment variables.

I'm sorry to say it but without any new information, we're unlikely to ever reach an agreement. I think we should give this issue a few more days to give other people (especially members of @AutoFixture/core) an opportunity to chime in. After that, if we still can't agree on how we should run builds in the CI environment, we might as well close this for now and revisit it later.

However, we could still merge c0b23d0 and c34c6fa. That way we would, at least, be able to release by manually invoking the appropriate target in the build script.

zvirja:
> Why not?

Because such kind of functionality should be robust to accidental situations. There is also no good technical reason why such limitation should be present at all.

> Can't this be solved by looking at the generated version number? If it contains alpha, push to private feed, otherwise no.

> Again, couldn't we look at the version number?

I don't see how generated version number could solve the issue. If you have your custom branch based on `vNext`, the generated version will also contain the `-alpha` suffix - branch name doesn't participate in version generation.

> In any case, I feel we're going around in circles here. 

Yep, I've spotted that already [4 days ago](https://github.com/AutoFixture/AutoFixture/pull/755#issuecomment-298001430) :unamused:

------
> After that, if we still can't agree on how we should run builds in the CI environment, we might as well close this for now and revisit it later.

The situation we currently have and had during the latest PRs is very uncomfortable for me. I believe we reached the point when we should clarify the things and decide how we could collaborate further. I'll try to speak clearly and the only goal I pursue is to achieve the common agreement. Also I would like to emphasize that I appreciate all the efforts from your side on the code review and your time.

This project was dead for the last half of year. Recently we had a spike I suggested my help to make it more alive. It was agreed that I'll take the Release Manager responsibilities. As a result I was provided with all the keys and was added to the @AutoFixture/core group. It inspired me and I decided to participate in the project to make it shine again. There are a lot of things that should be done, so I started to do them. 

The first thing we need to do is to simplify the release management. Given that formally this is my responsibility, I started to apply changes which make releases easier. And it appeared that I immediately faced a huge amount of resistance from @ecampidoglio. It required large efforts to communicate each particular change. I do understand that all we have the different vision and discussion are absolutely fine, 
however I found you very skeptical to the ideas I suggest and approaches I follow. It was not the best PR experience I ever had :confused: Personally, I feel that all this situation forces me to loose the passion and inspiration I have for this project 😞 

If we look at this particular PR, you would likely agree that approach I suggested from the beginning is _viable and working_. It doesn't break the public agreement we have about releases and the whole discussion is just a matter of personal tastes and vision. It appeared to become a long and time consuming dispute about the implementation detail and the responsibilities of the FAKE and AppVeyor parts. I feel that the suggested implementation is OK and could be accepted. However, it took 6 days to communicate minor things and even finally we are speaking about declining this change.

I expect that OSS maintainers should be flexible enough people and would accept changes even if personally they would implement that in a bit different way. I spent huge amount of my personal time to design and implement the approach and feel that the feedback I met is unfair and frustrating. If we continue to follow this way, the project will not be healthy - I believe that contributors will not always have the high amount of inspiration and desire to fight. If you read Mark's post [about contribution](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/) (point 1 - "Make it small"), we will find him talking about appreciation of the contributor's efforts. This is exactly what I mean. I want to ask you (@ecampidoglio) to be more flexible and open, so that contribution is easier. There are a lot of upcoming changes we should apply because of .NET Core transformation and I'm scared to even imagine how it could be done following the current way (and how much time it will take).

Another point which seems to be unclear is my role in the project. After I accepted all the keys, was invited to the @AutoFixture/core group and started to take care of releases, I became a part of the maintainer's team. That gives me a right to vote for the decisions we make and make my voice equal to voice of any other @AutoFixture/core member. Currently we have a situation when @adamchester doesn't have time to review this PR, @moodmosaic submitted that he didn't have enough experience in this area (correct if I got it wrong), @ecampidoglio would implement that differently and me is offering the current approach. I'd expect that unless my approach have some critical implications (and it doesn't seem to have), it has _the same_ priority as @ecampidoglio's vision. Given that it's already implemented, I believe it's totally fine to accept it. @ecampidoglio's decision to deny changes in current situation looks strange to me :confused:

I'm definitely open to your feedback and want to continue improvement of my contributions :pensive: Hope, we will address the issues above our subsequent collaboration will satisfy everybody :thumbsup: :blush:

moodmosaic:
@zvirja, perhaps, when it comes to new features, it's because [speed kills](http://programmer.97things.oreilly.com/wiki/index.php/Speed_Kills) one of the reasons we're moving slow the past 7+ years:

* #317 by @ecampidoglio opened on Oct 16, 2014, merged by @ploeh into master on Jan 27, 2015
* #318 by @ecampidoglio opened on Oct 16, 2014, merged by @ploeh into master on Oct 2, 2015
* #158 by @gertjvr opened on Aug 25, 2013, merged by @ploeh into master on Sep 24, 2013
* ... there are others

Mark's [article](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests/) that you mentioned above writes it clearly:
>Don't make your reviewer work.
>The more you make your reviewer work, the greater the risk is that your Pull Request will be rejected.

IMO, it'd be nice if at least you and another member from the core team understands how the new release management works, so it can still be maintainable if you're not around (busy at work, vacations, and so on).

That being said, I don't think that the 6 days it took you to communicate minor things is a long period...

ecampidoglio:
In addition to what @moodmosaic [said](https://github.com/AutoFixture/AutoFixture/pull/755#issuecomment-298828444)—which I agree with—I'd like to bring up a couple of more points.

You, @zvirja, are in a rather unique position: you joined the core team of a multi-year and fairly sized open source project without having any prior history with it. You might be the _best programmer_ on Earth, but open source is a [meritocracy](https://en.wikipedia.org/wiki/Meritocracy), which means that if we don't know you, you'll have some more explaining to do to justify your actions. Nothing personal—that's just how things work.

Having said that, the first thing you did after joining @AutoFixture/core was applying some [pretty substantial changes](https://github.com/AutoFixture/AutoFixture/pull/749) to our release process without first discussing it with the rest of the team. Maybe you felt that was what the project needed the most, maybe you thought that was the best way you could contribute; either way, a project's release process is _critical_ and not something you change lightly.

Naturally, a lot of questions were raised—especially by me—and rightly so. I had to digest the scope of your proposed changes, at the same time trying to understand where you were headed with your vision and possibly even foresee any consequences it might have for the project down the line.

I have a busy schedule, so these things take a while. In the meantime, you grew more and more frustrated both with the questioning and the response times so you decided to put pressure, trying to rush things along so you could move on to the next thing.

I understand your frustration, but that's _not how we work_ in this project.

At this point, you have a few different options:

1. You can be patient and work with us as we go through your proposed changes, knowing that we'll eventually reach a solution.
2. Put the whole release process thing on hold for the time being and contribute to some other part of AutoFixture.
3. Walk away from the project altogether.

I certainly hope you don't go for the third option, but if you do, I'll understand. As I said, AutoFixture has a long tradition and if you don't like the way we do things, you'll be happier putting your efforts on another project.

zvirja:
Thank you for your comments @moodmosaic and @ecampidoglio 👍 Your intentions looks logical and I agree with them in general.

@ecampidoglio Unfortunately, I still don't see how this

> You are OK with splitting the deployment process between appveyor.yml and the build script while I want everything in the build script, even if it has to read AppVeyor-specific environment variables.
> ...
>  After that, if we still can't agree on how we should run builds in the CI environment, we might as well close this for now and revisit it later.

might be explained by

> ... a project's release process is critical and not something you change lightly.

> I had to digest the scope of your proposed changes, at the same time trying to understand where you were headed with your vision and possibly even foresee any consequences it might have for the project down the line.

It looked that we agreed on the idea of _what should be done_ and the question was only about implementation detail. However, you vote to reject my PR because my approach doesn't match yours. Could you please provide the _technical_ reasons why the currently implemented approach cannot be accepted?

ecampidoglio:
@zvirja That's a fair question. My latest comment about putting the PR on hold is due to fact that, after giving it some more thought, I'm not entirely convinced that the versioning scheme (and the way it's calculated) is the right one. Since the version number is directly related to where the package should be published, the whole thing starts to fall down.

At the moment, I don't have time to sit down and think this through; at the same time, I don't want to rush it just for the sake of having something automated. We've had a release process that has served us well for the past few years and it can continue to do so a little while longer. That's why I'd like to put this issue on hold, for the time being.

zvirja:
@ecampidoglio The whole version calculation logic for the `vNext` branches has been already removed from this PR [some time ago](https://github.com/AutoFixture/AutoFixture/pull/755#issuecomment-297957621). Currently PR handles only release tags and PRs - minimal set to enable releases for the `master`.

Given that now PR doesn't change the previous release process, do you have some other concerns?

zvirja:
@ecampidoglio I've just updated PR, so now:
- all the AppVeyor logic is located in FAKE script.
- I handle SemVer tag creation only to cover the process we had so far.

That should address all the concerns you had so far. Could you review the changes so we could proceed on this one?

Build.fsx
```diff
@@ -115,14 +116,14 @@ Target "TestOnly" (fun _ ->
     let nunit2TestAssemblies = !! (sprintf "Src/AutoFixture.NUnit2.*Test/bin/%s/*Test.dll" configuration)
 
     nunit2TestAssemblies
-    |> NUnit (fun p -> { p with StopOnError = false
-                                OutputFile = "NUnit2TestResult.xml" })
+    |> NUnitSequential.NUnit (fun p -> { p with StopOnError = false
+                                                OutputFile = "NUnit2TestResult.xml" })
```
ecampidoglio: This change seems to be unrelated to the new `AppVeyor` target. Could you please commit it separately and describe its purpose in the message?

Build.fsx
```diff
@@ -115,14 +116,14 @@ Target "TestOnly" (fun _ ->
...
 
     let nunit3TestAssemblies = !! (sprintf "Src/AutoFixture.NUnit3.UnitTest/bin/%s/Ploeh.AutoFixture.NUnit3.UnitTest.dll" configuration)
 
     nunit3TestAssemblies
-    |> NUnit3 (fun p -> { p with StopOnError = false
-                                 ResultSpecs = ["NUnit3TestResult.xml;format=nunit2"] })
+    |> Testing.NUnit3.NUnit3 (fun p -> { p with StopOnError = false
+                                                ResultSpecs = ["NUnit3TestResult.xml;format=nunit2"] })
```
ecampidoglio: This change seems to be unrelated to the new `AppVeyor` target. Could you please commit it separately and describe its purpose in the message?

Build.fsx
zvirja: It's required to make code compilable - we have names collision after open of the `Fake.AppVeyor` namespace. 

Given that this change is caused by other changes related to this commit, do you agree that it's OK to keep it as is? :)

Build.fsx
zvirja: The same comment as above.

Build.fsx
ecampidoglio: Yes, consistency is more important. 😉

zvirja:
@ecampidoglio Thanks for the collaboration :star2:

Given that @adamchester [doesn't have time](https://github.com/AutoFixture/AutoFixture/pull/755#discussion_r113824699) to review PRs now, @moodmosaic do you have any concerns regarding the current implementation? If not - I'll merge this PR and will create a note somewhere on Wiki describing the current release process.

moodmosaic:
@zvirja, I don't have any concerns regarding the current changes 👍 💯

zvirja:
@moodmosaic, @ecampidoglio, @klimisa I've described the current release procedure in [wiki](https://github.com/AutoFixture/AutoFixture/wiki/Internal-processes). Feel free to append or fix it if I missed something.

klimisa:
Sorry guys for the delay, I was really busy with work. I did have a thorough and careful read on this thread, trying to understand the arguments of the people involved, mainly @ecampidoglio and @zvirja. It was a painful read to be honest. 
 
I keep in mind that also things have progress in the direction since then, in the direction suggested by @ecampidoglio. 

I would like to tell my opinion on the subject, even knowing that I'm late. 
For me, @zvirja has suggested an automated release solution which,

- It does not break current consumer’s experience.
- It is so flexible that can serve the product for many more years with no changes.

The separation of concerns into triggers and pipelines at my humble opinion is excellent and if I were, about to design a CI I would had followed the same approach.
 
Tags solutions on the other hand instead of branches gives the flexibility to play well with any kind of release naming that might come up in the future and be consistent. The best part though, it is cheap to implement.
 
I also understand the argument that @zvirja started to do stuff without discussing them much. I agree, the release process weren’t in the plans to change, though that is not a strong reason not to do stuff and move forward. In addition, the role of a release manager is a discrete role and can move in parallel with the Core team, which is responsible for the approval of the PR’s.
 
My only concern, is, if can I run the build locally as I have still lot of stuff to catch up.

PR #759, opened 2017-05-12
Minor build script improvements

zvirja:
Small refactorings for the `Build.fsx` file to improve readability. The logic wasn't changed somehow.

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -17,7 +17,6 @@ let nuGetPackages = !! (nuGetOutputFolder @@ "*.nupkg" )
                     -- (nuGetOutputFolder @@ "AutoFixture.AutoFakeItEasy2.*" )
 let signKeyPath = FullName "Src/AutoFixture.snk"
 let solutionsToBuild = !! "Src/All.sln"
-let processorArchitecture = environVar "PROCESSOR_ARCHITECTURE"
 
 type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string }
 let calculateVersionFromGit buildNumber =
--- a/Build.fsx
+++ b/Build.fsx
@@ -244,13 +244,12 @@ Target "PublishNuGetAll" DoNothing
 "PatchAssemblyVersions" ==> "Build"
 "BuildOnly"             ==> "Build"
 
+"BuildOnly" 
+    ==> "TestOnly"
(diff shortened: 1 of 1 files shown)


Build.fsx
```diff
@@ -126,9 +125,9 @@ Target "TestOnly" (fun _ ->
                                                 ResultSpecs = ["NUnit3TestResult.xml;format=nunit2"] })
 )
 
-Target "BuildAndTestOnly" (fun _ -> ())
-Target "Build" (fun _ -> ())
-Target "Test"  (fun _ -> ())
+Target "BuildAndTestOnly" DoNothing
+Target "Build" DoNothing
+Target "Test"  DoNothing
```
ecampidoglio: I think it looks rather wasteful to have all these no-op targets laying around. Couldn't we remove them?

Build.fsx
```diff
@@ -244,14 +243,13 @@ Target "PublishNuGetAll" DoNothing
...
+    ==> "BuildAndTestOnly"
+
 "Build"    ==> "Test"
 "TestOnly" ==> "Test"
 
-"BuildOnly" ==> "TestOnly"
-
-"BuildOnly" ==> "BuildAndTestOnly"
-"TestOnly"  ==> "BuildAndTestOnly"
-
```
ecampidoglio: Are _all_ these targets really necessary? Could we boil them down to just `Build`, `CompleteBuild`, and `AppVeyor`?

zvirja: It depends on the usage scenario. From the `CompleteBuild` or `AppVeyor` perspective, they are not required. However, personally I find the `BuildOnly` and `TestOnly` targets quite useful. They allows to _quickly_ check whether project compiles now and tests pass (these targets don't have dependencies on cleanup and validations). I used them for a couple of times.

Also these targets were present since the script creation, so might be used by people. I'd vote to keep them given that they make sense and are not useless 😉

Build.fsx
zvirja: Currently they are needed. Let's discuss that below and if we agree to strip `BuildOnly` and `TestOnly`, I'll refactor them.

Build.fsx
ecampidoglio: I'm aware that they have been around for a while; I figured that, since we're refactoring the build script, we might as well do some spring cleaning while we're at it. 😉

I see your point about them possibly being useful to someone. Let's keep them.

PR #761, opened 2017-05-14
Add .editorconfig file

zvirja:
VS 2017 added a native support of the `.editorconfig` files. I've added a file with basic configuration applicable to this project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*.{cs,fs,fsx}]
+indent_size = 4
+indent_style = space
+
+[*.{sln,csproj,fsproj,config,xml}]
+indent_size = 2
+indent_style = space


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# top-most EditorConfig file
+root = true
+
+[*.{cs,fs,fsx}]
+indent_size = 4
+indent_style = space
+
+[*.{sln,csproj,fsproj,config,xml}]
+indent_size = 2
```
moodmosaic: Are you sure this is-or should be-`2`? 😕

zvirja: I used the current repository state as a style source. All the specified files currently have 2 spaces indentation. I decided to not change that and just explicitly describe that in configuration :pensive:

AFAIK, 2 spaces is Visual Studio default for `sln`, `csproj`, `fsproj` and `packages.config`, so we shouldn't override it.

moodmosaic: 👍 Sounds good then.

.editorconfig
```diff
@@ -0,0 +1,10 @@
+# top-most EditorConfig file
```
ecampidoglio: Do we need a comment here? The line below is pretty self-explanatory. 😉

zvirja: Removed that comment 👍

PR #762, opened 2017-05-14
Add auto deployment for the vNext branch

zvirja:
This PR enables push to the [private feed](https://www.myget.org/F/autofixture) on any commit to the vNext branch (`v4` in our case).

## Version calculation
The version calculation logic was updated to the following one:
1. Extract `vNext` major version from the `vNext` branch name (e.g. `4` for the `v4`).
1. Calculate version using the current [tag based logic](https://github.com/AutoFixture/AutoFixture/pull/753). Compare version's major number to `vNext` major version. If they are the same, use this version (e.g. use result if it's `v4.0.0-beta1.25` for `v4` branch).
1. If previous version is not applicable, generate version based on `vNext` major version and number of commits since the last release before the `vNext` fork. The generated version has the `-alpha` pre-release suffix.  For instance, for the graph below
```
| * vNext commit 6
| * merge master to vNext
 /|
/ |
* | master commit
* | master commit
* | master commit
| * vNext commit 5
| * vNext commit 4 (tagged as v4.0.0-beta1)
| * vNext commit 3
| * merge master to vNext
 /|
/ |
| * vNext commit
* | master commit (tagged as v3.50.2)
| * vNext commit 1
* | master commit
| /
|/
* master commit (tagged as v3.50.1)
* master commit
```
result would be:
- `vNext commit 1` - `v4.0.0-alpha.1`
- `vNext commit 2` - `v4.0.0-alpha.2`
- `vNext commit 3` - `v4.0.0-alpha.4`
- `vNext commit 4` - `v4.0.0-beta1`
- `vNext commit 5` - `v4.0.0-beta1.1`
- `vNext commit 6` - `v4.0.0-beta1.3`

## vNext branch configuration (updated)
No need to configure the vNext branch somehow - simply create a branch with the `vX` name, where X is the number and builds will be automatically pushed to the MyGet feed.

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -18,8 +18,9 @@ let signKeyPath = FullName "Src/AutoFixture.snk"
 let solutionsToBuild = !! "Src/All.sln"
 let bakFileExt = ".orig"
 
-type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string }
-let calculateVersionFromGit buildNumber =
+type BuildVersionCalculationSource = { major: int; minor: int; revision: int; preSuffix: string; 
+                                       commitsNum: int; sha: string; buildNumber: int }
+let getVersionSourceFromGit buildNumber =
     // Example of output for a release tag: v3.50.2-288-g64fd5c5b, for a prerelease tag: v3.50.2-alpha1-288-g64fd5c5b
     let desc = Git.CommandHelper.runSimpleGitCommand "" "describe --tags --long --match=v*"
 
@@ -27,18 +28,31 @@ let calculateVersionFromGit buildNumber =
                              @"^v(?<maj>\d+)\.(?<min>\d+)\.(?<rev>\d+)(?<pre>-\w+\d*)?-(?<num>\d+)-g(?<sha>[a-z0-9]+)$",
                              RegexOptions.IgnoreCase)
                       .Groups
+
     let getMatch (name:string) = result.[name].Value
 
-    let major, minor, revision, preReleaseSuffix, commitsNum, sha =
-        getMatch "maj" |> int, getMatch "min" |> int, getMatch "rev" |> int, getMatch "pre", getMatch "num" |> int, getMatch "sha"
+    { major = getMatch "maj" |> int
+      minor = getMatch "min" |> int
+      revision = getMatch "rev" |> int
+      preSuffix = getMatch "pre"
+      commitsNum = getMatch "num" |> int
+      sha = getMatch "sha"
+      buildNumber = buildNumber
+    }
 
-    
+type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string; 
+                          source: Option<BuildVersionCalculationSource> }
+let calculateVersion source =
+    let s = source
+    let (major, minor, revision, preReleaseSuffix, commitsNum, sha, buildNumber) =
+        (s.major, s.minor, s.revision, s.preSuffix, s.commitsNum, s.sha, s.buildNumber)
+
     let assemblyVersion = sprintf "%d.%d.0.0" major minor
     let fileVersion = sprintf "%d.%d.%d.%d" major minor revision buildNumber
     
     // If number of commits since last tag is greater than zero, we append another identifier with number of commits.
     // The produced version is larger than the last tag version.
-    // If we are on a tag, we use version specified modification.
+    // If we are on a tag, we use version without modification.
     // Examples of output: 3.50.2.1, 3.50.2.215, 3.50.1-rc1.3, 3.50.1-rc3.35
     let nugetVersion = match commitsNum with
                        | 0 -> sprintf "%d.%d.%d%s" major minor revision preReleaseSuffix
@@ -48,17 +62,23 @@ let calculateVersionFromGit buildNumber =
                       | 0 -> nugetVersion
                       | _ -> sprintf "%s-%s" nugetVersion sha
 
-    { assemblyVersion=assemblyVersion; fileVersion=fileVersion; infoVersion=infoVersion; nugetVersion=nugetVersion }
+    { assemblyVersion=assemblyVersion; fileVersion=fileVersion; infoVersion=infoVersion; nugetVersion=nugetVersion; 
+      source = Some source }
 
 // Calculate version that should be used for the build. Define globally as data might be required by multiple targets.
 // Please never name the build parameter with version as "Version" - it might be consumed by the MSBuild, override 
 // the defined properties and break some tasks (e.g. NuGet restore).
-let buildVersion = match getBuildParamOrDefault "BuildVersion" "git" with
-                   | "git"       -> calculateVersionFromGit (getBuildParamOrDefault "BuildNumber" "0" |> int)
-                   | assemblyVer -> { assemblyVersion = assemblyVer
-                                      fileVersion = getBuildParamOrDefault "BuildFileVersion" assemblyVer
-                                      infoVersion = getBuildParamOrDefault "BuildInfoVersion" assemblyVer
-                                      nugetVersion = getBuildParamOrDefault "BuildNugetVersion" assemblyVer }
+let mutable buildVersion = match getBuildParamOrDefault "BuildVersion" "git" with
+                           | "git"       -> getBuildParamOrDefault "BuildNumber" "0"
+                                            |> int
+                                            |> getVersionSourceFromGit
+                                            |> calculateVersion
+
+                           | assemblyVer -> { assemblyVersion = assemblyVer
+                                              fileVersion = getBuildParamOrDefault "BuildFileVersion" assemblyVer
+                                              infoVersion = getBuildParamOrDefault "BuildInfoVersion" assemblyVer
+                                              nugetVersion = getBuildParamOrDefault "BuildNugetVersion" assemblyVer
+                                              source = None }
 
 let addBakExt path = sprintf "%s%s" path bakFileExt
 
--- a/Build.fsx
+++ b/Build.fsx
@@ -340,16 +340,19 @@ Target "PublishNuGetAll" DoNothing
 type AppVeyorEnvironment with
     static member IsPullRequest = isNotNullOrEmpty AppVeyorEnvironment.PullRequestNumber
 
-type AppVeyorTrigger = SemVerTag | CustomTag | PR | Unknown
+type AppVeyorTrigger = SemVerTag | CustomTag | PR | VNextBranch | Unknown
 let anAppVeyorTrigger =
     let tag = if AppVeyorEnvironment.RepoTag then Some AppVeyorEnvironment.RepoTagName else None
     let isPR = AppVeyorEnvironment.IsPullRequest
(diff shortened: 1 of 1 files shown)


Build.fsx
```diff
@@ -18,28 +18,47 @@ let nuGetPackages = !! (nuGetOutputFolder @@ "*.nupkg" )
...
 let solutionsToBuild = !! "Src/All.sln"
 let processorArchitecture = environVar "PROCESSOR_ARCHITECTURE"
-
-type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string }
-let calculateVersionFromGit buildNumber =
+// === vNext config ===
+// To disable vNext branch set all the variables below to None
+let vNextBranchName = Some "v4"
+// Specify sha1 of the master branch commit the vNext branch was forked from
+let vNextForkedCommitSha = Some "a177824f05e911c70aac1a459b8da5bd206cf410"
```
moodmosaic: What happens if we forget to update this SHA-1?

moodmosaic:
I'm afraid I'd have to defer this one to @ecampidoglio or @adamchester as this stuff is (currently) not my core expertise.

Build.fsx
ecampidoglio: @zvirja You can get that value with the `git-merge-base` command, so there's no need to hard-code it here. Here's an example:

```sh
git merge-base master v4
```

This will return the SHA-1 of the _first common parent_ between `master` and `v4`, which is 9d7bdd4. Now, you just need to wrap that in a function and you're good to go. 😉

zvirja: @moodmosaic If you forget to update variable (and it will be `None`), version will not be updated to  `4.0.0-alpha.X`. I'll improve that later by creation of a record, if we agree on approach with Enrico.

@ecampidoglio I know about that command, but I don't see how it could help. I use this hash to calculate number of commits in the fork branch and [generate](https://github.com/AutoFixture/AutoFixture/blob/ee2214c05294cd72565ea6c7442cbce8e1d1f7d3/Build.fsx#L98) __unique__ and __steady__ build version. As you noticed, the `merge-base` output changes after each merge from `master`, so the generated build version will not be unique and steady.

Instead I looked at the `git rev-list --count HEAD ^master` command to generate constantly increasing numbers. However, this approach [has drawbacks](http://stackoverflow.com/questions/1527234/finding-a-branch-point-with-git#comment13238893_10174692) and isn't reliable because of them, so I decided to not use it.

If you still see that the approach you suggested is suitable, please describe the whole version generation algorithm - likely it differs from mine.

P.S. I've already reviewed a lot of approaches to calculate this commit automatically (like ones mentioned [here](http://stackoverflow.com/questions/1527234/finding-a-branch-point-with-git)), but none of them is suitable and works reliably for our repo.

zvirja: Rebased & refactored code to handle @moodmosaic's notice - now everything or nothing 😉

Build.fsx
```diff
@@ -18,27 +18,48 @@ let nuGetPackages = !! (nuGetOutputFolder @@ "*.nupkg" )
 let signKeyPath = FullName "Src/AutoFixture.snk"
 let solutionsToBuild = !! "Src/All.sln"
 
-type BuildVersionInfo = { assemblyVersion:string; fileVersion:string; infoVersion:string; nugetVersion:string }
-let calculateVersionFromGit buildNumber =
+type VNextInfo = { name:string; forkedCommitSha:string }
+// To disable vNext branch set the vNextInfo variable to None
+// Specify sha1 of the master branch commit the vNext branch was forked from
+let vNextBranchInfo:Option<VNextInfo> = Some { name="v4"; forkedCommitSha = "a177824f05e911c70aac1a459b8da5bd206cf410" }
+
```
moodmosaic: This means one would have to look inside the build script and modify the F# code? Is this the only way? (I'm just asking because I currently don't have that much of experience in build automation stuff.)

zvirja:
Yes, Release Manager will need to update a variable in the `build.fsx` file to enable private builds for the `vNext` branch. Given that it will be documented and should be done rarely, that should not be a problem. 

We don't have too many options as values should be configured somewhere. We could specify them in the `appveyor.yml` and pass as parameters, but that doesn't look better..

moodmosaic:
OK, as long as it's documented somewhere, and checked-in on the repository's README file or similar, it's fine by me.

Do we plan to have the [wiki](https://github.com/AutoFixture/AutoFixture/wiki) also on the repository as a file?

ecampidoglio:
It looks like it might be a good time to revisit [GitVersion](https://github.com/GitTools/GitVersion) and see if we can use that to do the version calculation.

I did a quick experiment with this configuration:

```yaml
branches:
  master:
    mode: ContinuousDelivery
    prevent-increment-of-merged-branch-version: true
  v[0-9]+:
    mode: ContinuousDeployment
    increment: Major
    tag: alpha
    prevent-increment-of-merged-branch-version: true
```

If you want to try it out, create it a file called `GitVersion.yml` and place it in the root directory. Then, install [GitVersion.CommandLine](http://www.nuget.org/packages/GitVersion.CommandLine/3.6.5) (you can do that via Chocolatey) and simply run `gitversion` in that same directory. Here's what I got:

- `master`: `3.50.3` (next minor version from `3.50.2`)
- `v4`: `4.0.0-alpha1990` (next major version from `3.0.0`)

What do you guys think?

moodmosaic:
That's interesting! I like the idea of having a specific `GitVersion.yml` for the version calculation. It makes things more obvious/explicit compared to searching inside a (not very small) build script.

ecampidoglio:
It's a good tool. It also takes care of patching the `AssemblyInfo` files, which is nice.

If we can configure it in a way that works for us, it should be fairly easy to [integrate it in our build script](http://fsharp.github.io/FAKE/apidocs/fake-gitversionhelper.html).

klimisa:
Filling up SHA it's an error prone process even if it's done very rare. 
We either,
1. Reconsider the implementation that @zvirja suggested [here](https://github.com/AutoFixture/AutoFixture/pull/755)
2. Use an external tool like GitVersion for the calculation suggested by @ecampidoglio 

In any case GitVersion is a candidate and should be evaluated. 💯

zvirja:
Hi guys,

Sorry for the belated reply. I was on the vacation and currently I'm investigating the possibility to use the GitVersion tool. There are some issues with this tool and it doesn't work as expected in particular scenarios. I'm trying to find the ways to solve that.

In the meanwhile, I've found that there is a limitation with the GitVersion tool - the `ContinuousDeployment` mode doesn't work if you have tags ending with numbers (e.g. `-beta1`, `-rc1`, `-rc2`). If we agree to use this tool, we should accept that we will not be capable of creation of such tags. Personally, I don't like that limitation and would refuse tool usage because of that. What do you guys (@ecampidoglio, @moodmosaic, @klimisa) think about that?

zvirja:
I've reworked the implementation so now:
- any branch with `v\d+` name is treated as `vNext`;
- the `-alpha.XXX` tag is calculated automatically for `vNext`, __without any need__ to modify configuration.

Thank you for the concerns regarding the previous approach - the never one is indeed better.

----
As for the `GitVersion` tool, I've spent 2 weeks, dig a lot into its source code, but wasn't able to make it work as I wish. Given that I found a way to fix the concerns above, I decided to proceed with current approach.

The main issues I observed with `GitVersion` tool are following (given [this](https://github.com/AutoFixture/AutoFixture/pull/762#issuecomment-301709213) configuration and latest beta):
1. If you create any tag on `vNext` (e.g. `v4.0.0-rc1`), it will ignore it and subsequent commit will still get `v4.0.0-alpha.XXXX` version. That is incorrect behavior as it should start numbering from the tag.
1. It ignores the number suffixes in the pre-release tag as it was mentioned [before](https://github.com/AutoFixture/AutoFixture/pull/762#issuecomment-306428303).

I've tried different ways (e.g tried to name branch `v4.0` instead), but none of them works fine and breaks at some point. I suspect that it's because tune is aligned for different branching and deployment models (e.g. GitFlow).

---
@ecampidoglio @moodmosaic Are you fine with the current approach after I tweaked it?

moodmosaic:
Like I wrote [here](https://github.com/AutoFixture/AutoFixture/pull/762#issuecomment-301374792), I'd have to defer this to @klimisa, @ecampidoglio or @adamchester.

klimisa:
I would personally need a bit more time for that issue to think about it more thoroughly. At the moment I'm being extremely busy with work. Therefore I humbly ask your understanding to postpone my answer a bit more.

mxa0079:
Can anything be done by the community to accelerate this PR. .NET Core is starting to be adopted more and more in production projects and this library is sorely missed.

Thanks!

zvirja:
@klimisa @ecampidoglio I've rebased my branch of top of fresh `v4` - nothing was changed and I fixed the merge conflicts only.

@ecampidoglio I noticed that you were absent  for the last month - is everything OK? 😕Do you know when you will be able to get back to this and some other threads?

@mxa0079 I'm actively working on migration of the `AutoFixture` project to .NET Standard. Notice, those changes are not related to this particular PR, as this one is about pre-release feed.
I appreciate your readiness to help, but ATM I don't see how we could parallelize the work.. I hope to fire a PR with draft of the `AutoFixture` migration in upcoming days.

ecampidoglio:
@zvirja Thank you for asking! 😊 I've been occupied with work for the past month and I'm now enjoying some time off. I've been watching the threads and plan to pick them up when I get back in a few weeks.

zvirja:
@ecampidoglio Hope you have a great time off :blush: Do you know whether you will be able to participate here in the nearby future? We have a set of half-approved PRs and need a hero to proceed further :smile:

zvirja:
I've updated the PR description to affect all the latest changes and how the things go. Also I addressed all the concerns that we had - no special actions are required to support `vNext` - simply create a branch 😉

@klimisa Given that @ecampidoglio has been absent for a while and we don't know when he returns, could you please take a look on this if you have free time?

zvirja:
#773 will be merged very soon and there are people [who want to participate](https://github.com/AutoFixture/AutoFixture/issues/404#issuecomment-322195877) it early testing, which is pretty exciting. However, to provide people with a convenient way to do this we need this branch to be merged, so have a custom automated feed with the most recent builds of vNext.

Given that I've solved all the concerns that we had so far and we don't have anybody available to review this again (and didn't have for a while), @moodmosaic do you agree to make one-time exception and merge this as it is? Later we could rework or improve this if we find some issues with the approach. Otherwise, I'll need to manage and publish all this stuff manually and I don't want to spend my time on that (and don't have too much of it).

moodmosaic:
Build automation is not my expertise, thus I can't be of any help here. If I were in your shoes, I'd try to reach @ecampidoglio and even try to discuss around [this](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-323019240) comment I wrote in #703.

zvirja:
@moodmosaic I've tried to reach out him several times in different threads, however no luck 😞 Ok, let's try to solve that in that thread.

zvirja:
I'm merging this one as it has been agreed [here](703#issuecomment-323188537).

Great day for our project - we finally have a feed with pre-release packages for the vNext. Congratulations!

PR #763, opened 2017-06-14
Exclude revision from AssemblyVersion

zvirja:
Revision shouldn't be a part of assembly version as API is backward compatible during the bug fixes.

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -33,7 +33,7 @@ let calculateVersionFromGit buildNumber =
         getMatch "maj" |> int, getMatch "min" |> int, getMatch "rev" |> int, getMatch "pre", getMatch "num" |> int, getMatch "sha"
 
     
-    let assemblyVersion = sprintf "%d.%d.%d.0" major minor revision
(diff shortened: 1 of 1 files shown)


zvirja:
I preserve the MAJOR.MINOR.PATCH versioning for all the other parts - NuGet version, File Version, Informational Version and alter it for the `AssemblyVersion` attribute only.

The `AssemblyVersion` is an assembly attribute that indicates the API version for .NET. If it's changed - .NET will refuse to bind the updated version of an assembly. To solve that you will have to either recompile your code or add [Assembly Binding Redirection](https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions). Given that PATCH update is a bug fix and doesn't introduce any breaking changes in the API, usually it's omitted from the AssemblyVersion and MAJOR.MINOR are used only.

This is a common practice and you can read about it e.g. [here](https://stackoverflow.com/a/3905443/2009373) (the "What to set" section). The reason why initially I included it - I've simply overlooked that :)

moodmosaic:
OK, that sounds good then 👍 💯

PR #764, opened 2017-06-15
Restore AssemblyInfo files to their initial state after build

zvirja:
We patch `AssemblyInfo` files as a part of the build process. I've added target to restore those files to their original state after the build.

We need this PR because after the build we have a lot of dirty files in the git tree and it's pretty annoying to reset their state manually.


zvirja:
Thanks for the review 👍

Issue #760, opened 2017-05-12
Dependencies strategy for .NET Core switch


zvirja:
Currently AutoFixture [follows](https://github.com/AutoFixture/AutoFixture/blob/master/CONTRIBUTING.md#dependencies) a strategy when all the dependencies were included to the project in the `Packages` folder. It works well for the Full Framework because we can use the platform assemblies "for free". However, .NET Core (and .NET Standard) follows a bit different strategy and all the API you use should be installed via NuGet. As result, you have a lot of platform dependencies.

I played with migration of the AutoFixture project to .NET Core and noticed that after the NuGet restore the `Packages` folder started to take much more size.
Currently: 94MB
After switch to .NET Core: 320MB.

The delta is 226MB. The new size is 3.4 times larger.

Most of the libraries are not specific to our project somehow and, rather, are related to the `netstandard1.5` API we are going to target to.

Therefore, it's a good time to revisit our strategy and decide whether we are going to continue the approach we had before. The worst thing is that after new versions of .NET Standard might be released, they might increase all the dependency versions and our repository will blow up again.

Another thing is that after we migrate our projects to the .NET Core build system, they will target both `.NET Framework` and `.NET Standard` simultaneously. It will be impossible to keep local cache for `.NET Framework` only - build will if any dependencies are missing. Therefore, our solution should cover both targets simultaneously.

Personally, I see the following ways to follow:
### 1. Don't maintain the repository state more

The main reason why we maintained the repository cache is to be able to work without Internet immediately after checkout. We can continue to support work without Internet, however you will need to trigger a build before, so that the required packages are cached in local NuGet cache. 

Of course, this solution is not perfect and not all the cases are covered. For instance, if you have a custom branch with different dependencies, you should prefetch packages for it as well.

The benefit is that we will not blow up repository and that will simplify maintenance - given than number of packages now is much larger (about 65 system packages), it might be hard to keep track all them.

### 2. Try to change the strategy to store `nupkg` files only

We could try to investigate whether it would be possible to store `.nupkg` files only. After that when you trigger a NuGet restore, it should unpack files locally. The packed size is much smaller and takes `34.2MB` compared to `320MB`. However, we should still realize that if we update packages for a couple of time, that will add 100MB to the repository size and next git clone will take much longer.

I'm also not sure that NuGet restore will work fine without Internet even if all the necessary packages are present - that should be investigated.

### 3. Keep as is
[...]

PR #765, opened 2017-06-15
Add NuGet restore and remove dependencies from repository

zvirja:
It has been in agreed in #760 that we remove all the dependencies from the repo. In this PR I add a target to restore NuGet packages and remove all the packages from the git tree. Also I've updated NuGet binary to v4.1.0, so it can handle the .NET Core projects.

I need this change to migrate our projects to .NET Core. It seems to be the last prep PR before that migration.


zvirja:
@adamchester I know that you worked on .NET Core - do you have any concerns? :) I'm about to push another .NET Core related PR and I need to use this one as a base.

adamchester:
LGTM

PR #769, opened 2017-06-26
Migrate all projects to the .NET Core SDK

zvirja:
In this PR I've migrated all the projects to the .NET Core SDK (new `.csproj` file format). The projects target the same version of .NET as they targeted before (i.e. mostly `net 45`), and the only thing that was changed - the build system.

This change is a one step forward our migration to the .NET Core. I decided to migrate all the projects at once to make them look unified.

The following projects were not migrated:
- AutoFoq
- AutoFoqUnitTest
- Idioms.FsCheck
- Idioms.FsCheckUnitTest

The reason is that VS 2017 [doesn't support](https://github.com/Microsoft/visualfsharp/issues/2362) .NET Core for F# and that support should be added in VS 2017 Update 3 only. 

-----
P.S. Usually, the migration of each project consisted of the following steps:
1. Rewrite project to a new .NET Core format.
1. Include the `CommonProperties.props` file which contains the common properties.
1. Add all the project and package references. Remove the `packages.config` file.
1. Alter the `AssemblyInfo.cs` file and remove attributes which are generated automatically now.
1. Alter the `build.fsx` file and fix path to the release artifacts (add framework subdir).
----

@adamchester Would like to explicitly ask you for the review as you played with .NET Core already.


moodmosaic:
Looking forward to this 👍

@adamchester, @ecampidoglio, @klimisa, the changes look good to me. I scanned the diff output and noticed also that the build status is green.

adamchester:
It's looking good. I had a quick read and things look sensible. I can probably do a slightly more in depth look on the weekend if you think it's worthwhile and  if you're ok to wait

zvirja:
@adamchester Well, it's always great to have somebody looking at your approach and sharing his vision. So if you can allocate some time on the weekend to review this one - it would be great! 😎

Will be looking to the result in that case.

zvirja:
I've added one more commit to improve the `CommonProperties.props`. It doesn't change the logic somehow, so should not be an issue.

adamchester:
Hey @zvirja this is looking really good 👍 🥇 

Should we add a [global.json](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json). I have lots of versions of dot net CLI tools installed, and it's not obvious to me which version I (or even msbuild/VS) should be using to build and test this. Can you please introduce a `global.json` and set the sdk version?

zvirja:
Thanks for the review and your hints @adamchester!

Currently we don't use any specific features of SDK, so likely any SDK should work. The things might change after we bind to features that appeared in a particular version of SDK only (e.g. due to #770).

I'll merge this PR for now and will suggest to get back to this question later.

BTW, do you know, if we are basically SDK agnostic, which [exact version](https://github.com/dotnet/core/blob/master/release-notes/download-archive.md) should we specify? The lowest or the higher one?

Kralizek:
Anything from 1.0 should work. The previews of 1.0 were still using the old project.json

zvirja:
@Kralizek If you are familiar to this, could you clarify what would be the behavior if I set value e.g. to `1.0.1`? Does that mean that everybody will need to install the _exactly_ same version of SDK locally to compile the project? What if more recent SDK like `1.0.3` is installed - do you still need to install the olnder `1.0.1`?

Kralizek:
@zvirja yes. If you lock a certain folder (regardless of solutions or projects) to a certain SDK version via global.json, you will force everybody on having that exact version. I don't think it's possible to use version ranges in the global.json sdk attribute, so I would rather not use it.

By not using a global.json file, we are implicitly assuming that
* the person pulling the code has at least a RTM version of the .NET Core SDK
* there is no SDK 2.0 with major breaking changes
* we are using all functionalities available across the full 1.0 spectrum of the SDK, i.e. `[1.0,2.0)` in NuGet terms

Issue #770, opened 2017-06-30
Implement SourceLink approach to enable simpler source code stepping


zvirja:
Our current approach of distribution symbols and source code stepping possibility is to produce two NuGet packages:
- `Product.nupkg` - package with binaries only.
- `Product.symbols.nupkg` - package with binaries, pdb files and source code.

Later we upload package to the product and symbol feeds.

While this approach usually works (I wasn't able to get it working for the `3.50.2` though), there are some caveats with it. The major drawback is that it requires complex configuration of VS: an an additional Symbol Server should be added and a set of checkboxes should be enabled. I believe that most of our consumers doesn't even know about such a possibility.

Instead, I'd suggest to use the [Source Link](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/source_link.md) feature. The idea is that PDB files embed information where to look for the source code. In our case we'll point out directly to the GitHub (i.e. link for particular file will look like `https://raw.githubusercontent.com/AutoFixture/AutoFixture/{COMMIT}/Src/AutoFixture/Fixture.cs`). Later if you want to step into AutoFixture code, VS will simply grab source code from the GitHub repo. This feature is supported by VS 2017 and AFAIK is enabled by default (even if not - it's a single tick).

To embed information to the PDB files we can use the [SourceLink](https://github.com/ctaggart/SourceLink) tool. I've already tested it and it seems to work. The good news is that previously tool worked for the Portable PDBs only, however starting from VS 2017 Update 3 it [will be available](https://github.com/dotnet/roslyn/pull/18539) for the Windows PDBs as well.

Therefore, I'd suggest to change our distribution for `v4` so:
- we use the `SourceLink` tool to extend PDB files with source information.
- we extend the `Product.nupkg` package to include both binaries and PDBs. That will allow to enable source stepping for AutoFixture out of the box. While it will increase size of the package, I believe it's totally fine as we are in 2017 😄
- we still produce and upload the `Product.symbols.nupkg` package with PDB files and symbols. That should enable Source Stepping in older VS versions.

This feature is already used by [xunit](https://github.com/xunit/xunit/blob/bcea8b324628c17af340e9664fa7183820afb9e0/src/common.props#L23) and it works like a charm 👍

@AutoFixture/core Do you have any concerns regarding this? 😉

moodmosaic:
I have no experience with SourceLink, I've rarely used PDB files also 🤠  However, this sounds good to me:

[...]

Issue #745, opened 2017-04-02
Discussion: The 'Ploeh' namespace prefix


zvirja:
By opening this ticket I want to understand our position regarding the 'Ploeh.AutoFixture' namespace we currently have in AutoFixture and what is our vision regarding it in future. This question [already arose](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-255551871) and we will definitely meet it in future again :)

There are 2 ways how we could behave - keep it as is or remove the `Ploeh` prefix. Each option has this pros and cons.

## Keep as is
The idea is to keep the `Ploeh` prefix in all the namespaces for now, despite the fact that Mark isn't the owner of the repo more.

__Pros:__

1. That will immortalize the @ploeh's investments into this project as they were pretty large.
1. That will create no breaking changes during the migration to the `v4`.

__Cons:__
1. Very unclear ownership. As Mark stated [here](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-275415331), a lot of people still think that he is the owner of the repository. If we keep the namespace prefix, this confusion will be still present. 
On other side, if we remove it - his name will not be present in all the imported namespaces more, so after some time confusion will disappear.

1. Confusion from consumers. It looks similar to the point 1, but the difference is that not everybody knows who is the `Ploeh`. It might be unclear to the new (and existing) consumers why do we have this unusual prefix :)

## Change the prefix
The option is to alter all the files in all the projects and remove the `Ploeh` prefix. Technically it's easy to do, so question is only about our decision.

The pros and cons are the same as for the _Keep as is_ option:

__Pros:__
1. Ownership. It will be clear that AutoFixture is a standalone organization that manages the project by itself. The Mark (Ploeh) doesn't lead the project more. Also less people will think that this project belongs to Mark.

1. Simplified namespace. That will remove unnecessary word from the namespace, so will make namespace imports less verbose.

__Cons:__
1. This will be a huge breaking change as everybody who migrates to `v4` will be forced to change namespace imports. Also, it could happen that full type names are hardcoded somewhere as strings. Of course, it will be done only once and it's easy to do, but there are people who don't track all this ownership stuff and just use the AF - it will pretty boring for them.

---

First of all I'd like to hear the @ploeh opinion about this change and which way he personally prefers. You, Mark, invested a lot here so your vision does matter.

Also I'd like to involve guys from the core team: @ecampidoglio, @moodmosaic and @adamchester.

[...]

Issue #691, opened 2016-08-12
Cast operators shouldn't be used as factory method during specimen  construction


zvirja:
AutoFixture uses public static methods to detect the factory method for object construction. However, it appears that it also detects implicit/explicit cast operators and tries to use them.

Consider the following sample:

``` csharp
    public abstract class BaseClass
    {
      public static implicit operator BaseClass(int value)
      {
        return new DefaultImpl(value);
      }
    }

    public class DefaultImpl : BaseClass
    {
      private int _value;

      public DefaultImpl(int value)
      {
        _value = value;
      }
    }
```

If you ask Fixture to create `BaseClass`, it will create `DefaultImpl` instance successfully.
That happens because of [this code](https://github.com/AutoFixture/AutoFixture/blob/a3d21c3f67e49948c18cc8bbdf9f92a820fddd89/Src/AutoFixture/Kernel/FactoryMethodQuery.cs#L43) - operators are returned as static methods.

Is that expected behavior? I'd say that no, because that is very tricky and unclear behavior. In my opinion, cast operator should not be used as a way to construct instance. Moreover, if that is an explicit cast operator.

zvirja:
@AutoFixture/core Guys, if you have a free time - could you please examine this issue and confirm that it's a bug? I'm about to mark it as a bug and fix 😉

moodmosaic:
I've never thought about this when adding support for Static Factory Methods. Interesting.

What's wrong with leaving it the way it currently is though?

After all, we might be actually doing the right thing, specially in these situations where one has to [deal with primitive obsession](https://lostechies.com/jimmybogard/2007/12/03/dealing-with-primitive-obsession/).

zvirja:
Initially I observed that issue using the [System.IO.Abstractions](https://github.com/tathamoddie/System.IO.Abstractions) library. If you check [FileInfoBase.cs](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/FileInfoBase.cs) implementation you will notice the following:
```csharp
public static implicit operator FileInfoBase(FileInfo fileInfo)
{
	return new FileInfoWrapper(fileInfo);
}
```

They provide a way to cast a real `FileInfo` into the abstractions wrapper. It might be useful if you would like to pass object from legacy code into the other code written on the abstractions.

Now consider the following test:
```csharp
[Theory, AutoConfiguredNSubstituteData]
public void SomeTest(FileInfoBase fileInfo)
{
[...]

PR #773, opened 2017-07-03
Support .NET Standard by the AutoFixture project

zvirja:
This PR introduces support for .NET Core world by the AutoFixture ✨ `AutoFixture` is the only publishable project which is altered by this PR. This is related to the #404 and consumes all the changes from #740.

The list of the required changes:

- [x] Add `.NET Standard 1.5` as a target framework for the `AutoFixture` project.
- [x] Ensure that Portable PDB files are generated for .NET Standard and Windows PDB files for the full framework.
- [x] Investigate and run unit tests for the `AutoFixtureUnitTest` and NetCoreApp target (we have the [Albedo](https://github.com/ploeh/Albedo/issues/94) and [Exude](https://github.com/GreanTech/Exude/issues/21) dependencies which doesn't support NET Standard and xunit2).
    - [x] Update to xunit 2.2
    - [x] Fix `Exude` - might be replaced by the [`PropertyData`/`MemberData`](http://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/) attributes.
    - [x] Fix `Albedo` - we use it in a single place only, so could use a plain reflection instead.
    - [x] Fix all the .NET Core incompatibilities.
    - [x] Tune test runner to run tests for .NET Core.
- [x] Run unit tests for the `AutoFixtureDocumentationTest` and NetCoreApp target
    - [x] Upgrade xunit to 2.3
    - [x] Fix all the .NET Core incompatibilities.
    - [x] Tune test runner to run tests for .NET Core.
- [x] Update the `AutoFixture.nuspec` file and the build scripts to pack for all frameworks. Register an issue to use the `dotnet pack` instead of the `nuspec` files.
- [x] Investigate how to run code analysis for the .NET Standard application and enable it back.
    It appeared that if you simply add the `<EnableCodeAnalysis>true</EnableCodeAnalysis>`, FxCop will fail for the `netstandard1.5` framework. I temporarily disabled the code analysis for NET Standard. I tried to install the different versions of the [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/) package, but none of them worked fine.
- [x] Register an issue to re-enable the suppressed Code Analysis rules if needed.
- [x] Test that package is consumable by .NET Standard library and symbols work.
    __Result:__ Symbols are loaded for both .NET Desktop and .NET Core targets, however sources are loaded for the Desktop only. That should not be the issue as we are going to use [Source Link](https://github.com/AutoFixture/AutoFixture/issues/770). 

@adamchester and @Kralizek Could you please assist me with the Code Analysis part if you can? Do you know how to enable the FxCop for the `.NET Standard` framework? Likely, I simply don't know how to cook the `Microsoft.CodeAnalysis.FxCopAnalyzers` package...

@moodmosaic - Am I right that the `Exude` might be replaced by the `MemberData` attribute?


moodmosaic:
Yes, you may replace Exude with xUnit.net data theory attributes (like the ones you mentioned above). It'll be less type-safe, but I guess we can't do better since Exude isn't available for .NET Core (?)

BTW, great work so far 👍👌

Kralizek:
@zvirja never worked on code analysis before this :(

zvirja:
@moodmosaic, @adamchester, @ecampidoglio, @klimisa, @Kralizek I'm done with the migration and AutoFixture is .Net Core ready now 🎉

The list of changes you can see above. Most of the changes are insignificant and were introduced to fix .NET Core incompatibilities. However, there are a few changes which you should pay attention to:
- I've updated test projects to xunit 2.3 beta 3 and updated build scripts to use `dotnet xunit` to run tests for those projects. This approach guarantees that tests are run for all the targets. See more details in the [commit](https://github.com/AutoFixture/AutoFixture/pull/773/commits/bb3ec2d42e3bf4538c53c32fa3386f22bde7c69c).
- Target framework for the tests is `netcoreapp1.1`. That is because is `netstandard` is an agnostic framework and cannot be used to define the runtime we want to run tests on.
- As a part of the changes I needed to alter the logic inside the AutoFixture's [TypeEnvy.cs](https://github.com/AutoFixture/AutoFixture/pull/773/commits/fa810ff9a5c7a6a566c7c1fb71cdea584fc33228). That change looks logically equivalent and more appropriate, but please verify.
- I switched us to a new type of code analysis. I've installed the `Microsoft.CodeAnalysis.FxCopAnalyzers` package as this approach works for both Full Framework and Net Core targets (and is much quicker BTW). After the installation it appeared that we have a lot of warnings (around 100). I [tuned](https://github.com/AutoFixture/AutoFixture/pull/773/commits/5e338df33fcbf21e17c9ae2f25df58981ea75837) our ruleset and disabled rules which failed. After we approve this PR, I'll create a task to review warnings and re-enable code rules which we want to enforce - it would be too much for this PR :)

Looking forward to your code review. Explicitly request code review from you @adamchester 😉

arishlabroo:
Hey @zvirja , Any concerns with making branch ```zvirja/migrate-autofixture-to-net-core``` public so that I can check it out? Completely understand it is not ready, just want to follow along with your migration.

DixonDs:
@zvirja 
>Target framework for the tests is netcoreapp1.1. That is because is netstandard is an agnostic framework and cannot be used to define the runtime we want to run tests on.

It might make sense to multi-target to full .Net Framework as well.

zvirja:
@DixonDs Thanks for the hint - actually this is how it's [currently implemented](https://github.com/AutoFixture/AutoFixture/pull/773/files#diff-68899be54a998695611e1eb5bcccc6d0R6) 😉

moodmosaic:
>zvirja requested review from moodmosaic, adamchester, ecampidoglio, and klimisa and removed request for AutoFixture/core a day ago

@zvirja, 👍 💯 🥇 for all your hard work! This looks promising, but unfortunately I don't currently have experience with .NET Standard to a point where I could effectively review this pull request. Perhaps @adamchester, @ecampidoglio, or @klimisa could help?

zvirja:
@moodmosaic Thanks for the reply. Most of the changes are not related to the .NET Core itself - they are about using API which is available in both frameworks. Therefore, if you have free time, it would be cool if you could briefly review changes from that perspective:
- Change in the `TypeEnvy` class
- Changes in the unit tests which remove `Albedo` and `Exude` dependencies and usage of the `AppDomainSetup` like classes (which are not present).

One more pair of eyes will help us to be more sure that something important is not overlooked. Do you have time for that? 😊 😳

Build.fsx
```diff
@@ -133,12 +134,28 @@ Target "TestOnly" (fun _ ->
     let parallelMode = if parallelizeTests then ParallelMode.All else ParallelMode.NoParallelization
     let maxThreads = if maxParallelThreads = 0 then CollectionConcurrencyMode.Default else CollectionConcurrencyMode.MaxThreads(maxParallelThreads)
 
+    let netCoreXunitProjects = [
+        "AutoFixtureUnitTest"
+        "AutoFixtureDocumentationTest"
+    ]
```
moodmosaic: What about the other unit test projects?

Src/AutoFixture.ruleset
```diff
@@ -1,236 +1,247 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<RuleSet Name="AutoFixture" Description="Rule set for all AutoFixture projects" ToolsVersion="10.0">
+<RuleSet Name="AutoFixture" Description="Rule set for all AutoFixture projects" ToolsVersion="15.0">
   <IncludeAll Action="Warning" />
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
-    <Rule Id="CA1000" Action="Error" />
```
moodmosaic: I would expect non of the Rule Ids to change. What's the motivation behind these changes?

Src/AutoFixture/EmailAddressLocalPart.cs
```diff
@@ -6,11 +6,19 @@
...
     /// <summary>
     /// Represents the local part of the email address, defined as everything up to, but not including, the @ sign.  
     /// Since EmailAddressLocalPart is used in constructing MailAddress, enforcement of rules on a valid email address
     /// is performed by <see cref="System.Net.Mail.MailAddress"/> and not EmailAddressLocalPart other than as noted. 
     /// </summary>
+#else
+    /// <summary>
+    /// Represents the local part of the email address, defined as everything up to, but not including, the @ sign.  
+    /// Since EmailAddressLocalPart is used in constructing MailAddress, enforcement of rules on a valid email address
+    /// is performed by System.Net.Mail.MailAddress and not EmailAddressLocalPart other than as noted.
```
moodmosaic: XML Documentation `cref`s not working in .NET Standard?

Src/AutoFixtureDocumentationTest/Extension/Constraints/ConstrainedStringGeneratorTest.cs
```diff
@@ -39,7 +39,7 @@ public void CreateAnonymousWillPrefixWithSeed()
             // Exercise system
             var result = sut.CreateaAnonymous(seed);
             // Verify outcome
-            Assert.True(result.StartsWith(seed));
+            Assert.StartsWith(seed, result);
```
moodmosaic: Good catch 👍

Src/AutoFixtureUnitTest/AbstractRecursionIssue/Repro.cs
```diff
@@ -65,7 +65,7 @@ public void Workaround()
                     typeof(ItemBase),
                     typeof(FunkyItem)));
 
-            Assert.DoesNotThrow(() => fixture.Create<FunkyItem>());
+            AssertEx.DoesNotThrow(() => fixture.Create<FunkyItem>());
```
moodmosaic: Have we tested whether `AssertEx` works correctly? Are we sure we're not running into [False Positives](http://xunitpatterns.com/false%20positive.html) or [False Negatives](http://xunitpatterns.com/false%20negative.html)?

Src/AutoFixtureUnitTest/DomainNameGeneratorTest.cs
```diff
@@ -54,7 +54,7 @@ public void CreateReturnsOneOfTheFictiousDomains()
             var result = sut.Create(typeof(DomainName), null);
             // Verify outcome
             var actualDomainName = Assert.IsAssignableFrom<DomainName>(result);
-            Assert.True(Regex.IsMatch(actualDomainName.Domain, @"example\.(com|org|net)"));
+            Assert.Matches(@"example\.(com|org|net)", actualDomainName.Domain);
```
moodmosaic: 👍 💯 🥇

Src/AutoFixtureUnitTest/EnumGeneratorTest.cs
```diff
@@ -103,7 +103,7 @@ public void RequestForFlagEnumTypeReturnsCorrectResult(Type enumType, int reques
             var sut = new EnumGenerator();
             // Exercise system
             var dummyContext = new DelegatingSpecimenContext();
-            var result = Enumerable.Repeat<Func<object>>(() => sut.Create(enumType, dummyContext), requestCount).Select(f => f()).Last();
+            var result = Enumerable.Repeat<Func<object>>(() => sut.Create(enumType, dummyContext), requestCount).Select(f => f()).ToArray().Last();
```
moodmosaic: Was `ToArray()` necessary?

Src/AutoFixtureUnitTest/EnumGeneratorTest.cs
```diff
@@ -56,7 +56,7 @@ public void RequestForEnumTypeReturnsCorrectResult(Type enumType, int requestCou
...
```
moodmosaic: Was `ToArray()` necessary?

Src/AutoFixtureUnitTest/Kernel/CriterionTests.cs
```diff
@@ -94,9 +94,9 @@ public void SutEqualsIdenticalValue()
         [InlineData(typeof(Version))]
         public void SutDoesNotEqualAnyObject(object other)
         {
-            var sut = new Criterion<PlatformID>(
-                PlatformID.Unix,
-                new DelegatingEqualityComparer<PlatformID>());
+            var sut = new Criterion<ConsoleColor>(
+                ConsoleColor.Cyan,
+                new DelegatingEqualityComparer<ConsoleColor>());
             var actual = sut.Equals(other);
```
moodmosaic: I guess, all these changes are done because some objects aren't available in .NET Standard?

Src/Empty.ruleset
```diff
@@ -0,0 +1,601 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
moodmosaic: What's the purpose of this new .ruleset?

Build.fsx
zvirja: I run these unit tests using `dotnet xunit` that is available only if particular packages are installed to the project. Currently only 2 ones are supported. Therefore, only selected projects are compatible ATM. I expect that eventually each test project will be compatible, so we could get rid of the list.

Src/AutoFixture.ruleset
zvirja: I marked all "Error" as "Warning" in [this](https://github.com/AutoFixture/AutoFixture/pull/773/commits/d75c05d479fe11bbb8ba4d1e3ffd857e0f1b270a) commit. The motivation is that only `Verify` build should fail if there are some code warnings and all other builds should pass. During the `Verify` build we treat warnings as errors, so we will still don't allow to merge code which violates rules.

Also notice, as I mentioned [here](https://github.com/AutoFixture/AutoFixture/pull/773#issuecomment-313640629) I suppressed some part of rules which started to fail after I switched to a new Code Analysis.

Src/AutoFixture/EmailAddressLocalPart.cs
zvirja: No, everything is fine with `cref` in .NET Standard. The issue is that `System.Net.Mail.MailAddress` class doesn't exist there and our build verification fails because of that ☺️Therefore, I made two versions of the comment - one with cref and other without.

Src/AutoFixtureDocumentationTest/Extension/Constraints/ConstrainedStringGeneratorTest.cs
zvirja: Thanks, but that's not me - those are XUnit [code analyzers](https://github.com/AutoFixture/AutoFixture/pull/773/commits/a451db59ca0708ea98cb346a4011307af701ed52) that are now present 😳

Src/AutoFixtureUnitTest/DomainNameGeneratorTest.cs
zvirja: Cudos to [xunit code analyzers](https://github.com/AutoFixture/AutoFixture/pull/773/commits/a451db59ca0708ea98cb346a4011307af701ed52) 👋

Src/AutoFixtureUnitTest/AbstractRecursionIssue/Repro.cs
zvirja: I've created this method because [DoesNotThrow](https://github.com/xunit/xunit/issues/188) method has been removed in xunit 2.0 and I wanted to quickly migrated all the code.

Do you want me to create a test for the `AssertEx` method? Also I could replace this with the following pair as advised [here](https://github.com/xunit/xunit/issues/188#issuecomment-86584802):
```
var ex = Record.Exception(() => sut.Foo());
Assert.Null(ex);
```

Src/AutoFixtureUnitTest/EnumGeneratorTest.cs
zvirja: Yes, see [this](https://github.com/AutoFixture/AutoFixture/pull/773/commits/764f10899fb8ee25ef20f38c9fc4041122c1a783) commit message.

Src/AutoFixtureUnitTest/EnumGeneratorTest.cs
zvirja: Yes, see [this](https://github.com/AutoFixture/AutoFixture/pull/773/commits/764f10899fb8ee25ef20f38c9fc4041122c1a783) commit message.

Src/AutoFixtureUnitTest/Kernel/CriterionTests.cs
zvirja: Yep. The `PlatformID` and a few other types are missing, so I tried to find a replacement for them. In this particular example I replaced one enum with other one (while that still doesn't matter for the test).

That was the most complex part of the migration - to ensure that replacements are relevant.

Src/Empty.ruleset
zvirja: I specify this ruleset for all our tests (see [here](https://github.com/AutoFixture/AutoFixture/pull/773/commits/d75c05d479fe11bbb8ba4d1e3ffd857e0f1b270a#diff-948806e36ff188d70085b17290b13be1R7)). That is the only way I found to disable code analysis for the test projects.

Previously, we controlled that via the `RunCodeAnalysis>true</RunCodeAnalysis>` csproj property, but that no longer works .

zvirja:
@moodmosaic Thank you very much for the time and review! 🙏I know that this PR is pretty large, so it wasn't easy 😅

I replied to your comments. Please mark the resolved ones with 🎉 mark :)

Src/AutoFixtureUnitTest/AbstractRecursionIssue/Repro.cs
moodmosaic: >Also I could replace this with the following pair as advised [here](https://github.com/xunit/xunit/issues/188#issuecomment-86584802):
>```
>var ex = Record.Exception(() => sut.Foo());
>Assert.Null(ex);
>```

I think that's better, more straightforward than adding `AssertEx` 😉 What about

```
Assert.Null(Record.Exception(() => sut.Foo()))
```
?

zvirja: Yep, agree. Replaced with what you suggested.
I created a standalone commit rather than rebase to not mess up the existing commits (and links to them).

zvirja:
@adamchester I know that you're probably busy these days, but will you have any time in the nearby future to check this PR? I want to migrate other projects as well, do this one is blocking me a bit... :-(

zvirja:
@adamchester Sure, no problem! Thanks ;-)

zvirja:
@adamchester You mentioned that you could probably have free time over the upcoming weekend to review this PR. In this case, could you please also review the other PRs as well, so I can proceed further (they are already approved by @moodmosaic but we agreed to require 2 approvals) 😊

Thanks in advance ☕️

AlexDenton:
Just here to offer some words of encouragement. I was working on a side project when I realized AutoFixture didn't work with dotnet core. I came here only and found you guys were seemingly days from having it ready. Thanks for doing what you do! You guys are great!

zvirja:
@adamchester It seems that you didn't have time the previous weekend to check this 😟 Do you know whether you will have time during the upcoming weekend? :flushed: By no means I want to rush you, rather just want to know whether it makes sense to wait for the review in the nearby future or not..

P.S. I'd promise you some beer, but am not sure how quickly I'll be in Australia 😓

Src/AutoFixture/Kernel/TypeEnvy.cs
```diff
@@ -11,17 +11,12 @@ internal static class TypeEnvy
...
-            if(parameter != null)
-            {
-                return parameter.GetCustomAttribute<TAttribute>();
-            }
-            return null;
+            var attributeProvider = candidate as ICustomAttributeProvider;
+
+            return attributeProvider?
+                .GetCustomAttributes(typeof(TAttribute), true)
+                .Cast<TAttribute>()
```
adamchester: I'm not exactly familiar with how this method is used now, but it's worth briefly considering the performance impact of LINQ statements.

Src/AutoFixture/TypeFilter.cs
```diff
@@ -1,5 +1,10 @@
 ﻿using System;
 using Ploeh.AutoFixture.Kernel;
+#if NETFULL
```
adamchester: Would it be better to control this with something a little more granular than `NETFULL`?

Src/CommonProperties.props
```diff
@@ -7,30 +7,51 @@
...
+        <IsFullFramework>false</IsFullFramework>
+        <IsNetCore>true</IsNetCore>
+        <DefineConstants>$(DefineConstants);NETCORE</DefineConstants>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <PropertyGroup>
+        <IsFullFramework>true</IsFullFramework>
+        <IsNetCore>false</IsNetCore>
+        <DefineConstants>$(DefineConstants);NETFULL;SYSTEM_NET_MAIL;SYSTEM_RUNTIME_SERIALIZATION</DefineConstants>
```
adamchester: I note that we've split some conditionals into namespace/area, and then we have a `NETFULL` which seems to mash together a lot of different namespaces/areas. Is there any logic behind that?

Src/AutoFixture/TypeFilter.cs
zvirja: Done!

Src/CommonProperties.props
zvirja: Good catch! I did that because it was simpler and there were a lot of standalone places that contained switches. But you are right - it's better to explicitly define which particular features we are switching on. I've split `NETFULL` to the more specific switchers.

Src/AutoFixture/Kernel/TypeEnvy.cs
zvirja: Well, I could rewrite this to avoid LINQ usage, but currently I don't see how this could affect performance, given that we don't have any closures here. Do you have some particular issue in mind?

P.S. We have also a similar usage [here](https://github.com/AutoFixture/AutoFixture/blob/96fab6bfd5c5e6b4967ff92cbf04e9a7c96c5e91/Src/AutoNSubstitute/SubstituteAttributeRelay.cs#L31-L40) and no issues were reported so far.

zvirja:
@adamchester Thanks for your time 👍 I've fixed some of your concerns and replied to other. Will be looking to your follow up and when you have time to finish the review.

MihaMarkic:
@zvirja linq usually carries some performance hit regardless of closures. If you want better performance, do use loops instead. If you want code readability, then use linq.

zvirja:
@MihaMarkic You are speaking about very negligible performance overhead, which should never be the case unless you write some performance-critical or hot parts of code and have profiling results proving that this is a bottleneck. AutoFixture uses reflection without caching all over the way, which impacts performance much more to worry about general LINQ slowness 😉 I thought that @adamchester has some particular case in mind.

P.S. Could you please answer under the appropriate thread in future, so it's easier to track conversations? 😊

Src/AutoFixture/Kernel/TypeEnvy.cs
adamchester: There were a number of core classes that were used a lot on the 'hot path'. Back in 
#218 we removed some of the LINQ statements in favour of less allocations. I'm not requesting that you should change this, but I thought it's worthwhile briefly considering if it might have any impact.

We left a few comments in various key places like [this one]( https://github.com/AutoFixture/AutoFixture/pull/218/commits/0d9d95f1e4ea0008a7f68d46a67756cd37035d52).

zvirja: @adamchester Thanks for the reply. Now I better understand your concerns given that you fixed the similar issues in past. 

I've profiled our unit tests and found that this code is pretty hot:
image

Therefore, I've refactored it to avoid Linq usage. I'm still not convinced that it worth it, but let's proceed with this premature optimization if we did something similar in past :)

sjuberman:
Are you planning on creating a pre-release once this issue has been resolved?

zvirja:
@sjuberman Likely, not yet. There are a bunch of other project that don't support .Net Standard yet, so we should migrate them as well. Instead, we [have a plan](762) to set up a pre-release feed on MyGet. Everybody will be able to consume the early builds of the `v4`.

zvirja:
After I merged #777 build is green again. @adamchester Looking forward to the rest part of your review to address all your concerns and finally merge this one 😉

zvirja:
@adamchester Today we has been [asked](https://twitter.com/borlip/status/895443044874334213) about this feature again 😏 Do you know whether you will have a luck to allocate some time in the nearby future and finalize your code review? 😊 We have 9 more projects, but I cannot proceed as I want to use this one as a sample..

Thank you.

adamchester:
@zvirja I haven't had time to get to this, and it's hard to make any promises regarding when I could look at it in the near feature. Please push forwards without my review. I tried to get to this one over the past few weeks, sorry I wasn't able to do more. Great work here btw!

zvirja:
@adamchester Thanks for the update and no worries :+1: Thank you for what you have done so far! 

@moodmosaic Given that you also reviewed this PR in past and the whole conversation above, what would you suggest? Do you agree to merge the PR, or you feel that it's not enough and we should wait till Adam has free time?

moodmosaic:
>@moodmosaic Given that you also reviewed this PR in past and the whole conversation above, what would you suggest? Do you agree to merge the PR, or you feel that it's not enough and we should wait till Adam has free time?

I think—but please correct me if I am wrong—that @adamchester, @ecampidoglio, and @klimisa aren't going to be around to review this. And I'm not  experienced with supporting .NET Standard 😃

So I would suggest that you get hold of the actual users who requested this feature. They probably have some experience/interest on this, since they've asked for it.

---

On a side note, I have a gut feeling that #703 must be revised, and move to a new model, probably similar to [the previous one we had](http://oss-watch.ac.uk/resources/benevolentdictatorgovernancemodel).

zvirja:
>  think—but please correct me if I am wrong—that @adamchester, @ecampidoglio, and @klimisa aren't going to be around to review this. And I'm not experienced with supporting .NET Standard 😃

Well, there is a chance that @adamchester will have free time in the future, however it's unclear when exactly 😟 In fact .NET Standard support is all about usage of the compatible API that is present in both worlds. Therefore, if you see that code changes are fine for you, we could probably proceed. There is very little about infrastructure and that has been [already approved](769).

> So I would suggest that you get hold of the actual users who requested this feature and do a review. 

Probably, we could invite @Kralizek to review this one and let us know if he sees any issues. If not - proceed with this one.

This feature is being constantly requested, so we should make a strategy how to move on. Other popular testing projects supports that (like xUnit, NUnit, NSubstitute, Moq, FakeItEasy) so we shouldn't be a blocker for our users (and to not lose our audience :)). We cannot postpone this too much, as we have other projects to migrate. I've already applied some transformations to our project (like #769, #765, #780) and I have a vision how our projects are organized, so I could ensure a consistency.

I'd suggest the following plan:
1. Ask @Kralizek to review the support whether he sees some issues.
2. Fix all the concerns (if any).
3. If everything is fine - merge the PR and continue with support of other projects.

@moodmosaic Do you agree with such a plan? 😉

Given that AutoFixture [is all about feedback](https://github.com/AutoFixture/AutoFixture/issues/691#issuecomment-311272410), I believe it's better to release how we see it and improve that via feedback, than simply hold on things because we don't have enough expertise. After all, we could release a beta after we finish, so users could come to us and let us know about issues they discovered - it's not possible to foresee all them.

--------
> On a side note, I have a gut feeling that #703 must be revised

Could you please describe how you see it in #703, so we could discuss it there? I also agree that it could be improved, because the existing model assumes that all the maintainers are active, while in fact we have only me, you and partially @adamchester. Given that I cannot self-approve own PRs, it's [sometimes](https://github.com/AutoFixture/AutoFixture/pull/762) very hard to proceed.

moodmosaic:
>I'd suggest the following plan:
>1. Ask @Kralizek to review the support whether he sees some issues.
>2. Fix all the concerns (if any).
>3. If everything is fine - merge the PR and continue with support of other projects.
>
>@moodmosaic Do you agree with such a plan? 😉

Looks reasonable to me.

---

>> On a side note, I have a gut feeling that #703 must be revised
>
>Could you please describe how you see it in #703, so we could discuss it there? I also agree that it could be improved, because the existing model assumes that all the maintainers are active, while in fact we have only me, you and partially @adamchester. Given that I cannot self-approve own PRs, it's [sometimes](https://github.com/AutoFixture/AutoFixture/pull/762) very hard to proceed.

I'd rather continue discussing about that _in #703_.

zvirja:
@Kralizek Will you be able to allocate some time in the nearby future and check this PR whether it looks fine from the .NET Standard support perspective? :)

Kralizek:
@zvirja I'll try to get some time in the next days for a final check. I have been lurking your work on the v4 during the latest months so I am not falling from the sky on this.
I'll get back to you latest on Monday (this is really a busy week)

Anyway thanks for the ping on Twitter :)

zvirja:
@Kralizek Awesome, happy to see your readiness and commitment 😌 Will be looking for the review to finally proceed with other projects as well 😉

Kralizek:
@zvirja excuse my infinite n00bness, should I just pull the v4 branch or something else?

zvirja:
@Kralizek No worries :) The simplest way would be to fetch this PR as described [here](https://help.github.com/articles/checking-out-pull-requests-locally/):
```
git remote add upstream git@github.com:AutoFixture/AutoFixture.git
git fetch upstream pull/773/head:pr_core_migration 
git checkout pr_core_migration
```

Will be looking to your feedback 😉

Kralizek:
@zvirja is it expected to build? I ran `.\build.cmd` and I got some errors
[output.txt](https://github.com/AutoFixture/AutoFixture/files/1234409/output.txt)

zvirja:
@Kralizek Yep, it's supposed to build. It seems that Code Analysis became broken after VS 2017.3 update - I can reproduce it locally now.

Please continue your investigation and remove the following lines in the `Src\CommonProperties.props` file:
```
  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.0.0-beta2" />
  </ItemGroup>
```

zvirja:
@Kralizek Fixed - it appeared that other version of package is required for VS 2017.3. Please re-fetch the PR.

Kralizek:
Broken in a different way now :S

[error.txt](https://github.com/AutoFixture/AutoFixture/files/1234899/error.txt)

zvirja:
Everything is fine on the CI, so its likely to be something environment specific. Ensure that git is present in the PATH variable (you can call it via Windows console). 

If that doesn't help, could you attach the full build log? I see a single message only. 

Also don't forget to run `git clean -fdx . ` after you re-fetch the PR.

zvirja:
@Kralizek Have you managed to make it work? Let me know if you still have some difficulties with local project setup and I'll gladly help you, so you can proceed with actual review. :wink:

Kralizek:
@zvirja different machine, same error. 
Sorry for the lack of information in the previous upload. I used "build.cmd > error.txt" and I didn't check the content. :)

Here is the file with the output from my home computer.
[error.txt](https://github.com/AutoFixture/AutoFixture/files/1236898/error.txt)

zvirja:
@Kralizek It appeared that your fork contains 2 obsolete tags that had invalid format (`v.2.16.0
` and `v.3.21.1`) and they break the version calculation logic. I've removed them from our repository some time ago (fixed it with valid value), however you forked the project earlier and have them.  I've extended the build script logic to not fail in such cases and simply ignore the dot. However, if you wish, you can delete them from your fork as described [here](https://stackoverflow.com/a/5480292/2009373).

Please re-fetch the PR and proceed - it should be fine this time.

zvirja:
@Kralizek Does everything works fine now? 😉 Do you know whether you will have time to the finish review in the nearby future?

Thank you a lot for your time!

Kralizek:
Finally got it to work with the help of @zvirja :)

I'll be spending some time this evening (I'm in Stockholm) for a more detailed review.
The first thing I noticed is that we have two variables `IsFullFramework` and `IsNetCore`.
I'd consider merging the two boolean variables into a single variable with well-known values. Something like `NetRuntime` which has values `NetFramework` and `NetCore`. This would allow in future to extend to other runtimes without too much hassle (e.g. Xamarin and Mono).

On the other hand, I got the feeling that in reality we will be targeting .NET Framework for legacy reasons and .NET Standard for the future. That being the case, I would change the variable from `IsNetCore` to `IsNetStandard`.

zvirja:
> Something like NetRuntime which has values NetFramework and NetCore.

Well, it's a good idea, however I'd probably call it `FrameworkFlavor` with values `full` and `core`.

> That being the case, I would change the variable from IsNetCore to IsNetStandard.

We set `IsNetCore` if target framework is either `NetStandard` (e.g. `AutoFixture` project) or `NetCoreApp` (e.g. `AutoFixtureUnitTest` project). Therefore, `IsNetCore` name sounded better to reflect the actual things.

Kralizek:
For me we're good to go :)
Project files look great and I was able to write few small programs with the assemblies generated from the build.

Just a question about the Packages/ folder. Are we still using it?

zvirja:
@Kralizek Thanks for the testing and your help! Happy to know that it works! 

Now lets see what @adamchester thinks about variable and we could go further. 

As for the `Packages`, we excluded it from the repo some time ago. Now this folder is created during the NuGet restore locally and is never committed.

adamchester:
I don't think that `IsNetCore` or `core` is even the right thing here. If we are targeting `netstandard` we should use that name. If we use net core as it currently stands, then in some cases we could be running on full framework with code conditionally included under the `netcore` name, and that could be quite confusing.

Regarding framework flavour, I think there are existing sdk variables we can reuse, but I can't find them right now.

zvirja:
@Kralizek @adamchester Thanks for your concerns. I've decided to completely remove the aforementioned variables and how I use `TargetFramework` variable directly - it's more clear.

zvirja:
Merging this one as that has been agreed [here](773#issuecomment-322164882).

mwhelan:
Hi, I am working on the TestStack.Dossier NuGet package which takes a dependency on AutoFixture. Thanks for putting the alpha versions on myget. It has helped us upgrade our package to netstandard1.5. In order to release it to NuGet though, it would be very helpful if you could make a pre-release version of AutoFixture 4 available on NuGet. Would that be possible? What would be the timeframe for publishing a version to NuGet? Thanks

zvirja:
@mwhelan Thanks for your interest to AutoFixture 😊

Publish of the pre-release version to the NuGet is a bit painful question to me. While we started to support .NET Core for [most of the projects](https://github.com/AutoFixture/AutoFixture/blob/0affdb3/README.md#net-platforms-compatibility-table), we have a plan to [change the default namespace](https://github.com/AutoFixture/AutoFixture/issues/745). This would be a huge breaking change for our clients and all the client code will stop to compile. This is the point when the confusion appears:
- if we release `alpha` with existing namespace and change namespace in between the `alpha` and RTM, it will confuse our clients as they already saw v4 with existing namespace. I want our clients to feel that v4 was always released with a new namespace as governance model was changed.
- if we release `alpha` with changed namespace, we'll be unable to merge `master` to `v4` branch without pain anymore. Currently we commit to both the branches, that's why I'd like to defer namespace change to the end. Another point is that our documentation is not currently ready, so people might simply not understand why all the namespace imports are invalid and that they need to simply run a text replace. That will make them scary and experience will not be as smooth as I wish.

The best way to solve this situation would be to not release `alpha` and release the RTM only. Unfortunately, I cannot provide you with precise ETA as it depends on my capacity (I'm working on this project in my free time) and availability of other contributors (my PRs should be reviewed 😉). In my head I imagine the release in a month or two and hope it will happen within this range. This project was dead for about 9 months due to governance model change - that's the main reason of such a delay in support.

Issue #687, opened 2016-08-03
Upgrading to NuGet 3.0: AutoFixture xUnit Theory Does Not Get Discovered


Mike-EEE:
I have recently upgraded my solution projects to NuGet 3.0 (from packages.config to project.json).  Everything seems to work with the exception of xUnit Theory with AutoFixture applied to it.  I have created a very simple example of this here:
https://github.com/Mike-EEE/Stash/tree/master/ReSharper.NuGet30

If you try building, you should, uh, discover that the tests do not get discovered. :stuck_out_tongue: 

If you [comment out these lines](https://github.com/Mike-EEE/Stash/blob/master/ReSharper.NuGet30/ReSharper.NuGet30/Class1.cs#L7-L11) and try building again, the tests get discovered.

Is this a known issue?  Am I doing something terribly wrong?  Maybe a little of both? :smile:

ecampidoglio:
[Answered](http://stackoverflow.com/a/38764154/26396) on StackOverflow.

However, it might be worth investigating whether NuGet _can) be configured to automatically add binding redirects without resorting to the installation script. Last time I checked it didn't but that was a while ago (#8).

Mike-EEE:
Yes thank you so much @ecampidoglio!  Much appreciated.  I am thinking of opening an issue on xUnit as well to maybe update their documentation as well.  I can't believe I am the first person to run into this. :stuck_out_tongue: 

I am curious if NuGet has a UserVoice?  This seems like it would be a very very popular request.

ecampidoglio:
No worries, happy to help :smile: 

This problem has nothing to do with xUnit. It's just that [AutoFixture.Xunit2](https://www.nuget.org/packages/AutoFixture.Xunit2) is compiled [against xUnit 2.0](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/packages.config#L3-L4), so you're going to need a binding redirect if you want to use it with [xUnit 2.1](https://www.nuget.org/packages/xunit/2.1.0).

[...]

PR #780, opened 2017-07-13
Use dotnet core pack to generate NuGet packages

### LINKED ISSUE #774: Use `dotnet pack` to generate NuGet packages
Currently we use `nuspec` files to generate NuGet packages. That is not very convenient as we need to duplicate dependencies in both `csproj` and `nuspec` files. The `dotnet pack` solves that issue as all the package metadata is extracted from the `csproj` file only.

zvirja:
Closes #774.

The main change is that now I use dotnet core pack (the `Pack` [MSBuild target](https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets)) to produce NuGet packages based on the `csproj` files, without need for the standalone `nuspec` files. That allows to simplify the maintenance as dependencies for all target frameworks are generated automatically.

The significant applied changes and notes:
- The `AutoFixture.AutoFoq`, `AutoFixture.Idioms.FsCheck.nuspec` and `AutoFixture.NUnit2.nuspec` projects were not migrated and still rely on the nuspec files. Reasons:
    - F# projects are not supported by .NET Core yet (should be fixed by VS 2017 Update 3). I've added an empty `Pack` target to them to not fail during the Pack.
    - `AutoFixture.NUnit2` package is non-trivial and includes compilation result of 2 projects. That is not [currently supported](https://github.com/NuGet/Home/issues/3891). I tried to work around the issue, but solution was too dirty, so I decided to keep as is.
- Nuspec files were deleted and their information was moved to the corresponding `csproj` files. Notice, if a description ended with a link to the AutoFixture repository, I stripped the link (e.g. deleted `Read more at http://autofixture.codeplex.com/`). That is because the Project URL is embedded to a package itself, so any UI allows to easily jump to the project site.
- I've extended the `Authors` list for each package with `AutoFixture` (e.g. before: `Mark Seemann`, after: `Mark Seemann,AutoFixture`). That is to reflect a new governance model.
- For some packages I tuned the `PackageReference` version. That is to match a version that was embedded to their nuspec files.
- I deleted the `Install.ps` file from all packages as it was agreed [here](https://github.com/AutoFixture/AutoFixture/issues/687#issuecomment-291772519).

Looking forward to the review results 😉


zvirja:
@moodmosaic Thanks 👍 Done, registered #782 and #783.

zvirja:
I've updated this PR and changed NUnit2 project generation as well. Now we have F# projects only that need `nuspec` files.

To simplify merge I'll complete this one after the #773.

PR #781, opened 2017-07-17
Fix one customization could affect other customizations

### LINKED ISSUE #321: Using Customize<T>(..) for defaults
I've been using AutoFixture for sometime now and only recently discovered `Customize<T>(..)`. Which, by my understanding can be used for setting build/generation rules on a higher level. To avoid having to set With and Without on each `Build<T>(..)`.

My question is:
If I customize with: `Fixture.Customize<MyObj>(x => x.Without(y => y.ID))`
and later I wish to set the ID property with a `Build<MyObj>().With(x => x.ID)`, it throws an exception saying "The specimen created for assignment is not compatible with System.Int64". Is this a bug or expected behavior? I can set the property with `Build<MyObj>().With(x => x.ID, fixture.Create<long>())`, but that means more typing ^^

Is there another way of setting default values?
Sorry if I am posting this on the wrong forum.

### LINKED ISSUE #531: Customizing one child class impacts other child classes
Consider this test code:

``` C#
public class MyBase
{
    public string Id { get; set; }
}

public class ChildA : MyBase { }

public class ChildB : MyBase { }

[Fact]
public void Repro()
{
    var fixture = new Fixture();
    fixture.Customize<ChildA>(c => c.With(x => x.Id, "foo"));

    var b = fixture.Create<ChildB>();

    Assert.NotNull(b.Id); // This assertion fails
}
```

Surprisingly, as the comment explains, the test fails.

As far as I can tell, this happens because [the implementation of `With`](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Dsl/NodeComposer.cs#L337-L368) creates an `Omitter` with a `MemberInfoEqualityComparer`.

`MemberInfoEqualityComparer` uses `DeclaringType` to compare two members, and since `Id` is declared on `MyBase`, this impacts all `MyBase.Id` properties, while it should have impacted only `ChildA.Id`.

While it'd seem that the most straightforward fix would be to instead use `ReflectedType`, apparently [this property isn't going to be in .NET Core](https://github.com/dotnet/coreclr/issues/926#issuecomment-101032799), so for the sake of future compatibility, we should look for other options.

### LINKED ISSUE #772: Build does not populate properties that have customizations elsewhere
Using Build along with customizations results in a an unset property on the built object. In this example, I would expect either "Cat" or Guid.NewGuid(), but not null.

Observe:

```
[TestClass]
public class AutoFixtureTest
{
    public class TestObject
    {
        public string A { get; set; }
        public string B { get; set; }
        public string C { get; set; }
    }

    [TestMethod]
    public void ShouldPopulateABC()
    {
        var fixture = new Fixture();
        fixture.Customize<TestObject>(x => x.With(p => p.A, "Dog"));
        fixture.Customize<TestObject>(x => x.With(p => p.C, "Cat"));

        var testObject = fixture.Build<TestObject>().With(x => x.A, "Zebra").Create();

        Assert.AreEqual("Zebra", testObject.A, "A is Zebra.");
        Assert.IsNotNull(testObject.B, "B is null.");
        Assert.IsNotNull(testObject.C, "C is null.");
    }
}
```
Results in: 
```
Assert.IsNotNull failed. C is null.
```

zvirja:
This fixes #321, fixes #531 and fixes #772.
Currently when we customize any property (e.g. using `.Without(x => x.Prop)` or `.With(x => x.Prop, value)`), we add a global `Omitter` for that property. This leads to the issue if property belongs to the parent class, so when an instance of sibling type is created that property is not filled.

I reworked the way, so that we don't create a global `Omitter` more. Rather, we customize a local AutoProperties PostProcessor and exclude a particular field/property from there. As result, customization is local to our graph and never interfere with other customizations (and the `Build()` graph).

There are modified tests because they assert the particular graph of nodes that was modified (in fact, they are structural tests). No behavioral tests failed.

----
Some internal changes
- I configure the `AutoProperties` command's specification to set the list of members that should not be filled.
- I create the `AutoProperties` postprocessor when it's needed for the first time to store information about skipped properties. If autoproperties are not enabled, the postprocessor is present in tree, but it's in disabled state (via `postProcessor.Specification == FalseRequestSpecification`).
- When we call the `OmitAutoProperties()` I don't remove the `AutoProperties` postprocessor more. Rather, I disable it via specification. See reason in code comments.

P.S. I decided to use `v4` as a base because I change the customization graph. If you feel that this is not a breaking change, I could rebase my PR on top of master for the quicker delivery 😉

### DIFF
--- a/Src/AutoFixture/Dsl/NodeComposer.cs
+++ b/Src/AutoFixture/Dsl/NodeComposer.cs
@@ -212,8 +212,7 @@ namespace Ploeh.AutoFixture.Dsl
         /// </returns>
         public IPostprocessComposer<T> Do(Action<T> action)
         {
-            var graphWithoutSeedIgnoringRelay =
-                this.WithoutSeedIgnoringRelay();
+            var graphWithoutSeedIgnoringRelay = WithoutSeedIgnoringRelay(this);
 
             var container = FindContainer(graphWithoutSeedIgnoringRelay);
             var autoPropertiesNode = FindAutoPropertiesNode(container);
@@ -229,19 +228,6 @@ namespace Ploeh.AutoFixture.Dsl
             return (NodeComposer<T>)graphWithDoNode.ReplaceNodes(
                 with: n => n.Compose(n.Concat(new [] { new SeedIgnoringRelay() })),
                 when: filter.Equals);
-        }
-
-        private static ISpecimenBuilderNode FindAutoPropertiesNode(
-            ISpecimenBuilderNode graph)
-        {
-            return graph.FindFirstNodeOrDefault(IsAutoPropertyNode);
-        }
-
-        private static bool IsAutoPropertyNode(ISpecimenBuilderNode n)
-        {
-            var postprocessor = n as Postprocessor<T>;
-            return postprocessor != null
-                && postprocessor.Command is AutoPropertiesCommand<T>;
         }
 
         private static NodeComposer<T> WithDoNode(
@@ -269,16 +255,22 @@ namespace Ploeh.AutoFixture.Dsl
         /// </returns>
         public IPostprocessComposer<T> OmitAutoProperties()
         {
-            var targetToRemove = FindAutoPropertiesNode(this);
+            var autoPropertiesNode = FindAutoPropertiesNode(this);
 
-            if (targetToRemove == null)
+            if (autoPropertiesNode == null)
                 return this;
 
-            var p = this.Parents(targetToRemove.Equals).First();
-
-            return (NodeComposer<T>)this.ReplaceNodes(
-                with: targetToRemove.Concat(p.Where(b => targetToRemove != b)),
-                when: p.Equals);
+            // We disable postprocessor rather than delete it.
+            // The reason is that it's the only holder of which properties/fields should not be populated.
+            // If user later decide to enable properties population, we'll need this information 
+            // (e.g. if user does smth like fixture.Build().Without(x => x.P).OmitAutoProperties().WithAutoProperties()
+            // the information from "Without" should not be missed)
+            return (NodeComposer<T>) this.ReplaceNodes(
+                with: n => new Postprocessor<T>(
+                    autoPropertiesNode.Builder,
+                    autoPropertiesNode.Command,
+                    new FalseRequestSpecification()),
+                when: autoPropertiesNode.Equals);
         }
 
         /// <summary>
@@ -332,8 +324,8 @@ namespace Ploeh.AutoFixture.Dsl
         public IPostprocessComposer<T> With<TProperty>(
             Expression<Func<T, TProperty>> propertyPicker, TProperty value)
         {
-            var graphWithoutSeedIgnoringRelay =
-                this.WithoutSeedIgnoringRelay();
+            var graphWithAutoPropertiesNode = this.GetGraphWithAutoPropertiesNode();
+            var graphWithoutSeedIgnoringRelay = WithoutSeedIgnoringRelay(graphWithAutoPropertiesNode);
 
             var container = FindContainer(graphWithoutSeedIgnoringRelay);
          
@@ -349,17 +341,8 @@ namespace Ploeh.AutoFixture.Dsl
                     }),
                 when: container.Equals);
 
-            return (NodeComposer<T>)graphWithProperty.ReplaceNodes(
-                with: n => n.Compose(
-                    new []
-                    {
-                        new Omitter(
-                            new EqualRequestSpecification(
-                                propertyPicker.GetWritableMember().Member,
-                                new MemberInfoEqualityComparer()))
-                    }
-                    .Concat(n)),
-                when: n => n is NodeComposer<T>);
+            var member = propertyPicker.GetWritableMember().Member;
+            return (NodeComposer<T>) ExcludeMemberFromAutoProperties(member, graphWithProperty);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


zvirja:
Let's wait what @ecampidoglio and/or @adamchester would say here ;-)

zvirja:
@moodmosaic I do also like the idea to merge this to `master`, because it's a quite confusing bug (see the "magic explanation" section [here](699#issuecomment-311072979)) . We don't seem to expose the internal builder graph somewhere, so it's indeed not a breaking change. The only concern is that it might break tests if people relied on the buggy behavior.. 😕 But that's always the case when you fix something that doesn't throw 😄

Given that @adamchester and @ecampidoglio seems to be busy to review this and put their opinion - what would you suggest to do? Are our 2 votes fine to merge this to the `master`, or you'd suggest to continue to wait?

zvirja:
I've rebased this branch on top of `master` and resolved all the conflicts. Currently build fails, because I use C# 7 features. I've fired #795 to enable C# 7 and make build green.

Notice, recently #794 PR was created by the first-time contributor and it should be put on hold until this one is merged. Therefore, this one should be prioritized :) @moodmosaic @adamchester You are the only hope these days 😄

zvirja:
After I enabled MSBuild 15 this PR is green again.

@moodmosaic Would it be possible for you to allocate a bit of time and clarify whether you are fine to merge it to the `master`? If so, I could complete this PR (as no breaking changes are present here and other reviewers are missing 😞).

moodmosaic:
@zvirja, I can't recall if [these changes in `NodeComposerTest`](https://github.com/AutoFixture/AutoFixture/pull/781/files#diff-68dade75f99c180341e9e21ce0f0d3be)  where there from the begging, but to play it safe, it'd suggest we merged to `v4` branch. After all, there *are* some changes in the existing tests.
👍

zvirja:
@moodmosaic These changes were present from the beginning, however they are actually [structural inspection](http://enterprisecraftsmanship.com/2016/07/21/unit-testing-anti-patterns-structural-inspection/). I expect that these tests exist not for _locking_ the behavior, but rather for confirming that picture looks like you imagine it 😉

Do we follow a principle that changes in _structural inspection_ are not allowed? If so, I'd vote to not have such tests at all, as they locks _internal_ state from being modified and it will be really hard to fix something without making them red..

zvirja:
The reason why I'm so worried about this is that we have other PRs where such tests [are being introduced](https://github.com/AutoFixture/AutoFixture/pull/697/files#diff-3b20e30bf8f7c8e5020d5f1e40f2907bR126).  I thought that it's fine for such tests to be modified within the same major release. However, if that isn't so - I'll ask to remove the test..

zvirja:
@moodmosaic Rebased again to use v4 as a target as you are right that we could play safely and to not break users' solutions if possible. This change could break solutions as people could rely on observable side effects. Please approve if you are fine now.

Also please reply about structural inspection tests - as I want to know whether it's fine to accept them 😕

moodmosaic:
>@moodmosaic Rebased again to use v4 as a target [..]
👍 

>Also please reply about structural inspection tests - as I want to know whether it's fine to accept them 

Yes, *any* change made in the [exercise SUT](http://xunitpatterns.com/exercise%20SUT.html) and/or [result verification](http://xunitpatterns.com/result%20verification.html) phase can be a breaking change...

zvirja:
@moodmosaic That means that if we have something like following, we shouldn't accept the test:

```sharp
class Sut 
{
  public IEntity Value { get; }
}

// and test
Assert.IsAssignableFrom<SpecificEntity>(sut.Value);
```

As this test asserts the internal state which we don't guarantee to not be changed. This is the [structural inspection](http://enterprisecraftsmanship.com/2016/07/21/unit-testing-anti-patterns-structural-inspection/) I mentioned above.

Even if such test helps to better understand what is going on, we still are not allowed to add them. Also we should delete all the test like [this](https://github.com/zvirja/AutoFixture/blob/a783d556feb1a4185a84b08f020eef26210f41f0/Src/AutoFixtureUnitTest/Dsl/NodeComposerTest.cs#L272), as we don't in fact guarantee that graph will be always exactly the same. Rather, we should test the _behavior_ only (perform a facade test).

Do I see it right? 😟

moodmosaic:
Yes, it *may* be a breaking change if you're modifying an existing test. It *can* make it [less trustworthy](http://blog.ploeh.dk/2013/04/02/why-trust-tests/), that's all.

zvirja:
@moodmosaic What would you suggest to do in such situations? Do you agree to not add such tests if we want to make them trustworthy and we don't want to assert internal behavior?

zvirja:
@moodmosaic Pinging you just for case you missed this one. This question is quite important as that defines which tests we should accept and which not..

moodmosaic:
I'm not sure what you mean with the above... We've been doing this a lot throughout the codebase. What are some good reasons for deviating it from it? 😃

Your question was more about modifying *existing* tests, IIUC...

zvirja:
My question is that we have structural tests in our code. Such tests usually asserts the internal state and that's not something we want to promise for users.

Previously I thought that it's fine to modify such tests within the same major version and we have this tests for _convenience_ only.

However, in current PR you asked to target vNext as we modify the existing tests. In such scenario, I'd say that it's better to not have such tests at all - so that tests doesn't limit us and we still can introduce changes (e.g. fix bugs) within the same major. On other hand, if we agree to have such tests, we must be fine to modify them and don't consider them as breaking changes.

That's what caused my confusion - we have tests about internal state and are not allowed to modify them..

moodmosaic:
>Such tests usually asserts the internal state and that's not something we want to promise for users.

No, they do not.

The reason that Structural Inspection is used throughout the AutoFixture codebase is to [test the behavior of each part and then verify that the structure of the composition is correct](http://blog.ploeh.dk/2013/04/04/structural-inspection/).

Issue #783, opened 2017-07-21
Use dotnet pack for the AutoFixture.NUnit2 project


zvirja:
The `AutoFixture.NUnit2` package consists of two standalone projects embedded into a single package. Currently `dotnet pack` doesn't allow to handle that gracefully.

Investigate a way to solve that.

@AutoFixture/core Do you know the reason why we have two standalone projects for the `NUnit2` integration? It doesn't seem that the `AutoFixture.NUnit2.Addins` is used somewhere else, so we could collapse that to a single `AutoFixture.NUnit2` project in `v4`.. @gertjvr you are the author, so probably you could share details of such a design decision 😕

zvirja:
Closing this one as it's already covered by #780 after I combined two projects in #785.

PR #785, opened 2017-07-22
Merge AutoFixture.NUnit2.Addins to AutoFixture.NUnit2

zvirja:
As it was mentioned in #783 there are issues in usage of `dotnet pack` for the `AutoFixture.NUnit2` project because package consists of artifacts from two projects. I've investigated this more and found that actually everything could be merged into a single project, making the maintenance simpler. I've tried to find explanation in the git history why we needed two projects, but @gertjvr [didn't leave](https://github.com/AutoFixture/AutoFixture/commit/387e98e56728023067453885cc3488bde2783385) the reason 😞 I have an assumption that previously project contained much more code so it made sense to split them.

I've tested the result using the console runner, R# runner and VS discovery and it seems to work fine (parameter is generated when I use the `AutoData` attribute). It appeared that code analysis was disabled for the `AutoFixture.NUnit2.Addins` project, so I needed to fix some warnings 😕 Also I preserved the original namespaces to make the upgrade easier.

After this PR is merged, I could rework the packaging to use `dotnet pack` (like in #780).


moodmosaic:
I never liked the idea of having *AutoFixture.NUnit2.Addins* so I'm all 💯 for this. Great job 👍

zvirja:
Merged as agreed that single approval is enough.

PR #788, opened 2017-08-01
AutoMoq: Fixes mocking get-only interface properties with recent Moq version.

### LINKED ISSUE #434: AutoConfiguredMoqCustomization and read-only properties
This issue relates to the problem described here: [AutoConfiguredMoqCustomization and unsettable properties](http://stackoverflow.com/q/32067727/857807)

**Repro**

``` csharp
public interface Interface
{
    string Property { get; }
}

var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
var a = fixture.Create<Interface>();
Assert.NotNull(a.Property);
```

Passes with Moq 4.2.1409.1722 or lower
Fails with Moq 4.2.1502.911 or greater

**Background**
The setup of `Interface` in the repro is roughly equivalent to [these 3 steps](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoMoq/AutoConfiguredMoqCustomization.cs#L60):

``` csharp
var a = new Mock<Interface>();

// MockVirtualMethodsCommand - sets up all non-void methods,
// including getters (e.g. the auto-generated `get_Property`)
a.Setup(x => x.Property).Returns("test");

// StubPropertiesCommand
a.SetupAllProperties();

// AutoMockPropertiesCommand - sets all properties
// - skipped since there are no settable properties - 

Assert.NotNull(a.Object.Property);
```

**Problem**

It appears that, before Moq 4.2.1502.911, the visible side-effects of `SetupAllProperties` were:
- Stub get-set properties
- Initialize get-set properties with a default value - thus, overriding any previous setups.

Since 4.2.1502.911, `SetupAllProperties` now does this:
- Stub get-set properties
- Initialize **all** properties with a default value - thus, overriding any previous setups.

I see this as a breaking change introduced by Moq.

Whether it was intentional or not, I don't know - but even if it is a bug, judging from past experience, I don't think it'll be addressed anytime soon since this is kind of an edge case.

Should we address this issue and provide a fix? (e.g., by reordering our setup steps)

nphmuller:
This PR fixes #434.

It is my first time contributing to this project, and first time looking at its source. So I'm sure there's some feedback. :)

Before this fix, get/set properties would get mocked by MockVirtualMethodsCommand and later overwritten by Moq's SetupAllProperties() in StubPropertiesCommand. Because of a breaking change of Moq now also setting up get-only properties, I had to swap both commands.

So now SetupAllProperties in StubPropertiesCommand is ran, and then MockVirtualMethodsCommand is ran.

Because MockVirtualMethodsCommand also mocks get/set properties, which overrides the stub, I have black-listed get/set properties from being mocked in MockVirtualMethodsCommand.

### DIFF
--- a/Src/AutoMoq/MockVirtualMethodsCommand.cs
+++ b/Src/AutoMoq/MockVirtualMethodsCommand.cs
@@ -24,7 +24,8 @@ namespace Ploeh.AutoFixture.AutoMoq
     /// 
     /// Notes:
     /// - Due to a limitation in Moq, methods with "ref" parameters are skipped.
-    /// - Automatic mocking of generic methods isn't feasible either - we'd have to antecipate any type parameters that this method could be called with. 
+    /// - Automatic mocking of generic methods isn't feasible either - we'd have to antecipate any type parameters that this method could be called with.
+    /// - To not interfere with other post-processors in the chain we skip getters for properties that have both getters and setters.
     /// </remarks>
     public class MockVirtualMethodsCommand : ISpecimenCommand
     {
@@ -114,6 +115,14 @@ namespace Ploeh.AutoFixture.AutoMoq
             var methods = type.IsInterface
                               ? type.GetInterfaceMethods()
                               : type.GetMethods();
+
+            // Skip properties that have both getters and setters to not interfere
+            // with other post-processors in the chain that initialize them later.
+            var getMethods = type.GetProperties()
+                                    .Where(p => p.GetGetMethod() != null &&
+                                                p.GetSetMethod() != null)
+                                    .Select(p => p.GetGetMethod());
+            methods = methods.Except(getMethods);
 
             return methods.Where(CanBeConfigured);
         }
--- a/Src/AutoMoqUnitTest/FixtureIntegrationWithAutoConfiguredMoqCustomizationTest.cs
+++ b/Src/AutoMoqUnitTest/FixtureIntegrationWithAutoConfiguredMoqCustomizationTest.cs
@@ -76,6 +76,19 @@ namespace Ploeh.AutoFixture.AutoMoq.UnitTest
         }
 
         [Fact]
+        public void GetOnlyPropertiesReturnValueFromFixture()
+        {
+            // Fixture setup
+            var fixture = new Fixture().Customize(new AutoConfiguredMoqCustomization());
+            var frozenString = fixture.Freeze<string>();
+            // Exercise system
+            var result = fixture.Create<IInterfaceWithGetOnlyProperty>();
+            // Verify outcome
+            Assert.Same(frozenString, result.GetOnlyProperty);
+            // Teardown
+        }
+
+        [Fact]
         public void IndexersReturnValueFromFixture()
         {
             // Fixture setup
@@ -97,6 +110,7 @@ namespace Ploeh.AutoFixture.AutoMoq.UnitTest
             // Exercise system
             var result = fixture.Create<Mock<TypeWithVirtualMembers>>();
             // Verify outcome
+            Assert.Equal(frozenString, result.Object.VirtualGetOnlyProperty);
             Assert.Equal(frozenString, result.Object.VirtualMethod());
             Assert.Equal(frozenString, result.Object.VirtualProperty);
             // Teardown
--- a/Src/AutoMoqUnitTest/MockVirtualMethodsCommandTest.cs
+++ b/Src/AutoMoqUnitTest/MockVirtualMethodsCommandTest.cs
@@ -129,13 +129,30 @@ namespace Ploeh.AutoFixture.AutoMoq.UnitTest
             // Fixture setup
             var fixture = new Fixture();
             var frozenString = fixture.Freeze<string>();
-            var mock = new Mock<IInterfaceWithProperty>();
+            var mock = new Mock<IInterfaceWithGetOnlyProperty>();
 
             var sut = new MockVirtualMethodsCommand();
             // Exercise system
             sut.Execute(mock, new SpecimenContext(fixture));
             // Verify outcome
-            var result = mock.Object.Property;
+            var result = mock.Object.GetOnlyProperty;
+            Assert.Equal(frozenString, result);
+            // Teardown
+        }
+
+        [Fact]
+        public void SetsUpVirtualPropertyGetters_ToRetrieveReturnValueFromContext()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var frozenString = fixture.Freeze<string>();
+            var mock = new Mock<TypeWithVirtualMembers>();
+
+            var sut = new MockVirtualMethodsCommand();
+            // Exercise system
+            sut.Execute(mock, new SpecimenContext(fixture));
+            // Verify outcome
+            var result = mock.Object.VirtualGetOnlyProperty;
             Assert.Equal(frozenString, result);
             // Teardown
         }
@@ -285,6 +302,36 @@ namespace Ploeh.AutoFixture.AutoMoq.UnitTest
             var sut = new MockVirtualMethodsCommand();
             // Exercise system and verify outcome
             Assert.DoesNotThrow(() => sut.Execute(specimen, context.Object));
+        }
+
+        [Fact]
+        public void IgnoresPropertiesWithGettersAndSetters()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var mock = new Mock<IInterfaceWithProperty>();
+
+            var sut = new MockVirtualMethodsCommand();
+            // Exercise system
+            sut.Execute(mock, new SpecimenContext(fixture));
+            // Verify outcome
+            var result = mock.Object;
+            Assert.Null(result.Property);
+        }
+
+        [Fact]
+        public void IgnoresVirtualPropertiesWithGettersAndSetters()
+        {
+            // Fixture setup
+            var fixture = new Fixture();
+            var mock = new Mock<TypeWithVirtualMembers>();
+
+            var sut = new MockVirtualMethodsCommand();
+            // Exercise system
+            sut.Execute(mock, new SpecimenContext(fixture));
+            // Verify outcome
+            var result = mock.Object;
+            Assert.Null(result.VirtualProperty);
         }
 
         [Fact]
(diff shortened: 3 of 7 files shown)


Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -115,6 +115,13 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
                               ? type.GetInterfaceMethods()
                               : type.GetMethods();
 
+            // Don't mock "GetSet" properties because they will be stubbed.
+            var getSetMethods = type.GetProperties()
+                                    .Where(p => p.GetGetMethod() != null &&
+                                                p.GetSetMethod() != null)
+                                    .SelectMany(p => new[] {p.GetGetMethod(), p.GetSetMethod()});
+            methods = methods.Where(m => !getSetMethods.Contains(m));
```
zvirja: This implementation isn't very optimal as expensive enumeration will happen for each time you call for `Contains`. It's better to use the `Except()` LINQ method for such cases.

Src/AutoMoqUnitTest/MockVirtualMethodsCommandTest.cs
```diff
@@ -129,13 +129,13 @@ public void SetsUpPropertyGetters_ToRetrieveReturnValueFromContext()
             // Fixture setup
             var fixture = new Fixture();
             var frozenString = fixture.Freeze<string>();
-            var mock = new Mock<IInterfaceWithProperty>();
+            var mock = new Mock<IInterfaceWithGetOnlyProperty>();
```
zvirja: Instead of changing the existing test to verify new behavior, could you please:
1. Change this test to assert that get-set properties are ignored.
2. Add a test that get-set properties are ignored for classes.
3. Add a standalone test that get-only properties are set up for both interfaces and classes.

P.S. Pay attention to the types from the `TestTypeFoundation` project - you could potentially reuse some of them 😉

Src/AutoMoqUnitTest/FixtureIntegrationWithAutoConfiguredMoqCustomizationTest.cs
```diff
@@ -76,6 +76,19 @@ public void PropertiesReturnValueFromFixture()
         }
 
         [Fact]
+        public void GetOnlyPropertiesReturnValueFromFixture()
```
zvirja: Could you please add a similar test for the class as well?

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -115,6 +115,13 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
                               ? type.GetInterfaceMethods()
                               : type.GetMethods();
 
+            // Don't mock "GetSet" properties because they will be stubbed.
```
zvirja: Could you please also extend the class documentation (on top of the file) to specify that it excludes the properties that has both setters and getters.

Also the current description doesn't look too much informative. I'd suggest to alter it to smth like "Skip properties that have both getters and setters to not interfere with other post-processors in the chain which initialize them."
or something better :)

zvirja:
@dcastro AFAIK you are the author of this integration. If you have free time, it would be cool if you could also review this PR and let us know whether you are fine with it. Thank you in advance 😉

Src/AutoMoq/MockVirtualMethodsCommand.cs
nphmuller: Done. Hope its clear enough now!

Src/AutoMoq/MockVirtualMethodsCommand.cs
nphmuller: Never really used `Except()`. Nice to know!. Changed.

Src/AutoMoqUnitTest/FixtureIntegrationWithAutoConfiguredMoqCustomizationTest.cs
nphmuller: Added the additional check to `VirtualMembersReturnValueFromFixture()`. Is that good enough, or is it better if I make it a different test?

Src/AutoMoqUnitTest/MockVirtualMethodsCommandTest.cs
nphmuller: Tests updated.

Src/AutoMoq/MockVirtualMethodsCommand.cs
zvirja: Thanks.

Src/AutoMoqUnitTest/FixtureIntegrationWithAutoConfiguredMoqCustomizationTest.cs
zvirja: I think it's fine.

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -115,6 +116,14 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
                               ? type.GetInterfaceMethods()
                               : type.GetMethods();
 
+            // Skip properties that have both getters and setters to not interfere
+            // with other post-processors in the chain which initialize them."
+            var getSetMethods = type.GetProperties()
+                                    .Where(p => p.GetGetMethod() != null &&
+                                                p.GetSetMethod() != null)
+                                    .SelectMany(p => new[] {p.GetGetMethod(), p.GetSetMethod()});
```
zvirja: Do we need both the `get` and `set` methods here? As far as I can see, the `CanBeConfigured` method allows to pick up the getters only (because the `set` methods are void), so this is redundant.
If that adds no value - it's better to keep the `get` methods only to make it simpler. Sorry that I overlooked that during my previous review 😊

P.S. If we agree to change this, feel free to use `git rebase -i` to edit previous commits and change this - it doesn't seem to add any value to the history 😉

nphmuller: You're right. Selecting `GetSetMethod()` was not necessary. Was just to be more clear about the intention, but I agree that its simpler without it (and probably performs a bit better).

I've also merged the improvements with the original commits. Thought it would be more clear what I improved, but if you don't think its necessary I'm fine with merging them.

zvirja: Thanks for that fix, it looks better now 👍 I think that these filters are internal to the class, so it's fine to rely that the subsequent filter will skip all the setters and extra safety is not required.

> Thought it would be more clear what I improved

It's often a dilemma what to keep in the history. Usually, I try to not pollute the history with intermediate decisions if I think they bring no value (e.g. I change name of the variable - nobody will be interested how it was initially named). In this case it could be considered the from different perspectives, that's why I _suggested_ to rebase - do what you feel is better 😉

BTW, the variable name `getSetMethods` is no longer valid as only `get` methods are present there..

nphmuller:
Damnit. I wanted to change it but totally forgot. Guess the coffee hadn't kicked in yet.  ;)

I'll see if I have some spare time during the weekend. Else I'll fix it monday. Such a small thing...

On 4 Aug 2017, at 14:30, Alex Povar <notifications@github.com<mailto:notifications@github.com>> wrote:


@zvirja commented on this pull request.

________________________________

In Src/AutoMoq/MockVirtualMethodsCommand.cs<https://github.com/AutoFixture/AutoFixture/pull/788#discussion_r131377365>:

> @@ -115,6 +116,14 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
                               ? type.GetInterfaceMethods()
                               : type.GetMethods();

+            // Skip properties that have both getters and setters to not interfere
+            // with other post-processors in the chain which initialize them."
+            var getSetMethods = type.GetProperties()
+                                    .Where(p => p.GetGetMethod() != null &&
+                                                p.GetSetMethod() != null)
+                                    .SelectMany(p => new[] {p.GetGetMethod(), p.GetSetMethod()});


Thanks for that fix, it looks better now 👍 I think that these filters are internal to the class, so it's fine to rely that the subsequent filter will skip all the setters and extra safety is not required.

Thought it would be more clear what I improved

It's often a dilemma what to keep in the history. Usually, I try to not pollute the history with intermediate decisions if I think they bring no value (e.g. I change name of the variable - nobody will be interested how it was initially named). In this case it could be considered the from different perspectives, that's why I suggested to rebase - do what you feel is better 😉

BTW, the variable name getSetMethods is no longer valid as only get methods are present there..

zvirja:
@nphmuller Don't worry, we are still waiting for the @dcastro review, so we have time. In any case, if no other comments appear and you become missing, we can always push commits to your branch (if enabled) or even merge without fix :wink:

Src/AutoMoq/MockVirtualMethodsCommand.cs
nphmuller: var name changed 👍

zvirja:
LGTM, however let's see what @dcastro and other @AutoFixture/core members think as I don't know this area too much.

moodmosaic:
I don't hold strong opinions on how the Auto-mocking Glue Libraries work<sup>*</sup>, as long as we're not introducing breaking changes.

---

<sup>*</sup> I wrote though AutoFixture.AutoFakeItEasy and AutoFixture.AutoFoq, so I may be able to be of more help regarding these two.

dcastro:
Apologies for the delay 😱.

Looks good to me! This is exactly what I had in mind way back then, including the modification to `MockVirtualMethodsCommand` to avoid overwriting the setups done by `StubPropertiesCommand`. Well done!

zvirja:
@dcastro Thank you for the spent time!

@moodmosaic @adamchester Could you please also review whether you are fine, so we can proceed?

zvirja:
@moodmosaic I've just noticed [your opinion](788#issuecomment-320794306) above. I believe we are fine as no breaking API changes are being introduced here. Therefore, I'm going to proceed and merge this one. Let me know if you have some final concerns 😉

zvirja:
@nphmuller Thanks for your contribution 🥇 Will be published soon.

P.S. I've slightly tuned comments in your commits, so don't be surprised 😉

nphmuller:
@zvirja No problem.

And thank you all for maintaining this awesome project. I've read a bit about the governance change and how you all handled it. Great job!

zvirja:
Thanks :)

PR #791, opened 2017-08-04
Define test service marker in each test project

zvirja:
For each test project VS insert a [special marker](https://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for) to quickly discover projects with tests. Previously I tried to normalize that and defined this option in the `CommonTestProperties.props` that is imported to each test project. However, it doesn't work as VS still inserts this nodes to the project file itself.

I decided to commit this to repo, so you no longer need to revert these changes each time you run the tests (which is pretty annoying BTW 😢).


zvirja:
> FWIW, I'm still a huge fan of the TestDriven.Net tool

Quickly went through the features, however wasn't able to find what it offers in addition to R#. The only nice addition is that it seems to be free 😄

zvirja:
Will merge it after #773 to simplify conflicts solve.

PR #793, opened 2017-08-08
Heavily improve customization performance

zvirja:
## Background
I've found that on my real solution it takes about 15 second for NCrunch to discover all the tests. After I profiled things a bit I found that slowdown is caused by the AutoFixture and it's customization performance. After I applied the fixes from this this PR speed become about 4.5 second - more then 3 times faster.

This issue is reproducible when you perform customization of the fixture instance. The more customizations you have, the more you feel the slow down.  I've found two main places to improve.
1. Multiple enumeration in the `ComposeIfMultiple()` method. This is a hot place when we recompose a graph. That multiple enumeration causes graph to be re-composed a lot of times during a single composition, which costs time.
2. The `SelectNodes()` method which is called recursively to find all the matching nodes. It creates state machines on each recursion layer and this costs. It appeared that actually we don't need all the matching nodes and use the first one only. That allowed to save some time.

## Measurements

I used the BenchmarkDotNet tool and the following code as a baseline:
```csharp
[ClrJob]
public class CustomizationSpeedTest
{
    [Benchmark]
    public void CreateFixture() => new Fixture().Customize(new TestCustomization());
}

class TestCustomization: ICustomization
{
    {
        fixture.Customize(new IncrementingDateTimeCustomization());
        fixture.Inject(42);
        fixture.Freeze<object>();
    }
}
```
Idea was to have a couple of customizations which is a regular things for the real projects.

Environment:
```
BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 1 (10.0.14393)
Processor=Intel Core i7-4770 CPU 3.40GHz (Haswell), ProcessorCount=8
Frequency=3320307 Hz, Resolution=301.1770 ns, Timer=TSC
  [Host] : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2101.1
  Clr    : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2101.1
```

### Before fix
```
        Method |     Mean |     Error |    StdDev |
-------------- |---------:|----------:|----------:|
 CreateFixture | 1.203 ms | 0.0231 ms | 0.0216 ms |
```

### First commit applied only (refactored `ComposeIfMultiple()`)
```
        Method |     Mean |    Error |   StdDev |
-------------- |---------:|---------:|---------:|
 CreateFixture | 202.3 us | 3.939 us | 3.684 us |
```

### First and second commits (final result)
```
        Method |     Mean |    Error |   StdDev |
-------------- |---------:|---------:|---------:|
 CreateFixture | 178.5 us | 3.543 us | 3.638 us |
```

This PR makes the test above about 6.75 times quicker! The main boost is provided by the first commit (~5.95 quicker). The second commit adds yet another ~12% of boost.

Also the boost can be observed for AutoFixture's xUnit tests. Previously all them took ~16,2 secs on my machine, now they take ~12.5 secs - about 30% quicker 😮

----
The most exciting thing here is that we can achieve this boost with relatively simple code changes.

@adamchester Given that you worked on [the similar task](218) in past, it might be interesting to you 😉 
@moodmosaic Review from you is also expected if possible 😄

P.S. It might be easier to review this PR commit by commit.

### DIFF
--- a/Src/AutoFixture/Kernel/CompositeSpecimenBuilder.cs
+++ b/Src/AutoFixture/Kernel/CompositeSpecimenBuilder.cs
@@ -103,11 +103,49 @@ namespace Ploeh.AutoFixture.Kernel
 
         internal static ISpecimenBuilder ComposeIfMultiple(IEnumerable<ISpecimenBuilder> builders)
         {
-            var isSingle = builders.Take(2).Count() == 1;
-            if (isSingle)
-                return builders.Single();
+            // This code is called very frequently during graph mutation (e.g. customization).
+            // It's optimized to not perform unnecessary allocations and do not enumerate builders more than once.
+            // Code is written expecting that single builder is the most common case.
+            // See more detail here: https://github.com/AutoFixture/AutoFixture/pull/793
+            
+            ISpecimenBuilder singleItem = null;
+            List<ISpecimenBuilder> multipleItems = null;
+            bool hasItems = false;
 
-            return new CompositeSpecimenBuilder(builders);
+            using (var enumerator = builders.GetEnumerator())
+            {
+                if (enumerator.MoveNext())
+                {
+                    singleItem = enumerator.Current;
+                    hasItems = true;
+
+                    while (enumerator.MoveNext())
+                    {
+                        // We are here only if more than one item is present.
+                        // When we create collection for multiple items, we add the already enumerated one as well.
+                        if (multipleItems == null)
+                        {
+                            multipleItems = new List<ISpecimenBuilder> {singleItem};
+                        }
+                        
+                        multipleItems.Add(enumerator.Current);
+                    }
+                }
+            }
+
+            if (!hasItems)
+            {
+                //It's very rare case and it does't make sense to optimize it
+                return new CompositeSpecimenBuilder();
+            }
+
+            //Single item is present only, so we return it.
+            if (multipleItems == null)
+            {
+                return singleItem;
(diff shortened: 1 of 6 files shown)


Src/AutoFixture/Kernel/CompositeSpecimenBuilder.cs
```diff
@@ -103,22 +103,60 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
...
+                        }
+                        
+                        multipleItems.Add(enumerator.Current);
+                    }
+                }
+            }
+
+            if (!hasItems)
+            {
+                //It's very rare case and it does't make sense to optimize it
```
moodmosaic: Nitpick: We usually add a space after `//` in code comments, and a period (`.`) at the end of each sentence.

Src/AutoFixture/Kernel/CompositeSpecimenBuilder.cs
```diff
@@ -103,22 +103,60 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
...
+                return new CompositeSpecimenBuilder();
+            }
+
+            //Single item is present only, so we return it.
```
moodmosaic: Nitpick: We usually add a space after `//` in code comments.

PR #796, opened 2017-08-14
Removed governance announcement from README

ploeh:
The linked issue has now been closed for over four months, so the new
governance model is no longer that new. Additionally, it seems that most
people don't read the announcement anyway, because I still get a fair
amount of tweets and such about AutoFixture...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,5 @@
 [![Build status](https://ci.appveyor.com/api/projects/status/qlmobf6rt05pmt7e/branch/master?svg=true)](https://ci.appveyor.com/project/AutoFixture/autofixture/branch/master)
 
-## Announcement ##
-
-[AutoFixture is getting a new governance model](https://github.com/AutoFixture/AutoFixture/issues/703). This most likely means that you can expect delays if you have questions, issues, or pull requests.
-
 ## Project Description ##
 
 Write maintainable unit tests, faster.


zvirja:
Could you clarify what is the reason why you fired this PR? Do you see any issue with this announcement?

Personally, I like that it's present as:
- people still can read it when they want to contribute;
- it indicates about delays in PR review - this is definitely the case at least for now;
- people could come here to find you, but will know that you are not here;
- we still actively discuss things there regarding the processes, so ticket is alive;
- this is a good shortcut at the landing page 😉

Probably, we could rephrase it instead of removal. To say smth like "Project has changed the governance model and currently is being maintained by the @AutoFixture/core team instead of @ploeh. For more details see `this` discussion." This news is still pretty hot :)

> because I still get a fair amount of tweets and such about AutoFixture...

Forward them all to [our twitter account](https://twitter.com/autofixture) 😅

ploeh:
You're welcome to reject the PR if you think it's a bad idea... I just thought it was strange that the README file was announcing that something is going to happen that actually happened four months ago.

zvirja:
> You're welcome to reject the PR if you think it's a bad idea... 

And what about an idea to re-phrase the message to still indicate that we have a new governance model?

Also let's see what other members think (@moodmosaic, @adamchester, @klimisa).

zvirja:
Ok, if you treat that announcement as a temp one indicating that project is in undetermined state, which isn't so now, than it's indeed could make sense to remove it. After all, we are actively answering to new issues and closing the PRs, so we are in the stable state.

PR #798, opened 2017-08-17
Exclude VS temp files from git

zvirja:
Exclude temp VS files from git.


zvirja:
@moodmosaic Thanks! Hope we'll be lucky and @adamchester will approve this soon, so no need to wait for 1 week.. 😕

zvirja:
Unexpectedly :fearful: :astonished: 😃

PR #799, opened 2017-08-17
Migrate FSharp projects to .NET Core format

zvirja:
This is a leftover after the #769. Previously, VS didn't support .NET Core projects for F#, so I postponed the migration. A few days ago VS 2017 Update 3 has been released with support of F# for .NET Core projects so we can proceed. Everything works fine with previous version of .NET Core SDK, so I haven't added the [`global.json`](769#issuecomment-312410005) file.

Now all the projects we have are .NET Core projects which is awesome. After this change is merged I'll be able to complete #780 and completely get rid of the pesky `.nuspec` files (I hope everything will go smoothly).

@moodmosaic @adamchester You both are invited here, as you know F# and .NET Core. Changes are identical to #769 which you already approved.


zvirja:
Thanks for the quick review and approval, I appreciate that!

> Just in case I don't get back to this in a timely fashion, I'm recording my approval Now.

Will one week be enough for you? That's to clarify the "timely fashion" ranges and be sure we don't misunderstood each other :blush:

adamchester:
That's right, if I don't get back to testing it by the agreed 1 week, it's totally fair to move on 👍 I'll do my best, as I think it's necessary at least one of us (probably me) should try this out.

Build.fsx
```diff
@@ -69,7 +69,7 @@ Target "PatchAssemblyVersions" (fun _ ->
         buildVersion.fileVersion
         buildVersion.infoVersion
 
-    let filesToPatch = !! "Src/*/Properties/AssemblyInfo.*"
+    let filesToPatch = !! "Src/*/Properties/AssemblyInfo.fs"
```
moodmosaic: Most likely an odd question, but what about other extensions (e.g. `.cs`)?

zvirja: See [this](https://github.com/AutoFixture/AutoFixture/pull/799/commits/643c9d1daf234f40380867f29f8c1293eadbaea1) commit and its message. In short - we don't need it more as there is nothing to patch in `AssemblyInfo.cs` files 😄

moodmosaic: We might want to add this as a comment then? So we can remember to remove in the future? 😃

zvirja: Deal, added 😉

moodmosaic: Yes, adding that comment can be helpful in the future because it [explains the *why*](https://ayende.com/blog/163297/the-importance-of-comments). (On the contrary, [explaining the *how*, can be an apology](http://butunclebob.com/ArticleS.TimOttinger.ApologizeIncode), of course.)

Src/AutoFoqUnitTest/AutoFoqUnitTest.fsproj
```diff
@@ -1,108 +1,39 @@
...
-    <None Include="app.config" />
   </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="xunit" Version="1.9.2" />
+    <PackageReference Include="xunit.extensions" Version="1.8.0.1549" />
+    <PackageReference Include="Unquote" Version="3.1.0" />
+
+    <!-- AutoFoq targets FSharp.Core 3.0.2, but we require 3.1.2 as Unquote was compiled against it. -->
+    <PackageReference Include="FSharp.Core" Version="3.1.2" PrivateAssets="All" />
```
moodmosaic: Wow... *F# 3.x*

---

Perhaps, we should consider targeting F# 3.1.2 in `v4` branch...

Build.fsx
zvirja: Well, we use commit messages for such information. You can always use `git blame` to find out information about particular line. In this particular case I've duplicated the commit message in the code. While I agree that it might look fine in this case, I don't feel like it might sense to do that always.

Src/AutoFoqUnitTest/AutoFoqUnitTest.fsproj
zvirja: Feel free to improve that via separate PR given that you worked on this in the past 😉
I've never used this AF component and don't know well all the history of F# SDK..

zvirja:
Rebased on top of fresh v4 and solved all the conflicts.

@adamchester Just a friendly reminder that you have 2 days left for the testing ☺️

PR #801, opened 2017-08-18
Add information about MyGet feed to ReadMe

zvirja:
Announcement about the private feed we have. Let me know if you have any concerns. [Preview link](https://github.com/AutoFixture/AutoFixture/tree/zvirja/add-private-feed-info-to-readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ The artifacts of the next major version are published to the MyGet feed:
 - `https://www.myget.org/F/autofixture/api/v3/index.json` (Visual Studio 2015+)
 - `https://www.myget.org/F/autofixture/api/v2` (Visual Studio 2012+)
 
-You can use this feed to early access and test the next major version of the product.
+You can use this feed to early access and test the next major version of the AutoFixture.
 
 __Notice__, this feed exists for the _preview purpose_ only, so use it with caution:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,9 +87,20 @@ AutoFixture is available via NuGet:
 * [AutoFixture.Idioms](http://nuget.org/packages/AutoFixture.Idioms)
 * [SemanticComparison](http://nuget.org/packages/SemanticComparison)
 
+### vNext feed
+The artifacts of the next major version are published to the MyGet feed:
+- `https://www.myget.org/F/autofixture/api/v3/index.json` (Visual Studio 2015+)
+- `https://www.myget.org/F/autofixture/api/v2` (Visual Studio 2012+)
+
+You can use this feed to early access and test the next major version of the product.
```
moodmosaic: I would replace *product* with *AutoFixture* 😃

zvirja: Thanks! I rephrased that a few times and didn't know how it's better to say. Indeed, "AutoFixture" looks great 👍

zvirja:
@moodmosaic @adamchester By the way, I've contacted MyGet via [this form](https://www.myget.org/opensource) and asked them about their OpenSource subscription, as we are OSS. Yesterday, they provided us with activation code. Now we have 1GB instead of 500MB storage. That should be enough to store about 300-400 releases. If we need more storage - we could reach out to them later :)

PR #802, opened 2017-08-20
Obsolete TargetType property in FreezeOnMatchCustomization (v4)

sergeyshushlyapin:
Same as #697 but targets AutoFixture v4 because of the breaking changes detected in the [previous pull request](https://github.com/AutoFixture/AutoFixture/pull/697#issuecomment-318410144).

### DIFF
--- a/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
+++ b/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
@@ -123,7 +123,7 @@ namespace Ploeh.AutoFixtureUnitTest
         }
 
         [Fact]
-        public void InitializeWithObjectRequestShouldSetTrueRequestSpecificationMatcher()
+        public void InitializeWithObjectRequestShouldSetEqualRequestSpecificationMatcher()
         {
             var request = new object();
             var sut = new FreezeOnMatchCustomization(request);
--- a/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
+++ b/Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
@@ -75,8 +75,8 @@ namespace Ploeh.AutoFixtureUnitTest
             // Exercise system
             var sut = new FreezeOnMatchCustomization(targetType);
             // Verify outcome
-            var matcher = Assert.IsType<ExactTypeSpecification>(sut.Matcher);
-            Assert.Equal(targetType, matcher.TargetType);
+            var matcher = Assert.IsType<EqualRequestSpecification>(sut.Matcher);
+            Assert.Equal(targetType, matcher.Target);
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


zvirja:
Thank you for firing this PR. Recently [I've been rejected](https://github.com/AutoFixture/AutoFixture/pull/781#issuecomment-323019704) to merge changes to the `master` because of changes to the structural tests. May I ask you to delete all such tests for the current customization, so that we will not be restricted in future? We don't want to guarantee that implementation type of the `IRequestSpecification` will not change.

Src/AutoFixtureUnitTest/FreezeOnMatchCustomizationTest.cs
```diff
@@ -77,6 +78,63 @@ public void InitializeWithMatcherShouldSetCorrespondingProperty()
...
+        [Fact]
+        public void InitializeWithSingleRequestShouldSetCorrespondingProperty()
+        {
+            var request = new object();
+            var sut = new FreezeOnMatchCustomization(request);
+            Assert.Same(request, sut.Request);
+        }
+
+        [Fact]
+        public void InitializeWithObjectRequestShouldSetEqualRequestSpecificationMatcher()
```
zvirja: It seems that you forgot to delete this one - this is also a structural test.

zvirja:
@sergeyshushlyapin Thanks for the change. It seems that you forgot about one more place - see my comment above.

Could you please also reorganize and squash some commits? Now we have 15 commits, while some of them contain intermediate changes that were later removed, so that brings no value to the history.

Except these changes, everything looks fine to me.

sergeyshushlyapin:
> Could you please also reorganize and squash some commits? Now we have 15 commits, while some of them contain intermediate changes that were later removed, so that brings no value to the history.

those commits are results of numerous conversations in #697. Squashing or rebasing will ruin the _existing_ visible history (provided by GitHub UI).

zvirja:
Thank you fixing the test!

> Squashing or rebasing will ruin the existing visible history (provided by GitHub UI).

I believe it's totally fine. Changes during the discussions often breaks GitHub UI and I don't see any issues with that. I believe that laconic and meaningful git history brings more value than storing trash simply to not break the GitHub discussion.

@moodmosaic What would you suggest? Currently we have history with commits where subsequent commits completely rewrite/remove some parts of the previous commits. My point is that it's better to reorganize commits to avoid _intermediate decisions_ (like "Add X" and later "Remove X") within the same PR as it pollutes the history and its hard to read it later.

moodmosaic:
@zvirja, I don't have strong opinions around this. FWIW, @ecampidoglio knows Git much better than I do,
 so he'll probably have an answer for this 😄

zvirja:
Given that @ecampidoglio is completely missing, it would long time before we get the answer 😅

@sergeyshushlyapin If you don't have really strong opinion regarding this, could you please reorganize the commits and make them contain only changes that are finally present?

sergeyshushlyapin:
> Now we have 15 commits, while some of them contain intermediate changes that were later removed

@zvirja, sorry, but I can see no such commits except probably the last one with redundant tests. The history looks good to me, it reflects the PR life cycle and all the decisions made. 

You're welcome to squash whatever you think is necessary. IMO, that's done.

zvirja:
I've rebased the PR on top of v4 for better history.
Merging this as it has my approval and nobody else followed up 😉

PR #804, opened 2017-08-22
Migrate semantic comparison to .NET Core

zvirja:
Migrate SemanticComparison library to support both .NET 45 and .Net Standard 1.5.

Project is small, so number of changes is very little.

@moodmosaic @adamchester Looking forward to your code review 😉


Src/SemanticComparison/ProxyGenerator.cs
```diff
@@ -173,13 +183,10 @@ private static void BuildMethodGetHashCode<TDestination>(TypeBuilder type)
             method.SetReturnType(typeof(int));
 
             int derivedGetHashCode = 135;
-            MethodInfo getHashCode = typeof(TDestination).GetMethod(
```
zvirja: Notice a change here - now only public method will be matched. That should be fine as `GetHashCode` is public.

Src/SemanticComparison/ProxyGenerator.cs
```diff
@@ -203,21 +210,15 @@ private static void BuildMethodEquals(TypeBuilder type, FieldInfo equalsHasBeenC
...
                 "Equals",
-                BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic,
-                null,
-                new[] { typeof(object) },
-                null
-                );
+                new[] {typeof(object)}
+            );
 
-            MethodInfo equalityComparerEquals = typeof(IEqualityComparer).GetMethod(
```
zvirja: Notice a change here - now only public method will be matched. That should be fine as method is public.

Src/SemanticComparison/ProxyGenerator.cs
```diff
@@ -203,21 +210,15 @@ private static void BuildMethodEquals(TypeBuilder type, FieldInfo equalsHasBeenC
             var methodAttributes = MethodAttributes.Public | MethodAttributes.Virtual | MethodAttributes.HideBySig;
             MethodBuilder method = type.DefineMethod("Equals", methodAttributes);
 
-            MethodInfo objectEquals = typeof(object).GetMethod(
```
zvirja: Notice a change here - now only public method will be matched. That should be fine as method is public.

Src/SemanticComparison/ProxyGenerator.cs
```diff
@@ -268,13 +269,10 @@ private static void BuildMethodEquals(TypeBuilder type, FieldInfo comparer)
             var methodAttributes = MethodAttributes.Public | MethodAttributes.Virtual | MethodAttributes.HideBySig;
             MethodBuilder method = type.DefineMethod("Equals", methodAttributes);
 
-            MethodInfo equals = typeof(object).GetMethod(
```
zvirja: Notice a change here - now only public method will be matched. That should be fine as method is public.

Src/SemanticComparison/ProxyGenerator.cs
moodmosaic: Are you *sure* it's working now? It's been a while since I worked on the ProxyGenerator (*yes*, we use our own 😄) but I can tell you that this code can be ...sensitive 😆

zvirja: Well.. I've run all the unit tests in both .NET Framework and .NET Core platforms and they didn't fail. Could we trust tests in this project?

The only change is that now I'm looking for the public method rather than for public & private. Please review and let me know whether it does matter. From what I see, it doesn't matter as we select public methods only.

moodmosaic: >Could we trust tests in this project?

Well, you've also changed the tests so it's hard to tell...

I can imagine that you changed the tests because the `OperatingSystem` that we're passing in to the existing tests isn't available(?) Probably we can create our own `OperatingSystem` class though and keep the changes to the minimum?

zvirja: Yes, the `OperatingSystem` type no longer exists. I'd prefer to not create our own `OperatingSystem`. Rather, we could create some other type instead that has properties which we need (e.g. implements some specific interfaces), however we should understand that properties.

In fact I modified the single test only. From what I got it seems that we don't use any properties and the replacement is totally fine (as we override how we compare object).

It would be good if you could analyze this test change in depth and see whether my replacement is fine or not. After all, you created this particular lib 😉

zvirja: @moodmosaic Will you be able to find some time in the nearby future and follow up on this one so we can proceed further? ☺️

moodmosaic: I'd rather define an `Operating System` with the properties of the old one just to make the pull request smaller. This pull request, as-is, in order to review it carefully, requires me to find a contiguous time, which unfortunately I won't have until the end of this year.

zvirja: @moodmosaic Done! Defined a copy of the `OperatingSystem` class, so now no changes in unit tests are present.

You are fine with the changes now to proceed further?

PR #807, opened 2017-08-24
Use dotnet pack for F# projects

### LINKED ISSUE #782: Migrate F# projects to .NET Core SDK and use dotnet pack for them
The following projects should be updated
- AutoFixture.AutoFoq
- AutoFixture.Idioms.FsCheck

That should be possible starting the VS 2017 Update 3.

zvirja:
Closes #782.

Those were the last `*.nuspec` files present in our solution. Given that I removed all the `nuspec` files, I've cleaned up the build scripts, so that we stop to support `nuspec` files. That allowed to simplify build scripts a bit.

This PR is very similar to the #780, that you already approved.

@moodmosaic @adamchester Please review and approve if everything looks fine to you 😉


Src/AutoFoq/AutoFoq.fsproj
```diff
@@ -21,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="FSharp.Core" Version="3.0.2" />
+    <PackageReference Include="FSharp.Core" Version="3.0.2" PrivateAssets="All" />
```
moodmosaic: This means it's going to exclude the FSharp.Core NuGet dependency, yes? If so, I think this change is fine, and thank you for actually doing it 🎉 

---

As per the official F# guidelines at https://fsharp.github.io/2015/04/18/fsharp-core-notes.html#do-not-bundle-fsharpcore-with-a-library:

> Do not include a copy of FSharp.Core with your library or package. If
> you do, you will create havoc for users of your library.
>
> The decision about which FSharp.Core a library binds to is up to the
> application hosting of the library. The library and/or library package
> can place constraints on this, but it doesn’t decide it.

zvirja: Yes, it does exactly what you say - excludes FSharp.Core 3.0.2 from our NuGet dependencies. I did that simply to mimic the behavior we had before - no dependencies.

> Do not include a copy of FSharp.Core with your library or package.

I though that message is about shipping the FSharp assemblies inside the package. It's unclear whether that actually applies to NuGet dependencies as still host decides which actual version to use..

So should actually our NuGet package has `FSharp.Core 3.0.2` as a _package dependency_?

moodmosaic: According to the F# guidelines linked on my previous comment, the 'right' way is to *not* have the NuGet Package reference FSharp.Core.

>Yes, it does exactly what you say - excludes FSharp.Core 3.0.2 from our NuGet dependencies. I did that simply to mimic the behavior we had before - no dependencies.

AFAICT, from what I see in the changes, this ought to be the right way 👍

PR #809, opened 2017-08-24
Migrate xUnit 2 glue library to .NET Core

zvirja:
Added .NET Core support to xUnit2 glue library. Almost nothing changes - mostly type info usage.
Different xUnit versions are required for different targets:
- 2.0.0 .NET Framework 4.5
- 2.2.0 .NET Standard 1.5

@moodmosaic @adamchester Looking forward to your review 🌮

### DIFF
--- a/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
+++ b/Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
@@ -3,17 +3,18 @@
   <Import Project="..\CommonTestProperties.props" />
 
   <PropertyGroup>
-    <TargetFramework>net45</TargetFramework>
+    <TargetFrameworks>net452;netcoreapp1.1</TargetFrameworks>
     <AssemblyTitle>AutoFixture.xUnit.net2.UnitTest</AssemblyTitle>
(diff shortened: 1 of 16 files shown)


Src/AutoFixture.xUnit.net2.UnitTest/AutoFixture.xUnit.net2.UnitTest.csproj
```diff
@@ -3,17 +3,18 @@
...
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="xunit" Version="2.0.0" />
+    <PackageReference Include="xunit" Version="2.3.0-beta3-build3705" />
 
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta3-build3705" />
+    <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.0-beta3-build3705" />
```
moodmosaic: I think these changes should be fine, since they only affect this particular *unit test* project.

zvirja: Exactly! You can see such changes in all other migrations - they affect test projects only to use runner that supports .NET Core. The publishable project is not being affected here.

PR #810, opened 2017-08-24
Migrate AutoMoq to net core

zvirja:
Migrated AutoMoq project to .NET Core. Different targets has different Moq dependencies:
- `4.1.1308.2120` for .NET Framework 4.5
- `4.7.0` for .NET Standard 1.5

Most of the changes are the regular migration noise. However, there is one change that you should be aware of. It appeared for the full framework runtime the proxy objects, generated by Castle, contains two fields: `__interceptors` and `__target`, while for .NET Core runtime only the `__interceptors` field is present. I've updated test to make that field optional and do not fail if field is missing. See [this](https://github.com/AutoFixture/AutoFixture/commit/96e15ede41f278b2b212156566b6b3ca8041c64b) commit for more detail. That shouldn't affect the logic somehow - we always ignore these fields, so if it's missing - we should do nothing 😅

@moodmosaic @adamchester Looking forward to your review. If you know somebody else who could also take a look in a timely fashion - please invite 😉

### DIFF
--- a/Src/AutoMoq/AutoMoq.csproj
+++ b/Src/AutoMoq/AutoMoq.csproj
@@ -12,6 +12,10 @@
     <PackageId>AutoFixture.AutoMoq</PackageId>
     <Title>AutoFixture with Auto Mocking using Moq</Title>
     <Description>This extension turns AutoFixture into an Auto-Mocking Container. The mock instances are created by Moq. To use it, add the AutoMoqCustomization to your Fixture instance. Read more at http://blog.ploeh.dk/2010/08/19/AutoFixtureAsAnAutomockingContainer.aspx</Description>
+
+    <!--  Suppress warning about invalid dependency version in Castle.Core.
+          That is Moq dependency and we cannot fix that somehow. -->
+    <NoWarn>$(NoWarn);NU1603</NoWarn>
   </PropertyGroup>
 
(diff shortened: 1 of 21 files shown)


Src/AutoMoq/AutoMoq.csproj
```diff
@@ -12,16 +12,24 @@
...
     <CodeAnalysisDictionary Include="CodeAnalysisDictionary.xml" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)'=='net45' ">
     <PackageReference Include="Moq" Version="4.1.1308.2120" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)'=='netstandard1.5' ">
+    <PackageReference Include="Moq" Version="4.7.0" />
```
moodmosaic: Isn't this a breaking change?

zvirja: First of all notice that we are in v4 branch :)

However, nope, this is not a breaking change. If you install this package to .NET Framework 4.5 you will still need Moq 4.1 as a dependency requirement.

For the .NET Standard project you will need Moq 4.7, however .NET Standard support is a _new feature_, so nothing existing was broken.

PR #811, opened 2017-08-24
Housekeeping - use verify configuration in test projects

zvirja:
Currently when we run verification build we use the `Verify` configuration for publishable projects and `Release` configuration for test projects. That required us to configure that in each solution (using the `Configuration Manager` dialog), which is quite error prone and confusing approach.

I made it simpler, so now the `Verify` configuration is used for all the projects. Later each project could decide what to do. E.g. test projects don't run FxCop.

Additionally, VS updated project type ID and file version so I accepted that changes as well.

@moodmosaic @adamchester Likely, no objections from your side will be present as this change is pure housekeeping, so looking forward to your approval to proceed 😉


moodmosaic:
FYI, I'm currently overwhelmed with a great amount of work and might take me a few weeks before I can look at this pull request. Perhaps @adamchester, @klimisa, @ecampidoglio, may be of help meanwhile.

zvirja:
@adamchester Could you please take a look? Actually, nothing important is changed here. The only change is that now I use `Verify` configuration for test projects as well when we build solution in `Verify` configuration. All the rest changes are the changes VS automatically did and I simply accepted that.

zvirja:
@klimisa Probably you can take a look, given that @adamchester seems to be absent..

I simply changed all the solution files to build test projects in the `Verify` configuration when we build solution in the Verify configuration. All the rest changes - that's how VS converted solution files and that is not I did manually. 

Given that CI is green, everything is right with such configuration.

zvirja:
Should be merged after #820

moodmosaic:
@zvirja, you should be able to merge stuff like this without requiring an approval, as I'm proposing [here](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-330116357).

zvirja:
Closed this one as it's a pure housekeeping.

Issue #747, opened 2017-04-03
Ensure that for each nuspec file we have set upper boundary for dependencies


zvirja:
We should ensure that for each dependency in our `nuspec` files the upper bound is set. For instance, if we support xunit2, we should check that our dependency is following: `[2.0, 3.0)`, rather than simple `2.0`.

That is to ensure that library will not become broken after our dependency releases a new major version.

zvirja:
After thinking more about this I'm not sure whether the cap should be present. If we don't update the particular library in time, it will be unavailable. On other hand, people will not install the AF when compatibility wasn't explicitly tested, so it will be clear that it's not supported yet..

@AutoFixture/core Guys, what do you think about this one?

moodmosaic:
I don't see why not 👍 

Some dependencies, however, *do* work even after releasing a major version. Take Moq as an example, where AFAICT we depend on Moq 3 but I'm pretty sure AutoFixture.AutoMoq works also with Moq 4+.

I think it's safer though to set an upper bound limit.

zvirja:
In this case, I'd suggest us to follow the next approach:

1. Set upper limit to all the external components we use. Usually that are the testing frameworks for our glue libraries.
1. For each supported version of library (e.g. Moq 2, Moq 3) add a test project like that is done [here](https://github.com/AutoFixture/AutoFixture/pull/702). 

With such approach we will ensure that we always have tests for the particular version of dependency and if we guarantee that it works.

moodmosaic:
`1.` is fine by me 👍 
`2.` might need maintenance  in the long run

---

I'm voting, for now, only for `1.`, I'd hold `2.` unless @AutoFixture/core has a different opinion.

klimisa:
Agreed.  Vote for 1, hold on 2. 👍

zvirja:
Fine, agreed with 1 👍 

@moodmosaic Could you clarify which exact maintenance dept you mean? The idea was that we create an empty project which imports all the `*.cs` files from the base test project. The only difference will be the newer package dependencies. If later you want to modify any particular test body, we'll need to modify it in a single place only.

If we don't have such tests - how can we be sure that we support each particular dependency version and it works? :worried:

moodmosaic:
@zvirja,

>The idea was that we create an empty project which imports all the *.cs files from the base test project. The only difference will be the newer package dependencies.

This works as long as the base test project uses Semantic Versioning, and with the hope that a major version won't have breaking changes (?)

[...]

PR #812, opened 2017-08-25
Migrate AutoFakeItEasy to .NET Core

zvirja:
Added support for .NET Core and AutoFakeItEasy. We have different dependencies depending on target:
- `1.7.4109.1` for .NET Framework 4.5
- `3.0.0` for .NET Standard 1.6

In our glue library I target .NET Standard 1.6 (instead of 1.5 we use everywhere else) as that's required by FakeItEasy.

Additionally I've added one more test project for AutoFakeItEasy to assert its support for FakeItEasy3. The reason is that previous versions (1 and 2) don't support .NET Core, so I needed at least one project where I can run tests under .NET Core runtime.

@adamchester @moodmosaic Your review is expected! 😉
@blairconrad If you have free time, it would be great if you look through and confirm everything looks fine to you ☺️

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -184,6 +184,9 @@ Target "TestOnly" (fun _ ->
         "SemanticComparisonUnitTest"
         "AutoNSubstituteUnitTest"
         "AutoMoqUnitTest"
+        "AutoFakeItEasyUnitTest"
+        "AutoFakeItEasy.FakeItEasy2UnitTest"
+        "AutoFakeItEasy.FakeItEasy3UnitTest"
         "AutoFixture.xUnit.net2.UnitTest"
     ]
 
--- a/Src/All.sln
+++ b/Src/All.sln
@@ -70,7 +70,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		CommonTestProperties.props = CommonTestProperties.props
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoFakeItEasy.FakeItEasy2UnitTest", "AutoFakeItEasy.FakeItEasy2UnitTest\AutoFakeItEasy.FakeItEasy2UnitTest.csproj", "{7F957CC0-79C3-48A1-9A3E-47BED7539248}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoFakeItEasy.FakeItEasy2UnitTest", "AutoFakeItEasy.FakeItEasy2UnitTest\AutoFakeItEasy.FakeItEasy2UnitTest.csproj", "{7F957CC0-79C3-48A1-9A3E-47BED7539248}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoFakeItEasy.FakeItEasy3UnitTest", "AutoFakeItEasy.FakeItEasy3UnitTest\AutoFakeItEasy.FakeItEasy3UnitTest.csproj", "{A7AE8091-3AC8-42A3-ACEB-32F49A8B05D1}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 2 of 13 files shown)


blairconrad:
@	zvirja, thanks for asking. The weekend's a little busy, but I can take a peek here and there. Early next week at the latest. Suits?

zvirja:
@blairconrad Thanks for the reply. Yep, will be absolutely fine. The changes in the project are very little, so its more a sanity check..

blairconrad:
@zvirja, I should've looked at the scope of the changes before I projected "early next week"!
As you say, very small change overall.

I think everything looks quite good. There's one comment I'd make, and it's more of a discussion point. I'm not even entirely sure where I stand on this. It's the AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest projects.

I had [a similar question recently in one of my own projects](https://github.com/blairconrad/SelfInitializingFakes/pull/29#issuecomment-319835591): how to balance the need to maintain support for older versions and newer versions against the testing effort. In that project, I opted to test against the earliest supported as well as a recent version of FakeItEasy. For you that would probably mean 1.7whatever and 3.0.0 and 4.0.0 (I know the latter works, because I tested it myself before releasing 4.0.0). 

So, you may want to consider adding a project to test against (what's now) 4.0.0. And if this becomes onerous, I'd suggest dropping tests against FakeItEasy 2.0.0, and just keeping those against 1.7, 3.0.0, and "latestish"

So. Just some rambling thoughts to consider. Really, I think everything you have here is great. :tada:.

zvirja:
@blairconrad Thanks for reviewing that and confirming that everything is fine!

> I had a similar question recently in one of my own projects: how to balance the need to maintain support for older versions and newer versions against the testing effort. 

Actually, I haven't thought about that yet 😅 At the moment it doesn't look too scary as we have 3 test projects only. That isn't hard to maintain them as they consist of [single file only](https://github.com/AutoFixture/AutoFixture/blob/183b18df8c2beb1ce6f44625f1aee0fe8f85c193/Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj). However, when we get more versions (e.g. enable FakeItEasy v4 support), we'll need one more..

Personally, I'd prefer to have tests for all the versions we support - who knows what will be changed in future.. That might be something that is broken in an intermediate version we support, but doesn't cover with tests 😕

If we end up having too much test projects, we could always create a sub-folder for them and move them
 all to dump as they are in fact empty 😄 However, that decision should be made later.

In any case thank you for raising this question.

@moodmosaic @adamchester Would be interesting to know your opinion about the concerns Blair raised 😉

blairconrad:
>  However, when we get more versions (e.g. enable FakeItEasy v4 support), we'll need one more..

I think FakeItEasy v4 is already supported. Or at least not specifically excluded, since you specify only [a single version (no range)](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#version-ranges-and-wildcards) for your dependency.
Perhaps I misunderstand.

In any event, thanks for listening to my comment. Nice work on the PR.

zvirja:
> Or at least not specifically excluded, since you specify only a single version (no range) for your dependency.

Currently it's being _implicitly_ supported, as the upper boundary is simply missing. We have the issue #747 to fix that and ensure that we _explicitly_ include supported versions in range. During that issue implementation it could happen that we'll need to add more (or adjust existing) test projects. That's what I mean saying "when we support it" 😊

Thank you one more time for confirming PR looks and your input.

blairconrad:
> We have the issue #747 to fix that and ensure that we explicitly include supported versions in range. 

Ah. I hadn't noticed #747. Thanks for pointing it out.

moodmosaic:
@blairconrad, thank you for taking the time to review this and letting us know about your thoughts! 👍
@zvirja, I think this PR could be merged as-is, if that's easier for you.

What @blairconrad suggests [here](https://github.com/AutoFixture/AutoFixture/pull/812#issuecomment-325116994) is interesting, but it's up to you, @zvirja, if you want to spend additional time on this PR.

zvirja:
@moodmosaic The Brair's idea is beyond this PR, therefore nothing should be changed here. Rather, we could take that into account when we explicitly [set version constrains](747) for `FakeItEasy`.

Issue #803, opened 2017-08-22
Repository line endings hell


zvirja:
I've started to work on the SourceLink integration (#770) which is [sensitive](https://github.com/ctaggart/SourceLink/wiki/Line-Endings) to line endings and found a bit of hell in our repository. It appeared that half of the files are being committed with `CRLF` endings, while other ones with `LF`.

That should be fixed, so that all files are being stored in the repo with `LF`. Later, when you checkout files on Windows, git will automatically convert them to `CRLF` during checkout and to `LF` during the commit (unless you intentionally disabled that). That is the common practice in repositories and it's quite strange to see a mess here.

We should fix files in both the `master` and `v4` branch as we constantly merge `master` to `v4`.

@moodmosaic @adamchester @klimisa Do you have any objections regarding the fix? If not, I'll fix line endings for all the files and make direct commits with that changes. Otherwise, it will be impossible to proceed with #770.

adamchester:
👍 From me

moodmosaic:
👍 From me, too.

zvirja:
Agreed! I'll try to find the right moment and will fire the commits to both branches.

zvirja:
Fixed

PR #813, opened 2017-08-29
Introduce .gitattributes file

zvirja:
This is a precursor for the #803.

In this commit I'm introducing the [`.gitattributes`](https://git-scm.com/docs/gitattributes) file that defines how line endings should be normalized on checkout/commit. I've defined all the known text and binary file extensions present in our repo to ensure that we always handle them properly.

Additionally, I've discovered that the `Src/AutoNSubstitute/GlobalSuppressions.cs` file had invalid encoding (and git was unable to handle it properly after checkout), so I fixed it as well.

Notice, I haven't touched all the already committed files with invalid file endings - I'll do that later _without_ PR as changes should be applied quickly and in the right moment of time (we don't have PRs that are about to be merged).

@moodmosaic @adamchester Inviting you here for review, please take a look 💛


.gitattributes
```diff
@@ -0,0 +1,37 @@
+# Enable automatic line ending conversion for text files on checkout and commit.
+# For all extensions that are not being explicitly defined below git will try to guess
+# whether file contains text and if so, will convert the line endings.
+* text=auto
+
+# Define all the file extensions for which we should always apply line ending covnersion
+*.cs      text diff=csharp
```
moodmosaic: `text diff=csharp` looks pretty cool 👍 What about the other extensions? `.fs`, `.xml`, etc? 😃

zvirja: `csharp` is the only supported language we need. See the full list [here](https://git-scm.com/docs/gitattributes#_defining_a_custom_hunk_header).

moodmosaic: I'm not surprised that I don't see `fsharp` on that list. FWIW, AutoFixture.AutoFoq and Idioms.FsCheck are written [*in*<sup>*</sup>](https://codeblog.jonskeet.uk/2008/04/23/programming-quot-in-quot-a-language-vs-programming-quot-into-quot-a-language/) F# 😄 

---

<sup>*Means, they're almost entirely written using [object programming](https://twitter.com/dsyme/status/843849405212098560) rather than functional programmin O.o</sup>

zvirja: @moodmosaic Just noticed that you haven't approved the PR - do you have some other concerns about it?

PR #818, opened 2017-08-30
Improve fixture customization performance by adding laziness to adapted collection

zvirja:
## Introduction

Usually Fixture is used together with other glue libraries like `AutoMoq` or `AutoNSubstitute`. Also often fixture is customized per project needs and later the customized fixture instance is used all over the project (e.g. via custom `AutoData` attribute).

For xUnit you have a single `AutoData` attribute per method, so during test discovery (and execution) fixture is created large number of times and is being customized on construction for each time. The customization performance affects the discovery speed and with runners like NCrunch (that re-run tests on each modification you make) we are sensitive to that speed (usually you want to see whether nothing was broken after the change, so you wait for results).

## Issue

I've profiled fixture customization and found a redundancy that helps to save some time. It appeared that each time a customization is added, we re-create the `Customizations` and `ResidueCollector` objects. If you apply a few customizations in a row, the e.g. `ResidueCollector` instances might be never touched. However there is an expensive lookup in object constructor which takes time even if value is never used later. I've fixed the issue by adding the lazy evaluation, so we perform expensive lookup only if that's indeed required.

Also I found that logic had redundancy and we performed lookup twice. That has been fixed as well.

## Measurements
I used BenchmarkDotNet tool to measure results. Here is my setup:
```csharp
[ClrJob]
public class CustomizationSpeedTest
{
    [Benchmark]
    public void CreateFixture() => new Fixture().Customize(new TestCustomization());
}

class TestCustomization: ICustomization
{
    {
        fixture.Customize(new IncrementingDateTimeCustomization());
        fixture.Customize(new NoDataAnnotationsCustomization());
        fixture.Inject(42);
        fixture.Freeze<object>();
    }
}
```
Before (v. 3.50.6):
```
BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 1 (10.0.14393)
Processor=Intel Core i7-4770 CPU 3.40GHz (Haswell), ProcessorCount=8
Frequency=3320309 Hz, Resolution=301.1768 ns, Timer=TSC
  [Host] : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2102.0
  Clr    : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2102.0

Job=Clr  Runtime=Clr

        Method |     Mean |    Error |   StdDev |
-------------- |---------:|---------:|---------:|
 CreateFixture | 213.9 us | 1.942 us | 1.621 us |
```

After:
```
BenchmarkDotNet=v0.10.9, OS=Windows 10 Redstone 1 (10.0.14393)
Processor=Intel Core i7-4770 CPU 3.40GHz (Haswell), ProcessorCount=8
Frequency=3320309 Hz, Resolution=301.1768 ns, Timer=TSC
  [Host] : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2102.0
  Clr    : .NET Framework 4.6.2 (CLR 4.0.30319.42000), 32bit LegacyJIT-v4.7.2102.0

Job=Clr  Runtime=Clr

        Method |     Mean |     Error |    StdDev |
-------------- |---------:|----------:|----------:|
 CreateFixture | 148.5 us | 0.6110 us | 0.5716 us |
```

The mean value changes from ~213us to ~149us - that is about 30% boost. Not too bad for such a small change :)

On my working solution NCrunch test discovery decreased from approx. 4.5 sec to 3.7 sec. Given that this is delay after each code change - result is great.

---

@adamchester This PR might be interesting to you as it's about performance.
@moodmosaic Your review is also expected if possible 😃

P.S. I've tuned the number after the initial PR creation - that's because I did measurements before I discovered one more place to improve.

### DIFF
--- a/Src/AutoFixture/SpecimenBuilderNodeAdapterCollection.cs
+++ b/Src/AutoFixture/SpecimenBuilderNodeAdapterCollection.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture
@@ -23,6 +22,28 @@ namespace Ploeh.AutoFixture
     {
         private readonly Func<ISpecimenBuilderNode, bool> isAdaptedBuilder;
         private IEnumerable<ISpecimenBuilder> adaptedBuilders;
+
+        private IEnumerable<ISpecimenBuilder> AdaptedBuilders
+        {
+            get
+            {
+                // We evaluate this value lazily for the performance reasons.
+                // Usually during the Fixture construction it's customized a lot of times.
+                // Some of the collections are not even touched after the construction and are immediately
+                // recreated again after a subsequent customization.
+                // To cover such scenarios we evaluate value on demand only saving the initialization time. 
+                if (this.adaptedBuilders != null) return this.adaptedBuilders;
+
+                // The intermediate "result" variable is needed to ensure that null value can be never returned
+                // in case of concurrency (we set field to null). While current collection implementation doesn't 
+                // seem to support concurrency, the additional guard adds more safety.
+                var result = this.adaptedBuilders = Graph.FindFirstNode(TargetMemo.IsSpecifiedBy);
+                return result;
+            }
+        }
+
+        private void InvalidateCachedAdaptedBuilders() => this.adaptedBuilders = null;
+
 
         /// <summary>
         /// Initializes a new instance of the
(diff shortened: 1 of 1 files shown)


zvirja:
@adamchester Given that this PR is all about performance, do you know whether you will be able to allocate some time in the nearby future and review it? As you can see, no tests failed, so logically nothing changed.

I'll gain a benefit from it in my on-day basis activity so I'm a bit impatient to release this stuff 😉

zvirja:
@moodmosaic  Adam seems to be absent - could you please take a look (unless you explicitly decided to skip this one) if possible?

moodmosaic:
@zvirja, I apologize for the being late here.

Perhaps, in addition to what I wrote [here](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-330116357), you could also go ahead and merge these perf-related changes on your own―as long as no breaking changes are introduced.

👍 :1st_place_medal: 💯 🐴 🚀

PR #819, opened 2017-08-31
Use MSBuild based NuGet restore

zvirja:
Currently we use NuGet.exe directly to restore packages for the project. However after we migrated all our projects to a new format (.NET Core SDK) that is no longer required and we can use built-in MSBuild restore.

Additionally, I've changed the restore to not specify the specific packages location and use the default one. That was done for a few reasons:
1. To not fight with VS. Currently if you build solution via FAKE and later open the solution, VS will re-restore the packages to the default location.
2. By default MSBuild restores packages to the machine-wide packages cache. When we restore to a custom location we in fact make a copy of packages. As result restore takes longer and project occupies unnecessary space on the hard drive.

I've checked project after the refactoring and can confirm that you can work in offline after you run the `build.cmd` - MSBuild uses the already cached packages and is able to successfully compile the project.

@moodmosaic @adamchester @klimisa Please review and confirm if you are fine.

### DIFF
--- a/Build.fsx
+++ b/Build.fsx
@@ -163,10 +163,10 @@ Target "CleanAll"           DoNothing
 Target "CleanVerify"        (fun _ -> clean "Verify")
 Target "CleanRelease"       (fun _ -> clean "Release")
 
-Target "RestoreNuGetPackages" (fun _ ->
-    solutionsToBuild
-    |> Seq.iter (RestoreMSSolutionPackages (fun p -> { p with OutputPath = nuGetRestoreFolder }))
-)
+// Configuration doesn't matter for restore and is ignored by MSBuild.
+let restoreNugetPackages() = build "Restore" "Release" [ "RestorePackagesPath", FullName nuGetRestoreFolder ]
+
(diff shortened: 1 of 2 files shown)


zvirja:
@moodmosaic If you agree with what @adamchester wrote [here](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-323188537), we could confirm this PR without review - currently build scripts are being maintained by me only 😉

moodmosaic:
@zvirja, yes, go ahead and merge that one if you've tested it and know that it works 👍 💯

zvirja:
@moodmosaic From the comment it looks like you approved it, however PR state is not approved.. Could you please clarify? 😉

moodmosaic:
Oh, I thought you were going to merge this PR as per [this](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-323188537) comment? 😃

zvirja:
@moodmosaic Well, I missed a point that we agreed there that I can adjust build scripts without approval..
 😅 Fine, so I'll merge it later today 😉

PR #820, opened 2017-08-31
Fix minor warnings in xunit tests

zvirja:
Fix a few minor issues in our unit tests:
1. Use valid assertion method.
2. Remove duplicated theory data. One more confirmation that copy-paste approach is error prone 😉

@moodmosaic @adamchester Please review and accept - changes are very tiny.

### DIFF
--- a/Src/AutoFixtureUnitTest/DataAnnotations/RangeAttributeRelayTest.cs
+++ b/Src/AutoFixtureUnitTest/DataAnnotations/RangeAttributeRelayTest.cs
@@ -6,7 +6,6 @@ using Ploeh.AutoFixture.DataAnnotations;
 using Ploeh.AutoFixture.Kernel;
 using Ploeh.AutoFixtureUnitTest.Kernel;
 using Xunit;
-using Xunit.Extensions;
 
 namespace Ploeh.AutoFixtureUnitTest.DataAnnotations
 {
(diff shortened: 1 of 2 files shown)


Src/AutoFixtureUnitTest/DataAnnotations/RangeAttributeRelayTest.cs
```diff
@@ -125,16 +124,12 @@ public void CreateWithRangeAttributeRequestReturnsCorrectResult(Type type, objec
         [InlineData("Property", "-2.2", "-1.1")]
         [InlineData("Property", 10.0, 20.0)]
         [InlineData("Property", -2.0, -1.0)]
-        [InlineData("Property", 10, 20)]
```
moodmosaic: Why are those removed?

zvirja: If you look around you'll notice that are duplicated 😉 You are the author of these lines so it seems that you mistakenly introduced that by copy-paste 😄

zvirja: image

moodmosaic: 👍 👍 😄

PR #817, opened 2017-08-30
Migrate NUnit3 project to support .NET Core

zvirja:
Migrated the NUnit3 project to .NET Standard. We depend on different NUnit versions for different frameworks:
- `3.0.1` for .NET Framework 4.5
- `3.7.0` for .NET Standard 1.5

Everything went smoothly and no important changes are present. Most of the changes are related to changing our build scripts to run NUnit3 tests using `dotnet test` and later uploading the results to AppVeyor. Also I deleted `NUnit.Console` v3 runner from the repository as we don't longer use it.

@moodmosaic @adamchester Looking forward to your review. That is the last important migration to .NET Standard (we have only `Idioms` and F# projects left and they are blocked by dependencies). Hope that will cheer you if you are tired from such PRs 😃

PR #824, opened 2017-09-03
Passive Customization Attributes are ignored when used in combination with [Frozen] (#637, NUnit3, v4)

sergeyshushlyapin:
Solves #637 for NUnit3.


zvirja:
Notice, the merge of this one should be postponed untill the #817 is merged. After that it should be rebased and fix the support for .NET Core.

zvirja:
I've rebased code on top of the latest `v4` and added a fix for multi-framework support.

PR #827, opened 2017-09-06
[Build] Refactor build invocation to add new features

zvirja:
Previous implementation didn't allow to:
- Run build without configuration (needed by NuGet Restore)
- Specify build verbosity

In this commit the build invocation was rewritten to use FAKE function with wider MSBuild configuration. Additionally, now build verbosity and configuration could be configured via build parameters.


zvirja:
Merging this one without code review as per [this](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-323188537) agreement.

PR #830, opened 2017-09-07
Use ThreadLocal<> instead of the ConcurrentDictionary

zvirja:
Usage of the `ThreadLocal<>` instead of custom dictionary allows to remove the package dependency on `System.Threading.Thread` and make the code cleaner.

Also it offers better performance as far as I briefly spotted in the profiler. I haven't tested that, though.

Unless you know about any special reason why you used the dictionary (likely, simply for historical reasons), I suggest to the the `ThreadLocal<>` - it has been designed for exactly such scenarios.

### DIFF
--- a/Src/AutoFixture/Kernel/RecursionGuard.cs
+++ b/Src/AutoFixture/Kernel/RecursionGuard.cs
@@ -1,8 +1,7 @@
 using System;
 using System.Collections;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 
 namespace Ploeh.AutoFixture.Kernel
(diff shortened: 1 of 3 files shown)


Src/AutoFixture/Kernel/RecursionGuard.cs
```diff
@@ -11,16 +10,16 @@ namespace Ploeh.AutoFixture.Kernel
...
+    [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", 
+        Justification = "The main responsibility of this class isn't to be a 'collection' (which, by the way, it isn't - it's just an Iterator).")]
+    [SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable",
+        Justification = "Fixture doesn't support disposal, so we cannot dispose current builder somehow.")]
     public class RecursionGuard : ISpecimenBuilderNode
     {
-        private readonly ConcurrentDictionary<Thread, Stack<object>>
-            _requestsByThread = new ConcurrentDictionary<Thread, Stack<object>>();
+        private readonly ThreadLocal<Stack<object>> _requestsByThread
+            = new ThreadLocal<Stack<object>>(() => new Stack<object>());
```
moodmosaic: It has been since 2006 that I have to deal with threading explicitly<sup>*</sup>. IIRC, `ThreadLocal<T>` is similar to what we're doing with the concurrent dictionary. Your proposed changes result to one less assembly reference, so I think we should do this 👍

<sup>*Even at that time, we were using [Jeffrey Richter's Power Threading Library](https://github.com/Wintellect/PowerThreading) and its AsyncEnumerator class, which in some way *is* the ancestor of C# async/await, despite being said that it was *only* influenced by F#'s `async` Computation Expression.</sup>

zvirja: Interesting.. And what was the reason you gone away from `async`? Have you found it redundant? I haven't see any `async` code this days - probably I'm simply missing something 😊

moodmosaic: If I write C# code I actually have my functions or method return by of type `async`. Why it's not used in AutoFixture, I don't know since I'm not the creator of it.

zvirja: Oh, I see it. I was confused by `we were using` - I thought that it's here, in AutoFixture 😅

Issue #784, opened 2017-07-21
NSubstitute dependency approach discussion


zvirja:
Currently the `AutoNSubstitute` project depends on [NSubstitute 1.5](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoNSubstitute/packages.config#L3). However, current integration suffers from significant issues and limitations (see #720, #592, #707 and #653). There is a way to fix them, however we need the API which appeared in NSubstitute [2.0.2](https://github.com/nsubstitute/NSubstitute/blob/master/CHANGELOG.txt) only. Therefore, we need to decide how we'll proceed with this glue library.

### Way 1 - Keep single `AutoNSubstitute` project
With this way we continue with current `AutoNSubstitute` project but increase NSubstitute dependency to `2.0.2`. We will introduce this change in v4 only and describe that on the Breaking Changes page, so that should not be an issue.

__Pros:__
- We'll have two less project to support (glue + tests) 😉
- Less confusion. We need a never version of lib to perform the bug fix and change the behavior (support generics). If we implement that in v2 only, it might be quite confusing that for different glue libraries the behavior is different. As far as I know, currently all the versions of a glue library offer the same functionality.

### Way 2 - Create the `AutoNSubstitute2` project
Usually we followed this way when we wanted to support yet another version of library. However, AFAIK in that cases presence of the project was caused by the breaking changes rather than functionality limitation (if I see that clearly).

__Pros:__
- We continue to support all the NSubstitute libraries
- No breaking changes for the original NSubstitute, so it would be easier to consume v4 (it will be possible to migrate solution step by step). Users will have ability to choose when they want to switch to NSubstitute 2 and a new glue library. 

__Cons:__
- The `AutoNSubstitute` library is still compatible with `NSubstitute 2.0.0`, so it would a little mess to have two compatible versions of the glue library. Of course, we could artificially set a cap for AutoNSubstitute v1, but that limitation will be a small lie as binary everything is fine 😞

----
Personally, I don't have a strong opinion here because I see a value in both approaches. However, if I'm forced to choose, I'd prefer the way 1 because it's easier and NSubstitute 2 [doesn't have](https://github.com/nsubstitute/NSubstitute/blob/master/CHANGELOG.txt) breaking changes, so that would not be a problem to migrate to it.

@AutoFixture/core What do you think? ;-)

moodmosaic:
Yes, the less stuff to maintain, the easier it can be to move forward. I'm voting for **1**.

[...]

PR #832, opened 2017-09-11
Rework NSubstitute integration approach

### LINKED ISSUE #592: AutoConfiguredNSubstituteCustomization not working correctly in XUnit theories with parallel tasks

zvirja:
## Description

Our current integration with NSubstitute is a bit hacky and suffers from issues and limitations:
- generic methods are not supported (#720);
- void methods are ignored due to the way we integrate with NSubstitute. As result the `ref` and `out` parameters are not being filled for the void methods;
- it's impossible to configure return value for methods with `ref` and `out` parameters (#707). That's because AutoFixture substitutes argument values with auto-generated ones and the resulting call specification becomes invalid;
- integration doesn't support concurrency well and integration could fail with exceptions or produce invalid results (#592);
- the property values are resolved via `Type` request instead of `PropertyInfo` which causes some confusion (#653).

To fix these issues I worked tightly with NSubsitute team and we added a better integration point to NSubsitute2 (see [here](https://github.com/nsubstitute/NSubstitute/pull/259)). That was [quite](https://github.com/nsubstitute/NSubstitute/pull/156) [a long](https://github.com/nsubstitute/NSubstitute/pull/234) way started by @mrinaldi some time ago and I finished what he started.
Starting from v2 NSubstitute provides an API to register a custom call handler to resolve call results. This is a "passive" integration compared to the current approach - we are asked about call results rather than we inject hooks into each method we discovered. This approach supports concurrency much better and shouldn't fail because of it. I was capable of fixing all the issues mentioned above simply by using that API.

I think the consumers of this new integration approach should be happy 😃

## Testing

After I enabled a new approach none of the existing scenario (Facade) tests failed except ones asserting that we ignore `void` and `generic` methods - that was a technical limitation rather than a feature. Also I tried the updated integration on my real solution that heavily relies on NSubstitute integration and only a couple of tests failed - those were rather related to obscure NSubstitute usage than to issues with integration. That contributed to my confidence that this integration should work fine.

## NSubstitute 2 dependency

As it has been agreed in #784 I changed the dependency so that now we require NSubstitute 2.0.3 for both .NET Framework and .NET Standard. I need that consume the new API I previously introduced.

## Final notes

I know that PR is pretty large as it changes most of the integration with NSubsitute. However, I don't see any way to split it - most of the changes are logically coupled and should be done together. I'll definitely owe you some beer after the review 🍺🍻😉

Inviting @moodmosaic @adamchester @ecampidoglio to review this PR.
@mrinaldi Would be cool if you could also take a look - you are the author of the current integration.

PR closes #592, #653, #707, #720 and #784.

### DIFF
--- a/Src/AutoNSubstitute/AutoConfiguredNSubstituteCustomization.cs
+++ b/Src/AutoNSubstitute/AutoConfiguredNSubstituteCustomization.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using NSubstitute.Core;
 using Ploeh.AutoFixture.Kernel;
 
 namespace Ploeh.AutoFixture.AutoNSubstitute
@@ -9,8 +10,6 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
     /// </summary>
     public class AutoConfiguredNSubstituteCustomization : ICustomization
     {
-        private readonly ISpecimenBuilder builder;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="AutoConfiguredNSubstituteCustomization"/> class.
         /// </summary>
@@ -27,9 +26,9 @@ namespace Ploeh.AutoFixture.AutoNSubstitute
         public AutoConfiguredNSubstituteCustomization(ISpecimenBuilder builder)
         {
             if (builder == null)
-                throw new ArgumentNullException("builder");
+                throw new ArgumentNullException(nameof(builder));
 
-            this.builder = builder;
+            this.Builder = builder;
         }
 
         /// <summary>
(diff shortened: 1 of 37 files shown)


zvirja:
@alexandrnikitin A bit later, but just found that I promised to mention you. If you have free time - feel free to take a look, so I could address your concerns in via subsequent request :wink:

PR #844, opened 2017-09-30
Set upper boundary for external dependencies

### LINKED ISSUE #747: Ensure that for each nuspec file we have set upper boundary for dependencies

zvirja:
Closes #747.

Reviewed all the projects and ensured that upper boundary for dependency is always set. In fact, we have only 2 noticeable changes:
1. Support FakeItEasy4 explicitly, therefore added the `AutoFakeItEasy.FakeItEasy4UnitTest` project. I prefer to have test projects for all the supported versions - it costs almost nothing to support them, but we can be sure that we suddenly don't break some intermediate version support.
1. Explicitly support `FsCheck` till v2 inclusively for `Idioms.FsCheck` library. I've added 2 test projects: `Idioms.FsCheck.FsCheck1UnitTest` and `Idioms.FsCheck.FsCheck2UnitTest`. The idea is the same as for the `FakeItEasy` - if we support library, we should have the covering tests.

@moodmosaic @adamchester Please let me know whether you have any concerns 😉

### DIFF
--- a/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="FakeItEasy" Version="2.0.0" />
+    <PackageReference Include="FakeItEasy" Version="[2.0.0]" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Src/All.sln
+++ b/Src/All.sln
@@ -73,6 +73,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoFakeItEasy.FakeItEasy2U
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoFakeItEasy.FakeItEasy3UnitTest", "AutoFakeItEasy.FakeItEasy3UnitTest\AutoFakeItEasy.FakeItEasy3UnitTest.csproj", "{A7AE8091-3AC8-42A3-ACEB-32F49A8B05D1}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AutoFakeItEasy.FakeItEasy4UnitTest", "AutoFakeItEasy.FakeItEasy4UnitTest\AutoFakeItEasy.FakeItEasy4UnitTest.csproj", "{B8B07C7C-E7C8-4272-AAA4-E4ED4CCD249C}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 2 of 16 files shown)


Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
```diff
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="FakeItEasy" Version="2.0.0" />
+    <PackageReference Include="FakeItEasy" Version="[2.0.0]" />
```
Kralizek: Wouldn't it be better to use "[2,3)" or "[2,2.1)" if you want only patch revisions to be let through.

zvirja: Logically it probably makes sense, however I prefer the strict specification. I want to test the initial version only, as SemVer guarantees that if the initial version is OK - any minor will be OK too. Actually, even the previous specification worked as it always resolved to the lower range version, but I wanted to be double sure and therefore altered it.

Given that this is a test project and no other project depends on it, I'd prefer to keep it as is.

Kralizek: I see your point, but what if I need a non-breaking feature introduced in the 2.1.0?

zvirja: Where do you need that feature? If you look more precisely, you will find that this project's folder is totally empty and project contains no own files. Rather, it imports files from other test project. Its the only purpose is to run our existing unit tests under the different test framework and verify that we haven't broken them for a newer supported version. This project is _never_  intended to have own files so it cannot use features from this particular version. Therefore, your point seems to be not applicable here, or I'm missing something. :confused:

Kralizek: Ouch, my bad :)
I was on mobile when I added this comment and I didn't see that you implemented the "[x,y)" approach in other projects. I had wrongly assumed that this was the approach for the whole PR.
Apologies.

zvirja: No worries! Thanks for bringing an extra-attention on these lines and thanks for monitoring our PRs - appreciate that. That's really cool! If I make a wrong decision (or simply overlooked something) you might save the world 😉

I'm happy that this time everything was fine.

zvirja:
@moodmosaic Yep, I can - however wanted to see whether you have any feedback on that. It's fine that no - so I'll simply proceed with this one 😉

PR #846, opened 2017-10-01
[Housekeeping] Update xUnit version to v2.3 beta5

zvirja:
Updated the xUnit version we use.

Most of the changes are to fix the xUnit analyzer warnings as there is a better way to assert test outcome. Also it appeared that for the `CreateWithPropertyDecoratedWithRangeAttributeReturnsCorrectResult` and `CreateWithFieldDecoratedWithRangeAttributeReturnsCorrectResult` tests we don't use the passed inline data, so we could simplify the logic.

@moodmosaic @adamchester Please take a look 😉


zvirja:
Closed [as agreed](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-333778338).

PR #847, opened 2017-10-01
[Housekeeping] Disable parallel test execution for RhinoMock integration tests

zvirja:
It appears that RhinoMock is not thread safe and our xUnit tests could fail due to parallel test execution. The issue is intermittent and outcome might be different - from exception to complete hang.
You can read more information about this issue here: http://blog.smithfamily.dk/post/Thread-safe-version-of-Rhino-Mocks

We use an old version of RhinoMock that is no longer supported, so the only option we have is to disable the parallel test execution.

Here is an example of such a failure: [https://ci.appveyor.com/project/AutoFixture/autofixture/build/1.0.323-admnsybb](https://ci.appveyor.com/project/AutoFixture/autofixture/build/1.0.323-admnsybb).

@moodmosaic @adamchester Please take a look 😉

### DIFF
--- a/Src/AutoRhinoMockUnitTest/Properties/AssemblyInfo.cs
+++ b/Src/AutoRhinoMockUnitTest/Properties/AssemblyInfo.cs
@@ -1,5 +1,6 @@
 ﻿using System.Reflection;
 using System.Runtime.InteropServices;
+using Xunit;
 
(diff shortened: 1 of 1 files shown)


zvirja:
Closed without review [as agreed](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-333778338).

PR #850, opened 2017-10-01
[Housekeeping] Remove Foq usage sample files as they no longer works

zvirja:
Currently NuGet dependencies are resolved to the global cache, therefore it's impossible to open reference in fsx files. As results, those sample files are broken and don't work out of box. Given that those files
were installed by the package and are not required by the compilation, I removed them.


zvirja:
Closed without review [as agreed](https://github.com/AutoFixture/AutoFixture/issues/703#issuecomment-333778338).

PR #855, opened 2017-10-03
Fix and enable suppressed code analysis rules

### LINKED ISSUE #797: Review the suppressed code analysis rules and fix important ones
In PR #773 we suppressed a set of Code Analysis rules that started to fail after we turned on the Roslyn based analyzers. Review the rules in the `AutoFixture.ruleset` file marked as "Info" and make them either warnings, or disable them at all.

zvirja:
Closes #797.

Reviewed all the previously suppressed warnings, fixed them and enabled. It was quite a long way though..

I've fixed and enabled almost all of the warnings and disabled only 3 ones:
- CA1825:Avoid zero-length array allocations - The API it suggests to use is only available since .NET 4.6.
- CA2243: Attribute string literals should parse correctly - It's a [false positive](https://connect.microsoft.com/VisualStudio/feedback/details/796535/the-ca2243-rule-attribute-string-literals-should-parse-correctly-should-not-be-applied-to-assemblyinformationalversionattrubute) and it prevents us from writing the arbitrary data to the `AssemblyInformationalVersion` attribute (sha of commit in our case) which is absolutely legal.
- CA1006: Do not nest generic types in member signatures - It was previously suppressed so I decided to not touch. It seems like a false positive in most cases as it mostly fails on `Expression<Func<T>>`. 

@adamchester @moodmosaic Please take a look 🐴I'd recommend to go commit by commit as it would be simpler to see what exactly I changed for which exact rule.

### DIFF
--- a/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using FakeItEasy;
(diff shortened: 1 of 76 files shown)


zvirja:
@moodmosaic It was just in case you have any concerns regarding the changes 😉 However I'll proceed and merge as it seems you don't have 😄

PR #860, opened 2017-10-05
Add tests and impl for a random DateTimeOffset generator

steve-codemunkies:
This generator is based on the existing RandomDateTimeSequenceGenerator.
By randomly setting the offset you are able to ensure you are dealing
correctly with data from different time zones.

We (On The Beach) needed to generate DateTimeOffsets to test some of our
code. We have been using this generator for a while now, and in cleaning
up our code-base decided it was time to contribute this back.

Thanks for the great tool!


zvirja:
@steve-codemunkies Could you please clarify what is the difference compared to the #710? If there is no any difference - could we close this one and proceed further in #710?

steve-codemunkies:
@zvirja I pushed this so you can see the difference between the two, and then if you're really adamant in proceeding with the `RandomDateTimeSequenceGenerator` version you can merge this. Please see my comment in #710.

Issue #670, opened 2016-07-05
Support for Seed


matkoch:
It would be nice to repeat tests and have the same dummy values provided from AutoFixture.

ploeh:
Could you elaborate?

moodmosaic:
Do you mean something similar to what QuickCheck (and clones) do?

matkoch:
@moodmosaic sorry I'm not aware of QuickCheck :) But will check out later.

@ploeh actually we had this discussion already [here](http://stackoverflow.com/questions/30078556/can-i-re-generate-random-values-in-autofixture-using-a-seed). I think the arguments still hold.

To summarize: for some (not all projects), it would be handy to repeat a test run, and let AutoFixture generate the exact same values, as in a failed run. For instance, a run that happened on CI server.

moodmosaic:
>it would be handy to repeat a test run, and let AutoFixture generate the exact same values, as in a failed run

AFAICT, that would require to change some of the built-in generators (for one, strings shouldn't be random GUIDs anymore).

ploeh:
While this would be nice to have, it'd require a major rewrite of the AutoFixture kernel. Although I haven't spent any significant time thinking this through, I'd be surprised if this was possible without introducing breaking changes.

In all the years I've been using AutoFixture, I've never _needed_ this feature. Likewise, I can't remember having received a similar feature request before now.

These days, I mostly write F# code, and for that, I use [FsCheck](https://fscheck.github.io/FsCheck), which _does_ have a replay feature. Even so, I never use that feature with FsCheck.

For all of those reasons, I consider the utility of this feature limited, and the ratio of impact to benefit to be unwarranted.

It's possible that I'm mistaken. If someone wants to prove me wrong with a series of [good pull requests](http://blog.ploeh.dk/2015/01/15/10-tips-for-better-pull-requests), I shall be happy to reconsider my position.

For now, however, I'm going to close this issue.

ploeh:
BTW, reproducing an issue that happened on a CI server shouldn't be that difficult. I've been in that situation more than once.

Often, the output from the test run should already enable you to understand what went wrong. Sometimes, I grant, this hasn't been the case for me. When that happens, I always take that as an indication that my assertion messages are insufficiently useful, and then I refactor the test in question to include better failure messages.

Later failures of Erratic Tests will then enable me to zero in on the problem using only the test log.

This, I find, improves the overall quality of my tests.

0xorial:
[...]

PR #865, opened 2017-10-06
[NUnit3] Make `FixedNameTestMethodBuilder` the default strategy

### LINKED ISSUE #862: NUnit3: Make `FixedNameTestMethodBuilder` the default strategy
Currently we have the [known issue](https://github.com/AutoFixture/AutoFixture/wiki/Known-Issues#test-name-strategies-for-nunit3) about manual actions for full NUnit integration support with VS and NCrunch runners. We should make the `FixedNameTestMethodBuilder` strategy the default one in v4, so everything works out-of-the-box. 

If later user find that it needs volatile names for better diagnostic, they could change the default strategy in the way described in the known issue.

zvirja:
Closes #862.

In this PR I make the `FixedNameTestMethodBuilder` strategy (recently introduced in #709) the default one. The reason behind that is to support the VS and NCrunch test runners out-of-the-box if user uses `AutoData` and `InlineAutoData` attributes and NUnit3.

The cons is that this strategy makes it a bit harder to troubleshoot if particular test failed:
Test name before: `SomeTest(42,"autogeneratedstring")`
Test name after: `SomeTest(auto<Int32>,auto<String>)`

Now we show the placeholders instead of actual values. If the particular test fails, you cannot look on the test input to understand why that happened. Rather, you need to rely on the [assertion methods output](https://github.com/AutoFixture/AutoFixture/issues/670#issuecomment-235198841), which not actually so bad..

I think that being able to work with all the runners by default is more important, than potentially decreased diagnostic opportunities. If user doesn't like how it works now, he could switch the name strategy back by following [the described approach](https://github.com/AutoFixture/AutoFixture/wiki/Known-Issues#test-name-strategies-for-nunit3).

@moodmosaic @adamchester Please let me know what you think about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,6 @@ Notice how we can reduce unit tests to state only the relevant parts of the test
 
 Using AutoFixture is as easy as referencing the library and creating a new instance of the Fixture class!
 
-#### NUnit3 Known Issue
-You might need to change the test name strategy for compatibility with specific test runners. Read more about that [on our wiki](https://github.com/AutoFixture/AutoFixture/wiki/Known-Issues#test-name-strategies-for-nunit3).
-
 ## .NET platforms compatibility table
 
 | Product            | .NET Framework            | .NET Standard            |
--- a/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/TestNameStrategiesTest.cs
@@ -17,7 +17,7 @@ namespace Ploeh.AutoFixture.NUnit3.UnitTest
             var sut = new AutoDataAttribute();
             // Verify outcome
             Assert.That(sut.TestMethodBuilder,
-                Is.TypeOf<VolatileNameTestMethodBuilder>());
+                Is.TypeOf<FixedNameTestMethodBuilder>());
             // Teardown
         }
 
@@ -29,7 +29,7 @@ namespace Ploeh.AutoFixture.NUnit3.UnitTest
             var sut = new InlineAutoDataAttribute();
             // Verify outcome
             Assert.That(sut.TestMethodBuilder,
-                Is.TypeOf<VolatileNameTestMethodBuilder>());
+                Is.TypeOf<FixedNameTestMethodBuilder>());
             // Teardown
         }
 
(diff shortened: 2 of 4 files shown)


frblondin:
If you plan to change this default behavior in a major release, you could also change the NUnit prerequisite to 3.5+ and remove this [hack](https://github.com/zvirja/AutoFixture/blob/4af18e124e0db73c860b406dda80cc805811e1a2/Src/AutoFixture.NUnit3/FixedNameTestMethodBuilder.cs#L75) as it is no longer required with recent versions of NUnit3.

frblondin:
Fair enough, @zvirja

Kralizek:
Is this baked in into a nightly build? I'm using the latest, 4.0.0-alpha.253, and I don't get the test names to be fixed and therefore discoverable by VS 2017 Live Tests

zvirja:
This PR wasn't merged yet, so no. Will be available in 5 days, after the PR is merged.

zvirja:
@Kralizek You can now check how VS Live testing works with the latest AutoFixture drop.

Kralizek:
@zvirja I don't have concrete evidences but I believe that using the AutoData attributes sometimes causes Live Testing to crash and need manual restart.

zvirja:
@Kralizek It's bad 😢 Unfortunately, I don't have VS Enterprise and I don't use NUnit to test this feature by myself. Would be awesome if you could try to collect some diag info to troubleshoot why that happens. That might be also a bug in VS.

@frblondin Since you used this approach for a while - have you observed something similar?

frblondin:
@Kralizek there can be multiple causes for test framework crashes. I personally encountered many of them:
 IFixture creation containing an exception, library loading exceptions depending on AppDomain management of test runner or parallel activation binding redirects... I don't know about live testing since I am using the pro edition.

Is it specific to live testing?
Do you have a callstack?
Have you tried running these tests using NUnit runner successfully? Note that if it crashes also with NUnit runner you can debug it from within Visual Studio. See this [stackoverflow question](https://stackoverflow.com/a/759878).

There is a 1-month trial for the enterprise edition so if you are stuck, the questions above don't help you, and have a sample project incl. a procedure I could try to debug it.

Kralizek:
NUnit tests works as they should
* dotnet test
* R#
* Visual Studio Test Explorer

I will try to repro and attach the data I got as soon as possible :)

PR #883, opened 2017-10-11
Create a standalone package with seed methods

### LINKED ISSUE #151: Move all Create overloads taking a seed to another library
In my experience, the overloads for `Create` and `Freeze` methods that take a seed as extra argument cause much confusion (see [this discussion](http://stackoverflow.com/q/18161127/126014) for just one example).

All extension methods that take a seed as argument should be moved out of AutoFixture and into a separate library. This removes them from the mainstream AutoFixture API, which, hopefully, should make AutoFixture easier to use.

For those users still needing this feature, the extension methods will exist in a separate library, available as a separate NuGet package.

We can monitor the download statistics for that NuGet package to see if anyone uses that feature at all, and then completely retire it if that's not the case.

Please note that this is a breaking change, so the move cannot be done until AutoFixture 4, and complete removal will have to wait until AutoFixture 5 (if ever).

Implementation:
- [x] Create a standalone package with seeded methods
- [x] Update documentation and add a remark about package
- [x] Update Readme with a link to NuGet package and compatibility table.

zvirja:
Implements #151.

Extracted all the `fixture.Create()` and `fixture.Freeze()` overloads with a `seed` parameter to a standalone package. The reason is that those overloads are confusing for new (and old) guys and might be not widely needed.

With current approach people could decide whether they actually need those API.

I decided to proceed with `AutoFixture.SeedExtensions` name rather than `AutoFixture.SeededCreation`, because this library contains methods not for `Create` only, but for the `Freeze` as well. This will make package name more consistent.

After this PR is merged we'll need to adjust our documentation cases that use seed and add a remark that 
`AutoFixture.SeedExtensions` library should be installed for that.

@moodmosaic @adamchester Please take a look. This is the last issue in our v4 list for now. The only leftover is namespace adjustment only - will be done after this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,26 +72,28 @@ Using AutoFixture is as easy as referencing the library and creating a new insta
 
 ## .NET platforms compatibility table
 
-| Product            | .NET Framework            | .NET Standard            |
-| ------------------ | ------------------------  | ------------------------ |
-| AutoFixture        | :heavy_check_mark: 4.5.2  | :heavy_check_mark: 1.5   |
-| AutoFixture.xUnit  | :heavy_check_mark: 4.5.2  | :heavy_minus_sign:       |
-| AutoFixture.xUnit2 | :heavy_check_mark: 4.5.2  | :heavy_check_mark: 1.5   |
-| AutoFixture.NUnit2 | :heavy_check_mark: 4.5.2  | :heavy_minus_sign:       |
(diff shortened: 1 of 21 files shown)


moodmosaic:
I think, this is also maintenance stuff, in the sense that no new code was actually added, so you should be able to go ahead and merge 🎉

zvirja:
Agreed, merged 😊👍

PR #903, opened 2017-10-29
Fix Ranged request fails if request has same min and max

### LINKED ISSUE #897: Ranged request fails if request has same min and max
Copied from twitter: https://twitter.com/robert_engdahl/status/922728353098067969

> Robert: Surprising that the minimum and maximum are not allowed to be equal in RangedNumberRequest @AutoFixture

> AutoFixture: Indeed. However am not sure about real case scenario. What is the value of field/property if it's expected to have the exact value only? 🤔

> Robert: The real scenario is the end-points of the range also being created by AutoFixture, and eventually then could have the same value.

--

Personally I'm finding it weird to have same Min and Max value for the Range (e.g. `[Range(5,5)]`), however technically it looks correct and we should handle that gracefully by ending up with a single value.

@moodmosaic Could you please share your opinion on that? Do you agree that we shouldn't fail if range consists of single element only?

zvirja:
Fixes #897.

Support `RangedNumberRequest` with same min and max value by resolving the constant value.

### DIFF
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel.DataAnnotations;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
@@ -6042,6 +6043,38 @@ namespace AutoFixtureUnitTest
 
             Assert.Equal(1, numberOfRequestOccurence);
 
+            // Teardown
+        }
+        
+        [Theory]
+        [InlineData(typeof(byte))] 
+        [InlineData(typeof(sbyte))] 
+        [InlineData(typeof(short))]  
+        [InlineData(typeof(ushort))]
+        [InlineData(typeof(int))]  
+        [InlineData(typeof(uint))]  
+        [InlineData(typeof(long))] 
+        [InlineData(typeof(ulong))] 
+        [InlineData(typeof(decimal))]
+        [InlineData(typeof(float))]  
+        [InlineData(typeof(double))]
+        public void Issue897_ShouldCorrectlyHandleRequestsWithSameMinimumAndMaximumValue(Type type)
+        {
+            // Fixture setup
+            var expectedValue = Convert.ChangeType(42, type, CultureInfo.InvariantCulture);
+            var fakeMember = new FakeMemberInfo(
+                new ProvidedAttribute(
+                    new RangeAttribute(type, "42", "42"),
+                    inherited: false
+                ));
+            
+            var sut = new Fixture();
+            
+            // Exercise system
+            var result = sut.Create(fakeMember, new SpecimenContext(sut));
+
+            // Verify outcome
+            Assert.Equal(expectedValue, result);
             // Teardown
         }
     }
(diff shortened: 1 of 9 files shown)


Src/AutoFixtureUnitTest/RangedNumberGeneratorTest.cs
```diff
@@ -166,6 +167,35 @@ public void CreateReturnsCorrectResultWhenRunOutOfNumbers(int minimum, int maxim
...
+        [InlineData(typeof(uint))]  
+        [InlineData(typeof(long))] 
+        [InlineData(typeof(ulong))] 
+        [InlineData(typeof(decimal))]
+        [InlineData(typeof(float))]  
+        [InlineData(typeof(double))]
+        public void ReturnsSameNumberWhenMinimumAndMaximumAreSame(Type requestType)
+        {
+            // Fixture setup
+            var range = Convert.ChangeType(42, requestType, CultureInfo.InvariantCulture);
```
moodmosaic: Perhaps, we could even parameterize the `42` magic number, but not that big of a deal.

Src/AutoFixtureUnitTest/RandomRangedNumberGeneratorTest.cs
```diff
@@ -312,6 +313,33 @@ public void CreationWithLimitsInBoundariesShouldReturnValueInRange(Type type, ob
...
+        [InlineData(typeof(uint))]  
+        [InlineData(typeof(long))] 
+        [InlineData(typeof(ulong))] 
+        [InlineData(typeof(decimal))]
+        [InlineData(typeof(float))]  
+        [InlineData(typeof(double))]
+        public void ShouldCorrectlyHandleRequestsWithSameMinimumAndMaximumValue(Type type)
+        {
+            // Fixture setup
+            var range = Convert.ChangeType(42, type, CultureInfo.InvariantCulture);
```
moodmosaic: Perhaps, we could even parameterize the `42` magic number, but not that big of a deal.

Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -6044,5 +6045,37 @@ public void ShouldNotDuplicateRequestPathTwiceInCaseOfRecursionGuardException()
...
+        [InlineData(typeof(float))]  
+        [InlineData(typeof(double))]
+        public void Issue897_ShouldCorrectlyHandleRequestsWithSameMinimumAndMaximumValue(Type type)
+        {
+            // Fixture setup
+            var expectedValue = Convert.ChangeType(42, type, CultureInfo.InvariantCulture);
+            var fakeMember = new FakeMemberInfo(
+                new ProvidedAttribute(
+                    new RangeAttribute(type, "42", "42"),
+                    inherited: false
```
moodmosaic: Perhaps, we could even parameterize the `42` magic number, but not that big of a deal.

zvirja: Yep, that was my original design. I didn't want to hardcode string and wanted to parameterize `42` in it's native type (e.g. `42L`, `42.0`, etc). However, it appeared that you cannot provide a constant for the `decimal` type 😕 Therefore, we should either extract all the values to a `TheoryData<>` member or perform a conversion from string.

I decided to proceed with the latest approach. In that light it doesn't have too much sense to hardcode that string value for each theory case. Therefore, let keep it as is.

If you see a better way - let me know and I'll follow that approach in future.

Src/AutoFixtureUnitTest/RangedNumberGeneratorTest.cs
zvirja: Cannot, because no way to specify a constant for the `decimal` type.

Src/AutoFixtureUnitTest/RandomRangedNumberGeneratorTest.cs
zvirja: Cannot, because no way to specify a constant for the `decimal` type.

PR #905, opened 2017-10-30
Return ISpecimenBuilderNode in ISpecimenBuilderTransformation

zvirja:
The current implementation of the `ISpecimenBuilderTransformation` used in `fixture.Behaviors` collection is following:
```c#
public interface ISpecimenBuilderTransformation
{
    ISpecimenBuilder Transform(ISpecimenBuilder builder);
}
```

While this interface looks cool, internally we implicitly expect that transformation returns an instance of `ISpecimenBuilderNode` rather than `ISpecimenBuilder`:
https://github.com/AutoFixture/AutoFixture/blob/6e1765747245d45c6659006f14e680568aec42e3/Src/AutoFixture/SingletonSpecimenBuilderNodeStackAdapterCollection.cs#L195-L197

and a test:

https://github.com/AutoFixture/AutoFixture/blob/6e1765747245d45c6659006f14e680568aec42e3/Src/AutoFixtureUnitTest/SingletonSpecimenBuilderNodeStackAdapterCollectionTest.cs#L318-L325

That point confused me each time I created my own behaviors because I got a feedback at run time only and needed to rework my code.

In this PR I'm changing the interface to transform `ISpecimenBuilder` to `ISpecimenBuilderNode`, so we have guarantee type safety at the compile time:

```c#
public interface ISpecimenBuilderTransformation
{
    ISpecimenBuilderNode Transform(ISpecimenBuilder builder);
}
```

This is a breaking change which will affect the existing user behaviors. Users will need to change the signature of the implementation in their code. However, that shouldn't be an issue as they returned `ISpecimenBuilderNode` instances in any case.

While it's a breaking change in API, I believe that it worth it as user experience will be much better.

We are totally fine to accept this breaking change as we are still working on the major version, so SemVer is not violated - we are still on the v3 -> v4 boundary.

@moodmosaic @adamchester What do you think about the suggested change? Let's discuss if you have objections.

### DIFF
--- a/Src/AutoFixture/Kernel/DisposableTrackingBehavior.cs
+++ b/Src/AutoFixture/Kernel/DisposableTrackingBehavior.cs
@@ -43,7 +43,7 @@ namespace AutoFixture.Kernel
         /// <see cref="Trackers"/> property.
         /// </para>
         /// </remarks>
-        public ISpecimenBuilder Transform(ISpecimenBuilder builder)
+        public ISpecimenBuilderNode Transform(ISpecimenBuilder builder)
         {
             var tracker = new DisposableTracker(builder);
             this.trackers.Add(tracker);
--- a/Src/AutoFixture/Kernel/ISpecimenBuilderTransformation.cs
+++ b/Src/AutoFixture/Kernel/ISpecimenBuilderTransformation.cs
@@ -1,7 +1,7 @@
 ﻿namespace AutoFixture.Kernel
 {
     /// <summary>
-    /// Transforms one <see cref="ISpecimenBuilder"/> instance into another.
+    /// Transforms <see cref="ISpecimenBuilder"/> instance into <see cref="ISpecimenBuilderNode"/>.
     /// </summary>
     public interface ISpecimenBuilderTransformation
     {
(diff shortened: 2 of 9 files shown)


moodmosaic:
I'm impressed, BTW, for all those things―big and small―you were able to spot in this fairly sized codebase!

:1st_place_medal:

zvirja:
> Otherwise, if we target this to v4, we could take this PR as-is...

This PR targets `v4`. Currently the `master` branch is for the already released `v4 RC1` and the upcoming `v4 RTM` - we no longer support releases of `v3`. The reason why I followed this way was the namespace change, so it would be a hell to merge from `master` to `v4`. Therefore, I kept only single branch - `master`, which corresponds to the `v4` now. So now it's technically impossible to fire a PR for the v3 😄

> I'm impressed, BTW, for all those things―big and small

Thanks 😊 It's awesome to see how great the AutoFixture becomes without all those annoying issues.

moodmosaic:
Yeah, as long as v4 isn't officially out, then, perhaps this one can be taken as-is, without `Obsolete` attributes.

PR #909, opened 2017-11-01
Obsolete non-uniform specimen.Create() method overloads

zvirja:
I've found that there are some builders with extra-overloads of the `Create()` method which are not being used. I've altered the code to deprecate them.

That allows to unify the code so that all the specimen builders looks the same unless that is required on purpose.

Notice, the overloads with the `Anonymous` word were deprecated completely to align with other builders - this API convention was removed since v3.

### DIFF
--- a/Src/AutoFixture/BooleanSwitch.cs
+++ b/Src/AutoFixture/BooleanSwitch.cs
@@ -27,6 +27,7 @@ namespace AutoFixture
         /// <see langword="true"/>, followed by <see langword="false"/> at the next invocation, and
         /// so on.
         /// </returns>
+        [Obsolete("Please use the Create(request, context) method as this overload will be removed to make API uniform.")]
         public bool Create()
         {
             lock (this.syncRoot)
@@ -69,7 +70,9 @@ namespace AutoFixture
                 return new NoSpecimen();
             }
 
+#pragma warning disable 618
             return this.Create();
+#pragma warning restore 618
         }
     }
 }
--- a/Src/AutoFixture/ByteSequenceGenerator.cs
+++ b/Src/AutoFixture/ByteSequenceGenerator.cs
@@ -23,6 +23,7 @@ namespace AutoFixture
         /// Creates an anonymous number.
         /// </summary>
         /// <returns>The next number in a consecutive sequence.</returns>
+        [Obsolete("Please use the Create(request, context) method as this overload will be removed to make API uniform.")]
         public byte Create()
         {
             lock (this.syncRoot)
@@ -58,7 +59,9 @@ namespace AutoFixture
                 return new NoSpecimen();
             }
 
+#pragma warning disable 618
             return this.Create();
+#pragma warning restore 618
         }
     }
 }
--- a/Src/AutoFixture/DecimalSequenceGenerator.cs
+++ b/Src/AutoFixture/DecimalSequenceGenerator.cs
@@ -23,6 +23,7 @@ namespace AutoFixture
         /// Creates an anonymous number.
         /// </summary>
         /// <returns>The next number in a consecutive sequence.</returns>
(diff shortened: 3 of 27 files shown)


Src/AutoFixture/BooleanSwitch.cs
```diff
@@ -27,6 +27,7 @@ public BooleanSwitch()
         /// <see langword="true"/>, followed by <see langword="false"/> at the next invocation, and
         /// so on.
         /// </returns>
+        [Obsolete("Please move over to using Create(request, context) as this method will be removed in the next release")]
```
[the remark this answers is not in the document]
zvirja: I've simply copied this message from e.g. here:
https://github.com/AutoFixture/AutoFixture/blob/3f0d36de8de3d8069153186bc531259f957ea715/Src/AutoFixture/BooleanSwitch.cs#L48-L49

But probably you are right. There we changed API globally and really wanted to enforce everybody to use a new API, while here it's more soft.

zvirja:
@moodmosaic Your concern has been addressed - now we use more polite message for the deprecated `Create()` method overload 😅

Src/AutoFixture/BooleanSwitch.cs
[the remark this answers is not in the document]
moodmosaic: What's already there isn't necessarily perfect, just saying. 😅

Src/AutoFixture/BooleanSwitch.cs
```diff
@@ -27,6 +27,7 @@ public BooleanSwitch()
         /// <see langword="true"/>, followed by <see langword="false"/> at the next invocation, and
         /// so on.
         /// </returns>
+        [Obsolete("Please use the Create(request, context) method as this overload will be removed to make API uniform.")]
```
moodmosaic: Perhaps, not a blocker, but `Create(request, context)` target only the C# developers. What about those using F# or VB.NET, etc?

Perhaps, we could reword this as *the Create method taking a request and a context as parameters*.

zvirja: And what is the issue with F# and those overload? Why it isn't suitable for F#?

zvirja: We have a lot of `ISpecimenBuilder` implementations in our project and all them contain a single `Create` method overload. Those ones in the PR due to some reasons contain extra overloads. My point is that we should unify them all. If that overload is not compatible with F# or VB, we should revisit that at the whole project level, rather than fixing it for these particular builders only.

However, I still don't understand what are the issues with compatibility..

moodmosaic: Well, `Create(request, context)` becomes `Create (request, context)` in F# (but both ways will actually compile fine) though I don't know how it looks for VB.NET. I guess, it's fine as-is.

zvirja: Well, that will indeed be a common problem for all the builders we have (and likely most of the fixture API). Therefore if you work with F#, it's better to create a set of own helpers to get benefits from F# syntax.

zvirja:
Merging as agreed above. Our API is far from ideal to support F# in a convenient way. I believe that it's not our "primary" target and we usually don't keep F# usability in mind when we design the particular features.

If we decide to change our vision later, it will require revisit of the whole API surface. But the question is whether you need AutoFixture for F# as you have e.g. FsCheck.

moodmosaic:
>But the question is whether you need AutoFixture for F# as you have e.g. FsCheck.

*Hey!*, or [Hedgehog](https://github.com/hedgehogqa/fsharp-hedgehog) 😉

zvirja:
Opps, sorry, I thought that it's named FsCheck 😳 Yep, of course the F# Hedgehog should be used!

moodmosaic:
:rofl:

PR #911, opened 2017-11-01
[xUnit/xUnit2] Loosen up InlineAutoDataAttribute constructor to take DataAttribute argument

zvirja:
Currently `InlineAutoData` attribute allows to pass a custom `autoDataAttribute` implementation as a constructor argument:
https://github.com/AutoFixture/AutoFixture/blob/7e864358be5b4822df0134dc412861777a5a8b49/Src/AutoFixture.xUnit.net2/InlineAutoDataAttribute.cs#L87-L88

The issue is that it requires instance of `AutoDataAttribute`, while internally it works with it like that is a `DataAttribute`.

In our projects we sometimes need to deeply customize the AutoData attribute, so we derive directly from the `DataAttribute`. In this case it's impossible to re-use the `InlineAutoDataAttribute` implementation as it doesn't allow to pass our own `DataAttribute` types to the constructor.

In this PR I'm changing signature to take `DataAttribute` as a parameter. That doesn't lead to the breaking changes as `AutoDataAttribute` is derived from the `DataAttribute` type. Also I believe that it shouldn't lead to the usability issues as this constructor is protected and is used only when you define your own inline data attributes.

### DIFF
--- a/Src/AutoFixture.xUnit.net.UnitTest/InlineAutoDataAttributeTest.cs
+++ b/Src/AutoFixture.xUnit.net.UnitTest/InlineAutoDataAttributeTest.cs
@@ -105,9 +105,10 @@ namespace AutoFixture.Xunit.UnitTest
             // Fixture setup
             var sut = new InlineAutoDataAttribute();
             // Exercise system
-            AutoDataAttribute result = sut.AutoDataAttribute;
+            DataAttribute result = sut.AutoDataAttribute;
             // Verify outcome
             Assert.NotNull(result);
+            Assert.IsType<AutoDataAttribute>(result);
             // Teardown
         }
 
@@ -146,7 +147,7 @@ namespace AutoFixture.Xunit.UnitTest
         private class DerivedInlineAutoDataAttribute : InlineAutoDataAttribute
         {
             public DerivedInlineAutoDataAttribute(
-                AutoDataAttribute autoDataAttribute,
+                DataAttribute autoDataAttribute,
                 params object[] values)
                 : base(autoDataAttribute, values)
             {
--- a/Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
+++ b/Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
@@ -10,11 +10,19 @@ namespace AutoFixture.Xunit
     /// </summary>    
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     [CLSCompliant(false)]
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", Justification = "This attribute is the root of a potential attribute hierarchy.")]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes",
+        Justification = "This attribute is the root of a potential attribute hierarchy.")]
     public class InlineAutoDataAttribute : CompositeDataAttribute
     {
-        private readonly AutoDataAttribute autoDataAttribute;
-        private readonly IEnumerable<object> values;
+        /// <summary>
+        /// Gets the attribute used to automatically generate the remaining theory parameters, which are not fixed.
+        /// </summary>
+        public DataAttribute AutoDataAttribute { get; }
+
+        /// <summary>
+        /// Gets the data values to pass to the theory.
+        /// </summary>
+        public IEnumerable<object> Values { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="InlineAutoDataAttribute"/> class.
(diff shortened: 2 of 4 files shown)


Src/AutoFixture.xUnit.net.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -143,10 +144,23 @@ public void DoesntActivateFixtureImmediately()
             // Teardown
         }
 
+        [Fact][Obsolete]
```
moodmosaic: Why is this test marked as obsolete?

Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
```diff
@@ -80,34 +77,33 @@ public InlineAutoDataAttribute(params object[] values)
...
         /// </code>
         /// </example>
-        protected InlineAutoDataAttribute(AutoDataAttribute autoDataAttribute, params object[] values)
+        protected InlineAutoDataAttribute(DataAttribute autoDataAttribute, params object[] values)
             : base(new DataAttribute[] { new InlineDataAttribute(values), autoDataAttribute })
         {
-            this.autoDataAttribute = autoDataAttribute;
-            this.values = values;
+#pragma warning disable 618
+            this.AutoDataAttribute = autoDataAttribute as AutoDataAttribute;
```
moodmosaic: Why not just use the `DataAttribute` that's being passed in?

Src/AutoFixture.xUnit.net/InlineAutoDataAttribute.cs
```diff
@@ -80,34 +77,33 @@ public InlineAutoDataAttribute(params object[] values)
...
         /// </remarks>
-        /// <seealso cref="InlineAutoDataAttribute(AutoDataAttribute, object[])"/>
-        public AutoDataAttribute AutoDataAttribute
-        {
-            get { return this.autoDataAttribute; }
-        }
+        /// <seealso cref="InlineAutoDataAttribute(DataAttribute, object[])"/>
+        [Obsolete("This member is obsolete as instance might be initialized with any arbitrary DataAttribute. " +
+                  "Please use this.Attributes collection instead to find the attribute.")]
+        public AutoDataAttribute AutoDataAttribute { get; }
```
moodmosaic: Perhaps it should be `DataAttribute` now?

Src/AutoFixture.xUnit.net2.UnitTest/InlineAutoDataAttributeTest.cs
```diff
@@ -168,11 +168,24 @@ public void PreDiscoveryShouldBeDisabled()
 
             // Teardown
         }
+        
+        [Fact][Obsolete]
```
moodmosaic: Why is this test marked as obsolete?

Src/AutoFixture.xUnit.net2/InlineAutoDataAttribute.cs
```diff
@@ -84,34 +82,33 @@ public InlineAutoDataAttribute(params object[] values)
...
         /// </code>
         /// </example>
-        protected InlineAutoDataAttribute(AutoDataAttribute autoDataAttribute, params object[] values)
+        protected InlineAutoDataAttribute(DataAttribute autoDataAttribute, params object[] values)
             : base(new DataAttribute[] { new InlineDataAttribute(values), autoDataAttribute })
         {
-            this.autoDataAttribute = autoDataAttribute;
-            this.values = values;
+#pragma warning disable 618
+            this.AutoDataAttribute = autoDataAttribute as AutoDataAttribute;
```
moodmosaic: Why not just use the `DataAttribute` that's being passed in?

Src/AutoFixture.xUnit.net2/InlineAutoDataAttribute.cs
```diff
@@ -84,34 +82,33 @@ public InlineAutoDataAttribute(params object[] values)
...
         /// </remarks>
-        /// <seealso cref="InlineAutoDataAttribute(AutoDataAttribute, object[])"/>
-        public AutoDataAttribute AutoDataAttribute
-        {
-            get { return this.autoDataAttribute; }
-        }
+        /// <seealso cref="InlineAutoDataAttribute(DataAttribute, object[])"/>
+        [Obsolete("This member is obsolete as instance might be initialized with any arbitrary DataAttribute. " +
+                  "Please use this.Attributes collection instead to find the attribute.")]
+        public AutoDataAttribute AutoDataAttribute { get; }
```
moodmosaic: Perhaps it should be `DataAttribute` now?

zvirja:
@moodmosaic Thanks for catching that up. I thought that it's a redundancy as base class already provides the collection of attributes. However, for the sake of consistency I reworked the approach to simply change the property type.

Please take a look 😉

Issue #257, opened 2014-03-07
Support for IReadOnlyCollection


ploeh:
.NET 4.5 introduces the `IReadOnlyCollection<T>` interface. AutoFixture is currently built on .NET 4.0, so it would be a breaking change to update it to .NET 4.5. We've already had to reject [one pull request](https://github.com/AutoFixture/AutoFixture/pull/256) because of this.

We should _consider_ building **AutoFixture 4** on .NET 4.5. If we decide to do that, we should add support for `IReadOnlyCollection<T>`. If @clausndk keeps his fork around, we may even be able to reopen his PR and pull it at that time.

abatishchev:
What is the release schedule for v4?

ploeh:
There's no schedule as such, but I'm keeping an eye on what's happening with .NET Core. When .NET Core is released and stable, I could imagine that a lot of people would like to have AutoFixture available on that platform as well. That'll require breaking changes, but also seems like an appropriate occasion for releasing a new major version.

Issue #404 tracks the effort to make AutoFixture compatible with .NET Core, but it's not something anyone is actively working on, to my knowledge.

abatishchev:
Thanks for headups!
I would personally prefer v4 to target .NET 4.6.1 instead and v5 to point .NET Core.
So projects targeting the latest full .NET would have AutoFixture with the latest full .NET features. And projects targeting Core would get it as soon as the platform matures and the porting onto it takes place.

adamchester:
Just FYI and AFAICT, we would target [`netstandard1.X`](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md) which is the set of APIs that would run on any platform that implements them.

abatishchev:
hi, any update on this please? so far we have to always add custom type relay for it. would love to see it bundled by default.

abatishchev:
Keeping in mind https://github.com/AutoFixture/AutoFixture/issues/703, I'd like to propose to ship a new minor version (v3.6) targeting .NET 4.6 and support Core in the next major version (v4.0).

PR #913, opened 2017-11-02
Add read-only collection relay by default

zvirja:
We [have a message](https://github.com/AutoFixture/AutoFixture/wiki/v4.0-Release-Notes#%EF%B8%8F-autofixture-resolves-ireadonlycollectiont-now) in the release notes that we support `IReadOnlyCollection<>` out-of-the-box. I didn't test that as was under the impression that #257 adds this support.

However, it appeared that we simply implemented a builder, without registering it by default.

Given that this is quite popular interface and to correspond to our release notes, I'm registering a relay for this interface by default.

### DIFF
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -76,6 +76,7 @@ namespace AutoFixture
                                         new ExactTypeSpecification(
                                             typeof(ISpecimenBuilder)))),
                                 new StableFiniteSequenceRelay(),
+                                new ReadOnlyCollectionRelay(),
                                 new FilteringSpecimenBuilder(
                                     new Postprocessor(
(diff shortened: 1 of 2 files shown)


abatishchev:
Hi, I just upgraded to 4.0-rc1 hoping to see it out-of-the-box but as you mentioned it's happening it isn't there. What are the plans to release rc2 hopefully containing these changes? Thanks!

zvirja:
@abatishchev RC2 or RTM will indeed contain this fix, as it has been already merged. As for the ETA, it's hard to say - but I'd expect to have in very soon, in a few weeks.

The feature is really small, therefore feel free to customize the fixture manually in the meanwhile. That should not be an usage blocker for you :wink:

Issue #919, opened 2017-11-08
Add support for string based properties decorated with the RangeAttribute


jcasale:
Consider the following:
```cs
public class MyModel
{
    [Required]
    [Range(1, 65535)]
    public string Port { get; set; }
}
var fixture = new Fixture();
var model = fixture.Create<MyModel>();
```
In 3.51.0 the first call the Create() worked, however subsequent calls threw, now in 4.0.0-rc1 (the version I am using), the first call throws `ObjectCreationException`  whereas the attribute itself offers support for the use case.

ploeh:
Originated from https://stackoverflow.com/q/47184400/126014

zvirja:
@jcasale Thanks for raising the issue.

Well, as far as I noticed, AutoFixture wasn't designed to handle gracefully the `Range` attribute applied to non-numeric members. Therefore, I'm basically not surprised as we already have [the same issue](722) for enums. They should be fixed together and I've already started to investigate that. 

I don't have too much experience with annotation attributes and at the first glance it looks a bit weird that you apply numeric range attribute to a property of `string` type. Could you please point to some documentation describing that such usage is perfectly valid (and what is the expected behavior)? Just want to ensure that we cover valid scenarios :blush:

P.S. :+1: for using AutoFixture v4 :wink:

jcasale:
To be honest, it's a convention I am used to from the ASP arena and I don't know if it's an abuse of the facility or intended. I will check out the docs and if not, the reference source for logic that indicates it's valid and expected and not simply a fragile coincidence.

jcasale:
Reference [source ](https://github.com/Microsoft/referencesource/blob/b31308b03e8bd5bf779fb80fda71f31eb959fe0b/System.ComponentModel.DataAnnotations/DataAnnotations/RangeAttribute.cs#L92) implies it will handle casting a string (which it does) and the [docs](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.rangeattribute(v=vs.110).aspx) illustrate an example where the the value being applied is coerced and in that case, boxed. I have a workaround and I admit the use case is rare, however it does appear to be valid but certainly low priority if any.

zvirja:
@jcasale Thanks for the confirmation. Well, it's indeed seems that string _could be_ a valid type. Also it looks like that `Range` attribute could support even other types, like `DateTime`. The question is whether we are going to support that in AutoFixture _out-of-the-box_ as we clearly cannot support all those types.

[...]

PR #920, opened 2017-11-09
Introduce RangedRequest to support non-number types

### LINKED ISSUE #722: Range attribute on Enum throws InvalidCastException

zvirja:
This is a precursor to fix issue #722 and potentially #919 (if we decide).

Currently the `RangeAttributeRelay` builder assumes that the `Range` attribute is applied to the members of the number types only. However, as you might see in the referenced issues, that is not always so. 

To fix this issue in a general way, I've introduced the `RangedRequest` type in addition to the `RangedNumberRequest` we have. Currently if any member is marked by the `Range` attribute, the chain is following:
- `RangeAttributeRelay` relays request to `RangedRequest`.
- If type is a number, `RangedRequestToRangedNumberRequestRelay` relays request to `RangedNumberRequest`.

Later we could add more `RangedRequest` handlers to cover enum and string requests.

This PR contains some changes in logic as now `RangeAttributeRelay` creates requests of other type. However, we are still working on major release, so such changes are fine (also no breaking changes in API).

@moodmosaic @adamchester Please review and share your feedback 😉

### DIFF
--- a/Src/AutoFixture/DefaultRelays.cs
+++ b/Src/AutoFixture/DefaultRelays.cs
@@ -28,6 +28,7 @@ namespace AutoFixture
             yield return new FieldRequestRelay();
             yield return new FiniteSequenceRelay();
             yield return new SeedIgnoringRelay();
+            yield return new NumericRangedRequestRelay();
             yield return new MethodInvoker(
                 new CompositeMethodQuery(
                     new ModestConstructorQuery(),
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -6011,11 +6011,11 @@ namespace AutoFixtureUnitTest
         public void Issue453_RangeAttributeShouldFailWithMeaningfulException()
         {
             // Fixture setup
-            var fixture = new Fixture();
+            var sut = new Fixture();
 
             // Exercise system and verify outcome
             var actualEx = Assert.ThrowsAny<ObjectCreationException>(
-                () => fixture.Create<Issue453_AnnotationWithOverflow>());
+                () => sut.Create<Issue453_AnnotationWithOverflow>());
 
             Assert.IsType<OverflowException>(actualEx.InnerException);
             Assert.Contains("To solve the issue", actualEx.InnerException.Message);
@@ -6027,6 +6027,30 @@ namespace AutoFixtureUnitTest
         {
             [Range(short.MinValue, long.MaxValue, ErrorMessage = "Id is not in range")]
             public long CustomerId { get; set; }
+        }
+
+        private class RangeAnnotationWithLargeStringBoundaries
+        {
+            [Range(typeof(long), "1", /* long.MaxValue */ "9223372036854775807")]
+            public long PropertyWithStringValueRange1 { get; set; }
+
+            [Range(typeof(long), /* long.MinValue */ "-9223372036854775808", "-1")]
+            public long PropertyWithStringValueRange2 { get; set; }
+        }
+
+        [Fact]
+        public void RangeAttributeWithLargeStringRangeShouldWork()
+        {
+            // Fixture setup
+            var sut = new Fixture();
+
+            // Exercise system
+            var result = sut.Create<RangeAnnotationWithLargeStringBoundaries>();
+            
+            // Verify outcome
+            Assert.NotEqual(0, result.PropertyWithStringValueRange1);
+            Assert.NotEqual(0, result.PropertyWithStringValueRange2);
+            // Teardown
         }
 
         [Fact]
--- a/Src/AutoFixtureUnitTest/Kernel/RangedRequestTest.cs
+++ b/Src/AutoFixtureUnitTest/Kernel/RangedRequestTest.cs
@@ -1,0 +1,306 @@
+﻿using System;
+using AutoFixture.Kernel;
+using Xunit;
+
+namespace AutoFixtureUnitTest.Kernel
+{
+    public class RangedRequestTest
+    {
+        [Theory]
+        [InlineData(null, typeof(int), 1, 2)]
+        [InlineData(typeof(int), null, 1, 2)]
+        [InlineData(typeof(int), typeof(int), null, 2)]
+        [InlineData(typeof(int), typeof(int), 1, null)]
+        public void ShouldFailWithArgumentExceptionForNullArguments(Type memberType, Type operandType, object min, object max)
+        {
+            // Exercise system and Verify outcome
+            Assert.Throws<ArgumentNullException>(() =>
+                new RangedRequest(memberType, operandType, min, max));
+
+            // Teardown
+        }
+
+        [Fact]
+        public void PropertiesShouldReturnPassedToConstructorValues()
+        {
+            // Fixture setup
+            var memberType = typeof(object);
+            var operandType = typeof(int);
+            var min = 42;
+            var max = 100;
+
+            // Exercise system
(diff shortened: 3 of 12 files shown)


Src/AutoFixture/DefaultRelays.cs
```diff
@@ -28,6 +28,7 @@ public virtual IEnumerator<ISpecimenBuilder> GetEnumerator()
             yield return new FieldRequestRelay();
             yield return new FiniteSequenceRelay();
             yield return new SeedIgnoringRelay();
+            yield return new RangedRequestToRangedNumberRequestRelay();
```
moodmosaic: The complexity will start showing up once we run into more cases, other than turning a ranged request to a ranged number request. Then, I could imagine a bloated `DefaultRelays`

```c#
...
yield return new FiniteSequenceRelay();
yield return new SeedIgnoringRelay();

yield return new RangedRequestToRangedNumberRequestRelay();
yield return new RangedRequestToXyzRequestRelay();
yield return new RangedRequestToAbcRequestRelay();

yield return new MethodInvoker(
                  new CompositeMethodQuery(
...
```

Instead, I could imagine this relay of relays being a [Composite](https://en.wikipedia.org/wiki/Composite_pattern):

```c#
yield return new RangedRequestRelay(
	new NumericRangedRequestRelay(),
	new XyzRangedRequestRelay(),
	new AbcRangedRequestRelay());
```

Perhaps, internally it'll use the [Specification](https://en.wikipedia.org/wiki/Specification_pattern) like we do in other places of this code base.

zvirja: Thanks for the review.

Well, I understand your concern and indeed after we add a set of builders in a row, they will "pollute" the chain a bit. However, I don't see too much options we have. 

Currently we have a situation when `RangedRequest` is relayed to the `RangedNumberRequest`. It might look that `RangedRequest` will be always relayed to the more specific requests only, but that is not necessarily so. For instance, I'm going to fix [the issue with enums](722) in a way, so we have the following path: `RangedRequest -> EnumRangedRequestRelay -> RangedNumberRequest`. In this path one "sibling" relays to other siblings. So all the builders should be able to intercept the request in case of recursive calls. 

Also we would like to allow users to add their own customizations to handle the `RangedRequest` of custom types, therefore `RangedRequestRelay` should perform a recursive request, instead of delegating to its children only.

Unfortunately, while I'm not sure I see how the suggested design allows to achieve all that stuff in a graceful way. Could you please describe that in more detail if you have in idea in mind?

Also we cannot simply wrap all the range related builders to the `CompositeBuilder` (to logically group them) because [we have a customization](https://github.com/AutoFixture/AutoFixture/blob/0babb907b0fd7524201e99d1d5e9189e22c24599/Src/AutoFixture/NoDataAnnotationsCustomization.cs#L30) which removes the `RangedRequestRelay` from the fixture (for the performance sake).

Would be really great if you see a better way, but currently it looks like that current way is the  most appropriate one..

zvirja:
@moodmosaic I've improved implementation a bit since the initial revision:
- Added a `MemberType` property to the `RangedRequest`. Now we have both `MemberType` and `OperandType`, so we could build the value more precisely (for instance when `object Prop { get; set; }` is decorated with `[Range(typeof(long), "1", "10")]` we know that `long` should be created).
- Renamed relay to `NumericRangedRequestRelay` - thanks for the idea 😉

As for the whole design - kept it the same. See [my reply](https://github.com/AutoFixture/AutoFixture/pull/920#discussion_r150395414) for reasons. Looking forward to your follow-up 🤝

Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -6029,6 +6029,26 @@ private class Issue453_AnnotationWithOverflow
...
+        public void RangeAttributeWithStringRangeShouldWork()
+        {
+            // Fixture setup
+            var sut = new Fixture();
+
+            // Exercise system
+            var result = sut.Create<RangeAnnotationWithStringBoundaries>();
+            
+            // Verify outcome
+            Assert.NotEqual(0, result.PropertyWithStringValueRange);
```
moodmosaic: We *might* get `0` here ;)

Src/AutoFixtureUnitTest/Kernel/RangedRequestTest.cs
```diff
@@ -0,0 +1,306 @@
...
+            var max = 100;
+
+            // Exercise system
+            var sut = new RangedRequest(memberType, operandType, min, max);
+
+            // Verify outcome
+            Assert.Equal(memberType, sut.MemberType);
+            Assert.Equal(operandType, sut.OperandType);
+            Assert.Equal(min, sut.Minimum);
+            Assert.Equal(max, sut.Maximum);
```
moodmosaic: You could replace the [Assertion Roulette](http://xunitpatterns.com/Assertion%20Roulette.html) here with plain old [Structural Equality](https://en.wikipedia.org/wiki/Relational_operator), like so:

*(Typed on a text-editor)*

```c#
var expected = new RangedRequest(typeof(object), typeof(int), 42, 100);
var actual = new RangedRequest(memberType, operandType, min, max);
Assert.True(expected.Equals(actual));
```
This is also formatted according to [this](http://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/) heuristic, so you don't need to put any comments on each test phase.

Src/AutoFixtureUnitTest/FixtureTest.cs
zvirja: Good catch 😄

Src/AutoFixtureUnitTest/Kernel/RangedRequestTest.cs
zvirja: Thanks for sharing that idea of structural equality. It's definitely a useful practice and I'm already using it in the tests around (driven by samples 😅). Unfortunately, it isn't applicable in this particular test. Here I verify that each value I passed to the constructor equals the corresponding property. The suggested approach wouldn't guarantee that.

As for the test formatting heuristic, I don't agree with that Mark's idea, especially for the OSS projects. In my opinion, in the OSS project all code should be consistent and unified. When somebody creates a PR, he looks at the code around to align the style. If one sees that part of our tests is decorated with comments, while other part is not, he will be confused and will not have a clear understanding of the rules. The arranged comments in a simple test makes much less harm than missing comments in more complex tests. Therefore, if we need to pick the single style, I'd definitely vote to arrange the comments.

Another story is the readability. Personally, I'm used to work with commented tests. When I see a test without comments (we have such ones in the project) I spend much more time to understand it, rather than when comments are present. Therefore, [that idea](http://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/) is a pure personal preference with pros and cons, rather than a strong rule (like tab vs spaces) 😉

P.S. We have a R# template for a test, therefore I don't write those lines manually - they are already present when I create a test.

PR #931, opened 2017-11-21
RangedRequest can now Handle TimeSpan, DateTime and DateTimeOffset

### LINKED ISSUE #930: Support TimeSpan type for the Range attribute
I have the following Property

        [Range(typeof(TimeSpan), "00:00:00", "10675199.02:48:05.4775807")]
        public TimeSpan Length { get; set; }

This causes Autofixture to fail because internally it uses `Convert.ToType("00:00:00", typeof(TimeSpan))` to get the actual value.

I expect this to work also with DateTime and DateTimeOffset.

wertzui:
Fixes #930

### DIFF
--- a/Src/AutoFixture/Kernel/RangedRequest.cs
+++ b/Src/AutoFixture/Kernel/RangedRequest.cs
@@ -133,7 +133,13 @@ namespace AutoFixture.Kernel
         {
             try
             {
+                if (conversionType == typeof(TimeSpan))
+                    return TimeSpan.Parse(attributeValue.ToString(), CultureInfo.CurrentCulture);
+                if (conversionType == typeof(DateTimeOffset))
+                    return DateTimeOffset.Parse(attributeValue.ToString(), CultureInfo.CurrentCulture);
+
                 return Convert.ChangeType(attributeValue, conversionType, CultureInfo.CurrentCulture);
+
             }
             catch (OverflowException ex)
             {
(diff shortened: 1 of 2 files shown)


zvirja:
Thank you very much for the contribution - happy to see it 😊 Sorry for the delay - needed some time to realize how it's better to implement this.

I don't like an idea of hard-coded handlers for the specific types. If we proceed this way, we'll end-up with  a lot of cases, while there might be still some uncovered ones. We could try to use the TypeConverter intead:
```c#
var converter = TypeDescriptor.GetConverter(targetType);
var result = converter.ConvertFrom(value);
```

It should perform the same under the hood, however will handle more wide range of values. That is exactly the way the `range.IsValid()` method uses under the hood.

Likely, we'll need to add the `System.ComponentModel.TypeConverter` as a dependency for the `AutoFixture` for .NET Standard. We should be fine with that as we are still working on v4.

Also we should update the comments for the `GetConvertedMaximum/Minimum` methods, removing a statement that we are using `Convert`. Likely, it's better to say smth like `... using the known converters`.

Would you be fine to apply those changes?

zvirja:
You mentioned that this PR fixes the #930 - is that indeed so? Does AutoFixture start to fill the `DateTime` or `TimeSpan` properties within the specified range? I haven't tested that, but I'd be really surprised if the suggested changes are enough to achieve that.. 😕

zvirja:
@wertzui Do you know if you have time in the nearby future to follow up on this?

zvirja:
Closing this PR as the suggested changes are covered by #949.

PR #933, opened 2017-11-21
Update ModestConstructorQuery.cs

Loppor:
added the extension so that only constructors without parameters of the supplied type are returned.

### DIFF
--- a/Src/AutoFixture/Kernel/ModestConstructorQuery.cs
+++ b/Src/AutoFixture/Kernel/ModestConstructorQuery.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -26,6 +26,10 @@ namespace AutoFixture.Kernel
         /// one returned.
         /// </para>
         /// <para>
+        /// Extension added so that only constructors who have no parameters of the supplied type,
+        /// are returned.
+        /// </para>
+        /// <para>
         /// In case of two constructors with an equal number of parameters, the ordering is
         /// unspecified.
         /// </para>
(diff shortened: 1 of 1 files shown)


Loppor:
is there any way to do this with some kind of source control, because i would've made the commit of the tests in one go with the class changes. This is my first open-source project I'm contributing on ;-)

zvirja:
Try referring to this set of guides: https://help.github.com/articles/proposing-changes-to-your-work-with-pull-requests/

Basically the workflow is following:
1. Create a fork.
2. Checkout the forked repository locally.
3. Create a new branch and perform the work. You are free to alter commits (like amend) as much time as you need - it's your local branch. Ensure to push after you finish (or force push if you amended commits).
4. Create a pull request.
5. If any changes are required after the code review, simply adjust your commits (or add new, depending on scenario) in the branch and push/force push.
6. After the PR is merged, delete the branch.

That's it 😉

Loppor:
Thank you for the reply :-). I'll make sure the fork is created today, and all the changes will be pushed onto that. Normally I'll have enough time to do this and probably see if I can tackle something else (got more than enough time on my hands). Oh, and I'll make sure to follow the guide and workflow you proposed ;-)

zvirja:
Another concern is that we have other `IMethodQuery` classes (like `ListFavoringConstructorQuery`) and probably to make the behavior consistent we should adjust them as well. It would be great to investigate whether we can do that in some nice and unified way (although avoiding the inheritance). Otherwise, adding condition to each query would be probably clumsy a bit..

Loppor:
I'll see what I can do. I think I'll make the changes for the proposed classes first and then see if I can push it higher up the hierarchy without compromising too much. Because, as you said, adding the condition to each query separatly would be very clumsy and not so easy to maintain.

Loppor:
I've modified the code locally on my own fork, will be pushing it towards my fork here on Github. Is there any unified way you've already had in mind to design this? At this moment the only way I'd see is with virtual method overriding but that way we would need inheritance and since you want to avoid that. Any tips would be welcome, honestly 😄

zvirja:
Closing this one in favor of #937

PR #934, opened 2017-11-21
Update ModestConstructorQueryTest.cs

Loppor:
modified test so that the var expectedConstructors is correctly filled
written extra test to check that none of the parameters are of the supplied type

### DIFF
--- a/Src/AutoFixtureUnitTest/Kernel/ModestConstructorQueryTest.cs
+++ b/Src/AutoFixtureUnitTest/Kernel/ModestConstructorQueryTest.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


zvirja:
Please consolidate all the changes withing the same PR.

I'd suggest you rework the tests and avoid conditions inside the tests. Instead, you should have 2 or more tests where each particular condition is met. For instance:
```c#
public void SkipsOverloadsWithParametersOfSelfType(....) { ... }
public void ReturnsOverloadsWithoutParametersOfSelftype(...) { ... } // a sample, create if feel it's needed
```

That allows to maintain tests more easily, rather than adding condition `.When(p => p.ParameterType != x)` to each test. Basically, there is a rule that [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of tests should be 1.

Issue #938, opened 2017-11-23
Use Fare package for regular expression generation


zvirja:
@moodmosaic I've noticed that you have extracted Xeger to a separate NuGet package: https://github.com/moodmosaic/Fare.

In AutoFixture we have a copy of that project. Does it make sense to actually depend on your package rather than have a copy-paste inside our code?

However, in order to achieve that we need you to add .NET Standard support and probably sign the assembly, as it we are signed. I saw that you are [reluctant](https://github.com/moodmosaic/Fare/issues/25) to sign it, however we could probably trigger Fare 2.0, so that change is not so breaking.. Also we could cheat a bit and sign everything targeting .NET Framework v4.5+ 😊

moodmosaic:
>Does it make sense to actually depend on your package rather than have a copy-paste inside our code?

[Fare](https://github.com/moodmosaic/Fare) should be the same to what we have inside AutoFixture, plus some minor changes here and there based on feedback I got from its users.

If you wish, you can take a dependency on it, and delete the appropriate files from AutoFixture. Once you do that, **AutoFixture will depend on Fare**, instead of having no other dependencies.

I don't think we should worry about anything, as I no longer work actively in Fare, but if that makes you feel safer, I can add you as a collaborator.

If you lean towards taking Fare as a dependency, you'd have to do what it takes (support .NET Standard, etc.) as I'm currently busy with other stuff and don't have the bandwidth to make changes there.

zvirja:
> AutoFixture will depend on Fare, instead of having no other dependencies.

Well, that is not the case - we already have dependencies for the .NET Standard framework. [See v4](https://www.nuget.org/packages/AutoFixture/4.0.0-rc1).

Today AutoFixture is distributed as a NuGet package, so it doesn't provide us with any benefits to not have external dependencies. That's why I don't think we should worry about that and it's not a goal to not depend on anything.

> If you lean towards taking Fare as a dependency, you'd have to do what it takes

Got it 😉 I'm not sure I also have much free time for this now, but will see whether I have it in future.

---

Ideally I prefer to depend on other packages rather than keep code inside out project. In this particular case I understand that _real_ benefits are not so great, except the more clear code organization. So if you don't mind, let's keep this as a potential improvement, but do not prioritize it.

moodmosaic:
>Today AutoFixture is distributed as a NuGet package, so it doesn't provide us with any benefits to not have external dependencies.

[...]

PR #939, opened 2017-11-23
Improve organization of the DataAnnotation related stuff

zvirja:
In this commit I perform a bit of cleanup to make the code structure more clear and consistent:
- all the Xeger stuff was moved to its own namespace;
- all the data annotations specific stuff has been moved under the `DataAnnotations` namespace. While some of the changes are broken, we are fine as those files were introduced in non-released v4;
- moved `NumericRangedRequestRelay` to Customization section for consistency with other builders. Now all the builders are located nearby;
- extended the `NoDataAnnotationsCustomization` to remove all the related builders.

All those changes are minor, but now it's more clear which particular stuff is specific to DataAnnotations.

### DIFF
--- a/Src/AutoFixture/Fare/Xeger.cs
+++ b/Src/AutoFixture/Fare/Xeger.cs
@@ -20,7 +20,7 @@
 using System;
 using System.Text;
 
-namespace AutoFixture.DataAnnotations
+namespace Fare
 {
     /// <summary>
     /// An object that will generate text from a regular expression. In a way, 
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -119,14 +119,17 @@ namespace AutoFixture
                                     new MethodInvoker(
                                         new ModestConstructorQuery()),
                                     new NullableEnumRequestSpecification()),
-                                new RangeAttributeRelay(),
-                                new StringLengthAttributeRelay(),
-                                new RegularExpressionAttributeRelay(),
                                 new EnumGenerator(),
                                 new LambdaExpressionGenerator(),
                                 CreateDefaultValueBuilder(CultureInfo.InvariantCulture),
                                 CreateDefaultValueBuilder(Encoding.UTF8),
-                                CreateDefaultValueBuilder(IPAddress.Loopback))),
+                                CreateDefaultValueBuilder(IPAddress.Loopback),
+                        
+                                /* Data annotations */
+                                new RangeAttributeRelay(),
+                                new NumericRangedRequestRelay(),
+                                new StringLengthAttributeRelay(),
+                                new RegularExpressionAttributeRelay())),
                         new AutoPropertiesTarget(
                             new Postprocessor(
                                 new CompositeSpecimenBuilder(
--- a/Src/AutoFixture/DataAnnotations/NumericRangedRequestRelay.cs
+++ b/Src/AutoFixture/DataAnnotations/NumericRangedRequestRelay.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using AutoFixture.Kernel;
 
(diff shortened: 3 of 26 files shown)


zvirja:
@moodmosaic The issue #938 is about Xeger only, while in this PR I perform other improvements as well. Those rest changes don't depend on our decision and still should be applied.

I'd vote to proceed with this PR. Even if later we decide to remove Xeger from our sources, I believe it will not happen in the very nearby future, as we should perform a lot of change there (port to .NET Standard, etc).

Src/AutoFixture/Xeger/Automaton.cs
```diff
@@ -34,7 +34,7 @@
 using System.Collections.Generic;
 using System.Linq;
 
-namespace AutoFixture.DataAnnotations
+namespace AutoFixture.Xeger
```
moodmosaic: There's no such thing as `AutoFixture.Xeger`. I think we should either revert this change, or change it to `namespace Fare`, until we're done with #938.

zvirja:
Fixed - moved the Xeger related stuff under the `Fare` namespace 😉

P.S. Sorry for the late follow-up - was (and still am a bit) sick.

Src/AutoFixture/Fare/Xeger.cs
```diff
@@ -18,15 +18,18 @@
...
-namespace AutoFixture.DataAnnotations
+namespace Fare
 {
     /// <summary>
     /// An object that will generate text from a regular expression. In a way, 
     /// it's the opposite of a regular expression matcher: an instance of this class
     /// will produce text that is guaranteed to match the regular expression passed in.
     /// </summary>
+    [SuppressMessage("Microsoft.Naming", "CA1724:TypeNamesShouldNotMatchNamespaces", 
+        Justification = "This namespace contains everything specific to this module implementation, so call it the same.")]
```
moodmosaic: Note sure why this suppression is needed.

Src/AutoFixture/Fixture.cs
```diff
@@ -119,14 +119,17 @@ public Fixture(ISpecimenBuilder engine, MultipleRelay multiple)
...
-                                new StringLengthAttributeRelay(),
-                                new RegularExpressionAttributeRelay(),
                                 new EnumGenerator(),
                                 new LambdaExpressionGenerator(),
                                 CreateDefaultValueBuilder(CultureInfo.InvariantCulture),
                                 CreateDefaultValueBuilder(Encoding.UTF8),
-                                CreateDefaultValueBuilder(IPAddress.Loopback))),
+                                CreateDefaultValueBuilder(IPAddress.Loopback),
+                        
+                                /* Data annotations */
```
moodmosaic: There must be a better way. :)

moodmosaic:
>P.S. Sorry for the late follow-up - was (and still am a bit) sick.

No worries! No rush at all. (Get well soon!) 👍 :1st_place_medal: 💯 🚀

Src/AutoFixture/Fare/Xeger.cs
zvirja: Nice catch 🏇 That was a left-over from the previous implementation and now it's indeed not required more. Removed it.

Src/AutoFixture/Fixture.cs
zvirja: Yep. That is a temporary approach (probably). Later when we have a few more implementations, we could see whether it makes sense to organize them to a kind of standalone builder or so. For now it isn't clear how to deal with that to not loose the flexibility of potential customization... Any way, that change is out of scope of this PR.

moodmosaic: I've opened https://github.com/AutoFixture/AutoFixture/issues/944 so we don't easily forget it.

PR #940, opened 2017-11-24
Do not skip range request from path

zvirja:
Currently we strip all the intermediate "system" requests from the request path (e.g. in case of exception):
https://github.com/AutoFixture/AutoFixture/blob/e9212dc4c527f7446418cefc0340b371bc4cea17/Src/AutoFixture/ObjectCreationExceptionWithPath.cs#L64-L71

If there are Range attributes, the processing path might be very non-obvious and it will be hard to troubleshoot why activation failed.

In this PR I preserve all the ranged related requests, so if it fails - it's clear what was the internal path.

### DIFF
--- a/Src/AutoFixture/Kernel/PreserveInRequestPathAttribute.cs
+++ b/Src/AutoFixture/Kernel/PreserveInRequestPathAttribute.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace AutoFixture.Kernel
+{
+    /// <summary>
+    /// A marker to indicate that request of this type should not be skipped in the request path.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    internal sealed class PreserveInRequestPathAttribute : Attribute
+    {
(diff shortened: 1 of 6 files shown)


Src/AutoFixture/Kernel/PreserveInRequestPathAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace AutoFixture.Kernel
+{
+    /// <summary>
+    /// A marker to indicate that request of this type should not be skipped in the request path.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    internal sealed class PreserveInRequestPathAttribute: Attribute
```
moodmosaic: Can you add a space also *before* the semicolon?

---

✘

```
internal sealed class PreserveInRequestPathAttribute: Attribute
```

✔

```
internal sealed class PreserveInRequestPathAttribute : Attribute
```

PR #949, opened 2017-12-11
Improve RangedRequest conversion and handle string range value for enum

zvirja:
Previously `RangedRequest` handled only numeric value conversion, while there are much more types supporting ranges (e.g. `TimeSpan`, `DateTime`). In this PR I improve the conversion approach to mimic the [`RangedAttribute`](https://github.com/Microsoft/referencesource/blob/b31308b03e8bd5bf779fb80fda71f31eb959fe0b/System.ComponentModel.DataAnnotations/DataAnnotations/RangeAttribute.cs#L140) logic. Additionally, I still use the `Convert.ChangeType()` fallback, as it appeared that type converters don't handle some primitive value conversions (e.g. `int -> decimal`).

Also I've improved `EnumRelay` to handle string specified ranges.

@moodmosaic Your review is welcome 🍷

### DIFF
--- a/Src/AutoFixture/DataAnnotations/RangedRequest.cs
+++ b/Src/AutoFixture/DataAnnotations/RangedRequest.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using AutoFixture.Kernel;
@@ -87,23 +88,23 @@ namespace AutoFixture.DataAnnotations
         }
 
         /// <summary>
-        /// Gets minimum value converted to the specified type using the <see cref="Convert"/> helper.
+        /// Gets minimum value converted to the specified type.
         /// </summary>
-        public object GetConvertedMinimum(Type typeToConvert)
+        public object GetConvertedMinimum(Type targetType)
         {
-            if (typeToConvert == null) throw new ArgumentNullException(nameof(typeToConvert));
+            if (targetType == null) throw new ArgumentNullException(nameof(targetType));
 
-            return GetConvertedRangeBoundary(this.Minimum, typeToConvert);
+            return GetConvertedRangeBoundary(this.Minimum, targetType);
         }
 
         /// <summary>
-        /// Gets maximum value converted to the specified type using the <see cref="Convert"/> helper.
+        /// Gets maximum value converted to the specified type.
         /// </summary>
-        public object GetConvertedMaximum(Type typeToConvert)
+        public object GetConvertedMaximum(Type targetType)
         {
-            if (typeToConvert == null) throw new ArgumentNullException(nameof(typeToConvert));
+            if (targetType == null) throw new ArgumentNullException(nameof(targetType));
 
-            return GetConvertedRangeBoundary(this.Maximum, typeToConvert);
+            return GetConvertedRangeBoundary(this.Maximum, targetType);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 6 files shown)


Src/AutoFixture/DataAnnotations/RangedRequest.cs
```diff
@@ -133,8 +135,28 @@ public override string ToString()
                 "Workaround for a bug in CA: https://connect.microsoft.com/VisualStudio/feedback/details/521030/")]
         private static object GetConvertedRangeBoundary(object attributeValue, Type conversionType)
         {
+            Debug.Assert(attributeValue != null, "Value cannot be null as it's asserted in constructor.");
```
moodmosaic: What does the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) (explicit is better than implicit) tell us here?In the rest of the codebase we use [Guard Clauses](http://wiki.c2.com/?GuardClause) in the form of `if cond then exp`... 😕

zvirja: Well, I had a special reasoning about this decision 😉

I always use Guard Clauses to verify the incoming _public_ data, which is untrusted and potentially could be anything. 

However in our case the API is private and we totally control the data we pass to this method. Logically, the value can be _never_ null, because it has been already verified. However, that might be not obvious (especially to R#) that `null` can be never passed to this method. That's why I leave the `Debug.Assert()` - they are not compiled and are used just to indicate some internal contract we have.

I prefer to use the Guard Clauses to verify the public non-trusted data, while use `Debug.Assert` to indicate some internal contract that could never fail (until you change the implementation, so it's broken). Usually I use [R# Code Annotations](https://www.jetbrains.com/help/resharper/Code_Analysis__Code_Annotations.html) to indicate such things so such assertions are not even needed, but here they are missing. Probably, later we could rely on [non-nullable references](https://blogs.msdn.microsoft.com/dotnet/2017/11/15/nullable-reference-types-in-csharp/) when they appear in C#.

Does this reasoning make sense to you, or you still persist on Guard Clause and don't want to distinguish between public and internal contact? ☺️

moodmosaic: I've never found `Debug.Assert` useful. That being said, I wouldn't use it also here. However, I think we have to draw a line; does it matter if I (dis)like it? *You* are doing all the hard work, after all...

zvirja: Well, I don't know 😅 I use `Debug` to assert a thing which could _never_ happen. On other hand, regular assertions describe a _potentially_ occurring thing, which is not expected. I prefer to distinguish between those cases. Therefore, if you don't strongly persist, I'd keep that as is.

It's not critically important to me, so I'll keep that up to you 😉

moodmosaic: Well, it's up to _you_. Personally, I never use it, and never found it useful. I just use `if-else` statements instead.

zvirja: Rewrote to a regular Null Guard, so we can proceed and merge this 😉

moodmosaic: I said it's up to _you_ though 🤣 🥇

zvirja: Yep, but you _haven't approved_ the PR, meaning that you expected some actions from my side 😉 Given that this discussion already took much time, I decided that change is not so important 😄

PR #952, opened 2017-12-12
Refactor predefined builders declaration to improve readability

zvirja:
Improved readability by extracting the verbose declaration to a helper methods. That makes it simpler to read those definitions and there are less chances to make a typo.

@moodmosaic Your review is welcome 😉

### DIFF
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -76,54 +76,50 @@ namespace AutoFixture
                                             typeof(IFixture)),
                                         new ExactTypeSpecification(
                                             typeof(ISpecimenBuilder)))),
+
                                 new StableFiniteSequenceRelay(),
-                                new FilteringSpecimenBuilder(
-                                    new Postprocessor(
-                                        new MethodInvoker(
-                                            new ModestConstructorQuery()),
-                                        new DictionaryFiller()),
-                                    new ExactTypeSpecification(typeof(Dictionary<,>))),
-                                new FilteringSpecimenBuilder(
-                                    new Postprocessor(
-                                        new MethodInvoker(
-                                            new ModestConstructorQuery()),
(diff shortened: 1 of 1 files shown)


zvirja:
@moodmosaic Fixed to make consistent 😉

Src/AutoFixture/Fixture.cs
```diff
@@ -76,54 +76,47 @@ public Fixture(ISpecimenBuilder engine, MultipleRelay multiple)
...
-                                new FilteringSpecimenBuilder(
-                                    new MethodInvoker(
-                                        new EnumerableFavoringConstructorQuery()),
-                                    new ExactTypeSpecification(typeof(SortedSet<>))),
-                                new FilteringSpecimenBuilder(
-                                    new MethodInvoker(
-                                        new EnumerableFavoringConstructorQuery()),
-                                    new ExactTypeSpecification(typeof(List<>))),
+
+                                MakeQueryBasedDictionaryBuilder(
```
moodmosaic: This is adding a level of indirection though, isn't it?

zvirja: Yep, it's adding. However that is not a big issue, at that heavily improves the readability. Instead parsing the whole graph to understand the supported types, you can simply look at the essential.

Previously we had the same for the "fixed builders":
https://github.com/AutoFixture/AutoFixture/blob/0e2075464a23f7cec5ddec92dfeaff5b5f32cf6e/Src/AutoFixture/Fixture.cs#L124-L126

I've extended that idea more and extracted repeated graph to a sub-method.

zvirja:
@moodmosaic I've reworked the approach a bit (renamed a few methods), so please take a look. Also [I've replied](https://github.com/AutoFixture/AutoFixture/pull/952#discussion_r158291041) to your concern about indirection.

PR #954, opened 2017-12-15
Migrate Idioms.FsCheck to support .NET Standard

### LINKED ISSUE #880: Support .NET Core by Idioms.FSharp
Currently there is a technical limitation in F# SDK and you cannot target multiple frameworks simultaneously (see [here](https://github.com/AutoFixture/AutoFixture/issues/404#issuecomment-334730885) for more detail). We should add .NET Core support when this issue is solved in future versions of VS.

zvirja:
Closes #880.

Visual Studio 2017 Update 5 improved .NET Core support for F#, allowing to migrate the leftover FsCheck project to F#.

Nothing changed in the code, except added dependency.

### DIFF
--- a/Src/Idioms.FsCheckUnitTest/Program.fs
+++ b/Src/Idioms.FsCheckUnitTest/Program.fs
@@ -1,0 +1,1 @@
+﻿module Program = let [<EntryPoint>] main _ = 0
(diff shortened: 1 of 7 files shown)


Src/Idioms.FsCheckUnitTest/Program.fs
```diff
@@ -0,0 +1 @@
+﻿module Program = let [<EntryPoint>] main _ = 0
```
moodmosaic: I'm curious to find out why we have to do this for a Class Library project.

zvirja: As far as I got it, test project for the .NET Core is not a library, rather it's an application. The key point is that the default template for the .NET Core test project includes this file.

You can find more detail here (that is the error I met):
https://github.com/Microsoft/visualfsharp/issues/2669
https://github.com/dotnet/netcorecli-fsc/issues/79

PR #957, opened 2017-12-27
Group all the DataAnnotation related builders together

### LINKED ISSUE #944: Graph composition inside Fixture.cs
Perhaps that would allow us to get rid of this pesky comment:

https://github.com/AutoFixture/AutoFixture/blob/d576fdc25ed88a35f874f9ef128b678c72752991/Src/AutoFixture/Fixture.cs#L128

zvirja:
Closes #944.

In this PR I've added the `DataAnnotationsSupportNode` to decorate all the builders related to the data annotations. That allowed to significantly simplify the `NoDataAnnotationsCustomization` implementation, so now it could easily remove all the builders without knowing which exact types to remove.

Additionally, I've decided to move the `NoDataAnnotationsCustomization` type under the `DataAnnotations` namespace. While that is a breaking change, we are perfectly fine as we are building the major version. That will allow to have all the data annotation related stuff in the single place, as everything else is already there.

@moodmosaic Looking forward to your Christmas review 😅

### DIFF
--- a/Src/AutoFixture/Fixture.cs
+++ b/Src/AutoFixture/Fixture.cs
@@ -120,13 +120,14 @@ namespace AutoFixture
                                     Encoding.UTF8),
                                 MakeFixedBuilder(
                                     IPAddress.Loopback),
-                        
-                                /* Data annotations */
-                                new RangeAttributeRelay(),
-                                new NumericRangedRequestRelay(),
-                                new EnumRangedRequestRelay(),
-                                new StringLengthAttributeRelay(),
-                                new RegularExpressionAttributeRelay())),
+
+                                new DataAnnotationsSupportNode(
+                                    new CompositeSpecimenBuilder(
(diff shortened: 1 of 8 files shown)


Src/AutoFixture/Fixture.cs
```diff
@@ -120,13 +120,14 @@ public Fixture(ISpecimenBuilder engine, MultipleRelay multiple)
...
                                     IPAddress.Loopback),
-                        
-                                /* Data annotations */
-                                new RangeAttributeRelay(),
-                                new NumericRangedRequestRelay(),
-                                new EnumRangedRequestRelay(),
-                                new StringLengthAttributeRelay(),
-                                new RegularExpressionAttributeRelay())),
+
+                                new DataAnnotationsSupportNode(
```
moodmosaic: This is wicked stuff 🚀 🌔 💯 🥇

Issue #724, opened 2016-12-07
AutoFixture should fail with meaningful exception when nested properties are being configured


niksanla2:
I have:
```
        public class Response<T> : IResponse<T> where T : IResponseBody
        {
            public T Content { get; set; }
        }

        public class ApplayDeliveryResponseBody : IResponseBody
        {
            public bool? IsAuthenticated { get; set; }
        }
        public interface IResponseBody
        {
            bool? IsAuthenticated { get; set; }
        }

        public interface IResponse<out T> where T : IResponseBody
        {
            T Content { get; }
        }

```
And then I try to call:
```
var _fixture = new Fixture();
_fixture.Build<Response<ApplayDeliveryResponseBody>>()
    .With(x => x.Content.IsAuthenticated, true)
    .Create();
```

I catch an exception:

> "Object does not match target type."

>    at System.Reflection.RuntimeMethodInfo.CheckConsistency(Object target)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.SetValue(Object obj, Object value, Object[] index)
   at Ploeh.AutoFixture.Kernel.BindingCommand`2.TrySetValue(Object specimen, Object value, Type targetType, Action`2 setValue)
   at Ploeh.AutoFixture.Kernel.BindingCommand`2.Execute(Object specimen, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.RecursionGuard.Create(Object request, ISpecimenContext context)
   at Ploeh.AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
[...]

PR #971, opened 2018-01-20
Make expression validation logic more robust

### LINKED ISSUE #970: Generic extension method over parameter constrained to an interface fails as nested expression
Hi there,

First let me thank you for all the amazing work you do, AutoFixture is simply great!:)

Now to the problem. I will try to describe it as thorough as possible. Please ask more questions if needed.

I will start with the fact that the issue I will describe was not present in v3.50.2 and emerged in v4. As I use some customisations more often in the setup phase, I decided to create an extension method which does them for me. The extension method looks like this:

```c#
public static class AutofacExtensions
{
    public static IPostprocessComposer<T> WithUnknownData<T>(this IPostprocessComposer<T> composer, ISpecimenContext context, int dataLength) where T : IUnknownRecord
        => composer.With(rec => rec.UnknownData, context.CreateMany<byte>(dataLength).ToHexString());
}
```

As my "records" (e.g. classes I customise using this extension methods) do not have a specific base class containing the UnknownData property, I decided to go for an interface describing "unknown records" and being defined as this:

```c#
public interface IUnknownRecord
{
    string UnknownData { get; set; }
}
```

The extension method is then called in a lambda function like this:

```c#
(builder, context) => builder.WithUnknownData(context, 12)
```

instead of this:

```c#
(builder, context) => builder.With(rec => rec.UnknownData, context.CreateMany<byte>(12).ToHexString())
```

where `rec` is the "record" class. And here comes the issue. With v4, you implemented the fail-fast checking for nested expressions. I have quickly checked the code of the `VerifyIsNonNestedWritableMemberExpression` method and it seems that to support my scenario it would need to do another call to `TryUnwrapConversion` somewhere in the `if` block on line 29 in the same manner as on line 12/13 in the `GetWritableMember`. I don't want to guess if this can or cannot have some hidden side effects, but as of now this seems as a solution to me.

This only happens if the extension method is type-constrained to an interface. If I try to do the same with the base class (and customise a different property contained in the base class), there is no conversion, thus no `UnaryExpression`.

If anything, let me know!

All the best, Jan

zvirja:
Fixes #970 

In v4 we added the fail-fast guard if somebody tries to configure the nested properties (see #724). However, it appeared that there is an advanced valid scenario which is not recognized by the guard and AutoFixture fails. In this PR I made logic more robust by handling the possible implicit argument conversion. Additionally, I refactored the helper method for better re-usability.

None of the existing functionality was affected.

### DIFF
--- a/Src/AutoFixture/Kernel/ExpressionReflector.cs
+++ b/Src/AutoFixture/Kernel/ExpressionReflector.cs
@@ -9,24 +9,40 @@ namespace AutoFixture.Kernel
     {
         internal static MemberExpression GetWritableMember(this LambdaExpression propertyPicker)
         {
-            var me = propertyPicker.Body as MemberExpression;
-            if (me == null)
-                me = propertyPicker.TryUnwrapConversion();
-            if (me == null)
-                throw new ArgumentException("The expression's Body is not a MemberExpression. Most likely this is because it does not represent access to a property or field.", nameof(propertyPicker));
+            var bodyExpr = propertyPicker.Body;
 
-            var pi = me.Member as PropertyInfo;
-            if (pi != null && pi.GetSetMethod() == null)
+            // Method from C# may lead to an implicit conversion - e.g. if the property or field type is System.Byte,
+            // but the supplied value is a System.Int32 value. Since there's an implicit conversion, the resulting
+            // expression may be (x => Convert(x.Property)) and we need to unwrap it.
+            var memberExpr = bodyExpr.UnwrapIfConversionExpression() as MemberExpression;
+            if (memberExpr == null)
+                throw new ArgumentException(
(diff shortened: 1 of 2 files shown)


hambroz:
Hello there:) One quick and probably stupid comment. I have checked the source code and the new method name is misspelled: `UnwrapIfCovnersionExpression` instead of `UnwrapIfConversionExpression`. Hope it helps!

moodmosaic:
Good catch, @hambroz.

zvirja:
@hambroz Released v4.0.1 with the fix on board. Please update and enjoy 😉

Thanks again for spotting your issue and for your collaboration!

hambroz:
Thanks guys, after the update, everything works as expected! Have a great Sunday!

PR #972, opened 2018-01-22
Use Fare library as a package dependency

zvirja:
Remove the `Fare` implementation and switch to the `Fare` NuGet package instead. That will allow users to consume Fare fixes much quicker.

@moodmosaic Please review and approve 😉

P.S. I really love that feeling when you remove HUGE chunk of code 💃😊


moodmosaic:
@zvirja, I'm happy to sign it, if this isn't a breaking change for the existing users of the `Fare` package.

zvirja:
Sorry for a bit belated reply.

Well, in fact this is a breaking change to some degree. It depends how the assembly is used. The direct assembly users will not affected. However, if there are other packages that depend on the `Fare`, they will stop to bind to the assembly as its identity is changed.

I've found that you have a couple of packages which depend on you: [`EntroBuilder`](https://www.nuget.org/packages/EntroBuilder/), [`AntaniXml`](https://www.nuget.org/packages/AntaniXml/), [`SpecFlow.DSL`](https://www.nuget.org/packages/SpecFlow.DSL/), [`ggUnit`](https://www.nuget.org/packages/ggUnit/). All them will need to update to the never version of `Fare` to use the signed one. Otherwise, people will be unable to use the updated version of the `Fare` package. The good point is that they are not too popular, so this will not affect too much people.

Basically, I see we have 3 options here:
1. Keep this as is and don't sign the assembly. I'd not follow this way, as you automatically prevent the whole world of signed packages from using your library as a dependency.
2. Release next major version with a signed assembly. That is the breaking change, however as I pointed above this library is not too popular, so small number of people will be affected. Also this is one time change only.
3. Release a signed version of package using some other package ID (e.g. `Fare.Signed`). This is a workaround, but I don't like it too much.

Personally, I'd follow the way 2. After all the major versions exist to allow introduce the breaking changes 😅

moodmosaic:
>The good point is that they are not too popular, so this will not affect too much people.

Let's pretend that you never wrote this 😅 

---

It doesn't matter if they're popular or not; we shouldn't mess with their stuff! Also, I'm aware of at least seven organizations that use Fare, because they've contacted me on email asking about LICENSE (so at some point I've added a LICENSE file to it).

moodmosaic:
Thank you for making this analysis. I'll take some time to consider each one of these options, and perhaps come up with other options as well.

zvirja:
> Let's pretend that you never wrote this 😅

Sorry, didn't want to say anything bad about the product 😊 Moreover, would like to make it even more popular by using it in the AutoFixture 😉

>  I'll take some time to consider each one of these options, and perhaps come up with other options as well.

Sure. Will be looking forward to the decision from your side.

moodmosaic:
I think the safest option would be to go with `3`. Then, we should be able to reference the signed assembly.

zvirja:
> I think the safest option would be to go with 3

If we proceed with that way, we'll end up having 2 project importing the same files. Both project files will be the same, but one of them will contain the signing key. It could make the maintenance more complicated as the project will go. Also it could lead to confusion if you decide to create one more package in future (e.g. `Fare.Decade`) as it will be unclear whether it should depend on `Fare` or `Fare.Signed` (or to have two versions of package - `Fare.Decade` and `Fare.Decade.Signed`).

Therefore, want you to review the plan for one more time - we can simply release `Fare` 2.0 with signed assemblies and it's totally fine for the major version 😉 Let me know about your final decision and I'll proceed with that one :)

moodmosaic:
Perhaps, we could try something like [this](https://stackoverflow.com/a/331555/467754) and avoid messing with Fare?

zvirja:
Unfortunately, I don't see that the referenced approach is suitable here 😞 

**"Sign the third-party assembly."** - it would require us to ship AutoFixture together with the signed `Fare` assembly. Instead, I wish to have that assembly as a NuGet dependency, so clients could update independently.

**"Load the assembly dynamically"** - It's too scary to implement that, so it works reliably in different runtimes, supporting the .Net Standard. The xUnit framework is already solving that issue (as they load library with tests dynamically) and I often see issues on their bug tracker related to that.

Therefore, it seems the signed `Fare` NuGet package is the only reliable way we have 😕

moodmosaic:
I'm still not 100% sure if I should release Fare 2.0 just for the shake of strong-named assemblies.

Having to modify Fare (and any other project) just to satisfy a *single* consumer sounds horribly backwards to me.

Would Fare 2.0 mean the end of any other Fare consumer that don't go by strong-named signing? It means they will have to stick with Fare 1.x?

zvirja:
> I'm still not 100% sure if I should release Fare 2.0 just for the shake of strong-named assemblies.

Well, it's only about SemVer and you cannot release that as other version 😉 

> Having to modify Fare (and any other project) just to satisfy a single consumer sounds horribly backwards to me.

It's not a complete truth. By performing that change you enable _all_ the signed assemblies (in my opinion most of NuGet packages are signed) to consume your library. Currently we observe that issue with AutoFixture, however it could easily happen that in future somebody else will also meet it.

Consider this change as an enabler making your library more usable and more popular in the long term, rather than something specific to AutoFixture 😉 

> Would Fare 2.0 mean the end of any other Fare consumer that don't go by strong-named signing? It means they will have to stick with Fare 1.x?

Nope. Unsigned assemblies can use the signed ones (like you use .NET Assemblies which are signed), therefore there no usability issues at all.

The only thing is that if there are NuGet packages which depend on Fare 1.x, they should start to depend on Fare 2.x. Otherwise, their clients will be unable to update Fare to 2.x (as assembly binding redirection will not work) and consume new fixes/issues. Until they update, clients can continue to use the latest released Fare 1.x.

That is not a big issue. For instance, in v4 of AutoFixture.AutoNSubstitute we started to depend on NSubstitute 2.0 or higher. Therefore, all our clients were forced to update NSubstitute to 2.0 or higher. As soon as there are no breaking changes in the library (you will not touch the code), the update process will be seamless for the end clients.

moodmosaic:
FYI, I haven't got time yet to look carefully into this. It's on my list though.

zvirja:
@moodmosaic Thanks for the update 📯

moodmosaic:
@zvirja, I'm now almost positive that I won't have the necessary bandwidth to make this happen. 😢

zvirja:
> I'm now almost positive that I won't have the necessary bandwidth to make this happen

Is there something from my side I can help with? If we release a new major version that is signed, that will not take much of your time 😅

moodmosaic:
OK, I gave this another thought:

* Fare was created (primarily) because I needed to use it *inside* AutoFixture
* Fare since then received some stars and contributions
* AutoFixture, OTOH, has way higher activity on GitHub
* Thus, AutoFixture can be trickier to maintain compared to Fare

In favor of making AutoFixture always easier to maintain, and the time we spent around this issue (months), I can go ahead and increment the *major* version in Fare.

I don't know how to sign it though (I've never signed an assembly since 2005), and I don't have enough spare time to ramp up on the *proper* way of doing this, and any gotchas around this.

Would you care to sign Fare? 😊

zvirja:
@moodmosaic Just fired a PR for you to add the strong name 😉

zvirja:
@moodmosaic I tried to migrate to the Fare package, however found that some of the AutoFixture tests started to fail. Made a couple of PRs to Fare to fix those issues.

moodmosaic:
[Fare 2.1.0](https://www.nuget.org/packages/Fare/2.1.0) is published. Thank you!

zvirja:
@moodmosaic Sorry, it was my fault to not say that I wanted to add one more PR. Could you please publish one more version after the "ASCII only" PR?

Thanks.

moodmosaic:
https://github.com/moodmosaic/Fare/pull/39 should be released as Fare 2.1.1 soon. I'll keep you posted.

moodmosaic:
@zvirja, I just published Fare 2.1.1 on NuGet Gallery. It should be indexed in a couple of minutes.

zvirja:
@moodmosaic Please take a look. I'm using Fare 2.1.1 as min required version.

Issue #742, opened 2017-03-16
Unable to freeze a mocked delegate with AutoMoq


ecampidoglio:
AutoMoq has been able to create mocks for delegate types for a while now (see f98ae99).

However, _freezing_ a mocked delegate doesn't seem to work. 😢

Consider this failing test:
 
```csharp
[Fact]
public void FixtureCanFreezeMockOfFunc()
{
    var fixture = new Fixture().Customize(new AutoMoqCustomization());
    var frozen = fixture.Freeze<Mock<Func<string>>>();

    var actual = fixture.Create<Func<string>>();
    
    Assert.Same(frozen.Object, actual); ✖️
}
```

One would expect that a request for `Func<string>` would be satisfied by the frozen `Mock<Func<string>>` but, unfortunately, that's not the case.

## Background

Right now, the `AutoMoqCustomization` configures AutoFixture to provide instances of mocks in response to requests for [abstract classes and interfaces](https://github.com/AutoFixture/AutoFixture/blob/93d339bb63ef7f00d13d9373c4ce59f9419fc76b/Src/AutoMoq/MockRelay.cs#L115-L116) through the [`MockRelay`](https://github.com/AutoFixture/AutoFixture/blob/93d339bb63ef7f00d13d9373c4ce59f9419fc76b/Src/AutoMoq/MockRelay.cs):

```csharp
// MockRelay.cs
private class IsMockableSpecification : IRequestSpecification
{
    public bool IsSatisfiedBy(object request)
    {
        var t = request as Type;

        if (t == null)
        {
            return false;
        }

        return t.IsAbstract || t.IsInterface;
    }
}
```

Given that, the first solution that came to mind was to add delegates to that expression using the internal [`IsDelegate()`](https://github.com/AutoFixture/AutoFixture/blob/ab829640ed8e02776e4f4730d0e72ab3cc382339/Src/AutoMoq/MockType.cs#L46-L54) extension method:

```csharp
return t.IsAbstract || t.IsInterface || t.IsDelegate();
```

However, that alone didn't solve the problem. The reason is that the `AutoMoqCustomization` appends the `MockRelay` to the list of _residue collectors_; as the name implies, that list contains the specimen builders to be invoked _after_ all other builders in the pipeline have failed to satisfy the request. This makes sense because we don't want AutoFixture to provide mocks for interface types that _are_ supported, like for example `IEnumerable<T>`.

[...]

PR #986, opened 2018-02-21
Fake delegates with AutoFakeItEasy

### LINKED ISSUE #984: Treat delegates as Fakes with FakeItEasy
As far as I know, AutoFixture automatically treats interfaces and abstract types as mockables. I think it is a good idea to include the delegate types in this list as well.

blairconrad:
Fixes #984.

### DIFF
--- a/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
@@ -8,6 +8,7 @@
     <AssemblyTitle>AutoFakeItEasy.FakeItEasy2.UnitTest</AssemblyTitle>
     <AssemblyName>AutoFixture.AutoFakeItEasy.FakeItEasy2UnitTest</AssemblyName>
     <RootNamespace>AutoFixture.AutoFakeItEasy.UnitTest</RootNamespace>
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj
@@ -12,6 +12,8 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is FakeItEasy dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
+++ b/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
@@ -65,7 +65,7 @@ namespace AutoFixture.AutoFakeItEasy
 
             if (this.GenerateDelegates)
             {
-                fixture.Customizations.Add(new FakeItEasyRelay(new IsDelegateSpecification()));
+                fixture.Customizations.Add(new FakeItEasyRelay(new DelegateSpecification()));
             }
 
             fixture.Customizations.Add(
@@ -88,15 +88,6 @@ namespace AutoFixture.AutoFakeItEasy
                     nameof(GenerateDelegates),
                     minimumFakeItEasyAssemblyVersion,
                     actualFakeItEasyAssemblyVersion));
-            }
-        }
-
-        private class IsDelegateSpecification : IRequestSpecification
-        {
-            public bool IsSatisfiedBy(object request)
-            {
-                var type = request as Type;
-                return type.IsDelegate();
             }
         }
     }
--- a/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
+++ b/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Globalization;
+using System.Reflection;
 using AutoFixture.Kernel;
+using FakeItEasy;
 
 namespace AutoFixture.AutoFakeItEasy
 {
@@ -8,6 +11,8 @@ namespace AutoFixture.AutoFakeItEasy
     /// </summary>
     public class AutoFakeItEasyCustomization : ICustomization
     {
+        private bool generateDelegates;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="AutoFakeItEasyCustomization"/> class.
         /// </summary>
@@ -34,6 +39,23 @@ namespace AutoFixture.AutoFakeItEasy
         public ISpecimenBuilder Relay { get; }
 
         /// <summary>
+        /// When <c>true</c>, configures the fixture to automatically generate Fakes when a delegate is requested.
+        /// </summary>
+        public bool GenerateDelegates
+        {
+            get => this.generateDelegates;
+            set
+            {
+                if (value)
+                {
+                    AssertFakeItEasyCanFakeDelegates();
+                }
+
+                this.generateDelegates = value;
+            }
+        }
+
+        /// <summary>
         /// Customizes an <see cref="IFixture"/> to enable auto-mocking with FakeItEasy.
         /// </summary>
         /// <param name="fixture">The fixture upon which to enable auto-mocking.</param>
@@ -41,11 +63,41 @@ namespace AutoFixture.AutoFakeItEasy
         {
             if (fixture == null) throw new ArgumentNullException(nameof(fixture));
 
+            if (this.GenerateDelegates)
+            {
+                fixture.Customizations.Add(new FakeItEasyRelay(new IsDelegateSpecification()));
+            }
+
             fixture.Customizations.Add(
                 new FakeItEasyBuilder(
                     new MethodInvoker(
                         new FakeItEasyMethodQuery())));
             fixture.ResidueCollectors.Add(this.Relay);
+        }
+
+        private static void AssertFakeItEasyCanFakeDelegates()
+        {
+            var minimumFakeItEasyAssemblyVersion = new Version(1, 7, 4257, 42);
+            var actualFakeItEasyAssemblyVersion = typeof(A).GetTypeInfo().Assembly.GetName().Version;
+            if (actualFakeItEasyAssemblyVersion < minimumFakeItEasyAssemblyVersion)
+            {
+                throw new ArgumentException(string.Format(
+                    CultureInfo.CurrentCulture,
+                    "Option {0} was specified, but this requires FakeItEasy version {1} or higher, and {2} was detected. " +
+                    "Either remove the option or upgrade FakeItEasy.",
+                    nameof(GenerateDelegates),
+                    minimumFakeItEasyAssemblyVersion,
+                    actualFakeItEasyAssemblyVersion));
+            }
+        }
+
+        private class IsDelegateSpecification : IRequestSpecification
+        {
+            public bool IsSatisfiedBy(object request)
+            {
+                var type = request as Type;
+                return type.IsDelegate();
+            }
         }
     }
 }
--- a/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
+++ b/Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
@@ -9,34 +9,52 @@ namespace AutoFixture.AutoFakeItEasy
     /// <summary>
     /// Enables auto-mocking with FakeItEasy.
     /// </summary>
+    /// <remarks>
+    /// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
+    /// <br />
+    /// <code>new AutoFakeItEasyCustomization { GenerateDelegates = true }</code>
+    /// </remarks>
     public class AutoFakeItEasyCustomization : ICustomization
     {
         private bool generateDelegates;
+        private ISpecimenBuilder relay;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AutoFakeItEasyCustomization"/> class.
+        /// <para>
+        /// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
+        /// <br />
+        /// <code>new AutoFakeItEasyCustomization { GenerateDelegates = true }</code>
+        /// </para>
         /// </summary>
         public AutoFakeItEasyCustomization()
+#pragma warning disable CS0618 // Type or member is obsolete
             : this(new FakeItEasyRelay())
+#pragma warning restore CS0618 // Type or member is obsolete
         {
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="AutoFakeItEasyCustomization"/> class with a
-        /// <see cref="FakeItEasyRelay"/>.
+        /// Initializes a new instance of the <see cref="AutoFakeItEasyCustomization"/> class, using the specified
+        /// <paramref name="relay"/> instead of a <see cref="FakeItEasyRelay"/>.
         /// </summary>
         /// <param name="relay">The relay.</param>
+        [Obsolete("This constructor is obsolete and will be removed in the future version of the product. " +
+                  "Please use the AutoFakeItEasyCustomization() overload (without arguments) instead and set the Relay property.")]
         public AutoFakeItEasyCustomization(ISpecimenBuilder relay)
         {
             this.Relay = relay ?? throw new ArgumentNullException(nameof(relay));
         }
 
         /// <summary>
-        /// Gets the relay that will be added to <see cref="IFixture.ResidueCollectors"/> when
+        /// Gets or sets the relay that will be added to <see cref="IFixture.ResidueCollectors"/> when
         /// <see cref="Customize"/> is invoked.
         /// </summary>
-        /// <seealso cref="AutoFakeItEasyCustomization(ISpecimenBuilder)"/>
-        public ISpecimenBuilder Relay { get; }
+        public ISpecimenBuilder Relay
+        {
+            get => this.relay;
+            set => this.relay = value ?? throw new ArgumentNullException(nameof(value));
+        }
 
         /// <summary>
         /// When <c>true</c>, configures the fixture to automatically generate Fakes when a delegate is requested.
--- a/Src/AutoFakeItEasy/FakeItEasyType.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyType.cs
@@ -30,5 +30,10 @@ namespace AutoFixture.AutoFakeItEasy
         {
             return type.GetTypeInfo().GetGenericArguments().Single();
         }
+
+        internal static bool IsDelegate(this Type type)
+        {
+            return type != null && type.GetTypeInfo().IsSubclassOf(typeof(Delegate));
+        }
     }
 }
--- a/Src/AutoFakeItEasy/FakeItEasyType.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyType.cs
@@ -30,10 +30,5 @@ namespace AutoFixture.AutoFakeItEasy
         {
             return type.GetTypeInfo().GetGenericArguments().Single();
         }
-
-        internal static bool IsDelegate(this Type type)
-        {
-            return type != null && type.GetTypeInfo().IsSubclassOf(typeof(Delegate));
-        }
     }
 }
--- a/Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
@@ -12,6 +12,8 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is FakeItEasy dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
+++ b/Src/AutoFakeItEasy/FakeItEasyMethodQuery.cs
@@ -34,7 +34,7 @@ namespace AutoFixture.AutoFakeItEasy
             }
 
             var fakeType = type.GetFakedType();
-            if (fakeType.GetTypeInfo().IsInterface)
+            if (fakeType.GetTypeInfo().IsInterface || fakeType.IsDelegate())
             {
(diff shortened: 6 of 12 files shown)


blairconrad:
I took the suggested approach, augmenting the MethodQuery and adding an additional customization.
I couldn't find precedent for some of the test implementation choices I made.
The 3rd commit, extracting a base class so we could run all tests against all fixtures, could be dropped entirely, at your option.
And I used a compile constant to avoid trying to fake delegates when we're testing with the oldest of the FakeItEasy versions.

I'm happy to change absolutely anything, and/or squash before merge, if you'd like. No comment is too small!

zvirja:
@blairconrad Thanks a lot for your contribution. Let me think a bit about this feature and get back to you with feedback when I finish.

blairconrad:
A pleasure, @zvirja. Take your time. 

While you're thinking, I didn't want to push my agenda earlier, since I'm not an AutoFixture user, but I've always thought that once someone decides to use one of the mocking plugins such as AutoFakeItEasy, they might want to use the mocker to create all objects that it can, so all the delegates and interfaces (including the collections that are built into AutoFixture's core) and so on. If you think that would be a better way to go, I could adapt this solution. Or not!

Happy pondering. 🤔

zvirja:
I've finally reviewed the PR and would like to share my thoughts. 

The introduced implementation contributes a couple of drawbacks to API:

1. Now we have `EnableAutoFakeItEasyForDelegatesCustomization` that enables redirection of the delegate requests to `Fake<TDelegate>`. However, the issue is that this customization doesn't work properly if you don't apply `AutoFakeItEasyCustomization` before. It's very unclear from the user's perspective and it's really confusing.

2. If we look e.g. at NSubstitute, we'll see that we currently have two customizations: `AutoNSubstituteCustomization` and `AutoConfiguredNSubstituteCustomization`. The difference is that the former customization simply redirects interface and abstract classes to the mocks, while the latter one additionally configures all the fields/properties/methods to return automatically configured value. It could happen that we decide to have the similar customization for the FakeItEasy in future and we'll start to observe inconveniences (in fact, we already have them if we decide to follow the same approach with Moq). The major issue is that delegate generation feature adds a dimension, so number of customizations (like `AutoFakeItEasyWithDelegatesCustomization`) should be doubled (i.e. `abstract`, `delegates + abstract`, `configured abstract`, `configured  delegates + abstract`). I don't like that at all 😢 

I see a few ways how to we could work with that keeping the consistency of our API across all the glue libraries.

## Way 1 - Simpler

Given that delegates produced by AutoFixture are not that special, we can unconditionally redirect delegate generation to the `FakeItEasy`. Meaning, that it's a part of the `AutoFakeItEasyCustomization` now.

I'm not sure whether it's a breaking change or not. In fact, it changes the behavior, as currently delegates return the automatically generated value, while faked delegates will return the default value. In theory it could break the existing tests, however it shouldn't be a big problem as the risk is small.

## Way 2 - Make customization customizable :)

This way is a bit more generic and allows to specify how exactly you would like to customize the fixture. In fact, the `AutoFakeItEasyCustomization` (and similar customizations we have in other glue libraries) is a facades around the bricks shipped with the glue library. Therefore, this class already contains multiple responsibilities. I don't see it a big problem if we add some flags, allowing to specify how exactly this customization behaves.

API might be like following:

```c#
fixture.Customize(new AutoFakeItEasyCustomization
                  {
                      GenerateDelegates = true
                  });
```

It means that we should align our other glue libraries as well and obsolete the `AutoConfigured...` customizations in favor of the flags:
```c#
fixture.Customize(new AutoMoqCustomization
                  {
                      ConfigureMembers = true
                  });
```

It's a larger change, but it's a positive one, as it provides much better extensibility model. Also now we have the single "entry point" for the glue library.

Given all the stuff around, it's a preferable way to me (I really like idea of having the single entry point for glue libraries).

------

> they might want to use the mocker to create all objects that it can, so all the delegates and interfaces (including the collections that are built into AutoFixture's core) and so on.

While this behavior looks aligned and logical, I have a bit different vision about it. For me, the top priority is the convenience during the test creation. When you have collection interfaces, in most cases you expect them to behave like collections (i.e. they could be enumerated; the added element is returned later). If we switch to the mocks, we'll need to configure all that behavior manually each time we write the test. It might be really annoying (and it will be, given my experience of test writing).

Therefore, I'd prefer to keep that as is. Kernel intercepts the basic collections only, all other interfaces redirected, so behavior is still pretty obvious.

> The 3rd commit, extracting a base class so we could run all tests against all fixtures, could be dropped entirely, at your option.

At the first glance I'd prefer to remove that commit as it makes it harder to read the tests. However, it seems that we'll not need that reorganization at all after we solve the concerns mentioned above.

----

What I've shared is only my vision. @moodmosaic @blairconrad Would be happy to know how you see the situation and which way you prefer.

P.S. @blairconrad Thanks for raising the PR and triggering this discussion - sooner or later we must have had to decide 😄

blairconrad:
Thanks for the response, @zvirja. There's a lot to take in. For now

> Now we have `EnableAutoFakeItEasyForDelegatesCustomization` …

Yeah, I didn't care for that either. I had thought that you were suggesting I emulate the implementation in #742, so I attempted to do so. I'm happy to throw all that away.

And your point about behaviour of generated collection classes is excellent.

> as currently delegates return the automatically generated value, while faked delegates will return the default value

A faked delegate would actually return a [Dummy](http://fakeiteasy.readthedocs.io/en/stable/dummies/#how-are-the-dummies-made), but your point that the behaviour would be different is still valid. _However_, I believe we could return a faked delegate that by default returns the automatically generated value. Then users would never know that anything changed, unless they were examining the type of returned delegate (which seems silly), or they attempted to configure it as a Fake, in which case I'd think they'd be pleased with the result.

Me, I think this would be a good change. I wish I'd thought of it earlier. As a bonus, it shouldn't require any of the configuration changes you discussed above, not that I'm against a  consistent API over the glue libraries; implementing such a thing sounds like a good step, just not one that's needed here.

I may run off and try to implement my "faked delegate that acts like an AutoFixture delegate" idea, just to see if it will actually work. But I'm keen to hear what you think of it.

zvirja:
For me the top priority is the API consistency across our glue libraries. Currently we suggest two types of customizations (e.g. for `Moq` and `NSubstitute`):
- Simple (without using AutoFixture for the result)
- Configured (properties/fields/method results are generated by AutoFixture).

The FakeItEasy integration provides a simple customization only - the one which doesn't generate the property/field/method result value using the AutoFixture. And we should keep that customization to behave so for the sake of consistency.

> I believe we could return a faked delegate that by default returns the automatically generated value. 

> I may run off and try to implement my "faked delegate that acts like an AutoFixture delegate" idea

That's exactly what I _don't want_ to do. We should not make "simple" customization to behave like "configured" one for the delegates. Rather, this customization is expected to return delegates, for which we don't handle the result type (so it's up to the Mock library). However, the issue is that applying this change in minor release _might_ be considered as a breaking change (because it can really break the existing tests).

Therefore, we have the following options:
1. Tune the customization to fake the delegates by default. While the change is potentially breaking, we can cross our fingers and hope that nothing is broken 😅 That's the Way 1 mentioned above.
2. Add a way to enable that behavior optionally:
    1. Use the way you follow in this PR (have one yet customization). I don't like it, because it's unclear that customization works only if you apply another customization (see more info above).
    1. Follow the Way 2 described earlier.

For me it seems that actually Way 1 or Way 2 are the only choices we have. Do you see some other paths?

@blairconrad So do I see that clear that you in fact suggest the Way 1, which is potentially breaking?

@moodmosaic Could you please also share you thoughts regarding my concerns above and your way to solve the issue? Personally, I feel that Way 2 is win-win.

blairconrad:
> Currently we suggest two types of customizations
> * Simple (without using AutoFixture for the result)
> * Configured (properties/fields/method results are generated by AutoFixture).

Ah. I didn't understand the categorization there. That's helpful.

> The FakeItEasy integration provides a simple customization only - the one which doesn't generate the property/field/method result value using the AutoFixture. And we should keep that customization to behave so for the sake of consistency.

I don't want to quibble, but the FakeItEasy customization currently, when a naked delegate (not a `Fake<Delegate>`) is requested, is bypassed, so a delegate is constructed whose return value (if any) is produced by AutoFixture. For example, I just requested a delegate that returns an int, and that delegate gave me 64 when I Invoked it.

My suggestion, returning a Faked delegate whose default behaviour would be to return the same value that would be provided by AutoFixture, had been intended to preserve the behaviour users would see today, while still giving those, such as @OnurGumus, who wanted to configure their delegate, the ability to do so. This is effectively what I think you are calling Way 1 above, "Tune the customization to fake the delegates by default.", just with the added specification that the delegates will be fakes, but fakes that act as the AutoFixture-produced delegates do today, unless configured otherwise.

I had not considered the two types of customizations that you mentioned above. Still, in this situation, my proposal would be transparent to consumers who request a naked delegate and don't try to further configure it using FakeItEasy's Fake configuration mechanisms. The only possible breakage that I can see here would be if consumers are requesting delegates from AutoFakeItEasy and then performing a `GetType` on them or similar: the types would be different than before, even though the objects would behave in the same manner when Invoked.

I hope I've clarified my idea and position. Then again, maybe all this was clear before and it's just not compatible with the goals of AutoFixture. I'm not trying to be fractious, and ultimately I'll comply with whatever you want. I just thought this was a way forward that would satisfy @OnurGumus without breaking existing usages of AutoFixture.

moodmosaic:
>@moodmosaic Could you please also share you thoughts regarding my concerns above and your way to solve the issue? Personally, I feel that Way 2 is win-win.

I'm not sure I can follow this issue (I've also gone through #984).

* Why would we want FakeItEasy, Moq, etc., to handle delegates automatically? What is the scenario? Are we, for example, trying to pass delegates in the constructor(s)? 
* I haven't followed the *AutoConfiguredXyzCustomizations* work, which I see being mentioned above.

Happy to provide feedback, but most likely I don't have much to say regarding this issue and the linked one (#984).

blairconrad:
> Why would we want FakeItEasy, Moq, etc., to handle delegates automatically? What is the scenario? Are we, for example, trying to pass delegates in the constructor(s)?

That would be my guess. The behaviour would be useful if the system under test took a delegate as a constructor parameter, and the client wished to configure the delegate's behaviour or verify the SUT's interaction with the delegate.

moodmosaic:
@zvirja, I think, if we should go forward with this, we should add a way to enable that behavior optionally. I don't think it's *common* to pass delegates as constructor arguments in languages like C# and VB.NET (and even F#).

(We may turn this into a non-optional feature, say, in AutoFixture 5.)

blairconrad:
@moodmosaic, this is anecdotal, but I often see `Func`s or `Lazy`s passed around. Imagine if, in the AutoFixture codebase, a `Predicate<object>` were used instead of a `IRequestSpecification`. (I know, that doesn't provide the same level of information to the reader, but it is done.)

I'd proposed providing fake delegates that wrapped the AutoFixture-generated ones, so they'd perform exactly as today's AutoFixture-generated delegates do _until the client configured them_. 
@zvirja preferred optional activation of the feature, as you suggest, via a property `GenerateDelegates`. I accidentally pushed some work-in-progress a few minutes ago, but my implementation of his suggestion will be coming in very soon.

blairconrad:
Pushed. I think this matches what @zvirja had asked for. I'm sure there will be plenty of comments, and I'm happy to make whatever changes are requested.

I'd like to point out one slightly dodgy area. In `AutoFakeItEasyCustomization, if `GenerateDelegates` is `true`, we check the version of the FakeItEasy library to see if delegates are supported. I generally prefer to detect the presence of features rather than rely on version numbers, but there were no API-level changes when FIE gained the ability to fake delegates.

Also, there are currently no tests to ensure that we can fake a delegate when the FIE version is at least 1.7.4257.42—the tests currently bind to 1.7.4109.1, which cannot fake delegates, or 2.0,0, 3.0.0, or 4.0.0, which all can. Options, I think:
* leave the tests and production code as they are; it'll probably be fine (and it's fine, I did manually check with 1.7.4257.42. I'd take this option)
* add yet another test project targeting 1.7.4257.42, or
* artificially raise the minimum FIE required for AutoFakeItEasy to fake a delegate. We could require, for example, 2.0.0 (or 3.0.0 or 4.0.0), and then the production code is aligned with the tests. It does deny the feature to those who keep a very old FakeItEasy version, but it also helps us move them along to more modern FIE implementations, reducing the ongoing burden of maintaining AutoFakeItEasy's integration with so many versions of FIE.

blairconrad:
Oh, another thought. I added the `GenerateDelegates` property to turn on delegate generation, but this is not so discoverable. If you wanted another option, and envision just feature toggles, a `Flags` enum as a parameter to the customization is a possibility. Then users would see the extra constructor.

Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
```diff
@@ -41,11 +63,40 @@ public void Customize(IFixture fixture)
...
+                throw new ArgumentException(String.Format(
+                    CultureInfo.CurrentCulture,
+                    "Option {0} was specified, but this requires FakeItEasy version {1} or higher, and {2} was detected. Either remove the option or upgrade FakeItEasy.",
+                    nameof(GenerateDelegates),
+                    minimumFakeItEasyAssemblyVersion,
+                    actualFakeItEasyAssemblyVersion));
+            }
+        }
+
+        private class IsDelegateSpecification : IRequestSpecification
```
zvirja: Please move this type to the `AutoFixture` assembly, `AutoFixture/Kernel` namespace. It's quite generic specification and it might sense to have it in the common library.

Also please add a few tests for it, if possible.

Src/AutoFakeItEasy/FakeItEasyType.cs
```diff
@@ -30,5 +30,10 @@ internal static Type GetFakedType(this Type type)
         {
             return type.GetTypeInfo().GetGenericArguments().Single();
         }
+
+        internal static bool IsDelegate(this Type type)
+        {
+            return type != null && type.GetTypeInfo().IsSubclassOf(typeof(Delegate));
```
zvirja: See the comment [here](https://github.com/AutoFixture/AutoFixture/blob/9121cedad8b8eddc8993552ba87a06fbc92f97be/Src/AutoMoq/MockType.cs#L52). Should we use `MulticastDelegate` instead?

Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
```diff
@@ -41,11 +63,40 @@ public void Customize(IFixture fixture)
...
+            }
+
             fixture.Customizations.Add(
                 new FakeItEasyBuilder(
                     new MethodInvoker(
                         new FakeItEasyMethodQuery())));
             fixture.ResidueCollectors.Add(this.Relay);
         }
+
+        private static void AssertFakeItEasyCanFakeDelegates()
```
zvirja: The cool fail-fast approach, thanks! I completely agree with this way, as we cannot upgrade dependency for now.

Probably, it makes sense to register an issue for v5 to upgrade the min version (e.g. to v2) and remove all the legacy stuff we have.

zvirja:
> Oh, another thought. I added the GenerateDelegates property to turn on delegate generation, but this is not so discoverable.

I'm two hands up for this approach as it's scalable. If later we decide to add more features, we could easily introduce one yet property and handle it's value. Also properties are not limited to `bool` value (as flags imposes), but could have different types.

@blairconrad Huge thanks for the work you are doing here and please take my apologies for taking so much time to review the PR.

blairconrad:
> I'm two hands up for this approach as it's scalable. If later we decide to add more features, we could easily introduce one yet property and handle its value. Also properties are not limited to bool value (as flags impose), but could have different types.

Fair enough. Will continue with the properties. Another option, just while we're making the change, would be to add a new type as constructor argument: the customization options. This type could have all the properties you describe. It's an extra level of indirection, and _slightly_ more work than just providing properties on the customization, but would better highlight the fact that there _are_ options. Not pushing for it, but if you decide you like it, I can do.

> Huge thanks for the work you are doing here and please take my apologies for taking so much time to review the PR.

I'm happy to help. Please do not apologize. This is a hobby for us all, and sometimes we just want/need to spend our time doing other things! :wink:

zvirja:
> Another option, just while we're making the change, would be to add a new type as constructor argument: the customization options.

Well, let's see. If we follow the current approach, the API usage will look like following:
```c#
var fixture = new Fixture();
fixture.Customize(new AutoFakeItEasyCustomization
{
    GenerateDelegates = true,
    Relay = new CustomRelay()
});
```

If we have yet another level of indirection:
```c#
var fixture = new Fixture();
fixture.Customize(new AutoFakeItEasyCustomization(new CustomizationOptions
{
    GenerateDelegates = true,
    Relay = new CustomRelay()
}));
```

While the second way is not too much verbose, I still prefer the first way as it's simply much easier to support it and it looks more concise.

To fix the issue that users might not know about the possibility to configure features, we need to update the xml doc for the class and for the default constructor:

```c#
/// <summary>
/// Enables auto-mocking with FakeItEasy.
/// </summary>
/// <remarks>
/// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
/// <br />
/// new AutoFakeItEasyCustomization { GenerateDelegates = true }
/// </remarks>
public class AutoFakeItEasyCustomization: ICustomization
{
    /// <summary>
    /// Initializes a new instance of the <see cref="AutoFakeItEasyCustomization"/> class.
    /// <para>
    /// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
    /// <br />
    /// new AutoFakeItEasyCustomization { GenerateDelegates = true }
    /// </para>
    /// </summary>
    public AutoFakeItEasyCustomization() { }
}
```

Now when you type, as always see a hint in XML documentation:
image
or
image

Of course, people could still overlook it, however the chances are much less 😉 Additionally, after we create a doc site, we'll mention about this way there as well.

I do understand that documentation is a workaround here, however for me it looks like a reasonable trade-off (it's actually not that easy to make a decision) 😕.

blairconrad:
Thanks for the response, @zvirja. I'd for sure add pretty documentation anyhow. I agree that it's a tough choice.

Src/AutoFakeItEasy/FakeItEasyType.cs
blairconrad: Will do.

Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
blairconrad: Will do.

Src/AutoFakeItEasy/AutoFakeItEasyCustomization.cs
blairconrad: Done. #996

blairconrad:
Pushed 2 commits to (I think) address your requested changes, @zvirja. If you're unhappy, I'll keep pushing. If you're happy, I can squash.

zvirja:
Everything looks awesome to me, thanks! 👍 I have a few minor notes , see [this](https://github.com/blairconrad/AutoFixture/pull/1) PR.

> If you're happy, I can squash.

I'd prefer to keep all 3 commits - changes are quite nicely distributed across the commits.

blairconrad:
@zvirja, I incorporated your changes. Sadly, I'm not git-savvy enough to have done it in a way that allowed you to keep credit. I'm happy to reassign some of the commits to you, if you like.

zvirja:
> @zvirja, I incorporated your changes. Sadly, I'm not git-savvy enough to have done it in a way that allowed you to keep credit

Thanks a lot for the spent time! Me neither know. No worries about the credit - I'll be present in the PR history forever, so it's enough 😅 

----

PR looks completely fine for me!

@moodmosaic Do you want to take a look? If not - I'm happy to merge it as is.

Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
```diff
@@ -8,6 +8,7 @@
     <AssemblyTitle>AutoFakeItEasy.FakeItEasy2.UnitTest</AssemblyTitle>
     <AssemblyName>AutoFixture.AutoFakeItEasy.FakeItEasy2UnitTest</AssemblyName>
     <RootNamespace>AutoFixture.AutoFakeItEasy.UnitTest</RootNamespace>
+    <DefineConstants>CAN_FAKE_DELEGATES</DefineConstants>
```
moodmosaic: What is the purpose of `CAN_FAKE_DELEGATES` constant?

I haven't followed this pull request from the beginning, but @zvirja [asked](https://github.com/AutoFixture/AutoFixture/pull/986#issuecomment-371804702) if I could take a look, so hence the question. 😅

zvirja: The minimum supported version of FakeItEasy doesn't support the delegate generation we are trying to enable. However, we added the feature as more recent versions of FakeItEasy support it. 

We have unit tests for different version of the `FakeItEasy` library. The purpose of the constant is to specify whether the delegates feature is supported for the version of FakeItEasy we depend on.

Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj
```diff
@@ -12,6 +12,8 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is FakeItEasy dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <DefineConstants>CAN_FAKE_DELEGATES</DefineConstants>
```
zvirja: BTW, just noticed - it's better to write like this:
```xml
<DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
```

We already define some global constants in the [Common.props](https://github.com/AutoFixture/AutoFixture/blob/9121cedad8b8eddc8993552ba87a06fbc92f97be/Src/Common.props#L25), so this line will rewrite them, rather than append.

Probably, it doesn't make difference for this project as we don't use other constants, but it would be safer to replace that..

Let me know if you wish me to force push and do that 😉

blairconrad: :blush: You shouldn't have to. I'll make the change.

blairconrad: Wow. That "blush" emoji looks odd.

Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
zvirja: @moodmosaic Did you have a chance to look into this? 😊 I'd like to add a few more related changes before the release, so it might take a while before the next release.. 😟

moodmosaic: @zvirja, I've just quickly gone through the changes. Feel free to move on with this 👍

zvirja:
@blairconrad Thanks a lot for your contribution and for all the changes you applied during the review 👍 It's an ideal example of PR collaboration for me 😊 

I will release this change soon in a new version of AutoFixture product 😉

blairconrad:
Thank you, @zvirja. I also had a good time. I'm glad my "but what if?"s were not _too_ annoying!

zvirja:
> I'm glad my "but what if?"s were not too annoying!

They definitely were not, rather they provoked a very good discussion! It's always cool to ensure you take into account the multiple views! Therefore, do not hesitate to ask all the questions you have in future 😉

PR #998, opened 2018-03-12
[WIP]Wrap delegate

### LINKED ISSUE #984: Treat delegates as Fakes with FakeItEasy

blairconrad:
An alternative PR that fixes #984.

@zvirja, @moodmosaic, this is the approach that I'd originally advocated in https://github.com/AutoFixture/AutoFixture/pull/986#issuecomment-368730314:

>  I believe we could return a faked delegate that by default returns the automatically generated value. Then users would never know that anything changed, unless they were examining the type of returned delegate (which seems silly), or they attempted to configure it as a Fake, in which case I'd think they'd be pleased with the result.

@zvirja has already indicated that this approach is not desirable, although I thought that the reasoning was mostly around a perceived clash with what's normally called "Configured" customizations where the faked (mocked/substituted/rhinoed) objects are configured to return values from AutoFixture. I still think that this approach doesn't conflict, since it's all about changing the abilities of specimens that would previously have been supplied by AutoFixture kernel, not by AutoFakeItEasy.

All that's a bit of an aside, as I'm mostly pushing this because for a long time I couldn't even figure out how to accomplish this effect, and now that I have, I was hoping for a critique or this approach and/or discussion of an alternative approach.

All the juicy changes are in my second commit; the first one just grabs DelegateSpecification as was created in #986. But in 291e3ee, the approach I took was to, when FakeItEasy supports faking of delegates, to introduce a `FakeItEasyRelay` into customizations that would relay delegate requests to `Fake<TDelegate>`, as well as a builder that would create a `Fake<TDelegate>` by using `DelegateGenerator` to produce a regular AutoFixture-supplied delegate specimen, and then wrapping that delegate in a Fake (so the behaviour still looks like the AutoFixture delegate until you configure it).
Explicitly referencing the `DelegateGenerator` felt a little dirty, but I couldn't otherwise figure out how to pass the request for the delegate along to the AutoFixture kernel without having be picked up by the FakeItEasyRelay (or falling prey to the recursion guards).

Would my approach be considered reasonable, or is there a more idiomatic approach that I'm missing?

### DIFF
--- a/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
@@ -8,6 +8,7 @@
     <AssemblyTitle>AutoFakeItEasy.FakeItEasy2.UnitTest</AssemblyTitle>
     <AssemblyName>AutoFixture.AutoFakeItEasy.FakeItEasy2UnitTest</AssemblyName>
     <RootNamespace>AutoFixture.AutoFakeItEasy.UnitTest</RootNamespace>
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy3UnitTest/AutoFakeItEasy.FakeItEasy3UnitTest.csproj
@@ -12,6 +12,8 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is FakeItEasy dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
+++ b/Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
@@ -12,6 +12,8 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is FakeItEasy dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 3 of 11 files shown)


blairconrad:
@zvirja, I'm perfectly content to proceed as in #986 (which I see you've merged; thanks for that!). After all, your project, your rules. :wink:. I'm (mostly) just chasing information here (even though I retain an emotional attachment to the "wrapping" approach). To that end, if you have time to continue the conversation with me, I'd appreciate it. If not, say so, and I'll stop bothering you, and with no hard feelings!

> it might happen built-in delegates might not behave like faked ones (e.g. for the ref/out parameters), 

I suppose this is possible. I don't _think_ there'd be an issue, but it'd be easy to test.

> It might be unclear why after the update their own delegate builder suddenly stopped to work. 

Yes! This is the main thing I was trying to ask about. I puzzled over the AutoFixture API and could not figure out a way to provide a customization that would cause a Fixture to allow my new builder to wrap an otherwise-configured specimen value. I'm looking for a way to say in the builder "forward the request to the downstream builders (the next links in the chain of responsibility), and then act on the result". The closest concept I found within AutoFixture was the Pipeline, but even that only modifies values produced by the builder passed to the Pipeline object. Everything I tried (e.g. recording the fixture during customization and using it, or instead using `Resolve`) resulted in passing a request back up to the full fixture, where I'd invariably get a recursion error. Explicitly targeting `DelegateGenerator` was the only thing I could think of that would work, even though I knew it was wrong. Even though I expected the approach I pushed here to be rejected, it's become a puzzle, and I desperately want to know _how_ to do it, even if I shouldn't.

zvirja:
> I'm (mostly) just chasing information here (even though I retain an emotional attachment to the "wrapping" approach).

Given that we already merged the other approach, there is a gap - faked delegates do not return auto-values. If you wish, I'd like to ask you to implement the `Configured` mode for the FakeItEasy integration - it would be really awesome to have it. I love how it's done in the NSubstitute and would recommend that as a sample:
- automatically generate value as a method result/properties/fields. Return the same result for the same args;
- auto generate the `ref` and `out` parameters;
- the same for delegates;
- fully support the open generic methods;

If you wish, please create a PR for this feature. Nobody requested it so far, but it would be cool to have it.

> Even though I expected the approach I pushed here to be rejected, it's become a puzzle, and I desperately want to know how to do it, even if I shouldn't.

Indeed, if you put your builder into the `Customizations` or `ResidueCollectors` collections, your builder will suffer from the recursion issues if you perform `context.Resolve(request)`.

Probably, as a workaround, you can the `Behaviors` collection. In fact, it acts like as a pure Chain Of Responsibility pattern. Behavior allows you to proceed with the request and later somehow interact with the result. In your case, you can allow fixture to create the delegate using the default builder, wrap result into the faked delegate and return it. That is the best I could advise with the current kernel implementation and I don't see other easy options. If you try it, I'm almost confident it will work.

However, even with that approach, I'm still stick to the already merged way - I love the consistency it provides 😌

blairconrad:
> If you wish, please create a PR for this feature. Nobody requested it so far, but it would be cool to have it.

Sure. I'll warn you, though. I'm going on vacation in about a week and may not be able to submit any code until early April. If that's not a problem, I'll proceed.

zvirja:
> If that's not a problem, I'll proceed.

Definitely not a problem. We never had this feature - so it's fine to wait more.
Have a happy vacation 💃 

Closing this PR as we are not going to merge it.

PR #999, opened 2018-03-14
Add option to generate delegates via Moq

### LINKED ISSUE #742: Unable to freeze a mocked delegate with AutoMoq

zvirja:
Fixes #742

The main goal of this PR is to add an option to allow generate delegates by Moq, rather than by AutoFixture kernel. That allows to later use Moq API to configure the result value for them. The final API looks like following:
```c#
var fixture = new Fixture().Customize(new AutoMoqCustomization
{
    GenerateDelegates = true, // Enable delegate generation via Moq
    ConfigureMembers = true // Enable configuration of Moq to return values from AutoFixture.
});
```
A few major changes were applied in this PR to achieve the desired goal.

### 1. Deprecated `AutoConfiguredMoqCustomization`
To make it consistent with #986, I've deprecated the `AutoConfiguredMoqCustomization` type. Instead, the `AutoMoqCustomization` and the `ConfigureMembers` property should be used to enable the feature:

```c#
var fixture = new Fixture().Customize(new AutoMoqCustomization
{
    ConfigureMembers = true
});
```

This change is required, as we added yet another configuration option - delegate interception. If we follow the old approach, we'll be required to have 4 different classes:
- `AutoMoqCustomization`
- `AutoMoqWithDelegatesCustomization`
- `AutoConfiguredMoqCustomization`
- `AutoConfiguredMoqWithDelegatesCustomization`.

Obviously, this approach isn't scalable and is an overkill. Therefore, the introduced approach with properties should work better.

Additional benefit is that now glue library has a single facade entry point with properties to configure each possible variants of customization.

### 2. Add option to generate interfaces via Moq

Now you can tune configuration to additionally intercept the delegate requests. It required some changes in the code to ensure that delegates work fine with `ConfigureMembers` feature, so that AutoFixture generate result values for them as well.

-----
@moodmosaic Please review the PR commit by commit, so it will be simpler to you to understand the applied changes.

### DIFF
--- a/Src/AutoMoq/AutoMoqCustomization.cs
+++ b/Src/AutoMoq/AutoMoqCustomization.cs
@@ -48,6 +48,12 @@ namespace AutoFixture.AutoMoq
         public bool ConfigureMembers { get; set; }
 
         /// <summary>
+        /// If value is <c>true</c>, delegate requests are intercepted and created by Moq.
+        /// Otherwise, if value is <c>false</c>, delegates are created by the AutoFixture kernel.
+        /// </summary>
+        public bool GenerateDelegates { get; set; }
+
+        /// <summary>
         /// Gets or sets the relay that will be added to <see cref="IFixture.ResidueCollectors"/> when
         /// <see cref="Customize"/> is invoked.
         /// </summary>
@@ -82,6 +88,11 @@ namespace AutoFixture.AutoMoq
 
             fixture.Customizations.Add(mockBuilder);
             fixture.ResidueCollectors.Add(this.Relay);
+
+            if (this.GenerateDelegates)
+            {
+                fixture.Customizations.Add(new MockRelay(new DelegateSpecification()));
+            }
         }
     }
 }
--- a/Src/AutoMoq/AutoMoqCustomization.cs
+++ b/Src/AutoMoq/AutoMoqCustomization.cs
@@ -6,13 +6,27 @@ namespace AutoFixture.AutoMoq
     /// <summary>
     /// Enables auto-mocking with Moq.
     /// </summary>
+    /// <remarks>
+    /// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
+    /// <br />
+    /// <code>new AutoMoqCustomization { ConfigureMembers = true }</code>
+    /// </remarks>
     public class AutoMoqCustomization : ICustomization
     {
+        private ISpecimenBuilder relay;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="AutoMoqCustomization"/> class.
+        /// <para>
+        /// NOTICE! You can assign the customization properties to tweak the features you would like to enable. Example:
+        /// <br />
+        /// <code>new AutoMoqCustomization { ConfigureMembers = true }</code>
+        /// </para>
         /// </summary>
         public AutoMoqCustomization()
+#pragma warning disable 618 // Type or member is obsolete
             : this(new MockRelay())
+#pragma warning restore 618 // Type or member is obsolete
         {
         }
 
@@ -21,30 +35,52 @@ namespace AutoFixture.AutoMoq
         /// <see cref="MockRelay"/>.
         /// </summary>
         /// <param name="relay">The relay.</param>
+        [Obsolete("This constructor is obsolete and will be removed in a future version of the product. " +
+                  "Please use the AutoMoqCustomization() overload (without arguments) instead and set the Relay property.")]
         public AutoMoqCustomization(ISpecimenBuilder relay)
         {
-            this.Relay = relay ?? throw new ArgumentNullException(nameof(relay));
+            this.relay = relay ?? throw new ArgumentNullException(nameof(relay));
         }
 
         /// <summary>
-        /// Gets the relay that will be added to <see cref="IFixture.ResidueCollectors"/> when
-        /// <see cref="Customize"/> is invoked.
+        /// Specifies whether members of a mock will be automatically setup to retrieve the return values from a fixture.
         /// </summary>
-        /// <seealso cref="AutoMoqCustomization(ISpecimenBuilder)"/>
-        public ISpecimenBuilder Relay { get; }
+        public bool ConfigureMembers { get; set; }
 
         /// <summary>
-        /// Customizes a <see cref="IFixture"/> to enable auto-mocking with Moq.
+        /// Gets or sets the relay that will be added to <see cref="IFixture.ResidueCollectors"/> when
+        /// <see cref="Customize"/> is invoked.
+        /// </summary>
+        public ISpecimenBuilder Relay
+        {
+            get => this.relay;
+            set => this.relay = value ?? throw new ArgumentNullException(nameof(value));
+        }
+
+        /// <summary>
+        /// Customizes an <see cref="IFixture"/> to enable auto-mocking with Moq.
         /// </summary>
         /// <param name="fixture">The fixture upon which to enable auto-mocking.</param>
         public void Customize(IFixture fixture)
         {
             if (fixture == null) throw new ArgumentNullException(nameof(fixture));
 
-            fixture.Customizations.Add(
-                new MockPostprocessor(
-                    new MethodInvoker(
-                        new MockConstructorQuery())));
+            ISpecimenBuilder mockBuilder = new MockPostprocessor(
+                                              new MethodInvoker(
+                                                 new MockConstructorQuery()));
+
+            // If members should be automatically configured, wrap the builder with members setup postprocessor.
+            if (this.ConfigureMembers)
+            {
+                mockBuilder = new Postprocessor(
+                    builder: mockBuilder,
+                    command: new CompositeSpecimenCommand(
+                                new StubPropertiesCommand(),
+                                new MockVirtualMethodsCommand(),
+                                new AutoMockPropertiesCommand()));
+            }
+
+            fixture.Customizations.Add(mockBuilder);
             fixture.ResidueCollectors.Add(this.Relay);
         }
     }
--- a/Src/AutoMoq/MockVirtualMethodsCommand.cs
+++ b/Src/AutoMoq/MockVirtualMethodsCommand.cs
@@ -29,6 +29,8 @@ namespace AutoFixture.AutoMoq
     /// </remarks>
     public class MockVirtualMethodsCommand : ISpecimenCommand
     {
+        private static readonly DelegateSpecification DelegateSpecification = new DelegateSpecification();
+
         /// <summary>
         /// Sets up a mocked object's methods so that the return values will be retrieved from a fixture,
         /// instead of being created directly by Moq.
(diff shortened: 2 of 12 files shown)


blairconrad:
@moodmosaic wrote:

> I even think it's time to reconsider whether it was initially a good idea to offer all these auto-mocking glue libraries...

Maybe I'll slow down a bit on #1000. :wink:

zvirja:
@moodmosaic Thanks for sharing your opinion! 👍 

I do understand your concerns and indeed it might look that glue libraries started to handle a lot out-of-the-box. However, I like how it looks and I like the features we currently provide. For me the existing features of our glue libraries are the best place of the product and without it AutoFixture isn't that attractive. I really love that it requires me to write a few lines of the code to enable the desired behavior.

> In my opinion, we should offer just the basic customizations, and then it's up to the users to make their own complex, and even complicated, ones.

In fact, we don't offer too much features and handle the basic scenarios only:
- create a mock for the abstract/interface type;
- setup all the members to return values from the fixture;
- use mock framework to generate delegates as well.

All the advanced scenarios are up to the users and it's their responsibility to customize the rest.

Personally, I think that it's a great thing that we offer such features as a part of the product. Not all our clients know a lot how internals of the Moq/NSubstitute/FakeItEasy/AutoFixture are organized. Therefore, if they decide to integrate parts together on their own, they might meet a tremendous number of pitfalls, despair and decide to proceed without. I do like that we spend time and polish up the integrations to use the most efficient ways, so our clients could simply plug-and-play 😄 It does a lot for the marketing and when I advice to use AutoFixture, I always demonstrate how nicely things are integrated together and how little you should write to enable that. If decide to abandon that, product will not look so attractive and less people will use it.

After all, glue libraries are the separate packages, therefore it's totally fine to provide them with the functionality to cover the most popular use cases our clients have.

> Now, while your code changes, technically, look good, there is the question about what is going to follow for the rest of the auto-mocking glue libraries.

That is not a problem at all, as most of the work is already done. Currently we have the following glue libraries for mock frameworks:
- `AutoMoq`
- `AutoNSubstitute`
- `AutoFakeItEasy`
- `AutoFoq`
- `AutoRhinoMock`

Only the first 2 libraries provide the configured mode as they are the most popular libraries according to NuGet. Therefore, in order to unify the API we need to modify only those 2 projects, so API will be consistent across all the libraries. This PR is for the Moq integration and I was going to fire one more PR for the NSubstitute. That's all about the consistency 😉 

If you meant that some glue libraries luck the configured mode (like `AutoFakeItEasy`) for me that's totally fine. That doesn't break the API consistency, simply missing features for the less popular libraries.

-------

I believe that the glue libraries and the features they provide out-of-the-box is the heart of the AutoFixture, as they make our product much more friendly and useful. I use those features a lot (in all my projects I depend on the configured mode), therefore I'm a big fun of them too. Given that I'd like to keep them and invest into them - the benefit is huge to me.

zvirja:
@moodmosaic Did you have any chance to review my reply, so we can proceed further on this somehow?
Thanks 😉

moodmosaic:
@zvirja, I've read your reply and I appreciate it 😃 I think you should go for it, like I [wrote](https://github.com/AutoFixture/AutoFixture/pull/999#pullrequestreview-104130328) before:

>I understand the goals of this PR, and overall I think you should go for it 👍 

---

Nowadays, I'd use AF combined with QuickCheck/FsCheck and/or [Hedgehog](https://github.com/hedgehogqa), but for this, I first need to get back on track with blogging, GitHub samples and/or Gist snippets 😅

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -110,11 +112,16 @@ private static void SetupVoidMethod<TMock>(Mock<TMock> mock, Expression<Action<T
...
         private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
         {
-            // If "type" is an interface, "GetMethods" does not return methods declared on other interfaces extended by "type".
-            // In these cases, we use the "GetInterfaceMethods" extension method instead.
-            var methods = type.GetTypeInfo().IsInterface
-                              ? type.GetInterfaceMethods()
-                              : type.GetMethods();
+            IEnumerable<MethodInfo> methods;
+            if (DelegateSpecification.IsSatisfiedBy(type))
+                // If "type" is delegate, return "Invoke" method only. We don't need to configure the rest methods.
```
blairconrad: "the rest methods" → "the rest of the methods"

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -156,10 +163,16 @@ private static bool CanBeConfigured(MethodInfo method)
...
                 return null;
 
-            //e.g. "x.Method(It.IsAny<string>(), out parameter)"
-            var methodCall = Expression.Call(lambdaParam, method, methodCallParams);
+            Expression methodCall;
+            if (DelegateSpecification.IsSatisfiedBy(mockedType))
+                // e.g. "x(It.IsAny<string>(), out parameter)"
+                methodCall = Expression.Invoke(lambdaParam, methodCallParams);
+            else
+                //e.g. "x.Method(It.IsAny<string>(), out parameter)"
```
blairconrad: nit: inconsistent spacing. Elsewhere is "// e.g."

Src/AutoMoq/AutoMoqCustomization.cs
```diff
@@ -21,17 +35,33 @@ public AutoMoqCustomization()
         /// <see cref="MockRelay"/>.
         /// </summary>
         /// <param name="relay">The relay.</param>
+        [Obsolete("This constructor is obsolete and will be removed in the future version of the product. " +
```
blairconrad: grammar: "the future version" hits the ear funny. I'd prefer either something like "the next major version" or "a future version". I wish I'd caught this when adding the same note to AutoFakeItEasy.

Src/AutoMoq/AutoMoqCustomization.cs
```diff
@@ -21,17 +35,33 @@ public AutoMoqCustomization()
...
-        /// <seealso cref="AutoMoqCustomization(ISpecimenBuilder)"/>
-        public ISpecimenBuilder Relay { get; }
+        public ISpecimenBuilder Relay
+        {
+            get => this.relay;
+            set => this.relay = value ?? throw new ArgumentNullException(nameof(value));
+        }
 
         /// <summary>
         /// Customizes a <see cref="IFixture"/> to enable auto-mocking with Moq.
```
blairconrad: unrelated nit "a IFixture" should probably be "an IFixture".

Src/AutoMoqUnitTest/AutoMoqUnitTest.csproj
```diff
@@ -12,6 +12,9 @@
     <!--  Suppress warning about invalid dependency version in Castle.Core.
           That is Moq dependency and we cannot fix that somehow. -->
     <NoWarn>$(NoWarn);NU1603</NoWarn>
+
+    <!-- In the latest versions of Moq (consumed by .Net Core) they fixed issue with out parameters for delegates. -->
```
blairconrad: perhaps "they fixed an issue" or even remove the "they":

blairconrad: perhaps "they fixed an issue" or even remove the "they":

"The latest versions of Moq (consumed by .Net Core) fixed an issue with out parameters for delegates."

Src/AutoMoq/AutoMoqCustomization.cs
zvirja: Fixed. I didn't realized that `future version` is not exact enough to use the `the` article 😖

Src/AutoMoq/MockVirtualMethodsCommand.cs
zvirja: A good catch 👍 Fixed this and a few more similar places around.

Src/AutoMoqUnitTest/AutoMoqUnitTest.csproj
zvirja: Rephrased 😉

zvirja:
> Nowadays, I'd use AF combined with QuickCheck/FsCheck and/or Hedgehog, but in order to demonstrate this, I first need to get back on track with blogging, GitHub samples and/or Gist snippets

As usual, you are one step forward and use the functional languages/approaches to solve the issues we deal with using the imperative languages 😄 

@blairconrad Thanks for your review on my English - I definitely have gaps that should be improved 😥 I fixed all the places you have spotted 👍

blairconrad:
> @blairconrad Thanks for your review on my English

A pleasure. (Sadly, this is perhaps a little too literal. I enjoy words, and own [this t-shirt](https://goo.gl/images/uEmSy6).)

> I definitely have gaps that should be improved

As do we all. It's a process.

PR #1001, opened 2018-03-15
Fix spelling and grammar and tweak wording

blairconrad:
I noticed the "instanes" on line 46, then the other misspellings. Then "like that is" hit my ear funny, so I took the controversial step of replacing with "as was". Then I though someone might think "demanded" was harsh, so I took the super controversial step of softening it.

As usual, happy to make any amendments.


zvirja:
A good catch, thanks a lot!
I really miss the ReSpeller plugin for the Rider :( Some of those typos might be easily captured if it's present.

blairconrad:
Oh, I haven't tried Rider. Maybe I should. VS + ReSharper bring my laptop to its knees…

Thanks for the merge!

PR #1003, opened 2018-03-19
Add .NET Standard 2.0 direct support

### LINKED ISSUE #993: System.Net.Mail support in netstandard2.0 and netcoreapp2.0
Hi

It looks like System.Net.Mail is back in `netstandard2.0` and `netcoreapp2.0`. It would be great if the `MailAddress` generator could be included when targeting either platform. At the moment it looks like the `SYSTEM_NET_MAIL` conditional is [excluding it on both platforms](https://github.com/AutoFixture/AutoFixture/blob/dc7bd5ff68a146fb6027e09d6be82050e84ca620/Src/Common.props#L26) as it wasn't available prior to 2.0.

`System.Net.Mail` inclusion in 2.0 is documented here: https://github.com/dotnet/standard/blob/master/docs/versions/netstandard2.0.md

zvirja:
Closes #993 

Add .NET Standard 2.0 as one of the target frameworks to enable mail generation there, as API is available. Still keep the `netstandard1.5` support to be compatible with lower API versions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 indent_size = 4
 indent_style = space
 
-[*.{sln,csproj,fsproj,config,xml}]
+[*.{sln,csproj,fsproj,config,xml,props}]
 indent_size = 2
 indent_style = space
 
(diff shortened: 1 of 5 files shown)


zvirja:

PR #1004, opened 2018-03-21
Add support of delegate hook for NSubstitute

zvirja:
Perform the similar refactoring to AutoMoq and enable delegate generation support for the NSubstitute. I've obsoleted the `AutoConfiguredNSubsituteCustomization` in favor of `AutoNSubstituteCustomization`. Also enabled the delegate generation via NSubsitute (`GenerateDelegates` option).

@moodmosaic Please take a look 😉 I'll publish a new version of AutoFixture after this PR is merged.

### DIFF
--- a/Src/AutoNSubstitute/AutoConfiguredNSubstituteCustomization.cs
+++ b/Src/AutoNSubstitute/AutoConfiguredNSubstituteCustomization.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using AutoFixture.Kernel;
-using NSubstitute.Core;
 
 namespace AutoFixture.AutoNSubstitute
 {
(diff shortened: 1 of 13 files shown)


zvirja:
@moodmosaic Thanks for the review.

> I'd put all the auto-mocking glue libraries somewhere separately, outside of AutoFixture core.

But aren't they already put outside of the AutoFixture core? For me, AutoFixture is a product family. We have `AutoFixture` package - it's a core. Core is neat, clear and knows nothing about the rest. Everything else are the glue libraries we offer to provide integrations or extra features (like `Idioms`). The fact that all project are located under the same solution is more about the convenience rather than a requirement. It help to ensure we don't break them via changes and guarantee the API consistency. We could easily create a standalone repository for each glue library and maintain them there if we wish, but I believe that it will make the maintenance harder.

Therefore, the fact that glue libraries have a bit of logic should not make you sad - they are already decoupled from the kernel and are not a part of it.

moodmosaic:
You are right 👍 Perhaps taking those out of AutoFixture repo will make their maintenance more difficult. For some reason, when I think about AutoFixture my mind also *includes* the auto-mocking glue libraries 😅

PR #1006, opened 2018-03-27
Skip inherited properties in MockVirtualMethodsCommand

karoldeland:
Not the same behavior with ConfigureMembers = True on inherited interface properties. Base props were returning fixture instead of being setup correctly by Moq.

### DIFF
--- a/Src/AutoMoq/Extensions/TypeExtensions.cs
+++ b/Src/AutoMoq/Extensions/TypeExtensions.cs
@@ -24,11 +24,11 @@ namespace AutoFixture.AutoMoq.Extensions
         /// </summary>
         /// <param name="type">An interface type.</param>
         /// <returns>A collection of all properties declared by the interface <paramref name="type"/> or any of its base interfaces.</returns>
-        internal static IEnumerable<PropertyInfo> GetInterfaceProperties(this Type type)
+        internal static IEnumerable<PropertyInfo> GetAllProperties(this Type type)
         {
             return type.GetProperties().Concat(
                 type.GetInterfaces()
                     .SelectMany(@interface => @interface.GetProperties()));
-        }
+        }        
     }
 }
--- a/Src/AutoMoq/Extensions/TypeExtensions.cs
+++ b/Src/AutoMoq/Extensions/TypeExtensions.cs
@@ -8,27 +8,31 @@ namespace AutoFixture.AutoMoq.Extensions
     internal static class TypeExtensions
     {
         /// <summary>
-        /// Gets a collection of all methods declared by the interface <paramref name="type"/> or any of its base interfaces.
+        /// Gets a collection of all methods declared by the type or any of its base interfaces.
         /// </summary>
-        /// <param name="type">An interface type.</param>
-        /// <returns>A collection of all methods declared by the interface <paramref name="type"/> or any of its base interfaces.</returns>
-        internal static IEnumerable<MethodInfo> GetInterfaceMethods(this Type type)
+        internal static IEnumerable<MethodInfo> GetAllMethods(this Type type)
         {
-            return type.GetMethods().Concat(
-                type.GetInterfaces()
-                    .SelectMany(@interface => @interface.GetMethods()));
+            IEnumerable<MethodInfo> result = type.GetMethods();
+            
+            // If "type" is an interface, "GetMethods" does not return methods declared on other interfaces extended by "type".
+            if (type.GetTypeInfo().IsInterface)
+                result = result.Concat(type.GetInterfaces().SelectMany(x => x.GetMethods()));
+
+            return result;
         }
 
         /// <summary>
-        /// Gets a collection of all properties declared by the interface <paramref name="type"/> or any of its base interfaces.
+        /// Gets a collection of all properties declared by the type or any of its base interfaces.
         /// </summary>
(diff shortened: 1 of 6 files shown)


Src/AutoMoq/Extensions/TypeExtensions.cs
```diff
@@ -18,5 +18,17 @@ internal static IEnumerable<MethodInfo> GetInterfaceMethods(this Type type)
...
+        /// <summary>
+        /// Gets a collection of all properties declared by the interface <paramref name="type"/> or any of its base interfaces.
+        /// </summary>
+        /// <param name="type">An interface type.</param>
+        /// <returns>A collection of all properties declared by the interface <paramref name="type"/> or any of its base interfaces.</returns>
+        internal static IEnumerable<PropertyInfo> GetInterfaceProperties(this Type type)
+        {
+            return type.GetProperties().Concat(
+                type.GetInterfaces()
+                    .SelectMany(@interface => @interface.GetProperties()));
```
zvirja: Is it OK that we inspect the first level of the interfaces only? Should we go recursively here (for cases when you have interface inherited from other interface)?

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -125,15 +126,29 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
...
-                                    .Where(p => p.GetGetMethod() != null &&
-                                                p.GetSetMethod() != null)
-                                    .Select(p => p.GetGetMethod());
-            methods = methods.Except(getMethods);
+            methods = RemoveGetterSetterMethods(type, methods);
 
             return methods.Where(CanBeConfigured);
         }
 
+        private static IEnumerable<MethodInfo> RemoveGetterSetterMethods(Type type, IEnumerable<MethodInfo> methods)
```
zvirja: Probably, better name `SkipWritablePropertyGetters`:
- in fact, you not remove something, but filter that
- you don't filter out setters

Src/AutoMoq/MockVirtualMethodsCommand.cs
```diff
@@ -125,15 +126,29 @@ private static IEnumerable<MethodInfo> GetConfigurableMethods(Type type)
...
+        {
+            var getterSetterMethods = GetProperties(type)
+                .Where(p => p.GetGetMethod() != null &&
+                            p.GetSetMethod() != null)
+                .Select(p => p.GetGetMethod());
+            methods = methods.Except(getterSetterMethods);
+            return methods;
+        }
+
+        private static IEnumerable<PropertyInfo> GetProperties(Type type)
```
zvirja: I'd suggest to encapsulate all this logic inside the extension method you introduced above. We can name it `GetAllProperties()` and perform the necessary checks there. Otherwise, logic becomes a bit split across the different files 😕

Src/AutoMoq/MockVirtualMethodsCommand.cs
karoldeland: OK

Src/AutoMoq/Extensions/TypeExtensions.cs
karoldeland: GetInterfaces returns all the implemented interfaces "recursively". No need to do more here I think.

karoldeland:
You said that some tests in FixtureIntegrationWithAutoMoqCustomizationTest are failing. Could you point me which one?

Src/AutoMoq/MockVirtualMethodsCommand.cs
karoldeland: done

zvirja:
> You said that some tests in FixtureIntegrationWithAutoMoqCustomizationTest are failing. Could you point me which one?

Sorry for the confusing sentence 😕 I asked you to add a test that fails, so your changes fix it. This test will be an integration test, so it will help me to understand what exactly you are trying to solve via this PR. Hope that makes sense now.

Src/AutoMoq/Extensions/TypeExtensions.cs
zvirja: @karoldeland That's cool if that's so. Could you please add a test to verify that assumption? Just want to ensure that we are not relying on something specific to .NET Framework...

karoldeland:
Just add a test. Did I well understand your point?

zvirja:
@karoldeland Thanks a lot for the follow-up! Yep, you understood my question correctly.

I've added a few minor improvements to the code (mostly cosmetic) and pushed it to your branch. Hope you are fine with that 😉 

Thanks for the good catch in this PR and fixed bug. I'll release a new release in a few days with your fix on board. 💃 

P.S. I've noticed that you created PR on top of the `master` branch in your repository. I'd suggest to create a separate branch for each PR you make. That will allow you to:
- cleanup repository after the PR merge more easily (as not you'll need to run `git fetch --all;git reset --hard upstream/master`);
- have more than one ongoing PR, as each PR belongs to its own branch.

karoldeland:
Thanks for the advice. I'll try that next time. I'll stay tuned for the release.

zvirja:
@karoldeland Sorry for the belated release - published as 4.2.1. Should be available soon.

PR #1015, opened 2018-04-06
Fixed Fare to follow namespace standards

holm0563:
This global namespacing issue is preventing us from updating to the latest version of AutoFixture.  Other folders such as Dsl and DataNAnnotations follow the common namespacing pattern. For some reason Fare's does not. For any consuming projects with a model named "Fare" for example this can be problematic.

### DIFF
--- a/Src/AutoFixture/Fare/Automaton.cs
+++ b/Src/AutoFixture/Fare/Automaton.cs
@@ -34,7 +34,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 
-namespace Fare
+namespace AutoFixture.Fare
 {
     /// <summary>
     /// Finite-state automaton with regular expression operations.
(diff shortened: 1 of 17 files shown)


zvirja:
Hello. Thanks a lot for the contribution.

Before we proceed further on this, I would like to mention that we are going to remove all the code from the `Fare` namespace in the upcoming releases and switch to usage of the [`Fare` nuget package](https://www.nuget.org/packages/Fare/) instead. See more detail in #972. Therefore, probably it doesn't worth it to invest into this area much.

As for the "broken namespace", we made that inconsistency on purpose - all the classes inside that namespace are not a part of the AutoFixture product. Rather, that is a third-party library, embedded inside. See the related discussion here: https://github.com/AutoFixture/AutoFixture/pull/939#discussion_r153081669.

Out of curiosity, could you please clarify which exact issues you have with the `Fare` namespace? All the classes inside that namespace are internal and clients are not supposed to use it 😕

Thanks again for spending your time and your willing to improve the AutoFixture product! 🥇

zvirja:
Closing this one as #972 has been merged.

zvirja:
@holm0563 Just released v4.3.0 with the removed `Fare` namespace from the AutoFixture. Happy migration 😅

PR #1016, opened 2018-04-10
Support TimeSpan type for the Range attribute (#930)

### LINKED ISSUE #930: Support TimeSpan type for the Range attribute

damian-krychowski:
Hi, guys

I have added support for TimeSpan type in the Range attribute:

1. When the range is specified with strings, then they are simply parsed to TimeSpan (like "00:00:00" is TimeSpan.Zero and "12:00:00" is TimeSpan.FromHours(12))

2. When the range is specified with numbers, they are treated as milliseconds - I am not sure that it is desired behavior but it seemed very natural to me.

All remarks are highly appreciated! 

Closes #930

### DIFF
--- a/Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
+++ b/Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
@@ -1,0 +1,88 @@
+﻿using System;
+using System.Globalization;
+using AutoFixture.Kernel;
+
+namespace AutoFixture.DataAnnotations
+{
+    /// <summary>
+    /// Handles <see cref="RangedRequest"/> of TimeSpan type by forwarding requests
+    /// to the <see cref="RangedNumberRequest"/> with min and max TimeSpan as milliseconds values.
+    /// </summary>
+    public class TimeSpanRangedRequestRelay : ISpecimenBuilder
+    {
+        /// <inheritdoc />   
+        public object Create(object request, ISpecimenContext context)
+        {
+            if (context == null) throw new ArgumentNullException(nameof(context));
+
+            var rangedRequest = request as RangedRequest;
+
+            if (rangedRequest == null)
+                return new NoSpecimen();
+
+            if (rangedRequest.MemberType != typeof(TimeSpan))
+                return new NoSpecimen();
+
+            return CreateRangedTimeSpanSpecimen(rangedRequest, context);
+        }
+
+        private static object CreateRangedTimeSpanSpecimen(RangedRequest rangedRequest, ISpecimenContext context)
+        {
+            TimeSpanRange range;
+
+            if (rangedRequest.Minimum is string)
+            {
+                range = GetTimeSpanRangeFromStringValues(rangedRequest);
+            }
+            else if (rangedRequest.Minimum.GetType().IsNumberType())
+            {
+                range = GetTimeSpanRangeFromNumberValues(rangedRequest);
+            }
+            else
+            {
+                return new NoSpecimen();
+            }
+
+            return RandomizeTimeSpanInRange(range, context);
+        }
+
+        private static TimeSpanRange GetTimeSpanRangeFromStringValues(RangedRequest rangedRequest)
+        {
+            return new TimeSpanRange
+            {
+                MillisecondsMin = StringToMilliseconds((string)rangedRequest.Minimum),
+                MillisecondsMax = StringToMilliseconds((string)rangedRequest.Maximum)
+            };
+        }
+
+        private static double StringToMilliseconds(string serializedTimeSpan)
+        {
+            return TimeSpan.Parse(serializedTimeSpan, CultureInfo.CurrentCulture).TotalMilliseconds;
+        }
+
+        private static TimeSpanRange GetTimeSpanRangeFromNumberValues(RangedRequest rangedRequest)
+        {
+            return new TimeSpanRange
+            {
+                MillisecondsMin = (double)rangedRequest.GetConvertedMinimum(typeof(double)),
+                MillisecondsMax = (double)rangedRequest.GetConvertedMaximum(typeof(double))
+            };
+        }
+
+        private static object RandomizeTimeSpanInRange(TimeSpanRange range, ISpecimenContext context)
+        {
+            var millisecondsInRange = context.Resolve(
+                new RangedNumberRequest(typeof(double), range.MillisecondsMin, range.MillisecondsMax));
+
+            return millisecondsInRange is NoSpecimen
+                ? millisecondsInRange
+                : TimeSpan.FromMilliseconds((double)millisecondsInRange);
+        }
+
+        private class TimeSpanRange
+        {
+            public double MillisecondsMin { get; set; }
+            public double MillisecondsMax { get; set; }
+        }
+    }
+}
--- a/Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
+++ b/Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
@@ -28,21 +28,9 @@ namespace AutoFixture.DataAnnotations
 
         private static object CreateRangedTimeSpanSpecimen(RangedRequest rangedRequest, ISpecimenContext context)
         {
-            TimeSpanRange range;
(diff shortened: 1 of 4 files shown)


zvirja:
Hi Damian,

Thanks a lot for the great work you did! 🥇 

I didn't have a chance to review the code your posted in detail, but would like to preliminary answer to your question above.

> When the range is specified with numbers, they are treated as milliseconds - I am not sure that it is desired behavior but it seemed very natural to me.

In general we are trying to mimic the default `Range` attribute validation behavior. If you are able to specify ranges there as numbers, we should support them as well:

```c#
[Range(1, 10)]
TimeSpan Value { get; set; } = TimeSpan.FromMilliseconds(5)
```

If the code above doesn't fail the validation, we should also use numbers as milliseconds. Otherwise, if that doesn't work, we should review how value is parsed there and generate it in the same way

Would it be possible for you to review the `Range` attribute on the ReferenceSource and check how it behaves?

Thank you a lot!

I'll take a look at the code when I have a bit of free time.

damian-krychowski:
Sure, i will check this out as soon as I find some free time and let you know what is the actual behaviour of the Range attribute. In case it is different than I assumed I will correct the code to mimic it the right way.

zvirja:
@damian-krychowski Do you know if you find any time in the nearby future to follow-up on this one? Let me know if I can help somehow here.

Thanks.

damian-krychowski:
@zvirja I just checked, and find out that my assumption about numbers which will be treated as milliseconds is wrong (here is the repo I created with some tests to prove that https://github.com/damian-krychowski/ValidationAttributesPlayground/blob/master/Source/ValidationAttributesPlayground/ValidationAttributesPlayground/TimeSpanRangeAttributeTests.cs) 

I will remove that part of functionality

zvirja:
@damian-krychowski Thanks for the verification!

> I will remove that part of functionality

Probably, instead of removal numbers support, we should mimic the way `RangeAttribute` treats the numbers for `TimeSpan`. Likely, that's coded [here](https://referencesource.microsoft.com/#System.ComponentModel.DataAnnotations/DataAnnotations/RangeAttribute.cs,178).

When you update the tests, if possible, please add tests like [here](https://github.com/damian-krychowski/ValidationAttributesPlayground/blob/master/Source/ValidationAttributesPlayground/ValidationAttributesPlayground/TimeSpanRangeAttributeTests.cs):
1. Create type with TimeSpan range attributes
2. Ask `fixture` to create object
3. Verify it using `Validator`.

damian-krychowski:
@zvirja I have removed the handling for timespans ranged with numbers (and I think that mimics how the validation works for TimeSpan as it does not support conversion to int or double). I have also added the test checking if Fixture is able to generate the value which then passes the Validator validation. I hope that is what you meant. Please let me know I should correct anything else.

Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (rangedRequest.MemberType != typeof(TimeSpan))
+                return new NoSpecimen();
+
+            return CreateRangedTimeSpanSpecimen(rangedRequest, context);
+        }
+
+        private static object CreateRangedTimeSpanSpecimen(RangedRequest rangedRequest, ISpecimenContext context)
+        {
+            if (!(rangedRequest.Minimum is string) || !(rangedRequest.Maximum is string)) return new NoSpecimen();
```
zvirja: To follow the same styling please add line break (use one line for guard clauses only).

Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -5998,5 +5998,37 @@ IPostprocessComposer<T> ConfigurePropertyField<T>(IPostprocessComposer<T> compos
                 return composer.With(x => x.Value, "42");
             }
         }
+
+        private class TypeWithRangedTimeSpanProperty
+        {
+            [Range(typeof(TimeSpan), "00:00:00", "12:00:00")]
```
zvirja: Could you please start range from e.g. `02:00:00`? I simply afraid that minimum value looks too default, so if code doesn't work, test will still pass 😅

Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
```diff
@@ -0,0 +1,67 @@
...
+                new RangedNumberRequest(typeof(double), range.MillisecondsMin, range.MillisecondsMax));
+
+            return millisecondsInRange is NoSpecimen
+                ? millisecondsInRange
+                : TimeSpan.FromMilliseconds((double)millisecondsInRange);
+        }
+
+        private class TimeSpanRange
+        {
+            public double MillisecondsMin { get; set; }
```
zvirja: Better operate with `TimeSpan` type here. To me that's `RandomizeTimeSpanInRange()` implementation detail that it uses `TotalMilliseconds`. It will make code more natural 😉

Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
```diff
@@ -0,0 +1,67 @@
...
+            return TimeSpan.Parse(serializedTimeSpan, CultureInfo.CurrentCulture).TotalMilliseconds;
+        }
+
+        private static object RandomizeTimeSpanInRange(TimeSpanRange range, ISpecimenContext context)
+        {
+            var millisecondsInRange = context.Resolve(
+                new RangedNumberRequest(typeof(double), range.MillisecondsMin, range.MillisecondsMax));
+
+            return millisecondsInRange is NoSpecimen
+                ? millisecondsInRange
```
zvirja: Better for consistency and readability:

```c#
if (millisecondsInRange is NoSpecimen)
    return new NoSpecimen();

return TimeSpan.FromMilliseconds((double)millisecondsInRange);
```

It's minor, but everywhere else we follow the similar pattern, so it's quite unexpected to see ternary operator here 😉

Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        private static object CreateRangedTimeSpanSpecimen(RangedRequest rangedRequest, ISpecimenContext context)
+        {
+            if (!(rangedRequest.Minimum is string) || !(rangedRequest.Maximum is string)) return new NoSpecimen();
+
+            var range = GetTimeSpanRangeFromStringValues(rangedRequest);
+            return RandomizeTimeSpanInRange(range, context);
+        }
+
+        private static TimeSpanRange GetTimeSpanRangeFromStringValues(RangedRequest rangedRequest)
```
zvirja: Just a suggestion: `ParseTimeSpanRange()` name might look better, but current name is fine too.

Src/AutoFixture/DataAnnotations/TimeSpanRangedRequestRelay.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            var millisecondsInRange = context.Resolve(
+                new RangedNumberRequest(typeof(double), range.MillisecondsMin, range.MillisecondsMax));
+
+            return millisecondsInRange is NoSpecimen
+                ? millisecondsInRange
+                : TimeSpan.FromMilliseconds((double)millisecondsInRange);
+        }
+
+        private class TimeSpanRange
```
zvirja: Make a struct? 😉

zvirja:
@damian-krychowski Merged! Thanks a lot for your work! :pray:

I'm going to merge a few more things before making the next minor release. Preliminary ETA week or two. Are you fine with that?

damian-krychowski:
@zvirja the pleasure was all mine :) Could you please suggest me some more issues I could play with?

Issue #996, opened 2018-03-09
Drop support for older FakeItEasy versions


blairconrad:
Suggested by @zvirja in https://github.com/AutoFixture/AutoFixture/pull/986#discussion_r173122352.
Assumes that #986 is merged.

Supporting very old versions of FakeItEasy increases development and maintenance costs, and complicates the code quite a bit. If we dropped support for older FIE's, we could simplify the codebase as follows:

* Change minimum supported to 1.7.4257.42
    * Can remove the `CAN_FAKE_DELEGATES` compiler constant - all versions of FakeItEasy after this can fake delegates
    * Can remove `AutoFakeItEasyCustomization.AssertFakeItEasyCanFakeDelegates`
* Change minimum supported to 2.0.0
    * Can remove `AutoFakeItEasyUnitTest`
    * `FakeItEasyMethodQuery.FakeMethod<T>.Invoke` can be  drastically simplified, becoming `return new Fake<T>(b => b.WithArgumentsForConstructor(parameters));`, as it used to be
    * Can remove `FakeObjectCall.cs` bridge class, as it exists only to paper over differences between pre-2.0.0 and 2.0.0+ call representations. Usages can be replaced with `FakeItEasy.Core.IFakeObjectCall`, or `FakeItEasy.Core.IInterceptedFakeObjectCall` if necessary.
* Change minimum supported to 3.0.0
    * Can remove `AutoFakeItEasy.FakeItEasy2UnitTest` (or `AutoFakeItEasy.FakeItEasy3UnitTest` if restricting to FIE 4.0.0+, etc.)

Benefits are cumulative, of course.

zvirja:
@blairconrad Thanks for the detailed explanation. I've added that to v5 scope. I have no clue when we are going to start that work, however hope one day we'll apply those changes 😅

blairconrad:
Given #1155, it may be a good time to at least reconsider this issue. Dropping some older FakeItEasys may afford an opportunity to simplify the AutoFakeItEasy. If there's interest, I would be happy to take this on, given guidance about the lowest FakeItEasy you'd like to support.

mattiasnordqvist:
How about adding support for fakeiteasy 8?

blairconrad:
@mattiasnordqvist, I'm not a maintainer, but I think that would be best raised as a separate issue.

aivascu:
closed in #1438

PR #1025, opened 2018-04-25
Add ConfigureMembers to AutoFakeItEasyCustomization

### LINKED ISSUE #1000: Add Configured mode for AutoFakeItEasy
In https://github.com/AutoFixture/AutoFixture/pull/998#issuecomment-373170972, @zvirja suggests

> Given that we already merged the other approach, there is a gap - faked delegates do not return auto-values. If you wish, I'd like to ask you to implement the Configured mode for the FakeItEasy integration - it would be really awesome to have it. I love how it's done in the NSubstitute and would recommend that as a sample:
> 
> * automatically generate value as a method result/properties/fields. Return the same result for the same args;
> * auto generate the ref and out parameters;
> * the same for delegates;
> * fully support the open generic methods;

blairconrad:
Fixes #1000.


blairconrad:
This is a partial implementation. No pressure, but I'd be happy to have feedback while I continue the work.

I tried implementing ConfigureMembers against all versions of FakeItEasy, but found the problem quite tricky—the API changes that FakeItEasy's undergone since 1.7.4109.1 really interfered, so instead I started as if we were just supporting FIE 4. When that worked, I moved back to 3, and then 2 (as it turns out, no changes were required). So I'm sending that code up for perusal while I try to adapt to FakeItEasy 1.7.4109.1.
I figure if I succeed, we could potentially leave that commit separate from the others and then later if AutoFixture drops FakeItEasy 1.x support, the commit could be reverted and we'll have most of a working (but nicer) solution again. Notes on this implementation:

1. First commit is me ripping off the AutoMoq tests and TestTypes. I made things compile and changed the language to be a little more FakeItEasyish, but that's about it.
2. The second commit has all the magic. There are some differences from the Moq implementation, which should show up by seeing the changes to `FixtureIntegrationTest.cs`. Highlights of the implementation:
    1. AutoFakeItEasy supports [2.0.0,5.0.0)
    1. I drop AutoFakeItEasyUnitTest from All.sln. Mostly so `build.cmd` doesn't complain. This would need to be reversed. Unless we don't merge this feature until we drop support for FakeItEasy 1.x. (And if we did wait until then, we'd need to make more extensive changes, including those from #996.)
    1. Faked (configurable/virtual/whatever you want to call them) read/write properties are configured not by setting a value in, but by configuring a return value. This way if the user only gets the property, and asks if the setter had been called, the answer is "no", as it should be.
    1. Indexers will return the same values whenever gotten with the same arguments.
    1. `ref` parameters are supported.
    1. Properties with circular dependencies are supported
    1. Generic methods are supported.

Thanks!

blairconrad:
Okay, I've added the implementation that covers all versions of 1.7.4109.1 and up. As alwasy, I'm happy to make any changes,

Src/AutoFakeItEasy/MethodRule.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        private void SetOutAndRefValues(FakeObjectCall fakeObjectCall)
+        {
+            var parameters = fakeObjectCall.Method.GetParameters();
+            for (int i = 0; i < parameters.Length; i++)
+            {
+                var parameterParameterType = parameters[i].ParameterType;
+                if (parameterParameterType.IsByRef)
```
zvirja: One day we'll need to start to support the `in` modifier and do not mutate value. But for now let's proceed with this simple check.

Src/AutoFakeItEasy/PropertyRule.cs
```diff
@@ -0,0 +1,127 @@
...
+                       this.argumentTypes.SequenceEqual(call.argumentTypes) &&
+                       this.arguments.SequenceEqual(call.arguments);
+            }
+
+            public override int GetHashCode()
+            {
+                unchecked
+                {
+                    var hashCode = -712421553;
+                    hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(this.methodName);
```
zvirja: Any benefit over `this.methodName.GetHashCode()`? I thought that you expect that value might be null, but in the `Equals()` method above you don't check that...

Src/AutoFakeItEasy/PropertyRule.cs
```diff
@@ -0,0 +1,127 @@
...
+                returnValueAction.ApplyToCall(fakeObjectCall);
+            }
+        }
+
+        private static bool IsProperty(MethodInfo method) =>
+            method.IsSpecialName && (method.Name.StartsWith("get_", StringComparison.Ordinal) ||
+                                     method.Name.StartsWith("set_", StringComparison.Ordinal));
+
+        private static bool IsSetter(FakeObjectCall fakeObjectCall) =>
+            fakeObjectCall.Method.ReturnType == typeof(void);
```
zvirja: Given that this method is very tricky, please duplicate the `method.IsSpecialName && method.Name.StartsWith("set_", StringComparison.Ordinal)` check here to make it more robust. Otherwise, later somebody might miss to verify for `IsProperty` first and we'll get very weird bugs 😅

Src/AutoFakeItEasy/MethodRule.cs
```diff
@@ -0,0 +1,57 @@
...
+    internal class MethodRule : IFakeObjectCallRule
+    {
+        private readonly ISpecimenContext context;
+
+        public MethodRule(ISpecimenContext context)
+        {
+            this.context = context;
+        }
+
+        public int? NumberOfTimesToCall { get; } = null;
```
zvirja: Why do we have it? Did it come from `IFakeObjectCallRule` ? If so, then please:
- make it lambda;
- add a comment 😉 

Same for `PropertyRule`.

Src/AutoFakeItEasyUnitTest/AutoFakeItEasyCustomizationTest.cs
```diff
@@ -63,6 +63,16 @@ public void InitializeWithGenerateDelegatesOptionThrowsIfNotSupported()
         }
 #endif
 
+        [Fact]
+        public void ConfigureMembersIsDisabledByDefault()
```
zvirja: Could you please add a set of kinda structural tests like this: https://github.com/AutoFixture/AutoFixture/blob/f2378ec82e92f124ddac8468a545fa04bf9eddfe/Src/AutoNSubstituteUnitTest/AutoNSubstituteCustomizationTest.cs#L178

To unit-test the customization. The test below are "facade" or "integration" ones.

Src/AutoFakeItEasy.FakeItEasy2UnitTest/AutoFakeItEasy.FakeItEasy2UnitTest.csproj
```diff
@@ -8,15 +8,16 @@
     <AssemblyTitle>AutoFakeItEasy.FakeItEasy2.UnitTest</AssemblyTitle>
     <AssemblyName>AutoFixture.AutoFakeItEasy.FakeItEasy2UnitTest</AssemblyName>
     <RootNamespace>AutoFixture.AutoFakeItEasy.UnitTest</RootNamespace>
-    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES</DefineConstants>
+    <DefineConstants>$(DefineConstants);CAN_FAKE_DELEGATES;HAS_A_CALL_TO_SET</DefineConstants>
```
zvirja: I'd rename to `HAS_A_CALL_TO_SET_TESTER` (or whatever this kind of methods is called in your API), because first time I read it I was quite confused (probably, that's because I don't use FakeItEasy).

zvirja:
@blairconrad Thanks for your work! 🏅 🙏  Had a chance to review it briefly and left a few comments. Will try to make a deeper look later, but am not sure it's required.

One more thing to consider. In NSubstitute we have a feature called ["Partial substitution"](http://nsubstitute.github.io/help/partial-subs/). If configured, it calls the parent method as a fallback. I'm not sure about FakeItEasy, but please ensure that:
- it's still possible to return value from parent call;
- the overlapping behavior is reasonable (e.g. by default parent/fixture is returned always).

The reason is that for NSubsitute that story isn't perfect and I noticed that recently only. It might make sense to add a couple of tests for that. It will be a shame if they are already there and I simply overlooked them 😊

Src/AutoFakeItEasyUnitTest/FixtureIntegrationTest.cs
```diff
@@ -163,5 +164,423 @@ public void FixtureCanCreateAnonymousGuid()
...
+
+        [Fact]
+        public void WithConfigureMembers_MethodsWithRefParametersReturnValuesFromFixture()
+        {
+            // Arrange
+            var fixture = new Fixture().Customize(new AutoFakeItEasyCustomization { ConfigureMembers = true });
+            var frozenString = fixture.Freeze<string>();
+            // Act & Assert
+            IInterfaceWithRefMethod result = null;
+            Assert.Null(Record.Exception(() => result = fixture.Create<IInterfaceWithRefMethod>()));
```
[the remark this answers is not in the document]
blairconrad: Copied from `AutoMoqUnitTest\FixtureIntegrationWithAutoMoqCustomizationTest.cs`. I'd assumed that those tests reflected the desired style. Changing.

blairconrad:
Thanks for the comments, @zvirja! I think I've addressed everything. Generally I just made the requested changes.

The trickiest part was the comment about "partial substitution". Specifically 

> the overlapping behavior is reasonable (e.g. by default parent/fixture is returned always

I don't know what "parent/fixture is returned always" means. I added a test, `WithConfigureMembers_VirtualMemberConfiguredToCallBaseMethodReturnsValueFromBaseMethod`, to cover what I think is the reasonable behaviour.


> Sorry for being picky - I'm quite new to PR reviews and haven't learned to be permissive yet 😅

Don't be sorry for being picky. I welcome it, and in any event, it's your codebase. You have to maintain the project's standards. The important thing (in my opinion) is to be firm but kind. And I've never detected any rudeness from you. If it's not condescending, I think you're doing well.

> If you disagree with some of them - feel free to leave the reasoning 😉

Fret not, I would/will/do, where I feel strongly about something.

blairconrad:
Pushed. Not yet squashed. Given that you seem not to object to the "bridge" to allow the changes to work with the oldest FakeItEasy, I'd be inclined to just squash all the commits into one when you're happy. Or whatever.

zvirja:
Thanks for the follow-up, now it looks perfect 💃 

> Given that you seem not to object to the "bridge" to allow the changes to work with the oldest FakeItEasy

Yep, I'm completely fine with this. Probably it's not too efficient due to reflection, but given that we cannot produce breaking changes, that's the only way we have for now.

Could you please also update #996 with extra instructions? 😉 

> I added a test, `WithConfigureMembers_VirtualMemberConfiguredToCallBaseMethodReturnsValueFromBaseMethod`, to cover what I think is the reasonable behaviour.

That's exactly what I wanted 👍 Should we also assert that by default value is resolved by fixture?

A few more minor questions to clarify:
- I found that currently "property behavior" logic is implemented at integration part. Is there any way to delegate this logic to the FakeItEasy? For instance, in NSubstitute we add our handler in a way that we don't even "see" getter calls if they already have value set by setter. I simply afraid that it looks like we re-implement the FakeItEasy core logic and such things always scare me 😟
- Currently all "glue" libraries implement the "pure" method behavior, meaning that for the same method arguments we return the same result (i.e. `Method(1,2,3)` will generate result once and return the same value for each time; `Method(3,4,5)` will return different result). Do we have the same behavior here? If not, could you please implement that via a strategy (like in [NSubsitute](https://github.com/AutoFixture/AutoFixture/blob/9efbc7348a0c2c199f490ee7d72325e6c03589b5/Src/AutoNSubstitute/NSubstituteRegisterCallHandlerCommand.cs#L58) or in a similar fashion)? Also ensure that it works correctly for methods with `params`.

I believe that after we follow up on the last items, we could remove the "WIP" mark and proceed with a merge 💯

blairconrad:
> That's exactly what I wanted :+1: Should we also assert that by default value is resolved by fixture?

That's what `WithConfigureMembers_VirtualMembersReturnValueFromFixture` does.

> Currently all "glue" libraries implement the "pure" method behavior, […] Do we have the same behavior here?

We do not. I can give it a whirl.

> I found that currently "property behavior" logic is implemented at integration part (these lines). Is there any way to delegate this logic to the FakeItEasy?

I don't believe so. FakeItEasy's get/set autoproperty mechanism will happily defer to any other user-supplied rule that applies to the property, so even if we set a property value in, FIE's "return what we set" logic will be pre-empted by a rule that supplied a value to a getter. This is because the user-supplied rule has no way to tell that the "set" was called without keeping track of all the indices that were supplied to the property setters and we're all of a sudden re-implementing the current `PropertyRule`.

However, implementing pure method behaviour may allow me to alter how we treat the properties. Stay tuned.

blairconrad:
Oh, I forgot this

> Could you please also update #996 with extra instructions? :wink:

The second these changes are merged. Well, right after I get the notification.

blairconrad:
I think we might be done again!

Src/AutoFakeItEasy/MethodCallResult.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+
+namespace AutoFixture.AutoFakeItEasy
+{
+    internal class MethodCallResult
+    {
+        private readonly IList<PositionedValue> outAndRefValues;
```
zvirja: Probably that's a premature optimization, but I'd initialize it lazily, as in 99% it's not used. Don't use `Lazy<T>` - it's the same allocation 😉

Src/AutoFakeItEasyUnitTest/FixtureIntegrationTest.cs
```diff
@@ -163,5 +164,490 @@ public void FixtureCanCreateAnonymousGuid()
...
+            var result = fixture.Create<IInterfaceWithMethod>();
+            // Assert
+            Assert.Equal(frozenString, result.Method("hi"));
+        }
+
+        [Fact]
+        public void WithConfigureMembers_MethodsWithParametersReturnSameValuesWhenCalledWithSameArguments()
+        {
+            // Arrange
+            var fixture = new Fixture().Customize(new AutoFakeItEasyCustomization { ConfigureMembers = true });
```
zvirja: Please add scenario for the modified `ref` parameters:
```c#
int Method(ref int value);

int value = 42;
var result1 = Method(ref value); // here value is updated

value = 42;
var result2 = Method(ref value);

Assert.Equal(result1, result2);
```

blairconrad:
@zvirja, I believe I've addressed all your comments, except for 
https://github.com/AutoFixture/AutoFixture/pull/1025#discussion_r186121911, about hidden members.

By default, FakeItEasy will not intercept the hidden methods as defined on classes, and I am not sure how to configure it to do so. Honestly, this was a surprise to me. The methods can be configured if the defining types are interfaces. I can at least add tests (and change behaviour, if necessary) for interfaces. I'm discussing the lack of interception for class-based hidden methods with the rest of Team FakeItEasy and will provide more information here as it becomes available.

zvirja:
@moodmosaic Do you also wish to review the PR? It contains a bit of code, so one more pair of eyes will be helpful.

If you don't have time - let me know and I'll proceed as is. We already spent a lot of time working on it, so I believe no critical issues are left.

blairconrad:
@zvirja, I also had a good time. And learned things. The only downside is that I worry a little that the other glue libraries might not meet the same standard. The Moq one is at least missing some integration tests, even if the functionality is there (I modelled my original set of tests after Moq, and we've added a number of others since then).

blairconrad:
OH! The commits are a mess. We should remember to squash (or whatever you want - I think squashing to one commit would be fine) before merge. I can restore the `[WIP]` to prevent accidental merging, if you like. Or if you prefer to handle in some other way, that's cool.

zvirja:
@blairconrad I'm going to squash the commits 💥 There are a lot reworks in the approach so basically we have two options:
- Reorganize commits to the logical units, without the discarded code (i.e. `commit 2` removes half of the `commit 1` code because of different way).
- Squash all the code.

I don't see too much value in the option 1 as the feature is quite atomic. Personally, I'd probably try to follow it, but I don't believe it worth it to ask you about that 😜 So, don't worry about the mess and have a rest after the great word you done! 😅

blairconrad:
Personally, I'm with you on "atomic" and don't think it's worthwhile detangling. And FakeItEasy all of a sudden got busy, so if you want a merge soon, I say go for the 1 commit. If later when FIE things have calmed down, and there's been no merge, I'll try to remember and see if I can reorg.

zvirja:
@blairconrad Agreed 👍 

Let's wait for the @moodmosaic reply to understand the further plan better.

Src/AutoFakeItEasyUnitTest/TestTypes/IInterfaceWithNewMethod.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace AutoFixture.AutoFakeItEasy.UnitTest.TestTypes
+{
+    public interface IInterfaceWithShadowedMethod
+    {
+        //shadowed method
```
moodmosaic: Super nitpick: I think all of our comments are written like this:

```
// Shadows method.
```

Src/AutoFakeItEasyUnitTest/TestTypes/IInterfaceWithNewMethod.cs
```diff
@@ -0,0 +1,14 @@
...
+        string Method(int i);
+    }
+
+    public interface IInterfaceWithNewMethod : IInterfaceWithShadowedMethod
+    {
+        //new method
```
moodmosaic: Super nitpick: I think all of our comments are written like this:

```
// New method.
```

blairconrad:
Thanks, @moodmosaic. I'm always happy to address super nitpicks as relate to coding style. Or anything else, I suppose.

I'm sure you're aware of these tools, and probably eschew them for a reason, but have you considered using something like StyleCop to take the burden of talking about spacing away from the reviewers and authors? If I put a comment like the ones you commented on (oh, so meta! I'm commenting about your comment comments) in FakeItEasy code, the compilation gripes:

```
Creation\ProxyGeneratorSelector.cs(16,9): warning SA1005: Single line comment should begin with a space. [D:\Sandbox\FakeItEasy\src\FakeItEasy\FakeItEasy.csproj]
```

moodmosaic:
>but have you considered using something like StyleCop

The last time I used StyleCop must have been around 2007, but almost every single person I met kept saying I should just use ReSharper. (Last time I used ReSharper was 2005, IIRC.)

I liked StyleCop back then, but I've no idea how it looks like today...

blairconrad:
Like many of these tools, if configured poorly, they can lead to way too much harassment. However, if you have a few rules you care about, and want them enforced, I think it works well. The biggest concern I have about the tools is what to do with contributors. In a work setting, I'd set them so that violations broke the build, but I think that's a terrible thing to do to most OSS contributors. But even at warning level, some contributors will self-police and then not introduce violations, and if they _do_ introduce them, it's often an easier conversation for maintainers to have, ("do you mind making this adjustment to remove the new warning you introduced?") or even just a better way for maintainers to find the new violations and clean up themselves, if that's what's preferred.

I'm not trying to sell you on the tool or anything (I don't get paid if you use it). Personally, I like the way it helps keep the code tidy, but it's not for everyone.

zvirja:
@moodmosaic Thanks for your review!
@blairconrad Thanks for sharing your thoughts. Let's discuss the StyleCop question in the issue you created, thanks for that!

For now I'm proceeding further and merging this feature! Thanks for all the work done! 🙏 
Will release it in the nearby future. Probably, will combine it with a few more features.

Was a pleasure to work on this millennium issue 😅

zvirja:
@blairconrad Just a reminder about the promise you left [here](https://github.com/AutoFixture/AutoFixture/pull/1025#issuecomment-385466557) 😉

blairconrad:
Fulfilled!

blairconrad:
@zvirja, it was an interesting exercise. Thanks for the merge.

Issue #766, opened 2017-06-17
GreedyAttribute and Automoq don't play nice together


DixonDs:
I would be great if applying `[Greedy]` to a `Mock<>` parameter when `AutoMoqCustomization` is used would cause it to use `GreedyMockConstructorQuery` instead of `MockConstructorQuery`

zvirja:
While this behavior might look logical, I wouldn't implement it - we should not re-purpose the `[Greedy]` attribute. Instead, you can create your custom `[GreedyMock]` attribute which will do exactly what you say.

One more note is that `[Greedy]` attribute is defined in Test Framework specific glue libraries (e.g. `Xunit`, `Xunit2`) which are not aware of the Mocking Framework. Therefore:
- it's unlikely that we can re-purpose the existing `[Greedy]` attribute logic even if we want;
- each glue library contains its own `CustomizeAttribute` (e.g.
 [here](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/CustomizeAttribute.cs)) for attribute customizations, therefore, it's impossible to define `GreedyMock` alternative inside the `AutoMoq` project.

The only solution would be to create your custom attribute for your projects which reference both `Xunit/NUnit` and `AutoMoq`.

Reference @AutoFixture/core team, probably they could advise more.

DixonDs:
> One more note is that [Greedy] attribute is defined in Test Framework specific glue libraries (e.g. Xunit, Xunit2) which are not aware of the Mocking Framework

Correct me if I am wrong, but as far I remember, Autofixture effectively implements the chain-of-responsibility pattern, so you can probably implement this by having a proper order of customizations, i.e. AutoMoqCustomization can override and repurpose `[Greedy]` to make sense out of it when constructing instances of  `Mock<>` classes

zvirja:
Well, in theory it probably possible to implement that in a way you describe, but currently that doesn't fall withing the [existing `Greedy` implementation](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture.xUnit.net2/GreedyAttribute.cs) - it customizes the graph for exactly `Mock<T>` creation :disappointed:

Let's see what other guys say about the feature first.

zvirja:
[...]

Issue #1014, opened 2018-04-04
[NSubstitute] Add infrastructure for easy constructor customization for created substitutes


zvirja:
Like #1013, but for NSubsitutute

Issue #1013, opened 2018-04-04
[Moq] Add infrastructure for easy constructor customization for created mocks


zvirja:
Currently it's very hard to customize which exact constructor overload to choose when create a mock for the class. Rework the architecture to allow to customize that easy.

Suggestion: Introduce `BareMockRequest` type to activate the mocks. We could later easily customize that via `Customizations` collection.

PR #1033, opened 2018-05-02
Add GreedyMockAttribute (#766)

michalkowal:
I prepared also abstract MockCustomizeAttribute for any further attributes (like for xUnit, NUnit etc)

### DIFF
--- a/Src/AutoMoq/GreedyMockAttribute.cs
+++ b/Src/AutoMoq/GreedyMockAttribute.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Reflection;
+
+namespace AutoFixture.AutoMoq
+{
+    /// <summary>
+    /// An attribute that can be applied to parameters in an AutoDataAttribute-driven
+    /// Theory to indicate that the parameter value should be created using the most greedy
+    /// constructor that can be satisfied by an <see cref="IFixture"/>.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false)]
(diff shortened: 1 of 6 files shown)


michalkowal:
I also have proposal to add similar "greedy" functionalities into other mocking projects like Rhino, FakeItEasy etc. Maybe it is not very important functionality, but then most projects will have similar functionalities.
Of course I can do that

zvirja:
@michalkowal Thanks a lot for giving your attention here! 🥇 

I gave it a quick look and would like to share a few thoughts FWIW:
- The following customization implementation might not work with the `ConfigureMembers = true` customization mode:
    ```c#
    return new ConstructorCustomization(parameter.ParameterType, new GreedyMockConstructorQuery());
    ```
    If you wrap argument with `GreedyMock` attribute, the generated object will not be auto-configured.
- There is the ongoing issue #1013 to refactor the way we activate mocks. That should be an enabler for this PR (as you want to customize the mock activation without touching the further configuration actions). I have that change in my TODO list and hope to take a look in a week or two.
- If you are in progress, please ensure to add the `[WIP]` PR title prefix. This way I know that PR is not ready for review and you are still working on it 😉

> I also have proposal to add similar "greedy" functionalities into other mocking projects like Rhino, FakeItEasy etc

I'd say that `Rhino` is completely dead, but for `FakeItEasy` and `NSubstitute` it might make sense to consider the feature. However again, that could happen after the restructure (#1013, #1014) only.

michalkowal:
@zvirja thanks for your review.
Right now, as you suggest, I stop activity here and I will monitor status off this two issues

PR #1034, opened 2018-05-05
Min/MaxLength attributes relay added

### LINKED ISSUE #678: Feature: Create a string specimen with Min/MaxLength constraint

damian-krychowski:
Hi guys, 

I have implemented MinAndMaxLenghtAttributeRelay (I created that based on StringLengthAttributeRelay). I have decided to create one relay for both Min and Max attributes to be able to handle a situation when both of them are applied to a single property at once. 

All remarks will be highly appreciated! 

Closes #678


Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            if (maxLengthAttribute == null && minLengthAttribute == null)
+            {
+                return new NoSpecimen();
+            }
+
+            var minLength = minLengthAttribute?.Length ?? 0;
+            var maxLength = maxLengthAttribute?.Length ?? Int32.MaxValue;
+
+            return context.Resolve(new ConstrainedStringRequest(minLength, maxLength));
```
zvirja: According to the [MSDN](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.maxlengthattribute(v=vs.110).aspx) this attribute can control both string and array length. Therefore, we should check the member type and if that is an array, we should create `FiniteSequenceRequest`.

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -31,18 +41,117 @@ public object Create(object request, ISpecimenContext context)
...
+        {
+            return context.Resolve(new RangedNumberRequest(typeof(int), range.Min, range.Max));
+        }
+
+        private static object GetRandomCollection(ISpecimenContext context, Type elementType, object collectionCount)
+        {
+            return context.Resolve(new FiniteSequenceRequest(elementType, (int)collectionCount));
+        }
+
+        private static object ToArray(IEnumerable elements, Type elementType)
```
damian-krychowski: This function was copied from ArrayRelay. I would prefer to unify this logic somehow - @zvirja maybe some extension method for IEnumerable type? Or maybe you have some other suggestions?

Src/AutoFixture/DataAnnotations/IMemberTypeResolver.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace AutoFixture.DataAnnotations
+{
+    public interface IMemberTypeResolver
```
damian-krychowski: @zvirja I had to introduce this interface to be able to test the Relay - but I found it is something that was not practiced before - injecting dependencies into the relays. Unfortunately, type FakeMamberInfo does not provide a way to mock a MemberType, and I am not sure if that is even possible for it to do. Maybe you can figure out a better way to achieve this testability without introducing this interface?

damian-krychowski:
Hi @zvirja, will you find some time to review the changes?

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+    /// </summary>
+    public class MinAndMaxLengthAttributeRelay : ISpecimenBuilder
+    {
+        private readonly IMemberTypeResolver memberTypeResolver;
+
+        /// <summary>
+        /// Creates a new instance of MinAndMaxLengthAttributeRelay
+        /// </summary>
+        /// <param name="memberTypeResolver">MemberTypeResolver to determine member type of requested object</param>
+        public MinAndMaxLengthAttributeRelay(IMemberTypeResolver memberTypeResolver)
```
zvirja: Please use property injection here instead of constructor injection. We have a good default for the `IMemberTypeResolver`, so no need for client to pass it.

Example:
https://github.com/AutoFixture/AutoFixture/blob/11066d77bcee17e4175679cfa0e9133957695671/Src/AutoFixture.NUnit3/AutoDataAttribute.cs#L27-L31

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            var range = Range.TryGetFromAttributes(request);
+            
+            if (range == null)
+            {
+                return new NoSpecimen();
+            }
+
+            var memberType = memberTypeResolver.TryGetMemberType(request);
```
zvirja: Please update also [`RangeAttributeRelay`](https://github.com/AutoFixture/AutoFixture/blob/11066d77bcee17e4175679cfa0e9133957695671/Src/AutoFixture/DataAnnotations/RangeAttributeRelay.cs#L52) to use this interface. Use property injection pattern there as well.

Implementation might look like following:
```c#
private static Type GetMemberType(RangeAttribute rangeAttribute, object request)
{
    if (this.MemberTypeResolver.TryGetMemberType(request, out var memberType))
    {
        return memberType;
    }

    return Nullable.GetUnderlyingType(rangeAttribute.OperandType) ?? rangeAttribute.OperandType;
}
```

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
zvirja: Let's keep that as is for now 😉 Usually we try to follow the rule of 3 and unify it only in that case. For now it's unclear what is the best way to do that. The idea with extension method would probably work as well, but let's postpone it for now.

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+                    ? null
+                    : GetRange(minLengthAttribute, maxLengthAttribute);
+            }
+
+            private static Range GetRange(MinLengthAttribute minLengthAttribute, MaxLengthAttribute maxLengthAttribute)
+            {
+                var min = minLengthAttribute?.Length ?? 0;
+                var max = maxLengthAttribute?.Length ?? Int32.MaxValue;
+
+                //to avoid creation of empty strings/arrays
```
zvirja: Styling. Change to `// To avoid creation of empty strings/arrays.`

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            if (collectionCount is NoSpecimen)
+                return new NoSpecimen();
+
+            var resultCollection = GetRandomCollection(context, elementType, collectionCount);
+
+            return ToArray((IEnumerable)resultCollection, elementType);
+        }
+
+        private static object GetRandomNumberWithinRange(ISpecimenContext context, Range range)
```
zvirja: Change signature to return `int` as only this method knows the type of result.

Src/AutoFixtureUnitTest/DataAnnotations/MinAndMaxLengthAttributeRelayTest.cs
```diff
@@ -0,0 +1,273 @@
...
+            sut.Create(request, context);
+
+            //Assert
+            var expectedRequest = new ConstrainedStringRequest(expectedMin, max);
+
+            Assert.Equal(expectedRequest, actualRequest);
+
+        }
+
+        private class MockMemberTypeResolver : IMemberTypeResolver
```
zvirja: Please create `DelegatingMemberTypeResolver` like [this](https://github.com/AutoFixture/AutoFixture/blob/11066d77bcee17e4175679cfa0e9133957695671/Src/AutoFixtureUnitTest/Kernel/DelegatingSpecimenContext.cs#L6) and return `false` by default. I'd expect that each test would configure it rather than relying on default behavior - for better readability.

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeMemberTypeResolver.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Reflection;
+
+namespace AutoFixture.DataAnnotations
+{
+    /// <summary>
+    /// Member type resolver used by MinAndMaxLengthAttributeRelay
+    /// </summary>
+    public class MinAndMaxLengthAttributeMemberTypeResolver : IMemberTypeResolver
```
zvirja: Please add unit tests for this class.

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            return memberType?.IsArray ?? false;
+        }
+        
+       private static object ResolveString(ISpecimenContext context, Range range)
+        {
+            return context.Resolve(new ConstrainedStringRequest(range.Min, range.Max));
+        }
+
+        private static object ResolveArray(ISpecimenContext context, Type memberType, Range range)
```
zvirja: Probably it makes sense to rename `memberType` to `arrayType` 😉

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            var elementType = memberType.GetElementType();
+
+            var collectionCount = GetRandomNumberWithinRange(context, range);
+
+            if (collectionCount is NoSpecimen)
+                return new NoSpecimen();
+
+            var resultCollection = GetRandomCollection(context, elementType, collectionCount);
```
zvirja: Should we test against `NoSpecimen` reply here?

zvirja:
@damian-krychowski Thanks a lot for the follow-up! I've added a few minor tweaks to the code.

@moodmosaic Do you want to take a look as well? 😉

zvirja:
@moodmosaic Thanks for the quick check 😉 
@damian-krychowski I've merged the PR. Thanks for all your efforts and sorry for the belonged replies 🙏 🍷

PR #1045, opened 2018-06-04
Fix false positive in deferred method detection

### LINKED ISSUE #1042: GuardClauseAssertion incorrectly warns about deferred execution when return type is string
See also: #277 

# Steps to reproduce #
Assume this SUT:

```cs
public class MyClass
{
    public string GetString(string input)
    {
        return "Test";
    }
}
```
Write this test:
```cs
[Fact]
public void Demo()
{
    new GuardClauseAssertion(new Fixture()).Verify(typeof(MyClass));
}
```

# Expected result #
The test should fail, with the following warning:

`An attempt was made to assign the value null to the parameter "input" of the method "GetString", and no Guard Clause prevented this. Are you missing a Guard Clause?`

# Actual result #
The test fails with this warning:

`An exception of type 'AutoFixture.Idioms.GuardClauseException' occurred in AutoFixture.Idioms.dll but was not handled in user code: 'A Guard Clause test was performed on a method that may contain a deferred iterator block, but the test failed. See the inner exception for more details. However, because of the deferred nature of the iterator block, this test failure may look like a false positive. Perhaps you already have a Guard Clause in place, but in conjunction with the 'yield' keyword (if you're using C#); if this is the case, the Guard Clause is dormant, and will first be triggered when a client starts looping over the iterator. This doesn't adhere to the Fail Fast principle, so should be addressed.
See e.g. http://codeblog.jonskeet.uk/2008/03/02/c-4-idea-iterator-blocks-and-parameter-checking/ for more details.'
 Inner exceptions found, see $exception in variables window for more details.
 Innermost exception 	 AutoFixture.Idioms.GuardClauseException : An attempt was made to assign the value null to the parameter "input" of the method "GetString", and no Guard Clause prevented this. Are you missing a Guard Clause?
Method Signature: System.String GetString(System.String)
Parameter Type: System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e
Declaring Type: xunitaf01.UnitTest.MyClass, xunitaf01.UnitTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
Reflected Type: xunitaf01.UnitTest.MyClass, xunitaf01.UnitTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`

zvirja:
Fixes #1042.

I've found a way to make the deferred method detection even more robust. According to [MSDN](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/yield), return type for the methods with `yield` statement can only have 4 possible values. Therefore, testing for the precisely that types, rather than for any assignable one.

@moodmosaic Please take a look.
@adamchester Might be interesting to you, as you already tried to fix this issue in #278.

### DIFF
--- a/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
+++ b/Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
@@ -487,6 +487,7 @@ namespace AutoFixture.IdiomsUnitTest
         [InlineData(typeof(ClassWithEnumerableNonDeferredArrayListMissingGuard))]
         [InlineData(typeof(ClassWithEnumerableNonDeferredStackMissingGuard))]
         [InlineData(typeof(ClassWithEnumerableNonDeferredReadOnlyCollectionBaseMissingGuard))]
+        [InlineData(typeof(ClassWithEnumerableNonDeferredStringMissingGuard))]
         public void VerifyMethodWithNonDeferredMissingGuardThrowsExceptionWithoutDeferredMessage(
             Type type)
         {
@@ -629,6 +630,14 @@ namespace AutoFixture.IdiomsUnitTest
                     { "uniqueKey1", someString },
                     { "uniqueKey2", someString }
                 };
+            }
+        }
+
+        private class ClassWithEnumerableNonDeferredStringMissingGuard
+        {
+            public string GetValues(string someString)
+            {
+                return someString;
             }
         }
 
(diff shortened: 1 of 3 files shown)


Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
```diff
@@ -1626,37 +1637,49 @@ public NestedGenericParameterTestType(Func<T1, IEnumerable<IEnumerable<T2>>, T1[
 
         private class NonProperlyGuardedClass
         {
-            private const string InvalidParamName = "invalidParamName";
-
             public NonProperlyGuardedClass(object argument)
             {
-                if (argument == null) throw new ArgumentNullException(InvalidParamName);
+                if (argument == null) throw new ArgumentNullException("INVALID");
```
moodmosaic: Instead of just "INVALID" perhaps [a descriptive message is warranted](http://blog.ploeh.dk/2014/12/23/exception-messages-are-for-programmers/)?

Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
```diff
@@ -1626,37 +1637,49 @@ public NestedGenericParameterTestType(Func<T1, IEnumerable<IEnumerable<T2>>, T1[
...
             }
 
             public object Property
             {
                 get => null;
                 set
                 {
-                    if (value == null) throw new ArgumentNullException(InvalidParamName);
+                    if (value == null) throw new ArgumentNullException("INVALID");
```
moodmosaic: Instead of just "INVALID" perhaps [a descriptive message is warranted](http://blog.ploeh.dk/2014/12/23/exception-messages-are-for-programmers/)?

Src/IdiomsUnitTest/GuardClauseAssertionTest.cs
```diff
@@ -1626,37 +1637,49 @@ public NestedGenericParameterTestType(Func<T1, IEnumerable<IEnumerable<T2>>, T1[
...
                 }
             }
 
             public void Method(object argument)
             {
-                if (argument == null) throw new ArgumentNullException(InvalidParamName);
+                if (argument == null) throw new ArgumentNullException("INVALID");
```
moodmosaic: Instead of just "INVALID" perhaps [a descriptive message is warranted](http://blog.ploeh.dk/2014/12/23/exception-messages-are-for-programmers/)?

moodmosaic:
(TBC, this should be good to go even if you don't care about spending time at those small comments. :unicorn:)

zvirja:
@adamchester Thanks!
@moodmosaic Done, fixed! 👍

Issue #1050, opened 2018-06-22
Array property decorated with ```MinLength``` only will enter an infinite loop


Pvlerick:
This test will enter an infinite loop:

```csharp
private class TypeWithArrayPropertyWithMinLength
{
    [MinLength(5)]
    public string[] PropertyWithMinLength { get; set; }
}

[Fact]
public void ArrayDecoratedWithMinLengthAttributeShouldPassValidation()
{
    // Arrange
    var sut = new Fixture();

    // Act
    var item = sut.Create<TypeWithArrayPropertyWithMinLength>();

    // Assert
    Validator.ValidateObject(item, new ValidationContext(item), true);
}
```

Funny thing, with ```MaxLength``` only it works fine :-)

Pvlerick:
One more thing: this behavior was not present in 4.3.0 but appeared in 4.4.0

zvirja:
@Pvlerick Good catch! Thanks for reporting the bug!

Pvlerick:
As a matter of fact, it's not an infinite loop, but if there is no ```MaxLength``` property, it will be int.MaxValue, which means that the randomly selected range may request the creation of hundreds of millions of objects :-)

zvirja:
Fixed via #1051

PR #1051, opened 2018-06-23
If MaxLengthAttribute is not present, make it MinLengthAttribute + 10…

Pvlerick:
…0 instead of int.MaxValue to avoid to avoid creation of too many objects

Addressing #1050 

100 is completely arbitrary, could be a property on the Relay or in its ctor, but this basically adresses the issue.

### DIFF
--- a/Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
+++ b/Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
@@ -156,7 +156,7 @@ namespace AutoFixture.DataAnnotations
             private static Range GetRange(MinLengthAttribute minLengthAttribute, MaxLengthAttribute maxLengthAttribute)
             {
                 var min = minLengthAttribute?.Length ?? 0;
-                var max = maxLengthAttribute?.Length ?? int.MaxValue;
+                var max = maxLengthAttribute?.Length ?? min + 100;
 
                 // To avoid creation of empty strings/arrays.
                 if (max > 0 && min == 0)
--- a/Src/AutoFixtureUnitTest/DataAnnotations/MinAndMaxLengthAttributeRelayTest.cs
+++ b/Src/AutoFixtureUnitTest/DataAnnotations/MinAndMaxLengthAttributeRelayTest.cs
@@ -98,7 +98,7 @@ namespace AutoFixtureUnitTest.DataAnnotations
             var request = new FakeMemberInfo(
                 new ProvidedAttribute(minLengthAttribute, true));
 
-            var expectedRequest = new ConstrainedStringRequest(min, int.MaxValue);
+            var expectedRequest = new ConstrainedStringRequest(min, min + 100);
             var expectedResult = new object();
             var context = new DelegatingSpecimenContext
             {
--- a/Src/AutoFixtureUnitTest/FixtureTest.cs
+++ b/Src/AutoFixtureUnitTest/FixtureTest.cs
@@ -6190,5 +6190,43 @@ namespace AutoFixtureUnitTest
             // Assert
             Validator.ValidateObject(item, new ValidationContext(item), true);
         }
+
+        private class TypeWithArrayPropertyWithMinLength
+        {
+            [MinLength(5)]
+            public string[] PropertyWithMinLength { get; set; }
+        }
+
+        [Fact]
+        public void ArrayDecoratedWithMinLengthAttributeShouldPassValidation()
+        {
+            // Arrange
+            var sut = new Fixture();
+
+            // Act
+            var item = sut.Create<TypeWithArrayPropertyWithMinLength>();
+
+            // Assert
+            Validator.ValidateObject(item, new ValidationContext(item), true);
+        }
+
+        private class TypeWithArrayPropertyWithMaxLength
+        {
+            [MaxLength(5)]
+            public string[] PropertyWithMaxLength { get; set; }
+        }
+
+        [Fact]
+        public void ArrayDecoratedWithMaxLengthAttributeShouldPassValidation()
+        {
+            // Arrange
+            var sut = new Fixture();
+
+            // Act
+            var item = sut.Create<TypeWithArrayPropertyWithMaxLength>();
+
+            // Assert
+            Validator.ValidateObject(item, new ValidationContext(item), true);
+        }
     }
 }


zvirja:
@Pvlerick Thanks for working on this!
I see two ways to solve the situation if `MaxLength` is not specified:
- create `MinLength` items;
- create `MinLength + FixedBuffer` items.

In this PR you proceeded with the second approach with `FixedBuffer == 100`. 

Personally, I don't know which the approach is better.

@moodmosaic Could you please share you thoughts how you would solve the issue?

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
```diff
@@ -156,7 +156,7 @@ public static bool TryGetFromAttributes(object request, out Range range)
             private static Range GetRange(MinLengthAttribute minLengthAttribute, MaxLengthAttribute maxLengthAttribute)
             {
                 var min = minLengthAttribute?.Length ?? 0;
-                var max = maxLengthAttribute?.Length ?? int.MaxValue;
+                var max = maxLengthAttribute?.Length ?? min + 100;
```
moodmosaic: What's so special about 100? Why not 1, or 15, or 50? Why not 1337? Can we *ask* the context for a positive integer instead?

moodmosaic:
It's awesome to see you back @Pvlerick 🎉 I left a comment there.

Src/AutoFixture/DataAnnotations/MinAndMaxLengthAttributeRelay.cs
zvirja: @moodmosaic Here is the issue we are trying to fix by this PR. Currently to calculate the actual array size we already take a _random_ positive number in `[min;int.MaxValue)` range. But the point is that value might be very large and AutoFixture might generate the array for long period of time or even fail with OOM. We need to somehow restrict AutoFixture to not create arrays of crazy sizes.

Personally I also don't like `100` as a constant, as it isn't bound to the min value somehow. Thinking more about this PR I came up with an idea to consider the upper boundary (if not specified explicitly) as `min * 2`. This will allow to keep the comparable order for the actual size (for instance if `min` is 10, then capacity is in `[10;20]` range, while if `min` is 1000, the capacity is in `[1000;2000]` range). This way capacity will not be crazy unless `min` is not set to some really huge value - but in this case it's fine as we are enforced by the lower boundary anyway.

WDYT? 😉

moodmosaic: It looks like you're kind of hijacking the purpose of a *relay* that's all 😃 The purpose of relays is to *relay* a request for to one or more ISpecimenBuilder instances; in this case, I'd relay the request.

Then, it'd be the responsibility of some particular ISpecimenBuilder to decide what to do next. And MinAndMaxLengthAttributeRelay would keep having just one responsibility, instead of two.

zvirja: Yes, you are completely right! I'll fix that in my subsequent PR and will make it a pure relay.

Anyhow, the responsibility of this class is to define the range, so the change under discussion is still relevant.

moodmosaic: >Yes, you are completely right! I'll fix that in my subsequent PR and will make it a pure relay.

👍

>the responsibility of this class is to define the range, so the change under discussion is still relevant.

Why not just `byte.MaxValue` or `int16.MinValue` then, instead of the `100` magic number?

zvirja: @moodmosaic Please refer to [this](https://github.com/AutoFixture/AutoFixture/pull/1051#discussion_r198970259) comment. There I suggested to use `min * 2` and described my reasoning about that... I suggest to proceed with `min x2` rather than `100`.

`byte.MaxValue` and `int16.MaxValue` is not a universal constant, as it could always happen that `min` is greater than `byte.Max` or `int16.Max`...

moodmosaic: >`byte.MaxValue` and `int16.MaxValue` is not a universal constant, as it could always happen that `min` is greater than `byte.Max` or `int16.Max`...

Same with `min * 2`?

---

All I'm trying to say here is that we better be careful and make sure we're *not* putting a lipstick on a pig... Have we examined the [route cause (4.3.0 → 4.4.0)](https://github.com/AutoFixture/AutoFixture/issues/1050#issuecomment-399574161) of this issue?

zvirja: > Same with min * 2?

Nope, as `min * 2` is always greater than `min` 😅

> Have we examined the route cause (4.3.0 → 4.4.0) of this issue?

Yes. The problem is that currently array length is a _random_ number in `[min;int.MaxValue)` range. Often that capacity is __very__ huge and AutoFixture hangs trying to create such a large array.

To fix the issue we should not create huge arrays if `min` boundary allows that. If we pick random length in `[min;min * 2]` (or `[min;min+100]` as advised by @Pvlerick) range, it would perfectly solve the problem.

moodmosaic: AFAICT, the problem originates from https://github.com/AutoFixture/AutoFixture/pull/1051/?

zvirja: Yeah, that's the commit where the code has been introduced.

moodmosaic: Then we must at least let @damian-krychowski be aware of this. The change proposed by @Pvlerick makes sense for now, so let's move forward and merge this.

However, this really signals that we need to introduce the notion of *Range*, so that when there isn't an upper bound, you'll get a small *max* which scales constantly, or linearly, or exponentially.

A similar concept is used [here](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Internal/Range.hs) and [here](https://github.com/hedgehogqa/fsharp-hedgehog/blob/master/src/Hedgehog/Range.fs) in Hedgehog.

moodmosaic:
From my experience with DataAnnotations combined with AutoFixture, where they're used for providing fine-grained control over the scope of generated values, I have to say that they have a horrible API...

See my comments in https://github.com/AutoFixture/AutoFixture/issues/453 as an example. I'd consider putting them on a separate package (and keep the namespace as-is for now). We can make an issue for it if you think a change like this would be sensible.

zvirja:
@moodmosaic As far as I remember, [we have already discussed that](https://github.com/AutoFixture/AutoFixture/issues/930#issuecomment-346496083) previously and agreed to not separate them project will contains very few classes.

Do you feel that something changed from those times? :astonished:

moodmosaic:
The problem is that DataAnnotations keep causing confusing, aren't they?

zvirja:
Well, in this particular PR we are fixing the bug introduced recently, so I accept that fact as OK 😅 

Apart from that - well, not all the attributes are covered, but that looks fine IMO. We'll need to maintain the support in any case, so behavior might be still confusing and no matter whether the support is built-in, or a separate package is installed 😄

moodmosaic:
I don't know. I have mixed feelings for them as you can tell from this and #930.

Pvlerick:
Thanks for merging this. I agree that the 100 magic value is ugly but I'm not familiar enough with the code to find a better way that didn't involve asking the context.

PR #1057, opened 2018-07-08
Introduce ranged sequence request

zvirja:
As promised, created a new kind of request - `RangedSequenceRequest`. It's purpose is to activate a sequence withing the specified boundaries.

Currently I'm simply picking a random value withing a range. Later we might change the strategy if really needed.

### DIFF
--- a/Src/AutoFixture/EnumerableExtensions.cs
+++ b/Src/AutoFixture/EnumerableExtensions.cs
@@ -1,13 +1,20 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Reflection;
 
 namespace AutoFixture
 {
-    internal static class EnumerableList
+    internal static class EnumerableExtensions
     {
-        internal static int IndexOf<T>(this IEnumerable<T> items, T item)
+        private static readonly MethodInfo BuildTypedArrayMethodInfo =
+            typeof(EnumerableExtensions).GetTypeInfo().GetMethod(
+                nameof(BuildTypedArray),
+                BindingFlags.Static | BindingFlags.NonPublic);
+
+        public static int IndexOf<T>(this IEnumerable<T> items, T item)
         {
             int i = 0;
             foreach (var s in items)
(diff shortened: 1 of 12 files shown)


Src/AutoFixture/EnumerableExtensions.cs
```diff
@@ -20,17 +27,17 @@ internal static int IndexOf<T>(this IEnumerable<T> items, T item)
             return -1;
         }
 
-        internal static IEnumerable<T> Insert<T>(this IEnumerable<T> items, int index, T item)
+        public static IEnumerable<T> Insert<T>(this IEnumerable<T> items, int index, T item)
```
moodmosaic: You sure you wan't this public?

Src/AutoFixture/EnumerableExtensions.cs
```diff
@@ -20,17 +27,17 @@ internal static int IndexOf<T>(this IEnumerable<T> items, T item)
...
         {
             return items.Take(index).Concat(new[] { item }).Concat(items.Skip(index));
         }
 
-        internal static IEnumerable<T> RemoveAt<T>(this IEnumerable<T> items, int index)
+        public static IEnumerable<T> RemoveAt<T>(this IEnumerable<T> items, int index)
```
moodmosaic: You sure you wan't this public?

Src/AutoFixture/EnumerableExtensions.cs
```diff
@@ -20,17 +27,17 @@ internal static int IndexOf<T>(this IEnumerable<T> items, T item)
...
         {
             return items.Take(index).Concat(items.Skip(index + 1));
         }
 
-        internal static IEnumerable<T> SetItem<T>(this IEnumerable<T> items, int index, T item)
+        public static IEnumerable<T> SetItem<T>(this IEnumerable<T> items, int index, T item)
```
moodmosaic: You sure you wan't this public?

Src/AutoFixture/EnumerableExtensions.cs
```diff
@@ -41,5 +48,19 @@ internal static IEnumerable<T> SetItem<T>(this IEnumerable<T> items, int index,
 
             return a.Take(index).Concat(new[] { item }).Concat(a.Skip(index + 1));
         }
+
+        public static object ToTypedArray(this IEnumerable items, Type elementType)
```
moodmosaic: Perhaps it can be internal?

Src/AutoFixture/EnumerableExtensions.cs
zvirja: Notice, the `EnumerableExtensions` class is internal, so all the methods are still internal, even though they are marked as `public`.

Src/AutoFixture/EnumerableExtensions.cs
zvirja: Notice, the `EnumerableExtensions` class is internal, so all the methods are still internal, even though they are marked as `public`.

Src/AutoFixture/EnumerableExtensions.cs
zvirja: Notice, the `EnumerableExtensions` class is internal, so all the methods are still internal, even though they are marked as `public`.

Src/AutoFixture/EnumerableExtensions.cs
zvirja: Notice, the `EnumerableExtensions` class is internal, so all the methods are still internal, even though they are marked as `public`. This particular method is indeed internal, I don't want to make it public.

zvirja:
@moodmosaic From the design I have in my head, I definitely see value in all the requests/relays we currently have. All them cover the different areas:
- `MinAndMaxLengthAttributeRelay` - it knows about the data annotations attributes and how they should be applied.
- `RangedSequenceRequest/RangedSequenceRelay` - an abstraction for the AutoFixture core to create a sequence with the length within the specified range. It decides the exact length.
- `FiniteSequenceRequest` - request for the sequence of exact length.

I don't see the types we could strip. If you see some other design - please share your vision 👍

Issue #128, opened 2013-06-25
Customizing IEnumerable<SomeType> doesn't change the behavior of CreateMany<SomeType>


dhilgarth:
I am trying to configure what happens when many items of a particular type are created:

``` C#
f.Customize<IEnumerable<string>>(c => c.FromFactory(() => new[] { "a", "b", "c", "d" }));
```

This is used when I explicitly request an `IEnumerable<string>`:

``` C#
f.Create<IEnumerable<string>>() // returns { "a", "b", "c", "d" }
```

But it is not used when using `CreateMany`:

``` C#
f.CreateMany<string>() // doesn't return { "a", "b", "c", "d" }
```

Is this intended behavior? If so, what is the rational behind it?

PR #1058, opened 2018-07-08
Deprecate the MapCreateManyToEnumerable customization

zvirja:
Customization was added back in #128.

This customization leads to the recursion if client forgets to add a customization for the particular `IEnumerable<T>`. The reason is that the requests chain for the `IEnumerable<T>` request is following:
```
IEnumerable<T>
    MuiltipleRequest    
```

If customization is added, it leads to the cyclic requests.

This kind of behavior is really confusing to bundle it with the library. Instead, clients are advised to use the code snippet to customize the `MultipleRequest` requests for the desired sequence types only. This way we don't affect `IEnumerable` requests for other types.

Clients are also free to create the own extension methods inside their solutions, so the fact that the suggested chain is quite long shouldn't be a problem.

@moodmosaic Please take a look. The main motivation is to keep our library simple and handle only common scenarios. The scenario mentioned in #128 looks quite ad-hoc to keep the confusing customization for it.

### DIFF
--- a/Src/AutoFixture/Kernel/MultipleToEnumerableRelay.cs
+++ b/Src/AutoFixture/Kernel/MultipleToEnumerableRelay.cs
@@ -9,6 +9,9 @@ namespace AutoFixture.Kernel
     /// </summary>
     /// <seealso cref="MultipleToEnumerableRelay.Create(object, ISpecimenContext)" />
     /// <seealso cref="MapCreateManyToEnumerable" />
+    [Obsolete("This relay is obsolete and will be removed in future versions. " +
+              "The reason is that it causes recursion overflow for many requests " +
+              "if you forget to add IEnumerable<T> customization.")]
     public class MultipleToEnumerableRelay : ISpecimenBuilder
     {
         /// <summary>
--- a/Src/AutoFixture/MapCreateManyToEnumerable.cs
+++ b/Src/AutoFixture/MapCreateManyToEnumerable.cs
@@ -25,6 +25,20 @@ namespace AutoFixture
     /// </para>
     /// </remarks>
     /// <seealso cref="MultipleToEnumerableRelay" />
+    [Obsolete("This relay is obsolete and will be removed in future versions. " +
+              "The reason is that it causes recursion overflow for \"many\" requests " +
+              "if you forget to add IEnumerable<T> customization.\n" +
+              "Please use the following code snippet instead:\n" +
+              "fixture.Customizations.Add(" +
+              "  new FilteringSpecimenBuilder(" +
+              "    new FixedBuilder(" +
+              "      SEQUENCE_TO_RETURN)," +
+              "    new EqualRequestSpecification(" +
+              "      new MultipleRequest(" +
+              "        new SeededRequest(" +
+              "          typeof(T)," +
+              "          default(T))))));\n\n" +
+              "If you need that often, please create an extension method inside your project.")]
     public class MapCreateManyToEnumerable : ICustomization
     {
         /// <summary>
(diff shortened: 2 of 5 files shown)


Src/AutoFixture/Kernel/MultipleToEnumerableRelay.cs
```diff
@@ -9,6 +9,8 @@ namespace AutoFixture.Kernel
     /// </summary>
     /// <seealso cref="MultipleToEnumerableRelay.Create(object, ISpecimenContext)" />
     /// <seealso cref="MapCreateManyToEnumerable" />
+    [Obsolete("This relay is obsolete and will be removed in future versions. " +
+              "The reason is that it causes recursion overflow for many request if you forget to add IEnumerable<T> customization.")]
```
moodmosaic: request**s** (plural)

Src/AutoFixture/MapCreateManyToEnumerable.cs
```diff
@@ -25,6 +25,12 @@ namespace AutoFixture
     /// </para>
     /// </remarks>
     /// <seealso cref="MultipleToEnumerableRelay" />
+    [Obsolete("This relay is obsolete and will be removed in future versions. " +
+              "The reason is that it causes recursion overflow for many request if you forget to add IEnumerable<T> customization. " +
+              "Instead, please use the following code snippet:\r\n" +
+              "fixture.Customizations.Add(new FilteringSpecimenBuilder(new FixedBuilder(SEQUENCE_TO_RETURN), " +
+              "new EqualRequestSpecification(new MultipleRequest(new SeededRequest(typeof(T), default(T)))))).\r\n" +
+              "If you need that often, please create an extension method inside your project.")]
```
moodmosaic: It'd be nice if you format those code snippets to be <= 80 chars. Most output windows (and terminals) use that width by default and most chances are that the error message won't look pretty:

```c#
fixture.Customizations.Add(
    new FilteringSpecimenBuilder(
        new FixedBuilder(
            SEQUENCE_TO_RETURN),
    new EqualRequestSpecification(
        new MultipleRequest(
            new SeededRequest(
                typeof(T),
                default(T))))))
```

Src/AutoFixtureUnitTest/FixtureTest.cs
```diff
@@ -4904,6 +4904,25 @@ public void CustomizingEnumerableCustomizedCreateManyWhenCreateManyIsMappedToEnu
...
+        public void ItIsPossibleToMapManyRequestToCustomEnumerableInstance()
+        {
+            // Arrange
+            var sut = new Fixture();
+            var expected = new[] { "a", "b", "c", "d" };
+            sut.Customizations.Add(
+                new FilteringSpecimenBuilder(
+                    new FixedBuilder(expected),
+                    new EqualRequestSpecification(
+                        new MultipleRequest(new SeededRequest(typeof(string), default(string))))));
```
moodmosaic: `new SeededRequest` should be in its own line since all the other `new` calls are on their own lines too.

zvirja: Fixed!

Src/AutoFixture/MapCreateManyToEnumerable.cs
zvirja: Fixed!

Src/AutoFixture/Kernel/MultipleToEnumerableRelay.cs
zvirja: Thanks for spotting 😉

zvirja:
@moodmosaic Thanks for the review. I've addressed your remarks!
Ok, let's wait for @ploeh opinion on that.

ploeh:
I don't have any opinion on this, but keeping it simple sounds like a good motivation.

PR #1059, opened 2018-07-10
Add AsCustomization extension method

### LINKED ISSUE #1056: Add customization which appends builder to the `Customizations` collection
When you would like to build `CustomizeAttribute` instance (extension point for NUnit and xUnit), you should return the `ICustomization` instance:
```c#
public class CustomAttribute : CustomizeAttribute
{
    public override ICustomization GetCustomization(ParameterInfo parameter)
    {
        // Here I should return ICustomization
    }
}
```

I found that often all I need to do is to insert `ISpecimenBuilder` to the `fixture.Customizations` collection. However, to do that I need to create an extra class, implementing the `ICustomization` interface and adding my builder to the collection. It makes implementation polluted and it's hard to focus on the logic.

I suggest to add this type to the AutoFixture library, so we could write simpler customizations:
```c#
public class FortyTwo : CustomizeAttribute
{
    public override ICustomization GetCustomization(ParameterInfo parameter)
    {
        return new AddCustomBuilderCustomization(
            new FilteringSpecimenBuilder(
                new FixedBuilder(42),
                new EqualRequestSpecification(parameter)
            ));
    }
}
```

The only thing I'm unsure is the customization name. Usually we follow the convention to suffix customizations with `Customization` word, but here it might add noise. I see following candidates:
- `AddCustomBuilderCustomization`
- `AddBuilderToCustomizations`
- `AddBuilderToCustomizationsCustomization`
- `CustomizationAppender`
- `CustomizationAppenderCustomization`

None of the name looks great 😞 Also I don't like the "append" part, as later we might want to specify the position (first vs last), so append will be confusing.

@moodmosaic Do you see a nice name for this customization type?

zvirja:
Closes #1056

Add `AsCustomization` extension method which wraps the `builder` to the `ICustomization` which inserts the `builder` to the beginning of the `Customizations` collection.

### DIFF
--- a/Src/AutoFixture/CustomizationExtensions.cs
+++ b/Src/AutoFixture/CustomizationExtensions.cs
@@ -1,0 +1,39 @@
+using System;
+using AutoFixture.Kernel;
+
+namespace AutoFixture
+{
+    /// <summary>
+    /// A set of useful helpers to simplify work with fixture customizations.
+    /// </summary>
+    public static class CustomizationExtensions
+    {
+        /// <summary>
+        /// Create customization that inserts the <paramref name="builder"/> to the beginning of the
+        /// customizations collection.
+        /// </summary>
+        public static ICustomization ToCustomization(this ISpecimenBuilder builder)
+        {
+            if (builder == null) throw new ArgumentNullException(nameof(builder));
+
+            return new InsertCustomization(builder);
+        }
+
+        private class InsertCustomization : ICustomization
+        {
+            private readonly ISpecimenBuilder builder;
+
+            public InsertCustomization(ISpecimenBuilder builder)
+            {
+                this.builder = builder;
+            }
+
+            public void Customize(IFixture fixture)
+            {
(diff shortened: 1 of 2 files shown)


zvirja:
Renamed to `ToCustomization()`. I find this method more consistent in LINQ world.

moodmosaic:
There's a different in the semantics between *As()* and *To()*, that's why I picked *As()*. LINQ [uses both](https://msdn.microsoft.com/en-us/library/system.linq.parallelenumerable.asparallel.aspx).

zvirja:
@moodmosaic I also feel the difference between `To` and `As` semantics, and that's why I picked up `To`.. 😕  Could you explain your vision of the semantics?

Kralizek:
I see As being used in case of cast whilst To is used in case of conversion.

I'd say To is a better fit since you are wrapping the builder in some kind of façade and therefore instantiating a new object.

Issue #990, opened 2018-03-01
Specify desired amount of items when configuring a customization


Kralizek:
Hi,

I'm trying to test a piece of code that is sensible to the amount of items in a given set. Unluckily, I really can't avoid passing proper data to activate a specific path.

I would normally use `CreateMany<T>(int)` to specify the amount of items I need, but the set is in a property of a container object (btw, I am addicted to the new relay for `IReadOnlyList<T>`) like

```csharp
public class ContactIdList
{
    [JsonProperty("vids")]
    public IReadOnlyList<long> ContactIds { get; set; }

    [JsonProperty("vidOffset")]
    public long? Offset { get; set; }

    [JsonProperty("hasMore")]
    public bool HasMore { get; set; }
}
```` 

What I would suggest is to offer a way to specify the amount of items needed to be generated when customizing a type.

Ideally, something like this

```csharp
var list = fixture.Build<ContactIdList>()
                  .With(p => p.HasMore, false)
                  .Without(p => p.Offset)
                  .WithMany(p => p.ContactIds, 16)
                  .Create();
```

The parameters should be something like `(Expression<Func<T, IEnumerable<TProperty>> propertyPicker, int count)` but I'm honestly torn between naming it `WithMany<TProperty>` and adding another overload of `With<TProperty>`.

Obviously, I'm aware that I can create the list before and use `With<TProperty>(Expression<Func<T, TProperty> propertyPicker, TProperty value)` to pass the list.

zvirja:
Thanks for raising the scenario.

I do understand the scenario you have, however I don't like too much the suggested way to fix it. The scenario looks very specific to extend API for it.

Instead, it's much better to implement it as a part of the #708. If we add the `With()` method overload taking the factory method, so it can be solved like this:
```c#
fixture.Customize<ClassToCustomize>(c =>
    c.With(x => x.Values, () => fixture.CreateMany<int>(25).ToList())
);
```

This solution is way more generic as behavior is not limited to collections only.

@moodmosaic I'd be really happy to implement #708 as it's a real enabler for more specific scenario. Therefore, if you accept such a PR, I could implement it 😉

Kralizek:
@zvirja would it be ok to pass the fixture instance as argument to the factory method? That would allow us to write generator functions that don't depend on the captured context.

zvirja:
[...]

Issue #708, opened 2016-09-24
Add a factory method overload to the .With(...) customization method


abatishchev:
Hi,
Currently to assign property's value in runtime using a lambda we have to add the noise of `Without()` first:

``` csharp
var entity = fixture.Build<MyEntity>()
                    .Without(x => x.SomeProperty)
                    .Do(x => x.SomeProperty = i++)
                    .Create();
```

Without `Without` the lambda is ignored and random value is assigned instead.

Is it possible to drop `Without()`? Then it would be just:

``` csharp
var entity = fixture.Build<MyEntity>()
                    .Do(x => x.SomeProperty = i++)
                    .Create();
```

Much cleaner and shorter!

Real-world example:

``` csharp
fixture.Build<EarningLineItem>()
       .With(x => x.ProgramVersionId, programVersionId)
       .With(x => x.TransactionCurrencyCode, "USD")
       .With(x => x.EarningPeriodId, fixture.Create<int>())
       .With(x => x.UsagePeriodId, fixture.Create<int>())
       .With(x => x.EarningLineItemStatusId, EarningLineItemStatus.Eligible)
       .With(x => x.CorrelationId, Guid.NewGuid().ToString())
       .With(x => x.DataEarnDate, dataEarnData ?? DateTime.Now)
       .With(x => x.TenantId, Tenant.Dummy)
       .With(x => x.GLSplitCode, _glSplitCodes.Random())
       .With(x => x.PartnerType, fixture.Create<string>().Substring(0, 5))
       .With(x => x.AgreementNumber, fixture.Create<string>().Substring(0, 10))
       .With(x => x.RootAgreementNumber, fixture.Create<string>().Substring(0, 10))
       .With(x => x.ServiceCode, fixture.Create<string>().Substring(0, 5))
       .Without(x => x.TransactionDate).Do(x => x.TransactionDate = GenerateRandomDate())
       .Without(x => x.EarningFeeTypeId).Do(x => x.EarningFeeTypeId = (byte)fixture.Create<EarningLineItemFeeType>())
       .Without(x => x.PaymentCurrencyCode)
       .Without(x => x.PaymentAmount)
       .Without(x => x.PartnerProgramEnrollmentId).Do(x => x.PartnerProgramEnrollmentId = ppeIds[programVersionId].Random())
       .Without(x => x.CompanyCode).Do(x => x.CompanyCode = companyCodeIds[x.PartnerProgramEnrollmentId].Random());
[...]

PR #1072, opened 2018-10-17
[WIP] Add a factory method overload to the .With(...) customization method

Kralizek:
A work-in-progress PR to solve #708 and cases like #990

Ideally the new overload could be used like the following snippet.
```csharp
fixture.Customize<MyClass>(c =>
    c.With(x => x.Values, f => f.CreateMany<int>(25).ToList())
);
```

The PR is far from done, I'm just trying to wrap my head around what needs to go where.

I added a `PostprocessComposerExtensions` class in the `AutoFixture` namespace instead of `AutoFixture.Dsl` to facilitate the discovery of the extension methods.
Alternatively the file could be placed in the Dsl folder with the namespace not matching the folder. I guess it really is up to the maintainer!

The extension methods allow the user of the library to skip using the fixture to generate values.

```csharp
b.With(p => p.Property, () => DoSomething());

b.With<MyClass, MyPropertyType, string>(p => p.Property, (string a) => DoSomething(a));

b.With<MyClass, MyPropertyType, string, int>(p => p.Property, (string a, int b) => DoSomething(a, b));

b.With<MyClass, MyPropertyType, string, int, DateTime>(p => p.Property, (string a, int b, DateTime c) => DoSomething(a, b, c));

b.With<MyClass, MyPropertyType, string, int, DateTime, long>(p => p.Property, (string a, int b, DateTime c, long d) => DoSomething(a, b, c, d));
```

Note that type inference should kick in and allow the user to remove/skip the types from the method invocation. Something like

```csharp
b.With(p => p.Property, () => DoSomething());

b.With(p => p.Property, (string a) => DoSomething(a));

b.With(p => p.Property, (string a, int b) => DoSomething(a, b));

b.With(p => p.Property, (string a, int b, DateTime c) => DoSomething(a, b, c));

b.With(p => p.Property, (string a, int b, DateTime c, long d) => DoSomething(a, b, c, d));
```

### DIFF
--- a/Src/AutoFixture/Dsl/NodeComposer.cs
+++ b/Src/AutoFixture/Dsl/NodeComposer.cs
@@ -349,6 +349,19 @@ namespace AutoFixture.Dsl
             return (NodeComposer<T>)ExcludeMemberFromAutoProperties(member, graphWithProperty);
         }
 
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory)
+        {
+            var value = valueFactory();
+
+            return this.With(propertyPicker, value);
+        }
+
+        public IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory)
+        {
+
+        }
+
+
         /// <summary>
         /// Enables auto-properties for a type of specimen.
         /// </summary>
--- a/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
@@ -325,6 +325,19 @@ namespace AutoFixture.Dsl
                 when: n => n is NodeComposer<T>);
         }
 
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory)
+        {
+            var value = valueFactory();
+
+            return this.With(propertyPicker, value);
+        }
+
+        public IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory)
+        {
+
+        }
+
+
         /// <summary>
         /// Enables auto-properties for a type of specimen.
         /// </summary>
--- a/Src/AutoFixture/Dsl/IPostprocessComposer.cs
+++ b/Src/AutoFixture/Dsl/IPostprocessComposer.cs
@@ -66,7 +66,11 @@ namespace AutoFixture.Dsl
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "With", Justification = "Renaming would be a breaking change.")]
         IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, TProperty value);
 
-        /// <summary>
+        IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory);
+
+        IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory);
+
+            /// <summary>
         /// Enables auto-properties for a type of specimen.
         /// </summary>
         /// <returns>
--- a/Src/AutoFixture/Dsl/NullComposer.cs
+++ b/Src/AutoFixture/Dsl/NullComposer.cs
@@ -130,6 +130,17 @@ namespace AutoFixture.Dsl
             return this;
         }
 
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory)
+        {
+            return this;
+        }
+
+        public IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory)
+        {
+            return this;
+        }
+
+
         /// <summary>
         /// Does nothing.
         /// </summary>
--- a/Src/AutoFixtureUnitTest/Dsl/DelegatingComposer.cs
+++ b/Src/AutoFixtureUnitTest/Dsl/DelegatingComposer.cs
@@ -81,6 +81,17 @@ namespace AutoFixtureUnitTest.Dsl
             return this.OnWith(propertyPicker, value);
         }
 
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory)
+        {
+            return this.OnValueFactoryWith(propertyPicker, () => valueFactory());
+        }
+
+        public IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory)
+        {
+            return this.OnInputValueFactoryWith(propertyPicker, o => valueFactory((TInput)o));
+        }
+
+
         public IPostprocessComposer<T> WithAutoProperties()
         {
             return this.OnWithAutoProperties();
@@ -104,6 +115,8 @@ namespace AutoFixtureUnitTest.Dsl
         internal Func<IPostprocessComposer<T>> OnOmitAutoProperties { get; set; }
         internal Func<object, IPostprocessComposer<T>> OnAnonymousWith { get; set; }
         internal Func<object, object, IPostprocessComposer<T>> OnWith { get; set; }
+        internal Func<object, Func<object>, IPostprocessComposer<T>> OnValueFactoryWith { get; set; }
+        internal Func<object, Func<object, object>, IPostprocessComposer<T>> OnInputValueFactoryWith { get; set; }
         internal Func<IPostprocessComposer<T>> OnWithAutoProperties { get; set; }
         internal Func<object, IPostprocessComposer<T>> OnWithout { get; set; }
         internal Func<object, ISpecimenContext, object> OnCreate { get; set; }


Src/AutoFixture/Dsl/NodeComposer.cs
```diff
@@ -304,6 +304,37 @@ public IPostprocessComposer<T> OmitAutoProperties()
                 when: targetToDecorate.Equals);
         }
 
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<IFixture, TProperty> generator)
```
Kralizek: @zvirja I assume this is the place where most of the work should go. I hope I got it right :)

Src/AutoFixture/PostprocessComposerExtensions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Linq.Expressions;
+using AutoFixture.Dsl;
+
+namespace AutoFixture
+{
+    public static class PostprocessComposerExtensions
```
Kralizek: This class contains extension methods for providing additional functionalities.

The most important one is the one that offers the possibility to accept a parameterless function as a parameter.

```csharp
b.With(p => p.Property, () => DoSomething());
```

Additionally, overloads with up to 4 arguments are provided.

```csharp
b.With<MyClass, MyPropertyType, string>(p => p.Property, (string a) => DoSomething(a));
b.With<MyClass, MyPropertyType, string, int>(p => p.Property, (string a, int b) => DoSomething(a, b));
b.With<MyClass, MyPropertyType, string, int. DateTime>(p => p.Property, (string a, int b, DateTime c) => DoSomething(a, b, c));
b.With<MyClass, MyPropertyType, string, int. DateTime, long>(p => p.Property, (string a, int b, DateTime c, long d) => DoSomething(a, b, c, d));
```

Kralizek:
@zvirja sorry for not following [your suggestion](https://github.com/AutoFixture/AutoFixture/issues/708#issuecomment-430028361).

I find this way we can avoid adding 2 methods to the interface and eventually slim it by reimplementing `IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, TProperty value)` as
```csharp
public static IPostprocessComposer<T> With<TProperty>(
        this IPostprocessComposer<T> composer, 
        Expression<Func<T, TProperty>> propertyPicker,
        TProperty value)
{
    return composer.With(propertyPicker, f => value);
}
```

Being extreme, we could rewrite `IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker);` as 
```csharp
public static IPostprocessComposer<T> With<TProperty>(
        this IPostprocessComposer<T> composer, 
        Expression<Func<T, TProperty>> propertyPicker)
{
    return composer.With(propertyPicker, f => null);
}
```
and handle a `null` generator as a "give me anything" generator

zvirja:
Hi @Kralizek,

First of all thank you very much for taking a look at this - glad to see old good contributors making the product awesome 😊

I've reviewed the implementation draft and in general looks great! Would like to drop a few comments on that.

```c#
IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<IFixture, TProperty> generator);
```

This overload indeed looks universal, however it has a few major issues:
- The `IFixture` interface is very wide. It allows to do things which we don't want you to do in this scope (e.g. call `Customize()`, `Build()` API). While the common sense usually applies, we cannot rely that everybody has enough knowledge to understand what they are doing. More often people perform API exploration and try to call all the possible methods they have. We should have public API which guides users allowing them to perform only those actions which usually make sense in the current context.
- The `fixture.Create<T>()` method you use creates a new `ISpecimenContext` instance, which has different resolution root (especially valid in case of `Build<T>()` API usage) - normally we don't want to create new resolution scopes during the single request handle.
- This lookup is expensive, as you need to run the resolution request. Of course, it's very fast, but if we can avoid that, better do that.

To mitigate that I still suggest to expose the following API:
```c#
IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory);
```

To make it "universal", I'd do following:
- under the hood use the `BindingCommand` constructor overload taking the `Func<ISpecimenContext, TProperty> valueCreator` callback;
- allow to resolve `ISpecimenContext` as `TInput` (by simply returning argument from the callback);
- In the extension methods request `ISpecimenContext` and resolve the rest arguments you need using the `context.Resolve(typeof(TInput))` API.

I'd additionally expose the API without any input parameters:
```c#
IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory);
```

That is to make optimization and don't resolve any values if you don't need them.

Finally, as you stated above, we can re-implement the overload taking the `TProperty value` parameter as an extension method around overload without any input arguments.

I also wish to have as small public interface API as possible, but at the same time don't want to trade-off the performance. If we need extra method in public API to make code work faster - I'm completely up to have it.

Hope that makes sense to you!

Src/AutoFixture/Dsl/NodeComposer.cs
zvirja: I haven't tried to deeply dig into this method implementation (as the logic is really crazy here 😅), but consider a few things:
- I'd create a private method with `IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<ISpecimenContext, TProperty> generator)` signature. It should be an exact copy of the `public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, TProperty value)` method below. The only change is that you should pass callback to the `BindingCommand` constructor, rather than a scalar value.
- All the other `With()` implementations should delegate to it and either resolve `TInput` argument, or just return the fixed result.
- Don't activate any instances beforehand and rather activate them during the actual request (inside the callback you pass to the `BindingCommand`). That is because code might be invoked for a few times and you want to have different values for each time.

Src/AutoFixture/PostprocessComposerExtensions.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Linq.Expressions;
+using AutoFixture.Dsl;
+
+namespace AutoFixture
```
zvirja: Can we move it under the `Dsl` namespace? I believe that nowadays IDEs perform discovery well even for non-imported namespaces, so it should not be an issue.

Kralizek: Speaking about Visual Studio, whilst it's able to find an extension method if you use the correct types, I don't think that would be the case here because we use open types.

My fear is that these extension methods are difficult to be discovered for their "shape", from the user if not from the IDE. Having them readily available in the IntelliSense menu would mitigate the issue a bit.

Kralizek:
Hey! Thanks for your comments!

About the issues you highlighted,
* The interface is indeed very big (borderline a ISP violation), maybe we could extract Build and Create into a smaller interface that is then extended by IFixture? I am not sure if such a change is worth a major version change because only reflection-based usages would be broken.
* I found very dirty in resolving the IFixture myself to be able to use it. It was just for a quick proof of concept. Is there a way to simply forward to the delegate the current resolution scope?
* I expected the current fixture to be somehow available in the context, but maybe I failed to find it. 

About your other comments
* I'm not familiar with the interiors of autofixture, I really hacked things together to get things building and discuss method signatures and extension methods.  For some reason the delegate doesn't get invoked but I'll troubleshoot it next week.
* If we're not shy of extending the interface, I'd consider adding
  - Func<TProperty>
  - Func<TInput, TProperty>
  - Func<IFixtureSmall, TProperty> // I pulled the name out of my hat

My concern is really about discoverability, especially if we don't want people to use the full IFixture interface at postprocessing time, and resolving IFixture is as expensive as you say (I had no idea).

zvirja:
The `IFixtureSmall` interface is already present in the code and it's named `ISpecimenBuilder` 😄 It contains a single `Create()` method taking two arguments: `object request` and `ISpecimenContext context`. The issue is with the second parameter - this object is created once per whole object tree resolution and you don't have that object from the outside of the request. When you invoke `fixture.Create<T>()` or `specimenBuilder.Create<T>()` extension method, internally it creates a new context. That's the thing I really want to avoid in our current architecture as it might create pitfalls in future. I would like to keep the architecture simple and have only single context per the whole request.

Because of that we cannot expose neither `ISpecimenBuilder` nor `IFixture`. Ideally we could expose `ISpecimenContext`, however this interface is not designed for the external usage. It's pretty low-level and we don't have handy extension method infrastructure around it. Therefore, the best trade-off to me is to expose unbound overloads where you provide `T` you need (or none, or a few of them).

> specially if we don't want people to use the full IFixture interface at postprocessing time, and resolving IFixture is as expensive as you say (I had no idea).

I expect that those overloadы taking `T` input should cover the overwhelming majority of the use cases, so you never actually need the `IFixture` object. If you see the scenarios when `T` arguments are not enough and you require `IFixture`, please post them - will be quite interesting to see them.

Kralizek:
@zvirja thank you for your reply!

I'd be more interested into exposing `ICustomizationComposer<T> Build<T>();` in the "so called" `IFixtureSmall`. #990 is a good sample of where being able to use the Build/Create chain in a `With` call could be useful.

zvirja:
I'm not sure I fully got the context you are describing 😕 

In this PR we are working on the `ICustomizationComposer<T>` chain and the `With<T>()` callback we have in it. It will be used like you described above:

```
fixture.Build<T>
    .With<T>(x => x.Property, (int number) => number.ToString())
    .Create();
```

Why would you need the `Build<T>()` method inside the `With()` callback? 
image

----

I suggest us to stop on the following `With()` method overloads, as I suggested [above](https://github.com/AutoFixture/AutoFixture/pull/1072#issuecomment-431162349):
```c#
IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, Func<TInput, TProperty> valueFactory);
IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, Func<TProperty> valueFactory);
```

I don't see any good reason why you need `IFixture` instance inside the `valueFactory` callback. If you see it - please describe.

Kralizek:
As I mentioned before, this PR would help me solve the issue I have in #990 where I would like to do `fixture.Build<Something>().With(p => p.SomeStringArrayProperty, f => f.CreateMany<string>(25)).Create()`.

zvirja:
Oh, it seems I missed that point. For such cases we have the `Generator<T>` util, which is an infinite enumerable and allows to you generate custom sequences of variable length:

```c#
fixture.Build<Something>()
  .With(p => p.SomeStringArrayProperty, (Generator<string> strings) => strings.Take(25).ToArray())
  .Create();
```

Kralizek:
@zvirja My only blocker is the real life. I've been working on the budget for 2019 for my team. Trust me, I'd have rather worked on this PR a thousand times! 😉

Kralizek:
@zvirja pushed the new version of the methods as you suggested.

I'd need some help with [this method](https://github.com/Kralizek/AutoFixture/blob/94910947ede7e82bc7bd9d91bf1751a568eb5eac/Src/AutoFixture/Dsl/NodeComposer.cs#L359-L362) since it seemed you had a good idea on how to implement it.

zvirja:
Hi @Kralizek,

Thanks for your help on this. I found some free time this evening and implemented the feature in #1084. Was hugely inspired by our discussions and they helped me a lot! Thanks for driving the change 🤝 

Closing this one in favor of #1084.

PR #1075, opened 2018-10-31
- add ability to set properties with private set

NooNameR:
Add ability on build set Properties with private set

### DIFF
--- a/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
@@ -18,10 +18,11 @@ namespace AutoFixture.Dsl
     /// in the wrapped graph and applies the appropriate method to all matches.
     /// </para>
     /// </remarks>
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "The main responsibility of this class isn't to be a 'collection' (which, by the way, it isn't - it's just an Iterator).")]
-    public class CompositeNodeComposer<T> :
-        ICustomizationComposer<T>,
-        ISpecimenBuilderNode
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification =
+        "The main responsibility of this class isn't to be a 'collection' (which, by the way, it isn't - it's just an Iterator).")]
+    public class CompositeNodeComposer<T>
+        : ICustomizationComposer<T>,
+            ISpecimenBuilderNode
     {
         /// <summary>
         /// Initializes a new instance of the
@@ -115,8 +116,7 @@ namespace AutoFixture.Dsl
         /// An <see cref="IPostprocessComposer{T}"/> which can be used to
         /// further customize the post-processing of created specimens.
         /// </returns>
-        public IPostprocessComposer<T> FromFactory<TInput>(
-            Func<TInput, T> factory)
+        public IPostprocessComposer<T> FromFactory<TInput>(Func<TInput, T> factory)
         {
             return (CompositeNodeComposer<T>)this.ReplaceNodes(
                 with: n =>
@@ -145,8 +145,7 @@ namespace AutoFixture.Dsl
         /// An <see cref="IPostprocessComposer{T}"/> which can be used to
         /// further customize the post-processing of created specimens.
         /// </returns>
-        public IPostprocessComposer<T> FromFactory<TInput1, TInput2>(
-            Func<TInput1, TInput2, T> factory)
+        public IPostprocessComposer<T> FromFactory<TInput1, TInput2>(Func<TInput1, TInput2, T> factory)
         {
             return (CompositeNodeComposer<T>)this.ReplaceNodes(
                 with: n =>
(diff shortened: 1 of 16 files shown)


zvirja:
@NooNameR Thanks for working on this feature and spending your time! Happy to see new contributors 😉 

The private properties is quite a hot topic where people have different opinions on that. There are reasonable arguments on the both sides and it's a matter of experience and views you are trying to align with.

Currently the library is following a philosophy (at least that's how I see it) that Unit Test is also a code client, like any other production code you have. This vision is inspired by @ploeh's posts like [this](http://blog.ploeh.dk/2011/11/10/TDDimprovesreusability/). With that vision you shouldn't write to private properties in your tests - you wouldn't normally do that in the production code. This side has its own pros and cons, but currently we are following those rules.

The patch you suggest changes the philosophy and is more fundamental, than just ability to write to protected members. It will change a spirit of the library and might foster our clients to change their design approaches. Personally, I'm not sure that this change will be for the better.

With that in mind, I don't want to bring the suggested API to the core.

However, as a developer, I want our product be flexible enough to be used for variable purposes the clients might need. Therefore, I don't mind if the library contains an API to implement the similar features relatively easy, without hardcore diving into internal architecture. I admin that current design of `Build<T>()` API is not very friendly to external customization (or we just need to look at it from the proper angle 😉). This way you will be enabled to implement this feature in your project, if you and/or your team have different design preferences.

If the latter is still acceptable for you, we can discuss that after #1072 is merged (as it will change API a bit).

@moodmosaic @ploeh I'd appreciate if you could also provide your expertise on that. This way I can be more sure that I'm not mistaken and my position makes sense. Thanks!

ploeh:
I've handed over AutoFixture, so my opinion is just that: _an opinion_.

On the other hand, I've been doing TDD since 2004, and I no longer want to pretend to offer a balanced view on things. It may seem mature to say that there's pros and cons (and yes: _that's true_) but then, I'm post-mature.

As a comparison, I'm sure one can come up with pros and cons for writing assembly code in 2018, yet almost no-one does that. High-level languages are overwhelmingly better in all but extremely narrow niche scenarios. There's really no discussion left on that topic.

The same with the discussion about [unit testing internals](http://blog.ploeh.dk/2015/09/22/unit-testing-internals). I don't find it productive to pussyfoot around that discussion. It's just an overwhelmingly bad idea to unit test internals.

Kralizek:
I'd just add my 2 unrequested cents to the topic.

Often time we find ourselves having to test code using libraries that are not ours and therefore that we have no control upon.

Sometimes it happens that the aforementioned library is not designed with testing in mind and we are left with the short straw and the need to figure out how to write "good" tests.

Is it best practice to work with internals? Definitely not, yet sometimes we got to do it.

This is why I believe AutoFixture should provide a way to set non-public properties. I reckon I am biased because I also believe that AutoFixture could provide more out-of-the-box via extension packages.

That being said, there are many ways that it can be done. Moq supports the mocking of protected methods via a special API that you turn on only by importing a special namespace.

I'd personally go that way rather than polluting the main API.

ploeh:
> Often time we find ourselves having to test code using libraries that are not ours and therefore that we have no control upon.

Please see [GOOS](http://bit.ly/growingoos), particularly chapter 8.

ploeh:
Chapter 8 basically says _use Adapters_, or what [DDD](http://amzn.to/WBCwx7) calls an _anti-corruption layer_. The entire philosophy of the book, though, is that tests drive the design, which includes the choice of third-party libraries, frameworks, and tools. If those don't support testing, you choose something else.

I understand that this may not be possible in brownfield code, but I never designed AutoFixture to explicitly support brownfield development. Whether or not it should do so now, I'll leave up to the current maintainers...

zvirja:
Thanks guys for the follow-up and your thoughts on that! Appreciate that a lot! 👍

Given that @moodmosaic hasn't posted his opinion, I'll need to make the decision on my own.

My vision on the situation remains the same as it was initially:

> With that in mind, I don't want to bring the suggested API to the core.
> ...
> However, as a developer, I want our product be flexible enough to be used for variable purposes the clients might need. Therefore, I don't mind if the library contains an API to implement the similar features relatively easy, without hardcore diving into internal architecture. I admin that current design of Build<T>() API is not very friendly to external customization (or we just need to look at it from the proper angle 😉). This way you will be enabled to implement this feature in your project, if you and/or your team have different design preferences.

As a library maintainer I would like to stick to the current design and don't support private properties out of the box. I like spirit of the library we currently have and the code design the library enforces. I do understand that sometimes cruel reality might not suit very well into that, and those exceptional cases should be covered by customizations on project-to-project basis.

Huge thanks again @NooNameR for the time you spent on that and I'm sorry that I cannot accept the change you did 🙏 Hope that will not prevent you from future contributions!

P.S. Just a hint - if you are in doubt regarding the feature you would like to suggest, it might worth it to create an issue and discuss that beforehand, so you don't spend too much time.

djrhodes:
A bit late to the party and appreciate the issues is now closed, but in my instance we have a framework for binding POCOs to .net core configuration classes. As part of the binding process we perform some validation (for required fields, correct patterns etc). The POCOs are created as immutable and then are registered in DI and passed around as needed.

To test the validation, we are testing the public API - the Validate() method of the POCO - but to do this we need to create an instance of the POCO - but the setters are private. There are two approaches I can use :
1. Set up test config files and load them into config via the json provider for the test. This is fine, but is a bit cumbersome and also needs maintenance to keep the test config in sync with the POCO
2. Set up a POCO, then serialise it to json and load it  using the in memory provider. This is preferable as I don't need to create seperate config files for each validation scenario and also there's no maintenance since we're creating the config from the POCO. This is where I would like to use AutoFixture to set up the POCO by calling the private setters.

This seems a legitimate use case to me since:
1. I'm testing the public API (just requiring access to private members to do the arrange phase)
2. Following best practices for config POCOs by having them immutable.

Would be interested to hear any thoughts

PR #1079, opened 2018-11-22
Add compatibility with NSubstitute v4

zvirja:
NSubstitute v4 contains huge amount of binary breaking changes. Luckily, we don't use their API very intensively, so faced a couple of them only. Added binary compatibility shim to mitigate that. It's a good sign that in our next major versions it might make sense to consider dropping support of older versions...

While RC is released only, it's very unlikely they NSubstitute team is going to change API shape (I know that as I'm a maintainer of NSubstitute now and in fact all those breaking changes were done by me 😊). So I think it's right time to start working on support.

@moodmosaic Please take a look. 😉 It's my first PR since long pause.

### DIFF
--- a/Src/AutoNSubstituteUnitTest/NSubstituteRegisterCallHandlerCommandTest.cs
+++ b/Src/AutoNSubstituteUnitTest/NSubstituteRegisterCallHandlerCommandTest.cs
@@ -10,7 +10,7 @@ using Xunit;
 
 namespace AutoFixture.AutoNSubstitute.UnitTest
 {
-    public class NSubstituteRegisterCallHandlerCommandTest
+    public partial class NSubstituteRegisterCallHandlerCommandTest
     {
         [Fact]
         public void ShouldReturnValuesFromConstructor()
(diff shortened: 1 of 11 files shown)


Src/AutoNSubstituteUnitTest/NSubstituteRegisterCallHandlerCommandTest.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace AutoFixture.AutoNSubstitute.UnitTest
 {
-    public class NSubstituteRegisterCallHandlerCommandTest
+    public partial class NSubstituteRegisterCallHandlerCommandTest
```
moodmosaic: You *really* need `partial` classes?

moodmosaic:
:100: :confetti_ball:

Src/AutoNSubstituteUnitTest/NSubstituteRegisterCallHandlerCommandTest.cs
zvirja: Well, it might be rewritten without `partial`, but here it's very convenient usage. It's used only for compatibility tests, so to me it suits perfectly 😊 

C'mon, I learned about that feature from the very first day and all this time I waited for the chance to finally use it 😅  Let me do it at least once in my life 😊

moodmosaic: I mean, yeah, go for it, but I don't think it's reasonable as he or she who reads the code must remember to jump into the other part(s) of the partial type definition.

PR #1081, opened 2018-11-23
Fix missing support for generic method implementation in GuardClauseAssertion

frblondin:
When trying to implement generic methods, GuardClauseAssertion was not adding generic parameters and their contraints.

For example, the following interface:
```
public interface ISomeInterface
{
    ...

    TResult GenericMethod<TValue, TResult>(TValue argument)
        where TValue : ICloneable
        where TResult : class, ICloneable;
}
```

... was throwing the following error: `Method 'GenericMethod' in type 'ValueType5638d03332334b2b9d3caa1b60c1bf67' from assembly 'AutoFixture.DynamicProxyAssembly, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.`

The reason was that the emitted method did not have the same signature as the implemented interface.

This pull request intends to emit the same signature.

### DIFF
--- a/Src/Idioms/GuardClauseAssertion.cs
+++ b/Src/Idioms/GuardClauseAssertion.cs
@@ -821,6 +821,7 @@ See e.g. http://codeblog.jonskeet.uk/2008/03/02/c-4-idea-iterator-blocks-and-par
             private void ImplementMethod()
             {
                 this.DefineMethodBuilder();
+                this.DefineMethodGenericParameters();
                 this.DefineStaticSpecimenBuilderFieldBuilder();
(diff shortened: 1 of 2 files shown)


zvirja:
@frblondin Thanks for the fix!

zvirja:
@frblondin Yes, the MyGet feed is no longer in use, as we are not working on next major version ATM. I've just released a new version of AutoFixture, so feel free to take it from the NuGet feed 😉 

Thanks again for your contribution!

frblondin:
Neat, thank you @zvirja

PR #1084, opened 2018-12-07
Add With() overload taking factory method

### LINKED ISSUE #708: Add a factory method overload to the .With(...) customization method

zvirja:
Closes #708

Added overloads for the `With()` method taking action as a parameter. The following overloads were added:
```c#
With(x => x.Property, () => MakeValue());
With(x => x.Property, (object someInput) => MakeDesiredValueUsingInput(someInput))
```

Usually we have overloads [taking up to 4 input values](https://github.com/AutoFixture/AutoFixture/blob/c969f47c05e12abaadac44b8e9a6221abaa8eee5/Src/AutoFixture/Dsl/IFactoryComposer.cs#L152). Decided to not add those for now, as it's unclear whether they will be demanded in future.

Additionally, added a few more changes:
- deprecated the `CompositePostprocessComposer` as it's not used in our code
- used `interitdoc` whenever possible to not duplicate the parent documentation
- Reworked `With()` overload without parameters to unify graph when we perform `With()` change. The observable behavior hasn't changed and no tests failed, so I presume change might be considered as safe.

@moodmosaic Happy review!
CC @Kralizek

### DIFF
--- a/Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
@@ -10,6 +10,8 @@ namespace AutoFixture.Dsl
     /// Aggregates an arbitrary number of <see cref="IPostprocessComposer{T}"/> instances.
     /// </summary>
     /// <typeparam name="T">The type of specimen to customize.</typeparam>
+    [Obsolete("This class is deprecated and will be removed in future versions of AutoFixture. " +
+              "Please file an issue on GitHub if you need this class in your project.")]
     public class CompositePostprocessComposer<T> : IPostprocessComposer<T>
     {
         /// <summary>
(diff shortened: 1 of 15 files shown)


Kralizek:
Oh wow! I don't think I would have managed to implement this! Great job!

As for the overloads with multiple parameters, maybe 3 extension methods that request for a tuple behind the scenes?

Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
```diff
@@ -10,6 +10,8 @@ namespace AutoFixture.Dsl
     /// Aggregates an arbitrary number of <see cref="IPostprocessComposer{T}"/> instances.
     /// </summary>
     /// <typeparam name="T">The type of specimen to customize.</typeparam>
+    [Obsolete("This class is deprecated and will be removed in future versions of the product. " +
```
moodmosaic: "of AutoFixture" I'd rather say instead "of the product"

moodmosaic:
:+1: Looks reasonable. I would say go for it.

Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
zvirja: Done, thanks for advise 😉

zvirja:
@moodmosaic Merged. Thanks for the review 🤝

Kralizek:
@zvirja 
> Usually we have overloads taking up to 4 input values. Decided to not add those for now, as it's unclear whether they will be demanded in future.

I would suggest the extension method way for this kind of scenario.

After all, considering that the parameters are provided by us, we can say that generating N values to be fed into a delegate is equivalent to generate a tuple with N values and feed into the same delegate.

```csharp
// These two are equivalent
(T1 arg1, T2 arg2, T3 arg3, T4 arg4) => something(arg1, arg2, arg3, arg4)
((T1, T2, T3, T4) arg) => something(arg.Item1, arg.Item2, arg.Item3, arg.Item4)
```

So we could have something like

```csharp
public static IProcesscomposer<T> With<TProperty, TInput1, TInput2, TInput3, TInput4>(
    this IPostprocessComposer<T> composer,
    Expression<Func<T, TProperty>> propertyPicker,
    Func<TInput1, TInput2, TInput3, TInput4, TProperty> valueFactory)
{
    return composer.With(propertyPicker, ((TInput1 a, TInput2 b, TInput3 c, TInput4 d) arg) => valueFactory(arg.a, arg.b, arg.c, arg.d));
}
```

zvirja:
Hello @Kralizek,

Thanks for your input! It's a brilliant idea to implement extension method and relay request to a tuple! Thanks for that 👍 

I don't mind to have overloads taking a few arguments. I just don't want to implement them upfront as they might be never required. As soon as we get them requested in future for a few times, we might add them as well.

Do you find them required _in your projects_?

PR #1088, opened 2018-12-18
Support FakeItEasy 5.0.0+

blairconrad:
Moves FakeItEasy version restrictions up to `, 6.0.0)` and adds AutoFakeItEasy.FakeItEasy5UnitTest to ensure the integration works.

PR #1090, opened 2018-12-18
Build AutoFakeItEasy .Net Standard 2.0 package

blairconrad:


Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
```diff
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="FakeItEasy" Version="[4.0.0]" />
+    <PackageReference Include="FakeItEasy" Version="[4.9.2]" />
```
blairconrad: Not sure how you feel about this change. Me, I'd prefer all the tests to run against the latest release within the major version range, but I think this is only my opinion. In this case, the .NET Standard 2.0 package for FakeItEasy is not available until 4.9.0.

zvirja: Hi Blair,

Glad to see you 😊

I don't have strong opinion on that. Lower possible version is good because usually features are added up during a version evolution, so it could happen that we test functionality/behavior which is not available (or is different) for the lower versions.

On other hand, clients usually install the latest major version, which is especially true for the ancient versions.

To make it simple, let's prefer the consistency across the solution and not having exceptions in the particular projects. So may I ask you to please add a condition here and target `4.9.2` for `netstandard2.0` only?

If in future I end up with a different strategy, I'll update all the project at once, including this one.

Thank you a lot for your contribution! 🥂

Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
```diff
@@ -17,7 +17,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="FakeItEasy" Version="[4.0.0]" />
+    <PackageReference Include="FakeItEasy" Version="[4.0.0]" Condition=" '$(TargetFramework)'=='net452' " />
+    <PackageReference Include="FakeItEasy" Version="[4.0.0]" Condition=" '$(TargetFramework)'=='netcoreapp1.1' " />
+    <PackageReference Include="FakeItEasy" Version="[4.9.0]" Condition=" '$(TargetFramework)'=='netcoreapp2.0' " />
```
blairconrad: Hey, @zvirja!
You'd asked for 4.9.2 (which I can still do; just say the word) but 4.9.0 is the minimum, and I thought it would be more consistent to use that version.
Suits?

zvirja: @blairconrad Sure, I just messed up the things, sorry 😖 Please go with `4.9.0`.

blairconrad: I don't think _you_ messed the things up. I messed the things up (by deviating from your practices) and you are putting me back on track!

zvirja: Well, I'm not taking the whole responsibility😅 My contribution is limited by `4.9.2` vs `4.9.0` only, while all the rest is still supposed to be caused solely by you 🤣 

Merged the code, so you are enabled in #1088.

Good night 🌔

blairconrad: Thanks. g'night!

PR #1103, opened 2019-02-14
Update README.md

moodmosaic:
Replace old (*very* old, actually) location of @moodmosaic blog with the current one. The reason is, if I add a new AutoFixture post, it will show up in the new location, but won't show up in the old one.

---

(Sure I could add a redirect, but I just consider web to be currently broken anyways, [unless](https://ipfs.io/) ...[etcetera](https://ipfs.io/ipfs/QmNhFJjGcMPqpuYfxL62VVB9528NXqDNMFXiqN5bgFYiZ1/its-time-for-the-permanent-web.html)). :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ If you want to add your own testimonial to this list, we (the AutoFixture mainta
 
 * [Pluralsight course](https://www.pluralsight.com/courses/autofixture-dotnet-unit-test-get-started)
 * [ploeh blog](http://blog.ploeh.dk/tags/#AutoFixture-ref)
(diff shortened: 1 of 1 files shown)


zvirja:
Merged, thanks 👍
Instead of reading the PR, you made me spend half of the evening reading about IPFS :D

Kralizek:
I ended up doing the same 🤣

moodmosaic:
>reading about IPFS

Then you might also want to look into [this](https://ethereum.stackexchange.com/questions/375/what-is-swarm-and-what-is-it-used-for/388#388). :wink:

PR #1109, opened 2019-04-14
Added support for XUnit-ClassDataAttribute

muncinsky:
`ClassDataAttribute` was again added to XUnit2, [more details here](https://github.com/xunit/xunit/issues/359). This PR adds support for it by implementing new `ClassAutoDataAttribute`, so we can combine data comming from Autofixture with data comming from given Class of `ClassDataAttribute`.


zvirja:
@muncinsky Good day! Finally I got a chance to review your PR!

First of all many thanks for working on it and helping to make our AutoFixture library better :+1: It's awesome! Also thanks for the nice test coverage! I definitely see how this newly added attribute could make writing unit tests simpler.

I was thinking about that and a few things came up to my head. Let me share them with you, so we can together decide how to proceed further.

# SemVer and breaking changes

Current project is following [Semantic Versioning](https://semver.org/), meaning that breaking changes are not allowed withing the same major version. While the policy mostly focuses on API, I consider minimum supported version bump a breaking change as well. There are might be clients stuck with older versions of xUnit and they might get confused when they find they cannot use our next _minor_ version. Same will happen with our dependent libraries, which also expects us to not change dependencies.

# InlineAutoData

As of today, AutoFixture library ships `InlineAutoData` attribute, which is not covered. If we want to support `ClassData`, we need to decide what to do with `InlineAutoData` attribute as well.

# Other xUnit data attributes

`ClassData` attribute is only one among other `DataAttributes` shipped with xUnit. For instance, there is `MemberData` as well, which will be not covered. If we decide to support it in a same way, it might be a trap we are preparing for ourselves - later xUnit might add new attributes and we'll need to replicate those ones inside our library for consistency. That is not very generic approach, unfortunately.

Also we should consider the `InlineAutoData` attribute I mentioned above. Most likely there should be a version for that attribute as well, which will double amount of required classes :worried:

# Extensibility

From my experience `AutoData` and `InlineAutoData` are rarely used as is. It's because often the domain we are testing requires us to add some customizations (e.g. configure how to construct some specific types). Or it might be integration with mocking libraries like `Moq` or `NSubstitute`. Any way, it requires us to create our own data attribute derived from `AutoData`/`InlineAutoData` and use them instead.

Given this aspect the newly introduced `AutoClassDataAttribute` will have more limited applicability. Instead, people will need to customize it, so they will derive from it. If we consider a fact that `ClassData` is used in very rare cases by itself and the implementation of `AutoClassDataAttribute` is very primitive, it looks like it doesn't make a big difference whether you derive from `AutoClassDataAttribute` attribute, or directly from the `CompositeData`.

--------------------

Taking all these facts into account, I don't see a clear way how to proceed further. On one hand, I understand a motivation you have and the problem you are trying to solve. OTOH, probably we should re-consider the approach to make it more generic and solve all the concerns above.

@muncinsky Given that you already met this problem in your solution and likely played with it for some time already, do you see a possible way for us to go? :wink: Thanks!

muncinsky:
Hi @zvirja, thank you for your comprehensive review. Regarding your points:

**SemVer and breaking changes**
For this I have no other solution than conditionaly compile with `#if NETSTANDARD1_5`. When targeting net standard xunit.extensibility.core package is already conditionally restored with sufficient version 2.2.0. Not an ideal solution as it wouldn't be available for net452, but I have no other idea.

**InlineAutoData**
Here I don't understand what did you mean by "is not covered". XUnit comes with `InlineDataAttribute` and there is already related `InlineAutoDataAttribute` in AutoFixture. So I would say it is already covered. This brings some inconsistency, because `InlineDataAttribute` and  `ClassDataAttribute` are at the same hiearchy level.

**Other xUnit data attributes**
Regarding `MemberData`, I missed this one, but good idea we can support it as well :). I don't believe there might come something new in the future. Here we are talking about parametrized tests. We have just `InlineData` for primitive types, then `ClassData`and  `MemberData` for complex types and that's it.

**Extensibility**
The problem is that developers new to AutoFixture having XUnit background, they don't know they could create something derrived from `CompositeDataAttribute`. Instead they are searching for something related to their well known `ClassDataAttribute`, because for `InlineDataAttribute` there is already  `InlineAutoDataAttribute`.

Even though you are right in the fact, that developer most probably would create new derrived attribute, I disagree it would have limited applicability. Here we must separate two sets of types arranged for the test. First set includes types like mocked services and sut common for multiple tests, that can be arranged inside of derrived attribute. Second set includes data arranged for different scenarios of parametrized test, these data probably won't need to utilize features of Autofixture, but instead are manually specified in `ClassData`. In the end it means that I may have just one derived attribute and use it for multiple tests, but  with different `ClassData`, so I don't have to create new attribute for each and every test.

Regarding the generic approach, we can create one attribute for parametrized test which covers both primitive and complex types, but given the fact we have already  `InlineAutoDataAttribute`, it doesn't seem to be a consistent solution.

Thank you for your time

zvirja:
@muncinsky Many thanks for sharing your opinion - I appreciate that. Your explanation sounds reasonable and I see your point. At the end of the day we should respect the fact that it simplifies life for the end users and it's a tool for users, not the other way around.

The only thing which looks a bit ugly is that we cannot offer same the API for .NET Framework users 😟 

What if implement the `MemberData` attribute instead? I might be good substitute for the most scenarios. Later we might target `ClassData` in our next major version, when we have a chance to bump the dependency version.

Would this trade-off work for you? If not, could you please share your class with data, so I can better understand why you cannot rewrite it to member?

Thanks for your cooperation! 🤝

muncinsky:
@zvirja, yes `MemberData` should be sufficient for me. Should I create a new PR and close this one? 

Thanks!

zvirja:
Yes, let’s do that via new PR.

PR #1131, opened 2019-06-26
Allow factory methods returning interface to be selected

rexcfnghk:
This MR enhances `FactoryMethodQuery` to allow factory methods returning an interface to be selected.

### DIFF
--- a/Src/AutoFixture/Kernel/FactoryMethodQuery.cs
+++ b/Src/AutoFixture/Kernel/FactoryMethodQuery.cs
@@ -38,7 +38,7 @@ namespace AutoFixture.Kernel
             if (type == null) throw new ArgumentNullException(nameof(type));
 
             return from mi in type.GetTypeInfo().GetMethods(BindingFlags.Static | BindingFlags.Public)
-                   where mi.ReturnType == type &&
+                   where mi.ReturnType.GetTypeInfo().IsAssignableFrom(type) &&
                          !string.Equals(mi.Name, "op_Implicit", StringComparison.Ordinal) &&
                          !string.Equals(mi.Name, "op_Explicit", StringComparison.Ordinal)
                    let parameters = mi.GetParameters()
--- a/Src/AutoFixtureUnitTest/Kernel/FactoryMethodQueryTest.cs
+++ b/Src/AutoFixtureUnitTest/Kernel/FactoryMethodQueryTest.cs
@@ -86,5 +86,20 @@ namespace AutoFixtureUnitTest.Kernel
             // Assert
             Assert.Empty(result);
         }
+
+        [Fact]
+        public void
+            SelectMethodsFromTypeWithInterfacePropertyReturnsNonEmptyResult()
+        {
+            // Arrange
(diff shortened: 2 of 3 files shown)


zvirja:
Thanks for the contribution. I've reviewed the change, but, unfortunately, it leads to the following regression:

```c#
    public class Playground
    {

        [Fact]
        public void ShouldCreateInstance()
        {
            // Arrange
            var fixture = new Fixture();
            
            // Act
            var result = fixture.Create<SomeImplementation>();
            
            // Assert
            Assert.IsType<SomeImplementation>(result);
        }

        public interface ISomeInterface
        {
        }

        public class SomeImplementation : ISomeInterface
        {
            public static ISomeInterface CreateInterface () => new AnotherImplementation();
            
            private SomeImplementation() {}
        }

        public class AnotherImplementation : ISomeInterface
        {
        }
    }
```

Therefore, it's unsafe to make this change. Could you please highlight which issue you are trying to solve?

Thanks!

rexcfnghk:
Thanks for the reply. I was actually trying implement a singleton for a particular interface:

```c#
public interface IFoo
{
    void DoSomething();
}

public class SqlFoo : IFoo
{
    private readonly string _connectionString;

    public void DoSomething
    {
        using (var sqlConnection = new SqlConnection(_connectionString))
        {
            // Do something
        }
    }
}

// Singleton implementation of IFoo
public class SingletonFoo : IFoo
{
    private SingletonFoo() { }

    public static IFoo Instance { get; } = new SingletonFoo();
}

// Unit test, expecting Singleton.Instance to be called
[Theory, AutoData]
public void IsAssignableFromIFoo(SingletonFoo sut)
{
    Assert.IsAssignableFrom<IFoo>(sut);
}
```
But I guess I can always change the return type from `IFoo` to `SingletonFoo` to make the unit test pass.

zvirja:
Actually, the cast from `SingletonFoo.Instance` to `SingletonFoo` is unsafe, as there is no type guarantee that objects are convertible. Therefore, tools like AutoFixture should definitely prevent from doing that _implicitly_. Instead, you can easily customize AutoFixture to teach it _explicitly_ to do it:

```c#
// Unit test, expecting Singleton.Instance to be called
[Theory, AutoCustomData]
public void IsAssignableFromIFoo(SingletonFoo sut)
{
	Assert.IsAssignableFrom<IFoo>(sut);
}

public class AutoCustomData : AutoDataAttribute
{
	public AutoCustomData() : base(() =>
	{
		var fixture = new Fixture();
		fixture.Register<SingletonFoo>(() => (SingletonFoo) SingletonFoo.Instance);
		return fixture;
	})
	{
	}
}
```

Alternatively, if this pattern is common for your codebase, just add your own customization to `fixture.Customizations` to intercept all type requests when static `Instance` property exists.

You could also get inspiration from [this infrastructure](https://gist.github.com/zvirja/9366f23e28dd88e637d32bb52c025542) if you have an often need to apply test-specific customizations.

Closing this one as change is unsafe for average codebase. Thanks for the time spent on this!

PR #1137, opened 2019-09-18
automoq support for generic methods

riezebosch:
https://github.com/Moq/moq4/wiki/Quickstart#matching-generic-type-arguments-moq-413

### DIFF
--- a/Src/AutoMoq/AutoMoq.csproj
+++ b/Src/AutoMoq/AutoMoq.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\Common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netstandard1.5;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
     <AssemblyTitle>AutoMoq</AssemblyTitle>
     <AssemblyName>AutoFixture.AutoMoq</AssemblyName>
     <RootNamespace>AutoFixture.AutoMoq</RootNamespace>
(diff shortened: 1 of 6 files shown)


Src/AutoMoq/AutoMoq.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\Common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netstandard1.5;netstandard2.0</TargetFrameworks>
```
riezebosch: moq no longer supports `netstandard1.5` apparently: https://www.nuget.org/packages/Moq/4.13.0

zvirja:
Hi @riezebosch,

Thank you very much for the contribution to the library and your wish to make it better! 👍 

I've reviewed your changes and I found the following issues with the changes you suggest:
1. The `IsAnyType` util might be not compatible with the `T` type due to constrains. Consider the following signature:
	```c#
	int Get<T>() where T : IEnumerable<int>;
	```
    If you substitute `IsAnyType/IsSubtype/IsValueType` type matcher, you'll get runtime exception.

2. AutoFixture library is following [SemVer](https://semver.org/). Therefore, we are not allowed to apply breaking changes withing the same major version. In this PR you are changing the minimum required Moq version to `4.13.0`, meaning that users could get compilation issues if they still use older version. Practically, it means that we can change dependency versions only when we release next major version.

3. > moq no longer supports netstandard1.5 apparently:

    That is already handled by the conditional dependency versions. Check dependencies of the latest `AutoFixture.AutoMoq` package [here](https://www.nuget.org/packages/AutoFixture.AutoMoq/4.11.0). Most likely we'll drop netstandard 1.x support in our next major version, but unfortunately we cannot do that sooner 😟 

I've created issue #1139 to rework our integration approach, which might solve the problem in more elegant manner.

At this moment I suggest to close the PR as I cannot see what we can do giving the constrains we have.

Please don't let this rejection prevent you from further contributions to the library and sorry for how it happens 😟 

Let me know if I overlooked something and if I'm wrong somewhere in my assumptions.

riezebosch:
Value provider looks promising and handles the case also without requiring to update the dependency and generic type constraint. See my contribution in that issue.

PR #1154, opened 2019-12-06
Creation of AutoMoqDataAttribute.cs

CreepyGnome:
Seems like everyone has to create a basic AutoMoqDataAttribute to easily use Moq with AutoFixture so figured it would make it easier if one was provided in the package they are consuming already. This allows for the pass-through of  AutoMoqCustomization values and provides default ConfigureMembers to true if using the default constructor.


CreepyGnome:
Need to rework this I put it in the wrong place. This really should be a new package AutoFixture.AutoMoq.Xunit2 so not sure if a new package for one class is worth it at this time. Have to think more on this.

Kralizek:
The general idea about the issue is that normally a AutoDataAttribute requires more customizations than just adding AutoMoq.

This means that a glue package of glue packages would contain only one class that would be often inherited to allow further customizations.

That being said, to speed up the development in my team, I have created a dotnet-new template that adds references to all the relevant packages and a basic version of the AutoDataAttribute.

PR #1162, opened 2020-01-28
Add DateTime ranged relay

Tarig0:


Src/AutoFixture/DataAnnotations/DateTimeRangedRequestRelay.cs
```diff
@@ -0,0 +1,65 @@
...
+            var ticksInRange = context.Resolve(
+                new RangedNumberRequest(typeof(long), range.Min.Ticks, range.Max.Ticks));
+
+            if (ticksInRange is NoSpecimen)
+                return new NoSpecimen();
+
+            return new DateTime((long)ticksInRange);
+        }
+
+        private struct DateTimeRange
```
zvirja: Nit: We could also make it a `readonly struct`, as it's not supposed to be mutated.

Tarig0: Will the TimeSpanRange also need to be updated?

Kralizek:
Not a maintainer. 

Would it be possible to add a way to control which culture is used in the parsing process?

It could be helpful to be explicit in a multi-culture environment where developers have their machine with one locale (e.g. it-it) and the build server uses something else (e.g. en-us).

In this case parsing 1/12/1997 might work on a dev machine (where it's read as December 1st) but not on the build server (read as January 12th)

Tarig0:
Can you show me how the range attribute can be overwritten to parse the datetime explicitly?

Tarig0:
Looks like there is no way to override the parse for the range attribute at this time so the current implementation will cover all uses

Src/AutoFixture/DataAnnotations/DateTimeRangedRequestRelay.cs
```diff
@@ -0,0 +1,65 @@
...
+        private static object CreateDateTimeSpecimen(RangedRequest rangedRequest, ISpecimenContext context)
+        {
+            if (!(rangedRequest.Minimum is string) || !(rangedRequest.Maximum is string))
+                return new NoSpecimen();
+
+            var range = ParseTimeSpanRange(rangedRequest);
+            return RandomizeDateTimeInRange(range, context);
+        }
+
+        private static DateTimeRange ParseTimeSpanRange(RangedRequest rangedRequest)
```
alexbalser, line 38: Parse**DateTime**Range, right?

Issue #1142, opened 2019-09-27
MemberAutoData only uses first entry from the supplied enumerable


nirwall:
In `AutoFixture.xUnit2`, the `MemberAutoData` attribute only causes one test to be generated regardless of the size of the supplied data. You can see this in the test `AutoFixture.Xunit2.UnitTest.Scenario.MemberAutoDataSuppliesDataSpecimens`.

A similar test with `MemberData` and a member supplying a similar but "full" enumerable will generate two test cases:

```csharp
[Theory, MemberData(nameof(StringData))]
public void MemberDataSuppliesDataSpecimens(string s1, string s2, MyClass myClass)
{
    Assert.Equal("foo", s1);
    Assert.NotNull(s2);
    Assert.NotNull(myClass);
}

public static IEnumerable<object[]> StringData
{
    get
    {
        yield return new object[] { "foo", "bar", new MyClass() };
        yield return new object[] { "foo", "baz", new MyClass() };
    }
}
```

Ergamon:
The problem is that the implementation uses the CompositeDataAttribute base class. This one tries to merge the data from MemberDataAttribute and AutoDataAttribute, but the later always returns one row. So in your example MemberDataAttribute returns 2 rows, AutoDataAttribute only 1. The code merges this to the smallest number found, so only 1 row is generated.

I tried my best understanding of the AutoFixture source code and generated a (hopefully) working implementation:

` 
  using System;
  using System.Collections.Generic;
  using System.Linq;
  using System.Reflection;
  using System.Threading;
  using AutoFixture;
  using AutoFixture.Kernel;
  using AutoFixture.Xunit2;
  using Xunit;
  using Xunit.Sdk;

  public class MemberAutoDataAttribute : DataAttribute
  {
    private readonly Lazy<IFixture> fixture;
    private readonly MemberDataAttribute memberDataAttribute;

    public MemberAutoDataAttribute(string memberName, params object[] parameters)
      : this(memberName, parameters, () => new Fixture())
    {
    }

    protected MemberAutoDataAttribute(string memberName, object[] parameters, Func<IFixture> fixtureFactory)
    {
      if (fixtureFactory == null)
      {
        throw new ArgumentNullException(nameof(fixtureFactory));
      }
      
      memberDataAttribute = new MemberDataAttribute(memberName, parameters);
      fixture = new Lazy<IFixture>(fixtureFactory, LazyThreadSafetyMode.PublicationOnly);
    }

[...]

Issue #343, opened 2015-02-04
NUnit Values and AutoData doesn't work


scrous:
The AutoData overrides the [Values] tag of NUnit

```C#
[Test, AutoData]
public void Test(string name, [Values(1, -1)] int num)
```

num will be set by the AutoData and not from the Values

moodmosaic:
As a workaround, it is possible to use the [`Range`](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.rangeattribute%28v=vs.110%29.aspx) attribute on the test parameter(s):

```c#
[Test, AutoData]
public void Test(
    [System.ComponentModel.DataAnnotations.Range(-1, 1)]
    int actual)
{
    Assert.True(-1 <= actual && actual <= 1); // Passes.
}
```

scrous:
That seems to work, but only gets run once with -1. So System.ComponentModel.DataAnnotations.Range and NUnit doesn't work nicely together.

It doesn't work with NUnit's RangeAttribute ( http://www.nunit.org/index.php?p=range&r=2.6.3 ). 

You can also use Enums in Values, which you can't in Range, e.g:
```C#
        public enum TestData
        {
            One = 1,
            Two = 2,
            Three = 3,
            Four = 4
        }

        [Test, AutoData]
        public void Test_Enum(string name, [Values(TestData.Two, TestData.Three)] TestData testData)
        {
            Assert.IsNotNullOrEmpty(name);
            Assert.True(testData == TestData.Two || testData == TestData.Three);
        }
```

Another thing I have noticed is that AutoData doesn't pick a random enum, it always picks the first one.

adamchester:
Regarding `Range`, you should try with ’RandomRangedNumberCustomization`:

![image](https://cloud.githubusercontent.com/assets/570470/6125152/e5449cf2-b15e-11e4-93a8-832fbbd6a06e.jpg)

sergeyshushlyapin:
I spent some time investigating this issue. From the NUnit [docs](http://nunit.org/index.php?p=values&r=2.6.4)

> By default, NUnit creates test cases from all possible combinations of the datapoints provided on parameters - the combinatorial approach.

```csharp
	MyTest(1, "A")
	MyTest(1, "B")
	MyTest(2, "A")
	MyTest(2, "B")
	MyTest(3, "A")
	MyTest(3, "B")
[Test]
public void MyTest(
    [Values(1,2,3)] int x,
[...]

PR #1164, opened 2020-02-02
Enable MemberAutoDataAttribute to read multiple rows off a theory datasource

jbentus:
Currently, `MemberAutoDataAttribute` does not work with test cases with more than 1 row. The attribute always retrieves the first row regardless of how many elements your datasource has.

This fixes https://github.com/AutoFixture/AutoFixture/issues/1142.

It might be worth to mention a workaround for this limitation described [here](https://stackoverflow.com/a/18930198/4360304), but requires to have multiple `MemberAutoDataAttribute`s, which is not aligned with the expected behaviour of XUnit's `MemberDataAttribute`, which is what this attribute tries to reproduce. This change addresses that so you can have just one `MemberAutoDataAttribute` that groups under the same public static method, property or field all the test cases that make sense together.


jbentus:
@zvirja could you have a look at this PR ?

engenb:
I've also run into this issue so glad to see it's already been resolved!  Is there a release (or prerelease) on the horizon?

OkieCoder1:
Ugh YES!  Glad to see this is getting fixed.  Hopefully this makes it in soon =)  I ended up using the working "MemberAutoDataAttribute" implementation from https://stackoverflow.com/questions/38088639/use-autodata-and-memberdata-attributes-in-xunit-test.  Wondered why the one included in AutoFixture.Xunit2 doesn't work like the standard XUnit MemberDataAttribute...

Src/AutoFixture.xUnit2/MemberAutoDataAttribute.cs
```diff
@@ -67,5 +74,31 @@ protected MemberAutoDataAttribute(DataAttribute autoDataAttribute, string member
...
+
+            if (autoData == null) throw new InvalidOperationException($"{nameof(autoData)} cannot be null");
+
+            var combinedDataSet = new List<IEnumerable<object[]>>
+            {
+                memberData,
+                Enumerable.Repeat(autoData, memberData.Count)
+            };
+
+            return combinedDataSet.Zip(dataSet => dataSet.Collapse().ToArray());
```
ecampidoglio, line 101: This looks good, but I feel it can be made a bit more readable.

If you add an overload of the `Collapse` method that looks like this:

```csharp
internal static IEnumerable<T> Collapse<T>(this IEnumerable<T> sequence, params IEnumerable<T>[] others)
{
    return Collapse(new[] { sequence }.Concat(others));
}
```

Then, you can combine the values from the `MemberDataAttribute` with the ones generated by AutoFixture with a single LINQ expression:

```suggestion
            var memberData = this.MemberDataAttribute.GetData(testMethod);
            var autoData = this.AutoDataAttribute.GetData(testMethod).Take(1);

            if (autoData == null) throw new InvalidOperationException($"{nameof(autoData)} cannot be null");

            return from memberValues in memberData
                   from autoValues in autoData
                   let combinedValues = memberValues.Collapse(autoValues)
                   select combinedValues.ToArray();
```

This is simply the Cartesian product between the rows provided by `MemberDataAttribute` and the first row provided by `AutoDataAttribute`.

Src/AutoFixture.xUnit2.UnitTest/MemberAutoDataAttributeTest.cs
```diff
@@ -19,131 +19,33 @@ public void SutIsCompositeDataAttribute()
...
-
-            // Assert
-            Assert.Contains(result, a => a is AutoDataAttribute);
+            // Act & assert
+            Assert.Throws<ArgumentNullException>(() => sut.GetData(null));
         }
 
         [Fact]
-        public void AttributesAreInCorrectOrderWhenConstructedWithDefaultConstructor()
+        public void GetDataShouldReturnSameNumberOfRowsAsMemberDatasource()
```
ecampidoglio, line 37: A minor thing, but data source should be capitalized as `DataSource`.

Src/AutoFixture.xUnit2.UnitTest/MemberAutoDataAttributeTest.cs
```diff
@@ -275,6 +177,17 @@ public void PreDiscoveryShouldBeDisabled()
             Assert.Equal(expectedAssembly, actualAssembly);
         }
```
ecampidoglio, line 179: Shouldn't there also be a test that verifies that `MemberAutoDataAttribute.GetData` returns the values from `FooBarTestCase` and in the correct order?

Src/AutoFixture.xUnit2/MemberAutoDataAttribute.cs
aivascu: @ecampidoglio there's a minor oversight in the proposed implementation, `autoData` can never be `null`. In case the source collection is empty `Skip()` and `Take()` yield empty enumerables as well.
I'd prefer the second option, it doesn't use custom expressions.

ecampidoglio: @aivascu True that, we can remove the null check altogether. 👍

zvirja:
Hi guys,

Firstly many thanks @jbentus for working on this!

I reviewed the suggested change and see an issue with the suggested approach. Consider the following example:

```c#
public class Playground
{
    public static TheoryData<int> TestData => new TheoryData<int> { 1, 2, 4 };
    public class DemoCollection
    {
        private readonly List<int> items = new List<int>();

        public IEnumerable<int> Items => this.items;

        public void Add(int num) => this.items.Add(num);
    }

    [Theory, MemberAutoData(nameof(TestData))]
    public void ShouldContainAddedItemsOnly(int input, DemoCollection sut)
    {
        sut.Add(input);

        var itemsCount = sut.Items.Count();
        Assert.Equal(1, itemsCount);
    }
}
```

I believe that everybody would expect the test above to pass. However, it will fail on the second data row, as we re-use the `sut` instance between the tests. The scenario above is very simple, however imagine also usage of `[Frozen]` and multiple dependencies and the confusion it could bring.

To follow the well known Principle of Least Astonishment, I would expect `MemberAutoData` work in a way that the following two samples are equivalent:

```c#
public static TheoryData<int> TestData => new TheoryData<int> { 1, 2, 4 };

[Theory, MemberAutoData(nameof(TestData))]
public void Example1(int num, DemoCollection sut) {}
```
should be equivalent to
```c#
[Theory]
[InlineAutoData(1)]
[InlineAutoData(2)]
[InlineAutoData(4)]
public void Example2(int num, DemoCollection sut) {}
```

I would say that if `MemberAutoData` contains multiple rows, we should activate a new instance of AutoFixture context and generate new values for each row. Notice, the AutoFixture instance should be also new all the time, as otherwise mutable state of fixture (e.g. `Frozen`, `Inject` or other customizations) could lead to side-effects between tests.

Please share your opinion on this and advise whether you see scenarios when the present implementation is more favorable than the change I suggest.

Thanks!

Kralizek:
This is very similar to #343 (as in: let's let AutoFixture interact with attributes each library supports to generate more tests).

jbentus:
Thank you for your feedback @ecampidoglio, @aivascu, @zvirja.

I've applied suggestion no 2. After @zvirja comment I had to slightly change the optimized version, but if you have a way to optimize the current status further let me know @aivascu, I'll be happy to change it.

I think it makes a lot of sense what you said @zvirja, I've taken that into account and I've added a UnitTest based on your example to avoid a regression around this behavior.

On a side note, while working on this PR I've come across that the `Scenario.cs` file has grown a lot and was hard to navigate through this file. There might be an opportunity here to perhaps split out some of the code into a separate file. I don't think this is the right PR to execute this, but if you guys feel there's any value on it, I am happy to create an issue or PR around this. 

Let me know how do you guys find my latest commits.

Src/AutoFixture.xUnit2.UnitTest/MemberAutoDataAttributeTest.cs
jbentus: Hi @ecampidolgio, I was under the impression the `Scenario.cs` was the place holder for all the tests around data. It has a few tests to verify the data produced by the `MemberAutoDataAttribute`, which I believe may cover what you suggested.

Please let me know if you agree or you still want me to add more tests.

aivascu:
@zvirja you brought up a valid point. I did not consider this scenario.

@jbentus thank you for considering the feedback.
I have looked at the most recent version of your PR. I might be wrong but I think the scenario @zvirja mentioned is still reproducible.
What I think the suggestion was, is to create a new `Fixture` instance, for each row of theory data, so that that the generated instances are not shared between theory runs (i.e. when a specific type is `Frozen`).
I have updated the code snipped above to better exemplify the issue.
```csharp
public class Playground
{
    public static TheoryData<int> TestData => new TheoryData<int> { 1, 2, 4 };
    public class DemoCollection
    {
        private readonly List<int> items = new List<int>();

        public IEnumerable<int> Items => this.items;

        public void Add(int num) => this.items.Add(num);
    }

    [Theory, MemberAutoData(nameof(TestData))]
    public void ShouldContainAddedItemsOnlyFails(int input, [Frozen] DemoCollection sut)
    {
        sut.Add(input);

        var itemsCount = sut.Items.Count();
        Assert.Equal(1, itemsCount);
    }

    [Theory]
    [InlineAutoData(1)]
    [InlineAutoData(2)]
    [InlineAutoData(3)]
    public void ShouldContainAddedItemsOnlySucceeds(int input, [Frozen] DemoCollection sut)
    {
        sut.Add(input);

        var itemsCount = sut.Items.Count();
        Assert.Equal(1, itemsCount);
    }
}
```
image

jbentus:
Hi @aivascu thanks for pointing the missing case out.

I am not sure what's the best way to create a new `Fixture` off a `DataAttribute`, what do you and @zvirja find about the approach below:
```cs
public override IEnumerable<object[]> GetData(MethodInfo testMethod)
{
	// ...

	var autoDataType = this.AutoDataAttribute.GetType();

	return from memberValues in memberData
               from autoValues in CreateAutoData(autoDataType).GetData(testMethod).Take(1)
	       let combinedValues = memberValues.Concat(autoValues.Skip(memberValues.Length))
	       select combinedValues.ToArray();
}

private static DataAttribute CreateAutoData(Type autoDataType)
{
	return (DataAttribute)Activator.CreateInstance(autoDataType);
}
```

It passes the UnitTests and those mentioned previously. Is this a viable solution ?

aivascu:
@jbentus I think a better approach would be to pass in a `DataAttribute` factory delegate (i.e. `Func<DataAttribute>`) into the protected constructor, instead of the instance. The result would look like this:
```csharp
public override IEnumerable<object[]> GetData(MethodInfo testMethod)
{
    if (testMethod == null) throw new ArgumentNullException(nameof(testMethod));

    var memberData = this.MemberDataAttribute.GetData(testMethod).ToList();

    return from memberValues in memberData
           from autoValues in this.DataAttributeFactory.Invoke().GetData(testMethod).Take(1)
           let combinedValues = memberValues.Concat(autoValues.Skip(memberValues.Length))
           select combinedValues.ToArray();
}
```
@zvirja what do you think?

aivascu:
@jbentus @zvirja I have pushed some changes to the PR to explain my idea. Please let me know what you think.

Src/AutoFixture.xUnit2.UnitTest/MemberAutoDataAttributeTest.cs
jbentus: Hi @ecampidoglio. Would you be able to provide your blessing or reprimand on the above ? :)

jbentus:
@aivascu do you see any problem of this being a breaking change ? To me looks good as long as it's not problem to you. Probably the right thing to do in the long-run anyway. I was waiting on @zvirja to drop his insight on it.

aivascu:
@jbentus yes, from the [SemVer](https://semver.org/) perspective, this is a breaking change, since the signature of the protected constructor on `MemberAutoDataAttribute` has changed. Anyone using that constructor would get their code broken.
So normally this should go into the next major version. It's usually @zvirja that mentions these things, but he's not reachable at the moment, as well as most other maintainers.

I'm thinking I should just cut off the development on the `v4` and start working on the `v5`, so I can merge in all the breaking changes. The are several similar PRs and a lot of deprecations waiting in the pipeline anyway.

Src/AutoFixture.xUnit2/MemberAutoDataAttribute.cs
```diff
@@ -67,5 +74,25 @@ protected MemberAutoDataAttribute(DataAttribute autoDataAttribute, string member
...
+        /// Returns the composition of the theory data provided by MemberDataAttribute and AutoFixture.
+        /// </returns>
+        public override IEnumerable<object[]> GetData(MethodInfo testMethod)
+        {
+            if (testMethod == null) throw new ArgumentNullException(nameof(testMethod));
+
+            var memberData = this.MemberDataAttribute.GetData(testMethod).ToList();
+
+            return from memberValues in memberData
+                   from autoValues in this.DataAttributeFactory.Invoke().GetData(testMethod).Take(1)
```
ecampidoglio, line 93: It's a minor thing, but I feel that extracting this into a local variable called  `autoData` just below [L90](https://github.com/AutoFixture/AutoFixture/pull/1164/files#diff-8e03ae9ef1d63ccaa63555f169242392R90) would help clarify the relation between `memberData` and `autoData`.

Src/AutoFixture.xUnit2.UnitTest/MemberAutoDataAttributeTest.cs
ecampidoglio: You're right that the tests in `Scenario` verify the behavior; however, those are _end-to-end tests_ and, as such, they may fail for a number of different reasons. Adding a focused unit test for just this part would help narrow down any potential problems.

Src/AutoFixture.xUnit2/MemberAutoDataAttribute.cs
```diff
@@ -67,5 +74,25 @@ protected MemberAutoDataAttribute(DataAttribute autoDataAttribute, string member
...
+                   let combinedValues = memberValues.Concat(autoValues.Skip(memberValues.Length))
```
piotrzajac, line 94: I wonder if we could skip the values generation for these method arguments the memberData has already values for. With such improvement we should gain:
- performance boost (we do not generate values we already know we will skip)
- potential generation failure (it may happen that with the MemberAutoData we deliver such data that AutoDataAttribute would fail to generate)

I've already provided such behavior but with custom implementation of [AutoDataAttribute](https://github.com/ObjectivityLtd/AutoFixture.XUnit2.AutoMock/blob/master/src/Objectivity.AutoFixture.XUnit2.Core/Attributes/AutoDataAdapterAttribute.cs) in [AutoFixture.XUnit2.AutoMock](https://github.com/ObjectivityLtd/AutoFixture.XUnit2.AutoMock) packages.

Spider101:
Any ETA on when this might get merged? Looking forward to being able to properly use MemberAutoDataAttribute.

aivascu:
This PR will not be merged in v4, since it introduces breaking changes in the API. The support for `MemberAutoData` will be introduced in v5 as part of a different PR, along with the `ClassAutoData` attribute. I'll post updates on the feature in #1327.

Issue #625, opened 2016-05-20
Optionally disable caching for AutoMoq?


andreasnilsen:
When using an AutoConfiguredMoqCustomization, the value returned by the used fixture is cached, and the same value will be returned on each invocation (I'm not sure if this is done in MockType.ReturnsUsingContext or somewhere else, but I see a comment about it on line 104 there).

The only way I've found to override it is to pass a Frozen instance of the mock in question into the test method, and override the setup with a func invoking the fixture:

`someMock.Setup(it => it.SomeMethodReturningAString()).Returns(() => fixture.Create<string>())`

Is there any more generic/smarter way to configure the auto generated mocks to invoke the fixture each time a value is needed from AutoFixture, other than to manually setup each invocation?

dcastro:
The idea behind caching the result value was to make all methods [*pure*](https://en.wikipedia.org/wiki/Pure_function) by default, which is considered a good practice.

If the system under test expects a pure function, but is given an impure function, the test is likely to result in a false positive. E.g.

```csharp
// This method will return the expected result *if* `GetInt` is pure.
int Double(IDependency dep) {
    return dep.GetInt() + dep.GetInt();
}
```

On the other hand, if the system under test does not assume purity, giving it a pure or impure function should make no difference.

Hence, defaulting to pure functions made sense to me when I implemented this feature.

> Is there any more generic/smarter way to configure the auto generated mocks to invoke the fixture each time a value is needed from AutoFixture, other than to manually setup each invocation?

If you need to do this for *one* method, but across multiple test, I'd put it in a reusable `IConfiguration`.

If you want to do this for all methods, across multiple tests, then you'll need a bit more work:
* Define a new implementation of [`ReturnsUsingContext`](https://github.com/AutoFixture/AutoFixture/blob/98d7b4d00ba05e406796e32b560b5975f1f9af78/Src/AutoMoq/MockType.cs#L77) (IIRC, you can just copy the current implementation and remove line 104)
* Redefine [`MockVirtualMethodsCommand`](https://github.com/AutoFixture/AutoFixture/blob/98d7b4d00ba05e406796e32b560b5975f1f9af78/Src/AutoMoq/MockVirtualMethodsCommand.cs) and replace the call to `ReturnsUsingContext` to a call to your new method.
[...]

Issue #266, opened 2014-03-12
Online documentation


ploeh:
AutoFixture has rather decent documentation in the form of formal XML documentation comments in the source code, but unfortunately, few people discover it.

We should set up a site with on-line documentation, built from those XML comments, but we'd need an automated process for it. We should also consider how our rapid release cycle impacts the documentation; ideally, there should be a complete set of documentation files for each release - at least for every major and minor release.

Perhaps it would also be a good idea to ship the documentation with each NuGet package?

It's been quite a while since I last looked into the state of documentation generators for .NET, so I don't know what's good these days. Last time I looked, Sandcastle was the only viable alternative, and I found it quite difficult to work with.

We should keep in mind that we'd need to generate documentation from both C# and F# code bases.

moodmosaic:
Unless we can *easily* put this in our current build automation pipeline (e.g. using [F# Formatting](http://fsprojects.github.io/FSharp.Formatting/literate.html), or similar), should we close this?

PR #1175, opened 2020-04-20
Add convenience method to add AutoMoq to a fixture

Kralizek:
This pull request adds a convenient way customize a fixture with AutoMoq.

Here is an example of the intended usage:
```csharp
// default setup
var fixture = new Fixture().AddAutoMoq();

// enable members configuration
var fixture = new Fixture().AddAutoMoq(configureMembers: true);

// enable delagete generation
var fixture = new Fixture().AddAutoMoq(generateDelegates: true);

// enable both
var fixture = new Fixture().AddAutoMoq(configureMembers: true, generateDelegates: true);
```

The AddAutoMoq extension method has been added in the AutoFixture namespace to increase discoverability.

### DIFF
--- a/Src/AutoMoq/FixtureExtensions.cs
+++ b/Src/AutoMoq/FixtureExtensions.cs
@@ -1,0 +1,35 @@
+﻿using AutoFixture.AutoMoq;
+using AutoFixture.Kernel;
+
+// This class is placed in the AutoFixture namespace to improve discoverability.
+namespace AutoFixture
+{
+    public static class FixtureExtensions
+    {
+        /// <summary>
+        /// Applies the AutoMoqCustomization on the <see cref="IFixture"/>.
+        /// </summary>
+        /// <param name="fixture">The fixture upon which to enable auto-mocking.</param>
+        /// <param name="configureMembers">Specifies whether members of a mock will be automatically setup to retrieve the return values from a fixture.</param>
+        /// <param name="generateDelegates">Specifies whether delegate requests are intercepted and handled by Moq.</param>
+        /// <param name="relay">The relay that will be added to <see cref="IFixture.ResidueCollectors"/> when the customization is applied on the <see cref="fixture"/>.</param>
+        /// <returns>The <see cref="fixture"/> with the customization applied.</returns>
+        public static IFixture AddAutoMoq(this IFixture fixture, bool configureMembers = false, bool generateDelegates = false, ISpecimenBuilder relay = null)
+        {
+            var customization = new AutoMoqCustomization
+            {
+                ConfigureMembers = configureMembers,
+                GenerateDelegates = generateDelegates
+            };
+
+            if (relay != null)
+            {
+                customization.Relay = relay;
+            }
+
+            fixture.Customize(customization);
+
+            return fixture;
+        }
+    }
+}
--- a/Src/AutoMoq/FixtureExtensions.cs
+++ b/Src/AutoMoq/FixtureExtensions.cs
@@ -1,9 +1,13 @@
-﻿using AutoFixture.AutoMoq;
+﻿using System;
+using AutoFixture.AutoMoq;
 using AutoFixture.Kernel;
 
 // This class is placed in the AutoFixture namespace to improve discoverability.
 namespace AutoFixture
 {
+    /// <summary>
+    /// A class to collect extension methods of <see cref="IFixture"/>.
+    /// </summary>
     public static class FixtureExtensions
     {
         /// <summary>
@@ -12,10 +16,15 @@ namespace AutoFixture
         /// <param name="fixture">The fixture upon which to enable auto-mocking.</param>
         /// <param name="configureMembers">Specifies whether members of a mock will be automatically setup to retrieve the return values from a fixture.</param>
         /// <param name="generateDelegates">Specifies whether delegate requests are intercepted and handled by Moq.</param>
-        /// <param name="relay">The relay that will be added to <see cref="IFixture.ResidueCollectors"/> when the customization is applied on the <see cref="fixture"/>.</param>
-        /// <returns>The <see cref="fixture"/> with the customization applied.</returns>
+        /// <param name="relay">The relay that will be added to <see cref="IFixture.ResidueCollectors"/> when the customization is applied on the fixture.</param>
+        /// <returns>The fixture with the customization applied.</returns>
         public static IFixture AddAutoMoq(this IFixture fixture, bool configureMembers = false, bool generateDelegates = false, ISpecimenBuilder relay = null)
         {
+            if (fixture == null)
+            {
+                throw new ArgumentNullException(nameof(fixture));
+            }
+
             var customization = new AutoMoqCustomization
             {
                 ConfigureMembers = configureMembers,
--- a/Src/AutoMoqUnitTest/FixtureExtensionsTest.cs
+++ b/Src/AutoMoqUnitTest/FixtureExtensionsTest.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using AutoFixture.Kernel;
+using Moq;
+using Xunit;
+
+namespace AutoFixture.AutoMoq.UnitTest
+{
+    public class FixtureExtensionsTest
+    {
+        [Theory]
+        [InlineData(true, true)]
+        [InlineData(true, false)]
+        [InlineData(false, false)]
+        [InlineData(false, true)]
+        public void AddAutoMoqCustomizesFixture(bool configureMembers, bool generateDelegates)
+        {
+            var fixture = Mock.Of<IFixture>();
+
+            fixture.AddAutoMoq(configureMembers: configureMembers, generateDelegates: generateDelegates);
+
+            Mock.Get(fixture).Verify(p => p.Customize(It.Is<AutoMoqCustomization>(c => c.ConfigureMembers == configureMembers && c.GenerateDelegates == generateDelegates)));
+        }
+
+        [Fact]
+        public void AddAutoMoqCustomizesFixtureWithGivenRelay()
+        {
+            var fixture = Mock.Of<IFixture>();
+
+            var relay = Mock.Of<ISpecimenBuilder>();
+
+            fixture.AddAutoMoq(relay: relay);
+
+            Mock.Get(fixture).Verify(p => p.Customize(It.Is<AutoMoqCustomization>(c => ReferenceEquals(relay, c.Relay))));
+        }
+
+        [Fact]
+        public void AddAutoMoqReturnsSameFixture()
+        {
+            var fixture = Mock.Of<IFixture>();
+
+            var result = fixture.AddAutoMoq();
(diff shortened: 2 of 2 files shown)


aivascu:
@Kralizek if my understanding is correct this extension method is supposed to replace this code:
```csharp
fixture.Customize(new AutoMoqCustomization
{
    ConfigureMembers = true,
    GenerateDelegates = true,
    Relay = relayBuilder
});
```
I don't think the extension method will reduce the verbosity of the setup in a meaningful way. 
Besides, to be consistent, we would have to implement the same extensions for other frameworks (which might not always be possible) and we would have to ensure feature parity for both approaches to customizing the `Fixture` instance.

I feel that there might be another way to solve the issue you have. In my experience, convenience methods are implemented to reduce the amount of typing of something that is very common. In this particular case you could implement a specific customization that would contain all the customizations you add often to your tests.
```csharp
public class DomainCustomization : ICustomization // Maybe use a CompositeCustomization
{
    public virtual void Customize(IFixture fixture)
    {
        fixture.Customizations.Add(new AutoMoqCustomization {
           ConfigureMembers = true,
           GenerateDelegates = true,
           Relay = relayBuilder
        });
    }
}
```
This will reduce the amount of typing and you'll have a chance to give a descriptive name to your fixture setup `fixture.Customize(new DomainCustomization())`.

If you still like extension methods more, I could suggest publishing a package that would implement these convenience methods. The name prefix `AutoFixture.Extensions.` on NuGet is open for the community.

Kralizek:
Thank you for your feedback!

I'd say the intent is more to increase discoverability than just soaring keystrokes.

Fixture exposes different seams and they are quite confusing. Should I use Customize? Should I add a Customization? Is it a Behavior?

A simple AddAutoMoq extension would help users go past this.

Kralizek:
> If you still like extension methods more, I could suggest publishing a package that would implement these convenience methods. The name prefix `AutoFixture.Extensions.` on NuGet is open for the community.

Expanding on your comment, I don't think anybody would install an additional package just to add an extension method :)

aivascu:
@Kralizek sorry for the late reply. I suppose you are correct, that unlikely to happen. 😄 

Reading your comment made me realize I shared some of your confusion when starting out with AutoFixture.
There is a property on the `Fixture` class, called `Customizations` that acts like a collection of `ISpecimenBuilder`, however the name would imply it contains `ICustomization` instances. Reading the wiki on the [internal architecture](https://github.com/AutoFixture/AutoFixture/wiki/Internal-Architecture) made me understand how all the pieces fit together and now it doesn't make sense to me to have it any other way.

Considering this I think a more descriptive documentation and an API reference, on which I am working on at the moment #266 , would make the learning curve more shallow for newcomers.
Adding the extension method, I believe,  would add to the confusion around the setup since you would have more options and more things to consider, when doing it. Also it would mean more maintenance work while adding features like #625 .

@zvirja what do you think about this?

Kralizek:
@aivascu 

Sorry if I give the impression of being too pushy on this pull request, that's not the purpose. It's more about the principle and how it applies to the whole project.

In all honesty, in 2020 documentation sites are the second-last resort (right before posting a question on Stackoverflow).

Software frameworks should be self-discoverable. When a user adds a reference or the AutoMoq package is left with a feeling of "what now?". By exploring the namespace, they may find there is a "`AutoMoqCustomization`", where to plug it in?

Now look at other code bases like the `IServiceCollection` ecosystem.

Technically you could register your service as

```
//Syntax might not be compiler-correct, typing from phone
services.Add(ServiceDescriptor.Create(typeof (IService), typeof(MyConcreteService), Lifestyle.Transient);
```

Or, simply,

```
services.AddTransient<IService, MyConcreteService>();
```

Does `AddTransient` add more maintenance work, yes. But it also makes super easy and clear how to register a transient service.

With an extension like the one proposed in this PR, the user will install the package, go into the code, press `.` on a fixture and get `AddAutoMoq` in the intellisense list.

And yes, this would be just scratching the surface because the same logic could be added to many other methods. I am a strong supporter of the idea that AutoFixture needs a layer of sugar to help users take advantage of all its amazing capabilities.

aivascu:
@Kralizek I think I understand what you mean. You are aiming for a self-discoverable API like for example the configuration in asp.net core. This way you could compose different glue libraries by chaining methods when configuring the `Fixture`, like this.
```csharp
var fixture = new Fixture()
.AddMoq(c => {
   c.ConfigureMembers = true;
   c.GenerateDelegates = true;
})
.AddMockHttp()
.AddEntityFramework(c => {
   c.Provider = Providers.SQLite(x => {
       x.ConnectionString = 'temp.db';
   });
});
```
I guess this could be useful, in certain situations.

Kralizek:
> In any case I think the change would still have to be consistent across all libraries, so we don't alienate anyone.
> Could you update the PR to contain similar functionalities for the other mocking frameworks?

If there is an interest from the other maintainers, I see no issue in extending the scope of this PR to include other mocking frameworks.

ecampidoglio:
I like the idea of making AutoFixture's API more discoverable by adding a layer of [_convenience methods_](https://www.w3.org/TR/DOM-Level-2-Core/glossary.html#dt-convenience) on top. I usually do this myself in every project where I'm using AutoFixture, so it would be nice to have a few of them out of the box.

I think that the glue libraries for the various isolation frameworks is a good place to start. 👍

zvirja:
Hi guys!

I like an idea of having fresh and more concise fluent API for the AutoFixture. Especially, I like following about the idea:

- there is a clear entry point for the library; extension methods are intelli-sense friendly
- API could be inspired by ASP.NET Core using their patterns for customization. It should be friendly to modern developers. We could also use their guidelines to see how they solve typical issues

I think it's a big thing which we should carefully design. I would plan it for the next major version, so it can be one of the major features. By that time we could also play with it internally and try to convert the existing customizations to this API. This way we should see the general shape, challenges, etc and modify it before making it final.

I would not use `IFixture` as an entry point to not make intelli-sense even more cluttered. Instead, it could be a dedicated entry point, like:

```c#
fixture.Extend()
       .AddMoq(c => { c.ConfigureMembers = true })
       .AddRandomBooleanSequence();
```

This way after you typed `fixture.Extend()` you see only customizations you can apply, and you see all of them.

I suggest to abandon the current PR and create an issue for v5 instead. This way we could give it more considerations, rather than rushing it.

Thanks for the valuable idea, like it a lot!

aivascu:
I was thinking also that we might need some more time at the drawing board. The way the customization configuration looks now, would not fit the fluent asp.net-like API very well. To follow the asp.net configuration practices we would probably need configuration classes and maybe a couple overloads for each extension method.
Happy to continue the discussion in a dedicated thread.

Kralizek:
> Hi guys!
> 
> I like an idea of having fresh and more concise fluent API for the AutoFixture. Especially, I like following about the idea:
> 
> * there is a clear entry point for the library; extension methods are intelli-sense friendly
> * API could be inspired by ASP.NET Core using their patterns for customization. It should be friendly to modern developers. We could also use their guidelines to see how they solve typical issues
> 
> I think it's a big thing which we should carefully design. I would plan it for the next major version, so it can be one of the major features. By that time we could also play with it internally and try to convert the existing customizations to this API. This way we should see the general shape, challenges, etc and modify it before making it final.
> 
> I would not use `IFixture` as an entry point to not make intelli-sense even more cluttered. Instead, it could be a dedicated entry point, like:
> 
> ```cs
> fixture.Extend()
>        .AddMoq(c => { c.ConfigureMembers = true })
>        .AddRandomBooleanSequence();
> ```
> 
> This way after you typed `fixture.Extend()` you see only customizations you can apply, and you see all of them.
> 
> I suggest to abandon the current PR and create an issue for v5 instead. This way we could give it more considerations, rather than rushing it.
> 
> Thanks for the valuable idea, like it a lot!

`fixture.Extend()` (if needed) sounds more like something we could use in v4 to introduce the concept while we work at the big rework for v5.

> I was thinking also that we might need some more time at the drawing board. The way the customization configuration looks now, would not fit the fluent asp.net-like API very well. To follow the asp.net configuration practices we would probably need configuration classes and maybe a couple overloads for each extension method.
> Happy to continue the discussion in a dedicated thread.

zvirja:
@Kralizek I would rather prefer to not introduce concepts that are not well supported across the library. Otherwise, it might look like “yet another entry point” and will lead to further confusion. Rather, it’s cooler to introduce all at once. It’s just my opinion on this.

But technically speaking - yes, it could be introduced in v4 as well 😊

Kralizek:
> @Kralizek I would rather prefer to not introduce concepts that are not well supported across the library. Otherwise, it might look like “yet another entry point” and will lead to further confusion. Rather, it’s cooler to introduce all at once. It’s just my opinion on this.
> 
> But technically speaking - yes, it could be introduced in v4 as well 😊

My idea was to take v5 as the possibility to slim down `IFixture` and skip `Extend()` as a whole

zvirja:
@Kralizek Interesting idea, we could consider that. Not sure how it might work with breaking changes so it’s definitely requires thorough analysis and testing, but we should definitely consider it as one of the options!

aivascu:
As far as i can tell we have a common opinion that this feature should be postponed for `v5`.
@Kralizek would you like to then close this PR and open the issue for the proposed configuration API?
Since you proposed the idea, I think it is best if you expressed your vision on the implementation.

Kralizek:
> As far as i can tell we have a common opinion that this feature should be postponed for `v5`.
> @Kralizek would you like to then close this PR and open the issue for the proposed configuration API?
> Since you proposed the idea, I think it is best if you expressed your vision on the implementation.

Truth be told, my opinion is to take a leaner approach: 
* add the friendly extension methods to all glue libraries in `v4` 
* start designing a revamp of the public API (basically `IFixture`) in `v5`.

The idea is that v5 would be a breaking change for those using the old API but a code-compatible change (still a breaking one) for those using the new extension methods.

More work? probably, but also higher chances to get something concrete out to the market in the short term.

No offense to anybody involved, but this repo just woke up after a period of lethargy, so I'd start looking at what's achievable in the short term.

zvirja:
@Kralizek V5 shouldn't necessarily be a monster. It might be a version with pretty limited scope, as we can release v6 afterwards :)

We could also think of creating a dedicated branch and giving it a go and decide later whether it worth merging to v4.

P.S. There is no need to quote the whole previous answer all the time - it's clear that you reply to the message above 😊 It just clutters the thread and makes it harder to read 😟  Unless the goal is to add more weight to your messages by making them larger of course 😅

Kralizek:
@zvirja 

A wiseman once said, don't explain with malice what can be explain with stupidity.

Generally when a thread starts getting replies, I find it better to quote the message I'm replying to.

Off topic aside, I don't think adding the convenience methods right now would necessarily be "half baked". Take the extension method of this PR: I can't see how much better it can get 😅

Let's wait for @aivascu final call and we go from there.

aivascu:
@Kralizek I agree with @zvirja on this one. I think it is worth to giving this feature more thought and exploring more options.
Introducing a feature now, only to deprecate it a version later might make users question the reliability of the library.

Thank you for opening the issue. I will close this PR for now.
You can reopen it once we have settled on the details of the implementation details.

PR #1178, opened 2020-05-03
Optionally disable caching for AutoMoq

### LINKED ISSUE #625: Optionally disable caching for AutoMoq?

rus-art:
This PR resolves #625

The implementation is pretty simple. Please, tell me if we need more complex abstractions.
This is my first PR.

### DIFF
--- a/Src/AutoMoq/AutoMoqCustomization.cs
+++ b/Src/AutoMoq/AutoMoqCustomization.cs
@@ -54,6 +54,13 @@ namespace AutoFixture.AutoMoq
         public bool GenerateDelegates { get; set; }
 
         /// <summary>
+        /// If value is <c>true</c>, result of method call will be reused for following method calls.
+        /// Otherwise, if value is <c>false</c>, a separate result will be created for each method call.
+        /// The default value is <c>true</c>.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


aivascu:
Hello @rus-art.
Thank you for the contribution.
I've had a look at the PR and I'm thinking that the implementation could be improved by extracting the value resolution responsibility in the `MockTypes.ReturnsUsingContext()` and injecting a strategy that would handle setting the Return value, since it all boils down to either `setup.Returns(Func<TValue>)` or `setup.Returns(TValue)` being used.
I believe this is also more or less what was suggested in #625 .

Issue #1143, opened 2019-10-02
.NET Standard support in AutoFoq


tommyja:
Foq supports netstandard now so this should be possible to do.

aivascu:
This issue has been resolved in `v4.13.0`.

PR #1184, opened 2020-06-23
Add .net standard support to AutoFoq

josh-degraw:
In reference to #1143, this PR adds .net standard support for `AutoFoq`.

### DIFF
--- a/Src/AutoFoq/AutoFoq.fsproj
+++ b/Src/AutoFoq/AutoFoq.fsproj
@@ -2,7 +2,7 @@
   <Import Project="..\Common.props" />
 
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


aivascu:
Hello @josh-degraw! 
Thank you for the contribution.
It looks like the current build configuration ([visual studio 2015](https://www.appveyor.com/docs/windows-images-software/)) does not support .NET Core 3+. 
I'll see if the are any implications to the upgrade and get back to you.

josh-degraw:
@aivascu Okay, if it's easier I can just change the new target framework I for the unit tests to .NET Core 2.2

aivascu:
@josh-degraw I think changing the target framework, for now, might be the better plan.
I have tried to test the build under the VS 2019 environment, with no success.
I will upgrade the tests to use .NET Core 3, once I am more familiar with F# and FAKE.

zvirja:
Should we also update readme file to mark it netstandard compatible?

aivascu:
You are correct. I missed that.

PR #1185, opened 2020-06-27
Migrate to VS 2019 and maintain project

zvirja:
Maintain project by switching to the latest VS 2019 CI image. It required me to update FAKE, StyleCop and SourceLink, so I did additional maintenance of that.

Notice, checking the individual commits might help to more clearly see what exactly changed and why.


zvirja:
I've created a [PR](https://github.com/fsharp/FAKE/pull/2527) to FAKE to fix the issue. If they publish soon, then I could just upgrade to the latest FAKE and remove the workaround.

aivascu:
This looks awesome!

Src/AutoFixture/SpecimenBuilderNodeAdapterCollection.cs
```diff
@@ -113,6 +113,8 @@ private ISpecimenBuilderNode AdaptedBuilderNode
         /// <seealso cref="SpecimenBuilderNodeAdapterCollection(ISpecimenBuilderNode, Func{ISpecimenBuilderNode, bool})" />
         public int IndexOf(ISpecimenBuilder item)
         {
+            if (item == null) throw new ArgumentNullException(nameof(item));
```
aivascu, line 116: Will we need to update some tests for this change?

zvirja: Good catch 👍  I didn't do deep analysis around it. The reason I added this change is that the newest analyzers are failing about not checking the value. So it was not possible to run the build after I updated the analyzers.

I updated all the other places. Don't think it should be covered with unit tests, as in reality I believe it's only being used by our internal code - it's a collection => graph facade we use for tree mutation.

Let me know if you are fine with that.

zvirja:
@aivascu I updated the FAKE to latest (with my fix), please give it another look. Thanks!

zvirja:
@aivascu Let me do the next release, so I can confirm that CI pipeline works as expected after all the changes. After that please go ahead and take care of subsequent releases.

Let me know if you are fine ;)

aivascu:
Thank you for making the release @zvirja! 👍 

Of course, I'll take care of the next releases.
Hopefully you can still answer some questions if I am going to have any. 😅

Issue #1040, opened 2018-05-28
Proper random number generation


moodmosaic:
*This goes back to #2, one of the very first issues we've created here, on GitHub, after @ecampidoglio migrated the source code from CodePlex.* :rocket:

---

[RandomNumericSequenceGenerator](https://github.com/AutoFixture/AutoFixture/blob/e6293de7c8834d5dd42e1f81868af768d08f3438/Src/AutoFixture/RandomNumericSequenceGenerator.cs) is basically a wrapper on top of [System.Random](https://msdn.microsoft.com/en-us/library/system.random.aspx) with the addition of 64-bit integers and no duplicates (using an, inefficient, `HashSet<T>` internally for this).

I was looking at the source code of RandomNumericSequenceGenerator and I was wondering whether it's *random enough*. Overtime, several people had [issues](https://github.com/AutoFixture/AutoFixture/issues?q=is%3Aissue+random+label%3Abug) with it.

## Keep using RandomNumericSequenceGenerator 

We could keep using RandomNumericSequenceGenerator as long as it's random enough... How do we find out?

There are several statistical test suites we can run, e.g. [Dieharder](https://webhome.phy.duke.edu/~rgb/General/dieharder.php), [TestU01](http://simul.iro.umontreal.ca/testu01/tu01.html), [PractRand](http://pracrand.sourceforge.net/). I've  used Dieharder [to test the PRNG we use in Hedgehog](https://github.com/hedgehogqa/haskell-hedgehog/pull/185). I'd be tempted to try PractRand after seeing Melissa E. O’Neill's [post](http://www.pcg-random.org/posts/how-to-test-with-practrand.html).

## Switch to a proper PRNG

There are several out there already, some of them are battle-tested.

* Donald E. Knuth: *Subtractive random number generator* ([paper](https://dl.acm.org/citation.cfm?id=270146)) — used in .NET [System.Random](https://msdn.microsoft.com/en-us/library/system.random.aspx)
* Pierre L'Ecuyer: *Efficient and portable combined random number generators* ([paper](https://www.iro.umontreal.ca/~lecuyer/myftp/papers/cacm88.pdf)) — used in Haskell [random](http://hackage.haskell.org/package/random) and [QuickCheck](http://hackage.haskell.org/package/QuickCheck), and FsCheck v1 and v2.
[...]

PR #1187, opened 2020-06-30
Proposed fix for #1183 - multiple regex within short time might give identical strings

### LINKED ISSUE #1183: Creating multiple regular expression requests in rapid succession yields same result (.NET framework)
The test below (usually) fails in .NET framework fails in .NET framework, but not in core.

The underlying reason for this is that Xeger is used in autofixture without injecting an object of type Random, and Xeger uses `new Random()` - without injecting a seed.

The test passes, in .NET Core, because `new Random()` internally generates a unique seed.
However, as you probably know, in framework, it uses tick count as seed.

I haven't created a pr, because I didn't know where to do it to - does it make sense to create a pr with a failing test to master?

And I don't know if it should be fixed here in AF, or a pr should be created in the Fare repo.

What do you think? I'd like to fix this - but I'm not sure how to proceed...

```
        [Theory]
        [InlineData(@"^[A-Z]{27}$")]
        public void CreateTwoWithRegularExpressionRequestReturnsDifferentResults(string pattern)
        {
            // Arrange
            var sut = new RegularExpressionGenerator();
            
            var request = new RegularExpressionRequest(pattern);
            var dummyContext = new DelegatingSpecimenContext();
            // Act
            var result1 = sut.Create(request, dummyContext);
            var result2 = sut.Create(request, dummyContext);
            
            // Assert
            Assert.NotEqual(result1, result2);
        }
```

MagnusMikkelsen:
Included:
- Test that exposes the issue
- Proposed fix:
    - Use `Xeger(pattern, random)` constructor overload
    - Use the same instance of `Random` each time. I figured it should be enough to avoid the issue with identical seeds.

Fix #1183

### DIFF
--- a/Src/AutoFixture/RegularExpressionGenerator.cs
+++ b/Src/AutoFixture/RegularExpressionGenerator.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Text.RegularExpressions;
+using System.Threading;
 using AutoFixture.Kernel;
 using Fare;
 
@@ -8,9 +9,12 @@ namespace AutoFixture
     /// <summary>
     /// Creates a string that is guaranteed to match a RegularExpressionRequest.
     /// </summary>
-    public class RegularExpressionGenerator : ISpecimenBuilder
+    public class RegularExpressionGenerator : ISpecimenBuilder, IDisposable
     {
         private readonly Random random;
+        private readonly object syncRoot;
+        private readonly ThreadLocal<Random> threadLocalRandom;
+        private bool disposedValue;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RegularExpressionGenerator"/> class.
@@ -18,6 +22,8 @@ namespace AutoFixture
         public RegularExpressionGenerator()
         {
             this.random = new Random();
+            this.syncRoot = new object();
+            this.threadLocalRandom = new ThreadLocal<Random>(() => new Random(this.GenerateSeed()));
         }
 
         /// <summary>
@@ -49,7 +55,8 @@ namespace AutoFixture
             {
                 // Use the Xeger constructor overload that that takes an instance of Random.
                 // Otherwise identically strings can be generated, if regex are generated within short time.
-                string regex = new Xeger(pattern, this.random).Generate();
+                // Use shared random - but one for each thread to avoid threadsafety issues with Random.
+                string regex = new Xeger(pattern, this.threadLocalRandom.Value).Generate();
                 if (Regex.IsMatch(regex, pattern))
                 {
                     return regex;
@@ -65,6 +72,44 @@ namespace AutoFixture
             }
 
             return new NoSpecimen();
+        }
+
+        private int GenerateSeed()
+        {
+            lock (this.syncRoot)
+            {
+                return this.random.Next();
+            }
+        }
+
+        /// <summary>
+        /// Disposes  <see cref="ThreadLocal&lt;Random&gt;"/>.
+        /// </summary>
+        /// <param name="disposing">
+        /// <see langword="true"/> to release both managed and unmanaged resources;
+        /// <see langword="false"/> to release only unmanaged resources.
+        /// </param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!this.disposedValue)
+            {
+                if (disposing)
+                {
+                    this.threadLocalRandom?.Dispose();
+                }
+
+                this.disposedValue = true;
+            }
+        }
+
+        /// <summary>
+        /// Disposes  <see cref="ThreadLocal&lt;Random&gt;"/>.
+        /// </summary>
+        public void Dispose()
+        {
+            // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+            this.Dispose(disposing: true);
+            GC.SuppressFinalize(this);
         }
     }
 }
--- a/Src/AutoFixture/RegularExpressionGenerator.cs
+++ b/Src/AutoFixture/RegularExpressionGenerator.cs
@@ -10,6 +10,16 @@ namespace AutoFixture
     /// </summary>
     public class RegularExpressionGenerator : ISpecimenBuilder
     {
+        private readonly Random random;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RegularExpressionGenerator"/> class.
+        /// </summary>
+        public RegularExpressionGenerator()
+        {
+            this.random = new Random();
+        }
+
         /// <summary>
         /// Creates a string that is guaranteed to match a RegularExpressionRequest.
         /// </summary>
@@ -28,16 +38,18 @@ namespace AutoFixture
                 return new NoSpecimen();
             }
 
-            return GenerateRegularExpression(regularExpressionRequest);
+            return this.GenerateRegularExpression(regularExpressionRequest);
         }
 
-        private static object GenerateRegularExpression(RegularExpressionRequest request)
+        private object GenerateRegularExpression(RegularExpressionRequest request)
         {
             string pattern = request.Pattern;
 
             try
             {
-                string regex = new Xeger(pattern).Generate();
+                // Use the Xeger constructor overload that that takes an instance of Random.
+                // Otherwise identically strings can be generated, if regex are generated within short time.
+                string regex = new Xeger(pattern, this.random).Generate();
                 if (Regex.IsMatch(regex, pattern))
                 {
                     return regex;
--- a/Src/AutoFixture/RegularExpressionGenerator.cs
+++ b/Src/AutoFixture/RegularExpressionGenerator.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Text.RegularExpressions;
-using System.Threading;
 using AutoFixture.Kernel;
 using Fare;
 
@@ -9,12 +8,10 @@ namespace AutoFixture
     /// <summary>
     /// Creates a string that is guaranteed to match a RegularExpressionRequest.
     /// </summary>
-    public class RegularExpressionGenerator : ISpecimenBuilder, IDisposable
+    public class RegularExpressionGenerator : ISpecimenBuilder
     {
         private readonly Random random;
         private readonly object syncRoot;
-        private readonly ThreadLocal<Random> threadLocalRandom;
-        private bool disposedValue;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RegularExpressionGenerator"/> class.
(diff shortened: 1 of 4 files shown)


Src/AutoFixtureUnitTest/RegularExpressionGeneratorTest.cs
```diff
@@ -103,6 +103,31 @@ public void CreateWithRegularExpressionRequestReturnsCorrectResult(string patter
...
+            // Xeger(pattern) internally creates an instance of Random with the default seed.
+            // This means that the RegularExpressionGenerator might create identical strings
+            // if called multiple times within a short time.
+
+            // Arrange
+            var sut = new RegularExpressionGenerator();
+            var request = new RegularExpressionRequest(pattern);
+            var dummyContext = new DelegatingSpecimenContext();
+
+            // Repeat a few times - because the issue seen comes from Random class using tickcount as seed.
```
moodmosaic, line 120:
```suggestion
            // Repeat a few times to make the test more robust.
```

Src/AutoFixtureUnitTest/RegularExpressionGeneratorTest.cs
```diff
@@ -103,6 +103,31 @@ public void CreateWithRegularExpressionRequestReturnsCorrectResult(string patter
...
+            for (int i = 0; i < 10; i++)
+            {
+                var result1 = sut.Create(request, dummyContext);
+                var result2 = sut.Create(request, dummyContext);
+
+                // Assert
+                Assert.NotEqual(result1, result2);
```
moodmosaic, line 127: Wouldn't it be possible to avoid the `for-loop` and the [Assertion Roulette](http://xunitpatterns.com/Assertion%20Roulette.html) by doing something similar to what you did [here](https://github.com/moodmosaic/Fare/pull/56/commits/1563f8e5cea71bb58c85e7843498796d81975715#diff-a4691edefa1cec0b62875a993b25ed7c)? 👀

Src/AutoFixture/RegularExpressionGenerator.cs
```diff
@@ -10,6 +10,16 @@ namespace AutoFixture
...
     public class RegularExpressionGenerator : ISpecimenBuilder
     {
+        private readonly Random random;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RegularExpressionGenerator"/> class.
+        /// </summary>
+        public RegularExpressionGenerator()
+        {
+            this.random = new Random();
```
moodmosaic, line 20: This looks reasonable *if* RegularExpressionGenerator has also the same, transient, scope.

(A test similar to [this](https://github.com/moodmosaic/Fare/pull/56/commits/1563f8e5cea71bb58c85e7843498796d81975715#diff-a4691edefa1cec0b62875a993b25ed7c) one, which you did previously, should give the answer.)

Src/AutoFixtureUnitTest/RegularExpressionGeneratorTest.cs
MagnusMikkelsen: Yes - totally, just didn't know if introducing linq in this test was acceptable? Commit on the way.

Src/AutoFixture/RegularExpressionGenerator.cs
MagnusMikkelsen: ok - I've added a test in FixtureTests.

Src/AutoFixture/RegularExpressionGenerator.cs
```diff
@@ -28,16 +38,18 @@ public object Create(object request, ISpecimenContext context)
...
+        private object GenerateRegularExpression(RegularExpressionRequest request)
         {
             string pattern = request.Pattern;
 
             try
             {
-                string regex = new Xeger(pattern).Generate();
+                // Use the Xeger constructor overload that that takes an instance of Random.
+                // Otherwise identically strings can be generated, if regex are generated within short time.
+                string regex = new Xeger(pattern, this.random).Generate();
```
zvirja, line 52: As per my knowledge `Random` is not thread safe. See e.g. [here](https://stackoverflow.com/a/3049478/2009373). Previously we created a random instance per request (probably internally within Xeger), so we didn't have that problem.

Fixture by design can be used from multiple threads simultaneously, so we should take care of multi-threading.

MagnusMikkelsen: Ok, I'll look into it.
Should we use the code from https://codeblog.jonskeet.uk/2009/11/04/revisiting-randomness/ ? 

Basically doing `Xeger(pattern, ThreadLocalRandom.Instance)`

This way we could abstract threadsafety concerns out of RegularExpressionGenerator... 🤔, and maybe reuse it elsewhere.

moodmosaic: @MagnusMikkelsen you can just follow the pattern used in all the other generators, e.g.

https://github.com/AutoFixture/AutoFixture/blob/d3e14a818b933d948ea82464b197a5aeb152f070/Src/AutoFixture/RandomNumericSequenceGenerator.cs#L144

zvirja: @moodmosaic @MagnusMikkelsen If there is any way to do that without locks, please prefer that way. Introducing locking makes it hard to optimize performance later. When you have a bit suite, AutoFixture already takes substantial amount of time due to uncached reflection is uses everywhere - we shouldn't degrade it even more 😟 

In this particular case, it might be better to abstract out random generation and create generator, which uses e.g. `Interlocked`. Or, for instance, create thread-local random, but use interlocked to generate seed.

moodmosaic: Interlocked.CompareExchange or (IIRC) the Interlocked-extras by Jeffrey Richter in _CLR via C#_ can be candidates. 

- Unless you measure it, you can’t actually know how much ‘faster’ it’s going to be
- If speed matters, perhaps we can consider #928 — or [look into other PRNGs](https://github.com/AutoFixture/AutoFixture/issues/1040)

Though right now, I’d suggest we stick with `lock` as done in all other places and be consistent.

MagnusMikkelsen: Not sure how to use Interlocked.CompareExchange for this?

I did a couple of measurements, running the test below.
The times are measured using dotTrace, and are the total time and own time for `RegularExpressionGenerator.GenerateRegularExpression`.
The last test (ThreadLocal) took 12s. Seems to make sense, given the `GenerateRegularExpression` used some 85 seconds in total on 8 threads.

|					|Total Time [ms]  |	Own  [ms] |
|---------------------------|------------------|--------------|
|Current: shared `Random`|  83387               |  0                |
|Original: No injected `Random` | 88926 |	10 |
|New `Random` each time<br>locking on generate seed | 100843 |	0 |
| Sharing `ThreadLocal` <br> locking when generating `ThreadLocal` | 84848 |	27 |

(It is beside the point of this PR - but the `Xeger.ctor` seems to take most of this time - 81000ms total time. `Generate` is about 550ms total time. Might be an idea to cache `Xeger`? Fx `dictionary[pattern]=Xeger`)

``` cs
        [Fact]
        public void CreateManyAnonymousInParallelWithRegularExpressionValidatedTypeReturnsCorrectResult()
        {
            // Arrange
            var fixture = new Fixture();
            var result = Enumerable
                .Range(0, 10000)
                .AsParallel()
                .WithDegreeOfParallelism(8)
                .Select(_ => fixture.Create<RegularExpressionValidatedType>())
                .ToArray();

            // Assert

            Assert.All(result, r => Assert.Matches(RegularExpressionValidatedType.Pattern, r.Property));
            Assert.Equal(result.Distinct(), result);
        }
```

MagnusMikkelsen: > Interlocked-extras by Jeffrey Richter in _CLR via C#_ can be candidates.
Haven't got that book - so I don't it, do you have any links with examples? I'd like to learn new tricks 😁

MagnusMikkelsen:
> @MagnusMikkelsen, thank you for the investigation 👍 

You're welcome - I'm having fun 😁

>  just use `lock` and plan for perf improvements separately when needed.

I think I agree. In case there are many scenarios out there, that uses the same instance of Fixture to generate the same patterns many times - a major improvement would be to cache Xeger - much greater than those 2ms we gain by sharing the `Random`.

Just ran a small experiment in linqpad. Creating 1000 `Xeger` and generating regex takes 2 seconds. Reusing a `Xeger` to generate 1000 regex took ca 40ms.

zvirja:
@MagnusMikkelsen Thanks for the rework and fixing the concurrency issue. I fully agree that now we can proceed further with the current approach and merge the code. You are guarding by lock only random seed generation, so it sounds reasonable. My concern was about protecting the whole `Create()` method.

Just in case you are curious, my idea was to have something like:

```c#
static private int RandomSeed = Environment.TickCount;
...

new Xeger(new Random(Interlocked.Add(ref RandomSeed, 1)))

```

This way code is completely lock-free and you are always giving a new seed to random. I believe in this case it doesn't matter whether we just increment seed by one, or use completely random input - the `Random` class should randomize it anyway.

Anyway, it was just in case you are curious and I'm fine with the current implementation. Let me know when you want me to merge the code.

moodmosaic:
> Just in case you are curious, my idea was to have something like:
> 
> ```cs 
> new Xeger(new Random(Interlocked.Add(ref RandomSeed, 1)))
> ```

From a thread-safety viewpoint, this looks reasonable. In the context of pseudorandom number generators, this must be sampled in order to make sure it'll produce output that is uniformly distributed random and [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem)-free, at the very least.

MagnusMikkelsen:
Given @moodmosaic comment, shouldn't we just merge it as is @zvirja ?

zvirja:
@moodmosaic The same should be said about the current approach, shouldn't it? 😉 If none of the approaches was tested to prove the statistical quality of the resulting random sequence, we could just ignore this aspect, assuming that both approaches are equal. And with that we could solely focus on performance aspect.

Anyway, merging the code! Thanks again @MagnusMikkelsen for your help!

moodmosaic:
Yeah, but at least right we're following the 'pattern' so we're at least consistent.

zvirja:
@aivascu I suggest to drop a new release, so the fix could be consumed. Let me know if you need my assistance on this ;-)

moodmosaic:
>I suggest to drop a new release, so the fix could be consumed

:+1: :confetti_ball: :rocket: 

---

Thanks @MagnusMikkelsen for your work and @zvirja for the great feedback :beers:

aivascu:
@zvirja the release 4.13.0 has been deployed

PR #1194, opened 2020-08-26
Idiomatic assertions for `IEqualityComparer<T>`

Kralizek:
At my company we have ended up creating some implementations of `IEqualityComparer<T>` so I've worked on an idiomatic assertion to quickly check that they respect some basic requirements:
- `Equals(a, a) == true` (reflexive property)
- `Equals(a, b) == Equals(b, a)` (symmetric property)
- `Equals(a, b) == Equals(b, c) == Equals(a, c)` (transitive property)
- `Equals(a, null) == false`
- `Equals(null, null) == true`
- `GetHashCode(a) == GetHashCode(a)`

I added to this PR all the code produced right now. If there is an interest, I can continue working on this PR.

Here is a sample on how to use it:

Given the following class implementing `IEqualityComparer<string>`
```csharp
public class TestEqualityComparer : IEqualityComparer<string>
{
	public bool Equals(string x, string y)
	{
		if (x is null ^ y is null) return false;
		
		return string.Equals(x, y);
	}

	public int GetHashCode(string obj)
	{
		return obj.GetHashCode();
	}
}
```

Developers can use this assertion as simply as
```csharp
[Test]
public void IEqualityComparer_interface_is_implemented_correctly()
{
	var fixture = new Fixture();
	
	var assertion = fixture.Create<EqualityComparerAssertion>();
	
	assertion.Verify(typeof(TestEqualityComparer));
}
```

I haven't gone added the unit tests I have in my own solution and decorated methods and property with documentation comments. Also, I have used few features from C# 8 but I'm not sure if the current build setup supports them.

### DIFF
--- a/Src/Idioms/EqualityComparerAssertion.cs
+++ b/Src/Idioms/EqualityComparerAssertion.cs
@@ -29,13 +29,9 @@ namespace AutoFixture.Idioms
         private static IEnumerable<IIdiomaticAssertion> CreateChildAssertions(ISpecimenBuilder builder)
         {
             yield return new EqualityComparerEqualsSelfAssertion(builder);
-
             yield return new EqualityComparerEqualsSymmetricAssertion(builder);
-
             yield return new EqualityComparerEqualsTransitiveAssertion(builder);
-
             yield return new EqualityComparerGetHashCodeAssertion(builder);
-
             yield return new EqualityComparerEqualsNullAssertion(builder);
         }
     }
--- a/Src/Idioms/EqualityComparerAssertion.cs
+++ b/Src/Idioms/EqualityComparerAssertion.cs
@@ -1,16 +1,30 @@
-﻿using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Globalization;
-using System.Linq;
-using System.Reflection;
+﻿using System.Collections.Generic;
 using AutoFixture.Kernel;
 
 namespace AutoFixture.Idioms
 {
+    /// <summary>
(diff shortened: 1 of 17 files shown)


aivascu:
@Kralizek thank you for the contribution. I think this will make a nice addition to AutoFixture.

There is a [package](https://github.com/baks/EqualityTests) that implements these same assertions and some more.
I don't know if @baks would like to contribute those features to AutoFixture but at least it could serve as inspiration for this PR.

I think fixing the issues from the appveyor build log and adding the tests would be the first step to moving forward with the PR.

Kralizek:
@aivascu
Glad you like it! The package you linked doesn't have anything (AFAIK) for implementing IEqualityComparer, hence my PR.

Yep, appveyor went crazy on feedback. I'll work on it when I have some free time.

Kralizek:
@aivascu I fixed the build errors and added relevant tests so this PR is ready for a review. I looked ad @baks's code as suggested but I couldn't take much as it uses a lot of helper classes and I didn't know if we wanted to import them into `AutoFixture.Idioms.`

aivascu (approved):
@Kralizek The PR looks good, I've made only a few code style changes. Let me know if you're OK with them.

Kralizek:
No worries! Nice changes. I didn't think about putting that ugly pattern match into its own method.

zvirja:
@Kralizek Nice work, like how thorough we are testing it! Having more assertions of the quality you created could make library insanely useful! Thanks for your contribution! 🤝

Kralizek:
Glad you like it!

PR #1199, opened 2020-09-19
Add Injected attribute

Kralizek:
This PR adds a new attribute to be used when working with AutoData-based unit tests.

The `Injected` attribute lets the developer specify a value to be injected in the pipeline to be used whenever the same type is requested. Like for `Frozen`, developers can specify matching rules to restrict when the specified value should be used.

This is a simple use case of the new attribute

```csharp
[Test, AutoData]
public void InjectFirstParameterShouldReturnAssignedValue(
    [Injected(42)] int p1)
{
    Assert.AreEqual(p1, 42);
}

[Test, AutoData]
public void InjectFirstParameterShouldAssignSameInstanceToSecondParameter(
    [Injected(42)] int p1,
    int p2)
{
    Assert.AreEqual(p1, p2);
}

[Test, AutoData]
public void InjectParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected(42)] int value, 
    PropertyHolder<int> holder)
{
    Assert.AreEqual(value, holder.Property);
}
```

The PR is not finished
- [x] Add InjectOnMatchCustomization
- [ ] Add tests for InjectOnMatchCustomization
- [x] Add InjectedAttribute for NUnit 3
- [x] Add tests for InjectedAttribute for NUnit 3
- [x] Add scenario tests for InjectedAttribute for NUnit 3
- [ ] Add InjectedAttribute for Xunit
- [ ] Add tests for InjectedAttribute for Xunit
- [ ] Add scenario tests for InjectedAttribute for Xunit

### DIFF
--- a/Src/AutoFixture/InjectOnMatchCustomization.cs
+++ b/Src/AutoFixture/InjectOnMatchCustomization.cs
@@ -1,0 +1,93 @@
+using System;
+using AutoFixture.Kernel;
+
+namespace AutoFixture
+{
+    /// <summary>
+    /// A customization that injects a specific value of a specific <see cref="Type"/>
+    /// and uses them to satisfy requests that match a set of criteria.
+    /// </summary>
+    /// <remarks>
+    /// The criteria that determine whether the frozen specimen will be used
+    /// to satisfy a given request are specified at construction time
+    /// as a <see cref="IRequestSpecification"/> object.
+    /// Multiple criteria can be combined together as a boolean expression
+    /// by composing them into a <see cref="AndRequestSpecification"/>
+    /// or <see cref="OrRequestSpecification"/> object.
+    /// </remarks>
+    public class InjectOnMatchCustomization : ICustomization
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InjectOnMatchCustomization"/> class.
+        /// </summary>
+        /// <param name="valueToInject">The value to inject.</param>
+        /// <param name="request">The request used to create a specimen to freeze.</param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="valueToInject"/> or <paramref name="request"/> is <see langword="null"/>.
+        /// </exception>
+        public InjectOnMatchCustomization(object valueToInject, object request)
+            : this(valueToInject, request, new EqualRequestSpecification(request))
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InjectOnMatchCustomization"/> class.
+        /// </summary>
+        /// <param name="valueToInject">The value to inject.</param>
+        /// <param name="request">The request used to create a specimen to freeze.</param>
+        /// <param name="matcher">
+        /// The <see cref="IRequestSpecification"/> used to match the requests
+        /// that will be satisfied by the frozen specimen.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// <paramref name="valueToInject"/>, <paramref name="request"/> or <paramref name="matcher"/> is <see langword="null"/>.
+        /// </exception>
+        public InjectOnMatchCustomization(object valueToInject, object request, IRequestSpecification matcher)
+        {
+            this.ValueToInject = valueToInject ?? throw new ArgumentNullException(nameof(valueToInject));
+            this.Request = request ?? throw new ArgumentNullException(nameof(request));
+            this.Matcher = matcher ?? throw new ArgumentNullException(nameof(matcher));
+        }
+
+        /// <summary>
+        /// The value to inject.
+        /// </summary>
+        public object ValueToInject { get; }
+
+        /// <summary>
+        /// The request used to resolve specimens. By default that is TargetType.
+        /// </summary>
+        public object Request { get; }
+
+        /// <summary>
+        /// The <see cref="IRequestSpecification"/> used to match the requests
+        /// that will be satisfied by the frozen specimen.
+        /// </summary>
+        public IRequestSpecification Matcher { get; }
+
+        /// <summary>
+        /// Customizes the specified fixture.
+        /// </summary>
+        /// <param name="fixture">The fixture to customize.</param>
+        public void Customize(IFixture fixture)
+        {
+            if (fixture == null) throw new ArgumentNullException(nameof(fixture));
+
+            this.InjectSpecimenForMatchingRequests(fixture);
+        }
+
+        private void InjectSpecimenForMatchingRequests(IFixture fixture)
+        {
+            fixture.Customizations.Insert(
+                0,
+                new FilteringSpecimenBuilder(
+                    this.InjectSpecimen(),
+                    this.Matcher));
+        }
+
+        private ISpecimenBuilder InjectSpecimen()
+        {
+            return new FixedBuilder(this.ValueToInject);
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3.UnitTest/InjectedAttributeTest.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/InjectedAttributeTest.cs
@@ -1,0 +1,36 @@
+﻿using NUnit.Framework;
+
+namespace AutoFixture.NUnit3.UnitTest
+{
+    [TestFixture]
+    public class InjectedAttributeTest
+    {
+        [Test, AutoData]
+        public void SutIsAttribute(int value)
+        {
+            // Arrange
+            // Act
+            var sut = new InjectedAttribute(value);
+            // Assert
+            Assert.That(sut, Is.InstanceOf<CustomizeAttribute>());
+        }
+
+        [Test, AutoData]
+        public void GetCustomizationFromNullParameterThrows(int value)
+        {
+            // Arrange
+            var sut = new InjectedAttribute(value);
+            // Act & Assert
+            Assert.That(() => sut.GetCustomization(null), Throws.ArgumentNullException);
+        }
+
+        [Test, AutoData]
+        public void ValueToInjectReturnsConstructorParameter(int value)
+        {
+            // Arrange
+            var sut = new InjectedAttribute(value);
+            // Act & Assert
+            Assert.That(sut.ValueToInject, Is.EqualTo(value));
+        }
+    }
+}
--- a/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
@@ -331,6 +331,29 @@ namespace AutoFixture.NUnit3.UnitTest
             Assert.AreNotEqual(p1, p2.Field);
         }
 
+        [Test, AutoData]
+        public void InjectFirstParameterShouldReturnAssignedValue(
+            [Injected("Hello World")] string p1)
+        {
+            Assert.AreEqual(p1, "Hello World");
+        }
+
+        [Test, AutoData]
+        public void InjectFirstParameterShouldAssignSameInstanceToSecondParameter(
+            [Injected("Hello World")] string p1,
+            string p2)
+        {
+            Assert.AreEqual(p1, p2);
+        }
+
+        [Test, AutoData]
+        public void InjectParameterShouldAssignSameInstanceToPropertyHolder(
+            [Injected(42)] int value, 
+            PropertyHolder<int> holder)
+        {
+            Assert.AreEqual(value, holder.Property);
+        }
+
         [Test]
         [InlineAutoData(1, 2, 3)]
         public void InlineAutoDataTakesParameterValues(int p1, int p2, int p3)
--- a/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
+++ b/Src/AutoFixture.NUnit3.UnitTest/Scenario.cs
@@ -333,22 +333,22 @@ namespace AutoFixture.NUnit3.UnitTest
 
         [Test, AutoData]
         public void InjectFirstParameterShouldReturnAssignedValue(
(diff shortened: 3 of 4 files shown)


Kralizek:
@aivascu let me know if you like the idea and if I should proceed

aivascu:
@Kralizek I like the idea of having feature parity with our inline API and be able to `Inject` values just like we are able to `Freeze` them now. However I'd like to suggest a change to it's behavior.

Instead of providing the value literal inside the attribute constructor we could make use of the values injected by other attributes like `InlineAutoDataAttribute` or `MemberAutoDataAttribute`. This would allow injecting runtime evaluated values instead of being limited to compile time values.

In this case the `InjectedAttribute` would not be a `CustomizeAttribute` and would be used only as a marker for the values to be injected after they've been resolved by the testing framework.

A downside to this is that it might create some confusion when choosing between `[Frozen]` and `[Injected]`, but I think we can mitigate that.

```c#
[Test]
[InlineAutoData(42)]
public void InjectParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected] int value, 
    PropertyHolder<int> holder)
{
    Assert.AreEqual(value, holder.Property);
}

[Test]
[MemberAutoData(nameof(UserData))]
public void InjectParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected] User user, 
    PropertyHolder<User> holder)
{
    Assert.AreEqual(user, holder.Property);
}
```

Kralizek:
@aivascu great idea, I would consider both behaviors as they are not mutually exclusive and both solve some cases.
- parameterless constructor => the behavior you describe (which I don't know how to implement myself)
- value provided in the parameter => current implementation

aivascu:
@Kralizek I can't see a `MemberAutoDataAttribute` in either of the NUnit projects,
but I think it should be possible to implement since NUnit supports a similar behavior via `TestCaseSourceAttribute`.

Kralizek:
> @Kralizek I can't see a `MemberAutoDataAttribute` in either of the NUnit projects,
> but I think it should be possible to implement since NUnit supports a similar behavior via `TestCaseSourceAttribute`.

I might try implementing it after this one :)

ecampidoglio:
I think this would be a great addition to AutoFixture. I also like how it stays consistent with the `[Frozen]` attribute. 👌

Kralizek:
@aivascu and @ecampidoglio do you like the approach I suggested [here](https://github.com/AutoFixture/AutoFixture/pull/1199#issuecomment-696596024)?

aivascu:
@Kralizek I would focus on the first approach only, since it covers the second scenario as well and it might create confusion on which one takes priority when both approaches are used.

```c#
[Test]
[InlineAutoData(42)]
public void InjectParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected(13)] int value, // Here it's not clear whether it's going to be 42 or 13
    PropertyHolder<int> holder)
{
    Assert.AreEqual(value, holder.Property);
}
```

Src/AutoFixture/InjectOnMatchCustomization.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The criteria that determine whether the frozen specimen will be used
+    /// to satisfy a given request are specified at construction time
+    /// as a <see cref="IRequestSpecification"/> object.
+    /// Multiple criteria can be combined together as a boolean expression
+    /// by composing them into a <see cref="AndRequestSpecification"/>
+    /// or <see cref="OrRequestSpecification"/> object.
+    /// </remarks>
+    public class InjectOnMatchCustomization : ICustomization
```
zvirja, line 18: To me this name is misleading a bit. From the name I would expect that something would be injected somewhere when a matching request is coming. However, value is _already_ injected by a fact of adding this customization. From this perspective I would say `ReturnOnMatchCustomization` or `FixedValueOnMatchCustomization` is more clear for the functionality this class provides.

Another aspect is that customization itself is not a universal building block and I don't see how it can be used by clients. If they reference `AutoFixture` directly, then they would just use `fixture.Inject<T>()` method. Personally I think we should add only truly reusable building blocks to the core. Otherwise, it might become just a dump of junk customizations that are needed once in a while in very exotic situations 😟

I would just inline this one-liner in the `InjectedAttribute` directly:

```c#
fixture.Customizations.Insert(0, FilteringSpecimenBuilder(new FixedBuilder(value), this.Matcher));
```

Kralizek: I agree very much with you. I just mirrored how the `Frozen` attribute was implemented :)

Anyway, as per @aivascu comment, we're changing approach. So these changes will be removed in the next commit :)

zvirja:
Just to mention, there is one more way of re-using the existing `Frozen()` customization. We could introduce a new attribute that is just returning a constant value:

```#
void Test([Frozen][Const(42] int arg) {  }
```

This way we don't have to copy-paste logic from the `Frozen`. The downside is that this usage might be not obvious.

If we follow the approach @aivascu suggested above, then let's just re-use the existing `Frozen` attribute without introducing a new `Injected` one. It would still have a perfect semantic meaning:


```c#
[InlineAutoData(42)]
void Test ([Frozen] int value, PropertyHolder<int> boxedValue) { }
```

We might extend this attribute to implement one more interface needed by this feature. But eventually it will lead to the less confustion.

aivascu:
@zvirja I was thinking the same initially, they're both roughly speaking "singleton definitions" and the behavior is very similar but would that be consistent with our DSL?
In terms of  AutoFixture, "Freeze" means that the `Fixture` generates the value and then stores it for future resolutions, while "Inject" means the client provides an existing value to be later be used by the `Fixture` on subsequent resolutions.

zvirja:
@aivascu This is a very interesting perspective, I love this discussion! Personally, somehow I'm consistent in thinking that it's actually OK. We are giving contradicting commands to machine "Hey, please definitely use this constant value" and "Hey, create value and freeze it" while the first has stronger precedence, as it comes from the inline data. Therefore, the only way for machine to combine these contradicting commands would be to use constant and freeze it. There is no other sane way it could behave in this situation 😉 

By saying I'm consistent I mean that I had similar thoughts even couple of years ago, when I created `Injected` attribute in my private projects. I was thinking "damn, it's such an overkill, wish we could better integrate `Frozen` with InlineData`. But the only reason why I didn't do it was that I was afraid of technical challenge and how to make nice architecture around it.

Thinking of it today, I think that we could solve it actually in a beautiful way:
- We resolve `InlineData` values
- We inject customizations, saying that for particular `ParameterInfo` we should return concrete value, coming from point 1
- We create all the method input from AutoFixture, assuming, that it will return pre-configured values for first N parameters.

In this flow attributes like `Frozen` will behave naturally (AutoFixture indeed "created" the value), while creation affecting attributes like `[NoAutoProperties]` or `[Greedy]` would simply be omitted. This way "machine" wouldn't even have internal contradictions and will not feel discomfort about it 😅

Kralizek:
🍿🍿🍿

**Enjoying the conversation**

aivascu:
@zvirja I agree with the technical aspects that you described, but I was referring to whether we're consistent in how we call things and how the code would be read by the user.
Let's take the following example, where we inject a existing instance into the `Fixture`.
```c#
[Theory]
[MemberData(nameof(UserData))]
public void ParameterShouldAssignSameInstanceToPropertyHolder(User user)
{
    var fixture = new Fixture();
    fixture.Inject(user); // <-- we say "Inject" here
    var holder = fixture.Create<PropertyHolder<User>>();
    Assert.AreEqual(user, holder.Property);
}
```
Then refactor the test so the `fixture` is implicit and the `PropertyHolder<User>` instance is generated by the `Fixture`.
```c#
[Theory]
[MemberAutoData(nameof(UserData))]
public void FrozenParameterShouldAssignSameInstanceToPropertyHolder(
    [Frozen] User user, // here we say "Freeze" instead of "Inject",
                        // even though we'll use the data provided by `MemberDataAttribute`
    PropertyHolder<User> holder)
{
    Assert.AreEqual(user, holder.Property);
}
```
Do you see what I mean? When we use `Fixture.Freeze<T>()` the `T` is usually "generated" by the `Fixture` before it becomes a singleton.
It so happens that if the value was previously injected and then frozen, the fixture would still return the original injected value.
If we're to extend the `[Frozen]` attribute we'll be skipping to inform the user that the value was at some point "injected" into the fixture and we'd create an inconsistency where freezing would mean different things depending whether there is a matching parameter provided by the member or not.
```c#
[Theory]
[MemberAutoData(nameof(UserData))]
public void InjectedParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected] User user, // here we are explicitly stating that the value should be provided to the fixture
                          // and the fixture will make no effort in creating it if it was not
    PropertyHolder<User> holder)
{
    Assert.AreEqual(user, holder.Property);
}
```
I appreciate that the implementation we might end up with, could be strikingly similar for both `Frozen` and `Injected`, in the end the features **are** similar, but I think expressing the intent clearly is more important.

Kralizek:
What happens when you have this?

```
[Theory]
[InlineAutoData]
public void InjectedParameterShouldAssignSameInstanceToPropertyHolder(
    [Injected] User user,
    PropertyHolder<User> holder)
{
    Assert.AreEqual(user, holder.Property);
}
```

aivascu:
@Kralizek I see your point. This feature indeed opens up a can of worms. 
Now to answer your questions, I think the example you provided should not even compile since a complex type like `User` cannot be provided by the `InlineAutoDataAttribute`. We can achieve that using analyzers, similar to what xUnit does.
In case when we do not know how many parameters will be provided by the user (e.g. `MemberAutoDataAttribute`), we should fail the test saying `"A User instance was expected to be provided at position {0}, but none was found"`.
In case the users will use the `InjectedAttribute` outside the range of values provided by the `InlineData` or `MemberData` the test should also fail, saying `"The number of values provided was expected to be {0} but actually was {1}."` where `{0}` would be the position the last `InjectedAttribute` was found in the list of parameters.

Kralizek:
@aivascu I like your approach.

PS: it would compile as there is no `User` in the attribute :P

zvirja:
> Do you see what I mean? When we use Fixture.Freeze<T>() the T is usually "generated" by the Fixture before it becomes a singleton.

It's just a matter of the perspective you choose. Look at this from slightly different angle and you will see no contradictions 😉 
Let's take your sample:

```c#
[Theory]
[MemberAutoData(nameof(UserData))]
public void FrozenParameterShouldAssignSameInstanceToPropertyHolder(
    [Frozen] User user, // here we say "Freeze" instead of "Inject",
                        // even though we'll use the data provided by `MemberDataAttribute`
    PropertyHolder<User> holder)
{
    Assert.AreEqual(user, holder.Property);
}
```

First, we evaluate the `[MemberAutoData(nameof(UserData))]` statement. What is actually does can be pseudo translated to following:
- Configure Fixture to return the value from the `UserData` member for the first parameter: `ParameterInfo User user`
- Use Fixture to auto-generate input, taking into account the pre-configuration we did.

Later we evaluate the first parameter: `[Frozen] User user`. We tell Fixture to generate and pin the value for such type. However, there is another customization in place, which already pre-configured this parameter and returns fixed value. So eventually we get the same object being returned and pinned by criteria.

Look at this like injection happens not at `[Frozen]` attribute, but actually earlier inside the `MemberAutoData` attribute. Later `Frozen` is honestly resolving value from fixture, there is no change to its default behavior.

With the very rough approximation we have scenario like

```c#
    var fixture = new Fixture();
    fixture.Inject(user); // <-- Done by MemberAutoData* (*actually it pins concrete ParameterInfo)
    fixture.Freeze<User>(); // <-- Done by the [Frozen] attribute
    var holder = fixture.Create<PropertyHolder<User>>(); // <-- Done by attribute resolution
    Assert.AreEqual(user, holder.Property);
```

You might say that it's a bit weird to call `Freeze()` after `Inject()`, but so be it - it works 😊 

P.S. 
>  When we use Fixture.Freeze<T>() the T is usually "generated" by the Fixture before it becomes a singleton.

BTW, technically that's not necessarily true. If you have customization in place which is returning you customized value, freezing does not go through the full pipeline. What "freeze" actually does, it "resolves" value though the fixture - not necessarily generates something new.

> If we're to extend the [Frozen] attribute we'll be skipping to inform the user that the value was at some point "injected" into the fixture and we'd create an inconsistency where freezing would mean different things depending whether there is a matching parameter provided by the member or not.

You just saw revealed the dark side of `Freeze` - it never promised us to generate something new and pin it - it just promised to "resolve" it and pin it with full respect of other customizations. It never cared about previous injections and was never meant to tell users something about them.

Kralizek:
@aivascu did we kill this? 🤔

aivascu:
@Kralizek we definitely didn't.
Sorry for my absence. I had a few weeks of crunch at my day job.
Just starting to pick things up from where I left them.
I will get back on this PR later this week.

Kralizek:
@aivascu no worries. I was just curious if we had reached a consensus about this. I know real life can be a b***h sometimes. :)

aivascu:
@zvirja I have reviewed again the entire conversation and I think can see now the reason behind what you were saying.
You were viewing the method level attributes and the parameter level attributes as two distinct steps, while I was seeing it as a single step, so to speak.
I think we can go with reusing the `FreezeAttribute` for the purpose of this feature and then see how it goes.

@Kralizek I think answers the question if we have a consensus on the topic. 🙂 I will attempt my own implementation of it just to see if there are any roadblocks or if I have any ideas that might complement your implementation.

aivascu:
This has since been implemented in #1433 for XUnit 2 and will soon be implemented for NUnit 3 and 4.

PR #1198, opened 2020-09-17
Enables running NUnit2 tests from dotnet cli

aivascu:
- Enables running NUnit2 tests from VS 2019 and dotnet cli
- Upgrades the minimum NUnit2 version to `v2.7.1`
- Adds reference to `NUnitV2.Core` package
- Removes `References` directory

PR #1204, opened 2020-10-04
Revert "Enables running NUnit2 tests from dotnet cli"

aivascu:
Reverts AutoFixture/AutoFixture#1198 due to breaking SemVer policies


zvirja:
@aivascu Should we revert it entirely, or there is a way to still run it on .NET Core for that old version?

zvirja (approved):

aivascu:
@zvirja the older versions do not work with .net core cli. The test runner was updated recently for v2, just for that purpose.

zvirja:
@aivascu Can't we just update NUnit version for the unit test project only? It should be backward compatible and must work.

zvirja:
Anyway, don't want this to take much of your time. If you find it easier to just cherry-pick fix for v5, just go with it.

PR #1206, opened 2020-10-25
Update build script to support GitVersion and buid time versioning

aivascu:
- Replaces custom version parsing with GitVersion
- Refactors existing targets
- Changes the location of build artifacts


aivascu:
@zvirja could you have a look at this PR? I'd like to know your opinion on this.

zvirja:
Hi! I'll try to review it in upcoming days. I gave it a quick look and found that you rewrote the existing script a lot.

In the meanwhile could you please describe which problems you are solving with the current PR? For instance, I see you switched to GitVersion. Did you find any limitation with the current versioning schema? Are you enabling some new scenarios we are not currently supporting?

That will help me to better understand motivation and the changes themselves. Thanks!

aivascu:
@zvirja the intent behind this PR was to make the build script more maintainable (and less intimidating as it was for me 🙂) by removing as much infrastructure code as possible from the main script file.
As you can see I've moved some infrastructure to external files inside the `build` directory since it's name was more intuitive to look for partial build scripts. This means I had to move the build artifacts to another folder which is `artifacts` in this case.

On the `GitVersion` topic. I should say this is my personal preference, there was nothing particularly wrong with the old code.
Since we had a lot of code to calculate the version of the build, and since there are tools that already specialize in this, I considered it would be a good idea to replace the versioning with code we don't have to maintain.

Kralizek:
@aivascu if FAKe is not a requirement, over time I have reached quite a satisfying setup with Cake that you can look at. Check any of my public repos, they all have the same script (more or less). I also use AppVeyor so the setup is very easy to move here.

aivascu:
@Kralizek don't get me wrong the current build script is OK and FAKE is a decent tool, once you learn some F# and get used to the API. The current build script has served AutoFixture pretty well so far, so there's no reason to replace it.
With this PR I was just trying to fix a bug in the `vNext` builds and restructure the script a bit so it's easier to read.
In essence the script has mostly remained the same as before.
Now how successful I was in making the script better remains to be seen after the code review.
In any case, thank you for the offer! 🙂

zvirja (approved):
Sorry for the very late reply. It was very rude of me to make you wait that long. Was really hard to find passion to sit one evening and review it 😟 

I've skimmed through the changes and in general like the changes you did. After playing a bit more with Nuke in my other repos (thank you for the Nuke introduction 💙) now I think that it does not worth it to maintain these scripts in FAKE. Nuke is so convenient and handy, that FAKE can barely compete with it.  So if you have even a slight idea to switch to Nuke, I would pretty much support it! 👍 

The only concern I have is regarding the usage of GitVersion. I understand that it's a pretty much standard, but personally for me it always worked in a strange and unexpected way.

Examples:
- `master` branch, NuGet package version. Before: `4.14.0.3`, After: `5.0.0`
- `v5` branch, a few commits, NuGet package version. Before: `5.0.0-alpha.6`, After `5.0.0-preview0001`. After I added few more commits to `v5` : still `5.0.0-preview0001` (in contrast to old versioning where I got `5.0.0-alpha.9`)

The only situation where GitVersion works reasonable is when tag is checked out. In that case both versions are fine.

Back in days I spent a lot of time trying to make GitVersion in a way I want, but without luck. So I ended up with those 50 lines which do exactly what I want. Seems you did not proceed too much in that direction as well :( Therefore I suggest to either ensure that it works fine, or simply keep old version calculation logic.

aivascu:
@zvirja don't worry about it. 🙂 I really appreciate you took the time to review the PR.

I agree with you about Nuke. There are a lot of features, in Nuke, we could benefit from.
I am using Nuke in some of my own repositories so I'm comfortable using it.
I will make a proposal PR for the Nuke build script and tag you, once it's ready.

On the GitVersion topic. I do not deny it can act strange at times.
Most of the times for me, it was due to a missed configuration or a branch created out of place.
I'm fairly sure it should be possible to configure it, to act like the current versioning schema. 
If it cannot be configured that way, we can discuss the trade offs or implement the current schema in Nuke (C#).

I will close this PR for now.

Kralizek:
I would love to see the NUKE script in the making, I am personally a big fan of CAKE :)

zvirja:
@aivascu You can borrow the old versioning logic from [here](https://github.com/zvirja/git-sync-creation-date/blob/2ef6ceb180e1daad9a9db93260204ea9c2e8326a/build/GitBasedVersion.cs#L7), as I use it in a lot of repos.

> Most of the times for me, it was due to a missed configuration or a branch created out of place.
> I'm fairly sure it should be possible to configure it, to act like the current versioning schema.

Yes, that's were exactly my thoughts. But then I gave up after I spent a few days on that (and even debugging the GitVersion source code, I took it really seriously 😅). If you can make it - please show me an example 😃 

@Kralizek You can find an example e.g. [here](https://github.com/zvirja/git-sync-creation-date). The idea is really fascinating - your build script is just a regular .NET console app. It can be run as a console app. It's developed as a console app. It can be debugged as a console app. And of course we have targets and a lot of nice utils around it 😊

PR #1213, opened 2020-12-14
Make FSharp.Core references explicit

aivascu:
The latest version of `FSharp.Core` - `5.0.0` , has dropped support for `.NET Framework`.
All F# projects that reference the `.NET Framework`, even those that have explicit package references to older versions of `FSharp.Core`, are now broken.

This PR:
- Prevents projects from restoring the implicit `FSharp.Core` package (i.e. `v5.0.0`)
- Adds explicit `FSharp.Core` package references to F# projects
- Downgrades `AutoFoqUnitTests` target framework `netcoreapp2.2` to LTS `netcoreapp2.1`
- Adds target framework `netcoreapp2.1` to `FsCheck2UnitTest` project

### DIFF
--- a/Src/Common.props
+++ b/Src/Common.props
@@ -25,6 +25,10 @@
     <PackageIconUrl>https://raw.githubusercontent.com/AutoFixture/AutoFixture/79c882c3f4af3cf52ad43e5c95851f25d217ac17/AutoFixtureLogo200x200.png</PackageIconUrl>
     <PackageIcon>icon.png</PackageIcon>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+
+    <!--Prevents restoring the latest version of FSharp.Core (i.e. FSharp.Core 5.0.0)-->
+    <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
+    <DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
   </PropertyGroup>
   <ItemGroup>
     <None Include="$(MSBuildThisFileDirectory)..\AutoFixtureLogo200x200.png" Pack="true" PackagePath="icon.png"/>
--- a/Src/Common.props
+++ b/Src/Common.props
@@ -25,10 +25,6 @@
     <PackageIconUrl>https://raw.githubusercontent.com/AutoFixture/AutoFixture/79c882c3f4af3cf52ad43e5c95851f25d217ac17/AutoFixtureLogo200x200.png</PackageIconUrl>
     <PackageIcon>icon.png</PackageIcon>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-
-    <!--Prevents restoring the latest version of FSharp.Core (i.e. FSharp.Core 5.0.0)-->
-    <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
-    <DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
   </PropertyGroup>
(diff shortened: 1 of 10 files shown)


aivascu:
@zvirja could you have a look? This PR fixes the `master` branch.

Src/Idioms.FsCheck.FsCheck2UnitTest/Idioms.FsCheck.FsCheck2UnitTest.fsproj
```diff
@@ -20,10 +20,19 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="FsCheck" Version="[2.0.1]" />
     <PackageReference Include="Unquote" Version="4.0.0" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)'=='net452'">
+    <PackageReference Include="FSharp.Core" Version="[4.2.3]" PrivateAssets="All" />
```
zvirja, line 27: I found that we have a mix of "exact" (`[version]`) and "at least" (`version`) specified dependency versions. Is it on purpose? Could we just go with "exact" everywhere in unit tests?

Src/Common.props
```diff
@@ -25,6 +25,13 @@
     <PackageIconUrl>https://raw.githubusercontent.com/AutoFixture/AutoFixture/79c882c3f4af3cf52ad43e5c95851f25d217ac17/AutoFixtureLogo200x200.png</PackageIconUrl>
     <PackageIcon>icon.png</PackageIcon>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+
+    <!--Prevents restoring the latest version of FSharp.Core (i.e. FSharp.Core 5.0.0)-->
+    <DisableImplicitFSharpCoreReference>true</DisableImplicitFSharpCoreReference>
+    <DisableImplicitSystemValueTupleReference>true</DisableImplicitSystemValueTupleReference>
```
zvirja, line 31: I'm a bit worried that we add these properties for all the projects, not F# related only. These properties are very "magic" and it's a bit hard to remember specifics for this particular repo.

Are they strictly necessary for all the projects?

Src/Common.props
```diff
@@ -25,6 +25,13 @@
...
+
+    <!--Prevents failing cross target NuGet package restores-->
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
```
zvirja, line 34: Could you please elaborate a bit what exactly it changes and why we need it? Usually ItJustWorks without this prop 🤔

zvirja (approved):
Thank you for fixing it! I dropped a bit of comments, but feel free to merge as is if you think they are not too much useful and relevant 😊

Src/Common.props
aivascu: @zvirja as far as I can tell, there's a regression in the NuGet CLI deployed on AppVeyor. It manifests in projects that are cross targeting `netcoreapp1.x`. There's more details in this NuGet [issue](https://github.com/NuGet/Home/issues/7414).

Src/Common.props
aivascu: @zvirja I agree, it would make sense to create a props file specific for F# projects, it would make the fix more explicit. I will make the necessary adjustments.

Src/Idioms.FsCheck.FsCheck2UnitTest/Idioms.FsCheck.FsCheck2UnitTest.fsproj
aivascu: @zvirja I went for exact versions on `.NET 4.5` projects, to prevent NuGet from restoring a later version. I suppose we could use the minimum necessary version for each project.

aivascu:
@zvirja I've made the required adjustments.
- Moved the implicit dependency flags to a F# specific props file
- Moved the NuGet fallback configuration to the test specific props file
- Restricted the `FSharp.Core` to exact versions for all F# test projects

aivascu:
I've double checked, there are no changes to the NuGet package dependencies.

PR #1225, opened 2021-02-04
Move build script to NUKE

aivascu:
- [x] Remove FAKE build script
- [x] Add NUKE build script
- [x] Implement coverage report generation
- [x] Implement original versioning schema
- [ ] Implement AppVeyor artifact publishing
- [ ] Implement build validation


Src/AutoFakeItEasy.FakeItEasy4UnitTest/AutoFakeItEasy.FakeItEasy4UnitTest.csproj
```diff
@@ -19,7 +19,7 @@
   <ItemGroup>
     <PackageReference Include="FakeItEasy" Version="[4.0.0]" Condition=" '$(TargetFramework)'=='net452' " />
     <PackageReference Include="FakeItEasy" Version="[4.0.0]" Condition=" '$(TargetFramework)'=='netcoreapp1.1' " />
-    <PackageReference Include="FakeItEasy" Version="[4.9.0]" Condition=" '$(TargetFramework)'=='netcoreapp2.0' " />
+    <PackageReference Include="FakeItEasy" Version="[4.9.0]" Condition=" '$(TargetFramework)'=='netcoreapp2.1' " />
```
Kralizek, line 22: Since .NET Core 2.1 is out of support, wouldn't be better to use `netcoreapp3.1`?

aivascu: I think it's more a question which targets are supported by `FakeItEasy 4.9.0`. But you're right we should move the supported .NET Core versions up. Currently the .NET Core 1 is causing a lot of issues.

ecampidoglio:
I apologize if this piece of feedback comes in a bit late, but, as a member of the @cake-build project, I think it would be more advantageous to implement our build process in Cake. Putting aside any discussions about which one is "better" or "worse", if nothing else, I feel I can provide better support for Cake, being familiar with how it works.

If everyone in @AutoFixture/core agrees, I'll be happy to convert the existing build process to Cake.

aivascu:
@ecampidoglio this has been discussed in #1206 to some extent. Honestly it's hard to say anything without touching on the subject which is better.
The original intent here was to make the build script more accessible and avoid the bus factor. Moving to Cake I think creates more of the same issues we currently have.

ecampidoglio:
@aivascu I reviewed the conversation in #1206, but I couldn't find any direct comparison between Cake and Nuke.

> Moving to Cake I think creates more of the same issues we currently have.

Could you be more specific? Also, let me clarify that I'm not trying to be confrontational here. Just wanting to understand the perceived benefits of using Nuke.

aivascu:
@ecampidoglio don't get me wrong, Cake is good enough tool. I have used it in the past and I'm grateful for it, however my experience with it has not been so smooth. I haven't used Cake for almost a year so my opinions might be outdated.
One of the most impactful things for me is that is uses a special flavor of C#, which is not supported out of the box on any editors and the plugins on VS and VS Code were not working most of the time. NUKE on the other hand has native support on any editor since it uses vanilla C# projects.
Then there's the adapter libraries for things like GitVersion, that were most of the time outdated or had environment requirements. NUKE though has build in support for most tools and there's little need to seek third-party libraries.
Another thing is the documentation. It's true that both tools mostly rely on generating the documentation from source but NUKEs documentation seems more complete.
Finally there are some features that I haven't seen implemented implemented in other tools, like script generation for different CI/CD environments.

zvirja:
I also give my voice to usage of Nuke. I've been using all of them in different projects and Nuke is the best because you have native IDE support for it no matter where you open your solution. There is no need to learn any DSL or install any plugin - you just run it as a regular console app. The same if you need to debug - just put a breakpoint and that's it, as it's just a console app.

After giving it a try I cannot vote for anything else anymore 😊

appveyor.yml
```diff
@@ -1,25 +0,0 @@
-image: Visual Studio 2019
```
zvirja, line 1: Can we still keep the file? Even if this line is a simple proxy to `build.ps1`, it's cool to have it for versioning purposes.

build/Build.cs
```diff
@@ -0,0 +1,164 @@
...
+    IEnumerable<Project> TestProjects => TestPartition.GetCurrent(Solution.GetProjects("*Test"));
+
+    AbsolutePath SourceDirectory => RootDirectory / "src";
+    AbsolutePath ArtifactsDirectory => RootDirectory / "artifacts";
+    AbsolutePath TestResultsDirectory => ArtifactsDirectory / "testresults";
+    AbsolutePath CoverageDirectory => ArtifactsDirectory / "coverage";
+    AbsolutePath ReportsDirectory => ArtifactsDirectory / "reports";
+    AbsolutePath PackagesDirectory => ArtifactsDirectory / "packages";
+
+    Target Clean => _ => _
```
zvirja, line 58: Nice hard cleanup! 👍

build/Build.cs
```diff
@@ -0,0 +1,164 @@
+using System;
```
zvirja, line 1: I don't see AppVeyor specific targets here (e.g. upload of tests results, version action selection, etc). Are you planning to add it later?

zvirja (commented):
I did initial round of review! Looks awesome 🎉 
Please ping me when you fix CI, so I can give it another try.

appveyor.yml
aivascu: Of course I'll amend the commit that removes it.

build/Build.cs
aivascu: Yes, this is still in progress.

appveyor.yml
zvirja: @aivascu Sorry if I was not clear enough in my comment. I meant have this file in repository, instead of configuring build script on AppVeyor side via UI. So we use `appveyor.xml` and if we build e.g. old branch - it still works fine.

Was that what you mean by saying `amend`? 😊

ecampidoglio:
@aivascu @zvirja Thank you for sharing your opinions. I see your points.

As the appointed lead maintainers of AutoFixture, I think you have the right to choose the build automation tool you're most comfortable with. If that happens to be Nuke, then so be it. 🙂

Obviously, I would have rather seen that AutoFixture used Cake, but that's just my preference.

aivascu:
@zvirja I'm not that familiar with AppVeyor. Could you give me a hand with the test results upload?

Kralizek:
@aivascu you can check my repos (like [this one](https://github.com/Kralizek/AWSSecretsManagerConfigurationExtensions/)) where I combine Cake with AppVeyor :)

aivascu:
will reopen after more testing

PR #1236, opened 2021-02-21
Fixes ConstructorInitializedMemberAssertion false positive for bool arguments

### LINKED ISSUE #1235: Idioms: ConstructorInitializedMemberAssertion returns false positive in case of bool types
hello,

I have studied the issues related to ConstructorInitializedMemberAssertion on this repo and I think I found another bug:

given:
```c#
public class MyClass
    {
        public SomeEnum SomeEnumValue { get; }
        public bool IsSomeBool { get; }

        public MyClass(SomeEnum someEnumValue, bool isSomeBool)
        {
            SomeEnumValue = someEnumValue;
        }
    }
```
with the test:

```c#
        private IFixture _fixture;

        [SetUp]
        public void SetUp()
        {
            _fixture = new Fixture();
        }

        [Test]
        public void ReadonlyPropertiesShouldHaveConstructorParameterAssignedValues()
        {
            // Arrange
            var guardClauseAssertion = new ConstructorInitializedMemberAssertion(_fixture);

            var someEnumValueMember = typeof(MyClass).GetProperty(nameof(MyClass.QuickActionItemType));
            var isSomeBoolMember = typeof(MyClass).GetProperty(nameof(MyClass.IsTurnedOn));

            // Act / Assert
            guardClauseAssertion.Verify(typeof(MyClass));
            guardClauseAssertion.Verify(someEnumValueMember);
            guardClauseAssertion.Verify(isSomeBoolMember);
        }
```
the test passes but I would expect it to fail,

when debugging I observed that in both constructor verification call and members verification calls, 
* the first time the constructor is called the enum parameter value is set to "non first value" and the bool parameter is set to true, but only the enum type property is checked
* the second time the constructor is called the enum parameter value is again set to "non first value" but the bool parameter is now set to false and the bool property value is checked, since that default bool value is false then the test will pass 

if I switch the last to statements from the unit test and put the bool check first:

```c#
guardClauseAssertion.Verify(isSomeBoolMember);
guardClauseAssertion.Verify(someEnumValueMember);
```

then everything works as expected and the test fails, but I don't think you should care about the order of the statements

aivascu:
### Changes
- Fixes `ConstructorInitializedMemberAssertion` false positive results for `bool` constructor parameters occurring at even positions

fixes #1235

### DIFF
--- a/Src/Idioms/ConstructorInitializedMemberAssertion.cs
+++ b/Src/Idioms/ConstructorInitializedMemberAssertion.cs
@@ -105,8 +105,7 @@ namespace AutoFixture.Idioms
             if (constructorInfo == null) throw new ArgumentNullException(nameof(constructorInfo));
 
             var parameters = constructorInfo.GetParameters();
-            if (parameters.Length == 0)
-                return;
+            if (parameters.Length == 0) return;
 
             var publicPropertiesAndFields = GetPublicPropertiesAndFields(constructorInfo.DeclaringType).ToArray();
 
@@ -154,7 +153,7 @@ namespace AutoFixture.Idioms
 
             if (!matchingConstructors.Any())
             {
-                if (IsMemberThatRequiresConstructorInitialization(propertyInfo))
+                if (RequiresConstructorInitialization(propertyInfo))
                 {
                     throw new ConstructorInitializedMemberException(propertyInfo, string.Format(
                         CultureInfo.CurrentCulture,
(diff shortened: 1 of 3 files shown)


Src/Idioms/ConstructorInitializedMemberAssertion.cs
```diff
@@ -105,8 +105,7 @@ public override void Verify(ConstructorInfo constructorInfo)
             if (constructorInfo == null) throw new ArgumentNullException(nameof(constructorInfo));
 
             var parameters = constructorInfo.GetParameters();
-            if (parameters.Length == 0)
-                return;
+            if (parameters.Length == 0) return;
```
zvirja, line 108: I'd say that the previous style is better in terms of readability. Otherwise it's easy to miss that there is an eager return from the function.
The only place where I would use oneliners - assertions which are throwing exceptions.

That is though purely my opinion, so if you find this style reasonable to go - fine then ;-)

zvirja (approved):
LGTM 👍 Dropped a single comment about one-liner condition style in conjunction with `return` statements.

PR #1251, opened 2021-03-30
To not override ref/out if was modified to same value

zvirja:
Be more strict on detecting whether the boxed struct value was modified. If we find that user rewrote the value to self (or e.g. re-assigned complex struct which has shallow equality), we should not override the value with auto-generated.

Found this issue when was writing tests with intense usage of `ref/out`.

@aivascu Thank you for the review. It's a small and almost trivial fix, so I don't expect it to break things.


aivascu (approved):
LGTM

PR #1290, opened 2021-09-13
Add support for IParameterDataSource to AutoFixture.NUnit3

### LINKED ISSUE #343: NUnit Values and AutoData doesn't work

Kralizek:
Fixes #343


iamteamstar (changes requested):
Renamed test methods to include type names to avoid NUnit runner conflicts with overloaded method names.
 Aligned Values(...) parameters between AutoData and InlineAutoData versions for consistency.
 Verified [Frozen] + [Values] + [InlineAutoData] combination behaves as expected when no inline values are provided.

PR #1309, opened 2021-12-01
Fixed SelectMethods when exists more than one method with the same name

riwexoyd:
I found an exception when i was trying to add ReadonlyCollectionPropertiesBehavior to fill Protobuf properties.


riwexoyd:
Well, I found another case: if we use Dictionary<>, then my solution is not working. Previous solution is also not working. The problem is that Add(KeyValuePair<>) from ICollection<KeyValuePair<>> is explicit method.

PR #1331, opened 2022-02-10
Target framework upgrade

### LINKED ISSUE #1324: ✨ Add support for .NET Standard 2.1
### Description

Add support for .NET Standard 2.1 in order to enable newer language features.

### Example

N/A

### Tasks

- [x] Confirmed feature idea with maintainers in support forums

### More Information

Thanks!

aivascu:
### Description
- Raised minimum .NET Framework target version to v4.6.1
- Added netstandard2.1 target framework to AutoFixture
- Added netcoreapp3.1 target framework to AutoFixture test projects
- Removed test projects for AutoFakeItEasy v2-5 (inclusive)
- Raised minimum supported version of FakeItEasy to v6
- Added `Microsoft.Bcl.HashCode` as a dependency
- Fixed validation errors for `GetHashCode()` using `HashCode`

### Closed issues

fixes #1324

### Checklist


Kralizek:
Since you're open to bcl packages, you could also take this one to support IAsyncEnumerable in netstandard2.0

https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/

aivascu:
@Kralizek yes, this was one of the reasons to upgrade to net461 specifically and deprecate netstandard1.x. If I didn't, someone would have raised soon enough an issue asking for support of `IAsyncEnumerable` on net48.

aivascu:
@zvirja thank you for agreeing to review this PR, however this PR has been waiting for a while. I'm planning to merge it today in the evening, if all goes to plan.
If you'll have any comments on the PR please let me know. In case you'll find something critical we'll be able to revert the changes since this is a preview branch.

zvirja:
@aivascu finally got time to look at it. Looks great. Nice work 😊 

A couple of suggestions:

- I suggest to test agains `net6 + net461` only in unit tests. That's because `2.1` is already out of support and `3.1` will be out of support at the end of a year. So let's have NET6 to be up to date as long as possible.
- I suggest to also target `net6` for AutoFixture to enable new API and make dependency flow smaller when targeting framework directly

Kralizek:
@zvirja are you sure you're not mixing .NET Core 2.1 with .NET Standard 2.1? AFAIK standards have no support per se, hence they can never be out of support 🤔

aivascu:
@zvirja thank you for the review! 🎩

@Kralizek I think Alex means that tests should only reference net6 and net461. 

I would like to agree, but I'm not sure we could do it. Explicit target framework specification for referenced projects doesn't work in multi targeting projects, afaik.

So we either have to create copies of test projects that are targeting a single framework or give up on testing netstandard versions of the library.

Do you guys have any experience with that? I'd love to get rid of legacy frameworks.

Kralizek:
Ohh i totally misread it. Never comment on issues before the morning coffee 😜

My to go solution would have been explicit runtime framework but it seems you already tried it

zvirja:
@aivascu Okay, I believe I see what you mean. You use obsolete versions of runtime to test specifically `netstandard2.0` support.

We have a couple of scenarios here:

- Libraries like `AutoFixture.xUnit2` which target `net461 & netstandard2.0`. That is the majority of projects.
- Libraries like `AutoFixture` which multi-target `net461 & netstandard2.0 & netstandard2.1`

I also read this article to align my understanding of the domain: https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/.

With that I propose the following.

- Update all the libraries to additionally target `net6` explicitly. That will be 99% of all the use cases in the nearby future and it allow to simplify the dependency graph for those who target it explicitly.
- Given that[ .NET Framework 4.6.1 supports .NET Standard 2.0](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0) I suggest to drop explicit full framework support and provide it though .NET Standard. In practice it would not make any difference (does it?)
- I suggest to not expand our support to `netstandard2.1`. Instead focus solely on `net6` framework for new features (like `IAsyncEnumerable`). This way we support all modern platforms going forward and support will be easier.

This way in all our unit tests we could target `<TargetFrameworks>net461;net6</TargetFrameworks>` - so we test both `netstandard2.0` branch and `net6`.

Do you see issues with this approach?

Kralizek:
I just want to add that, by direct experience, `net461`'s support of .NET Standard 2.0 is a lie.

From your [same link](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0):
> The versions listed here represent the rules that NuGet uses to determine whether a given .NET Standard library is applicable. While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

That being said, not supporting `netstandard2.1` directly or indirectly via `netcoreapp3.1` (out of support) or `net5.0` (phasing out of support) might leave a lot of users of this library without support for runtime functionalities.

I was hit first hand because `IAsyncEnumerable` needs to be explicitly handled (see #1322 )

How about making AF5 support the latest supported of each branch?
- .NET Framework: i.e. `net48`
- .NET LTS: i.e. `net6`
- .NET Standard: i.e. `netstandard2.1`

PR #1402, opened 2023-09-05
ReadonlyCollectionPropertiesBehavior conversion bug fixed

doboczyakos:
Fixed InvalidCastException: Object must implement IConvertible. when using ReadonlyCollectionPropertiesBehavior


doboczyakos:
@dotnet-policy-service agree

PR #1404, opened 2023-09-05
ReadonlyCollectionPropertiesBehavior conversion bug fixed

doboczyakos:
Fixed InvalidCastException: Object must implement IConvertible. when using ReadonlyCollectionPropertiesBehavior

### DIFF
--- a/Src/AutoFixture/Kernel/EnumerableEnvy.cs
+++ b/Src/AutoFixture/Kernel/EnumerableEnvy.cs
@@ -9,7 +9,7 @@ namespace AutoFixture.Kernel
     {
         public static IEnumerable<object> ConvertObjectType(this IEnumerable<object> enumerable, Type type)
         {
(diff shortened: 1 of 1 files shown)


iamteamstar (approved):
Before Convert.ChangeType was always called, causing NullReferenceException for null values
and InvalidCastException when the object was already assignable to the target type.
After Added a guard condition — skip conversion if value is null or already assignable.

zvirja:
Hey! Thanks for the contribution!

Could you please add a small unit test to demonstrate the fix? Could you also rebase on top of `release/5.0.0`, as we are not going to publish new v4 versions for now.

thanks!

zvirja:
@doboczyakos I was trying to repro it, but struggled. So would appreciate if you drop a line with a scenario. The fix looks reasonable, I just want to have an example of the error

doboczyakos:
@iamteamstar @zvirja Sorry, I don't remember, it was years ago. I had this problem with one of our unit tests, so it had surely happened. Maybe try to add a property with type List< T > where T does not implement IConvertible

Issue #1421, opened 2023-11-29
✨ Accept `ISpecimentBuilder` as value generator for `With`


Kralizek:
### Description
I have to work with a class with a string property whose valid value is limited to few option (a missed `enum`).

I can use `ElementsBuilder<string>` to do so but the configuration is a bit cumbersome.

It would be great if `With` accepted `ISpecimenBuilder` directly.

### Example

```csharp
fixture.Customize<SomeProduct>(o => o
    .With(p => p.OperatingSystem, (IFixture f) => f.Build<string>().FromFactory(new ElementsBuilder<string>(new[]{"ios", "windows", "mac"})).Create())
);
```

could be written as

```csharp

// IPostprocessComposer<T> With<TProperty, TInput>(Expression<Func<T, TProperty>> propertyPicker, ISpecimenBuilder valueFactory);
fixture.Customize<SomeProduct>(o => o
    .With(p => p.OperatingSystem, new ElementsBuilder<string>(new[]{"ios", "windows", "mac"}))
);
```

aivascu:
Implemented in #1444. Available after AutoFixture v5.0.0-preview009

PR #1444, opened 2024-02-27
Adding ISpecimenBuilder as param on With method

luccanog:
### Description
Accepting `ISpecimentBuilder` interface as value generator for `With` method:

``` c#
  ISpecimenBuilder specimenBuilder = new ElementsBuilder<string>("ios", "windows", "linux");

  var foo = _fixture.Build<Data>()
                   .With(d => d.Name, specimenBuilder)
                   .Create();
```

### Closed issues
#1421

### Checklist

### DIFF
--- a/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
@@ -174,6 +174,15 @@ namespace AutoFixture.Dsl
         }
 
         /// <inheritdoc />
+        public IPostprocessComposer<T> With<TProperty>(Expression<Func<T, TProperty>> propertyPicker, ISpecimenBuilder builder)
+        {
+            return (CompositeNodeComposer<T>)this.ReplaceNodes(
+                with: n =>
+                    (NodeComposer<T>)((NodeComposer<T>)n).With(propertyPicker, (IFixture f) => f.Build<TProperty>().FromFactory(builder).Create()),
+                when: n => n is NodeComposer<T>);
+        }
+
+        /// <inheritdoc />
         public IPostprocessComposer<T> WithAutoProperties()
         {
             return (CompositeNodeComposer<T>)this.ReplaceNodes(
--- a/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositeNodeComposer.cs
@@ -178,7 +178,7 @@ namespace AutoFixture.Dsl
         {
             return (CompositeNodeComposer<T>)this.ReplaceNodes(
                 with: n =>
-                    (NodeComposer<T>)((NodeComposer<T>)n).With(propertyPicker, (IFixture f) => f.Build<TProperty>().FromFactory(builder).Create()),
+                    (NodeComposer<T>)((NodeComposer<T>)n).With(propertyPicker, builder),
                 when: n => n is NodeComposer<T>);
         }
 
--- a/Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
+++ b/Src/AutoFixture/Dsl/CompositePostprocessComposer.cs
@@ -82,6 +82,13 @@ namespace AutoFixture.Dsl
         }
 
         /// <inheritdoc />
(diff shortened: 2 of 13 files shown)


luccanog:
@dotnet-policy-service agree

luccanog:
@aivascu  The pipeline is failing due to previous .NET versions, and I've noticed recent PRs adding .NET 5 runtime like on #1443 being merged to the release branch.

May this PR be based on [release/5.0.0](https://github.com/AutoFixture/AutoFixture/tree/release/5.0.0) instead of the master branch? Couldn't find info about it on the code docs.

I was able to build, test, and pack it locally but the cmd script fails.

aivascu:
@luccanog the errors thrown by the pipeline are related to the SDK building the code. Looks like AppVeyor have updated their VS 2019 image. I'll add the errors to ignore list.

aivascu:
Looks like they outright removed .net core 1 and 2, and I've just finished re-balancing all versions on the release branch...
@luccanog could you rebase your PR on the release branch and target that instead of the master?

luccanog:
@aivascu The rebase is done. I've noticed an update on the `Readme` file, please let me know if I should undo those changes or if I should keep them.

aivascu:
@luccanog yes please undo the readme. I'll synchronize them in a separate PR.

luccanog:
@aivascu I came back with the old content, but for some reason git hub still thinks there is a difference between files. It might be irrelevant to the process

aivascu:
@luccanog I've rebased and squashed your commits on the release branch. I hope you don't mind.

I've had a brief look at your implementation, but I'm not done reviewing it for now.
I'm not convinced that piggybacking on the factory function overload is the way to go. As a workaround it might work fine, but there must be a more efficient way to compose the nodes than that.

Will let you know if I come up with anything.

luccanog:
@aivascu No problem at all, feel free to do any other rebase if needed.

About he piggybacking, do you mean the changes at the [NodeComposer](https://github.com/AutoFixture/AutoFixture/blob/7aa56019d6a070067d66afb48e70142719c1fac6/Src/AutoFixture/Dsl/NodeComposer.cs#L197) class? There I re-used an existent overload since the method has got what we need, but of course we have some other ways. As I see, we could call the `WithCommand` method directly at our new `With` method, and still use the same public [BindingCommand signature](https://github.com/AutoFixture/AutoFixture/blob/master/Src/AutoFixture/Kernel/BindingCommand.cs#L67).

In this way our new `With` method is going to be pretty similar to the [existent one](https://github.com/AutoFixture/AutoFixture/blob/7aa56019d6a070067d66afb48e70142719c1fac6/Src/AutoFixture/Dsl/NodeComposer.cs#L185), but with a conversion from `ISpecimenBuilder` to a `Func<TInput, TProperty>`.

Please let me know when you got any thoughts.

aivascu:
@luccanog yes, I meant that you reused the builder syntax, to create a factory that's using the specimen builder. 😅
Sorry I should be more explicit.

Using the BindingCommand sounds better however I propose that we add a new overload for the public constructor.

```cs
public BindingCommand(Expression<Func<T, TProperty>> propertyPicker, ISpecimenBuilder builder)
{
    if (propertyPicker is null) throw new ArgumentNullException(nameof(propertyPicker));
    if (builder is null) throw new ArgumentNullException(nameof(builder));

    this.Member = propertyPicker.GetWritableMember().Member;
    this.ValueCreator = c => (TProperty)builder.Create(typeof(TProperty), c);
}
```
This would allow to pass the context directly into the builder and avoid going through several layers of abstractions.
I tried it locally and it seems to be working fine.

Let me know if you'll try these changes, and cover them with tests.

luccanog:
I just implemented the new `BindindCommand` conStructor and added more tests as requested. Seems to work properly.
Now we don't have to re-use the `With` method overload, and we can directly construct the BindingCommand with a `ISpecimenBuilder`

aivascu:
@luccanog after some consideration, I decided I don't want yet to couple `BindingCommand` to `ISpecimenBuilder`, so I reverted the changes and used the existing constructor.

luccanog:
@aivascu alright. Thank you!

Issue #1489, opened 2025-02-13
Make NoSpeciment singleton


Griboedoff:
### Describe the change
Hi, i can't decide is that a bug or idea, so feel free to change issue type if you want to.
I've seen old(2014ish) issue about fixture perfomance optimizations. One of the idea was to use singleton instance of NoSpecimen, since it has no  meaningfull fields.

### Scenario
just use fixture in tests

### Expected Behavior
almost no NoSpecimen instances

### Tasks

- [ ] Updated all AutoFixture and related libraries
- [ ] Confirmed in support forums if behavior is expected

### Screenshots

I have up to 30% of NoSpecimen in all created by Fixture objects.
![Image](https://github.com/user-attachments/assets/45b395d0-1ad4-45f6-9051-6aa0a95fc1cd)

Thanks!

sensslen:
I would love to make use of this as I noticed something very similar happening in my tests as well. Is there a chance we can get a release with #1511?

PR #1505, opened 2025-04-15
Removed obsolete NoSpecimen(object) constructor overload

jzabroski:
### Description
NoSpecimen constructor has been marked as obsolete for 10 years.

### Closed issues
#475

### Checklist

### DIFF
--- a/Src/AutoFixture/Kernel/NoSpecimen.cs
+++ b/Src/AutoFixture/Kernel/NoSpecimen.cs
@@ -14,38 +14,12 @@ namespace AutoFixture.Kernel
     /// </remarks>
     public class NoSpecimen : IEquatable<NoSpecimen>
     {
-        private readonly object request;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="NoSpecimen"/> class with a null request.
         /// </summary>
         public NoSpecimen()
         {
         }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="NoSpecimen"/> class with the supplied
-        /// request.
-        /// </summary>
-        /// <param name="request">
-        /// The original request that prompts the creation of this instance.
-        /// </param>
-        [Obsolete("Use NoSpecimen() instead of NoSpecimen(object). The Request property, and this constructor that populates it, is being retired in future versions of AutoFixture, as it has turned out that no one uses it. You can still use the parameterless NoSpecimen constructor overload. The NoSpecimen class itself will remain. If you're seeing this warning in AutoFixture 3.x, and, despite expectations, have a real need to use the Request property, please provide feedback on https://github.com/AutoFixture/AutoFixture/issues/475 .", true)]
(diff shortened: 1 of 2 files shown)


Src/AutoFixture/Kernel/NoSpecimen.cs
```diff
@@ -93,10 +65,6 @@ public bool Equals(NoSpecimen other)
             {
                 return false;
             }
```
aivascu: this method should probably be
```c#
return other is { };
```

aivascu:
I renamed the PR to something more descriptive, since it's going to end up in the commit history.
Hope you don't mind.

jzabroski:
Will fix the build. Sorry, was moving between GitHub desktop and Github codespaces. Must've copy-pasted something wrong.

Src/AutoFixture/Kernel/NoSpecimen.cs
jzabroski: Will make it `return other is not null;` as I think that is more readable idiom.

jzabroski (commented):
Fixed the build; also changed Equals method to not add a reference equality given it is dominated by the pattern match check that the object being compared with is also a NoSpecimen.

jzabroski:
Ideally, NoSpecimen would be a singleton, to avoid extra allocations.

I would also like to submit a separate PR to remove net462 and net5.0 support, as those are no longer really supported by Microsoft and most developers do not install the SDKs for these on their machines. Anyone who needs to use AutoFixture on these older platforms can likely reference 4.x

aivascu:
@jzabroski the singleton `NoSpecimen` is already approved for development #1489. If you would like to raise a PR, it would be welcome, however the (non-obsolete) public parameterless constructor must remain available, at least for now.

As for the `net462` and `net5.0`, I generally agree but I'd like to publish at least one major release supporting these versions. There are several reasons to that:
- In my experience, there's a lot of enterprise software still targeting these versions.
- `net462` and to some degree `net5.0` are historically special. 
  - `net462` is the latest version of .NET Framework not properly supporting `netstandard2.0`.
  - `.net5.0` is the first non-core version that introduced major language features in C#.
- A few libraries that AutoFixture has integrations with, like Moq, and NSubstitute, still target `net462`.
- AutoFixture v5 contains some important bug-fixes and additions, meaning it's not as easily substituted with v4.
- Except for a couple preprocessor directives and a few Kb in the NuGet package they shouldn't be an issue.
- AutoFixture has no impact over the actual runtime version used in test projects/applications. To AutoFixture these versions are just contracts enabling some language features.

jzabroski:
> A few libraries that AutoFixture has integrations with, like Moq, and NSubstitute, still target net462

As my law professor once quipped 

Sometimes society pulls the Law along, and sometimes the Law pulls society along.

But, I understand the idea of AutoFixture 5.0 as you describe it.

Issue #1339, opened 2022-04-08
🐛 ReadonlyCollectionPropertiesBehavior does not select collection Add method correctly


ArturDorochowicz:
### Describe the Bug

`ReadonlyCollectionPropertiesBehavior` does not select collection `Add` method correctly.

The `ReadonlyCollectionPropertiesCommand` class tries to select any method called `Add`: 
https://github.com/AutoFixture/AutoFixture/blob/730b47884975d0c1256209073a719a9fe510d8a9/Src/AutoFixture/Kernel/ReadonlyCollectionPropertiesCommand.cs#L57-L62

This is incorrect in two ways: 
1) There may be more than one `Add` method which then causes an exception from `SingleOrDefault`
2) It finds any method called `Add`. It should likely specifically select `ICollection<>.Add` (though that's probably still not fully specific as the type may implement multiple `ICollection<>`)

### Scenario

In this scenario the readonly collection property is a `Dictionary<,>` which has multiple `Add` methods.

```csharp
using System.Collections.Generic;
using System.Linq;
using AutoFixture;
using Xunit;

public class Bugs
{
    class MultipleAddMethodsModel
    {
        public Dictionary<int, int> Items { get; } = new();
    }

    [Fact]
    public void MultipleAddMethods()
    {
        var fixture = new Fixture();
        fixture.Behaviors.Add(new ReadonlyCollectionPropertiesBehavior());

        // Throws
        fixture.Create<MultipleAddMethodsModel>();
    }
}
/*
System.InvalidOperationException
Sequence contains more than one element
   at System.Linq.ThrowHelper.ThrowMoreThanOneElementException()
   at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Boolean& found)
   at AutoFixture.Kernel.ReadonlyCollectionPropertiesCommand.Execute(Object specimen, ISpecimenContext context)
   at AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
   at AutoFixture.Fixture.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.SpecimenContext.Resolve(Object request)
   at AutoFixture.Kernel.SeedIgnoringRelay.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
[...]

Issue #1340, opened 2022-04-08
🐛 ReadonlyCollectionPropertiesBehavior does not account for collection property being null


ArturDorochowicz:
### Describe the Bug

`ReadonlyCollectionPropertiesBehavior` does not account for the collection property being `null`.

The code in `ReadonlyCollectionPropertiesCommand` retrieves the value of the collection property and  immediately passes it to the `InstanceMethodQuery` constructor which throws `ArgumentNullException` when the value is `null`.

https://github.com/AutoFixture/AutoFixture/blob/730b47884975d0c1256209073a719a9fe510d8a9/Src/AutoFixture/Kernel/ReadonlyCollectionPropertiesCommand.cs#L57-L63

### Scenario

```csharp
using System.Collections.Generic;
using System.Linq;
using AutoFixture;
using Xunit;

#nullable enable

public class Bugs
{
    class NullCollectionModel
    {
        public Dictionary<int, int> Items { get; } = default!;
    }

    [Fact]
    public void NullCollection()
    {
        var fixture = new Fixture();
        fixture.Behaviors.Add(new ReadonlyCollectionPropertiesBehavior());

        // Throws
        fixture.Create<NullCollectionModel>();
    }
}
/*
System.ArgumentNullException
Value cannot be null. (Parameter 'owner')
   at AutoFixture.Kernel.ReadonlyCollectionPropertiesCommand.Execute(Object specimen, ISpecimenContext context)
   at AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
   at AutoFixture.Fixture.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.SpecimenContext.Resolve(Object request)
   at AutoFixture.Kernel.SeedIgnoringRelay.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.Postprocessor`1.Create(Object request, ISpecimenContext context)
   at AutoFixture.AutoPropertiesTarget.Create(Object request, ISpecimenContext context)
   at AutoFixture.Kernel.CompositeSpecimenBuilder.Create(Object request, ISpecimenContext context)
[...]

PR #1508, opened 2025-07-04
Bugfix/1339+1340 readonly collection properties behavior fixes

bestsoftever:
### Description
This PR fixes issues related to `ReadonlyCollectionPropertiesBehavior`:
- selecting wrong `Add` method for generic dictionary;
- exception when collection is `null`.

### Closed issues
#1339 #1340 

### Checklist

- [x] Provided inline documentation comments for new public API N/A

### DIFF
--- a/Src/AutoFixture/Kernel/InstanceMethodByParametersCountQuery.cs
+++ b/Src/AutoFixture/Kernel/InstanceMethodByParametersCountQuery.cs
@@ -6,18 +6,18 @@ using System.Reflection;
 namespace AutoFixture.Kernel
 {
     /// <summary>
-    /// Selects a method by its name and parameters count. 
+    /// Selects a method by its name and parameters count.
     /// Takes into account all instance methods, also private ones.
     /// </summary>
-    public class MethodByParametersCountQuery : IMethodQuery
+    public class InstanceMethodByParametersCountQuery : IMethodQuery
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="MethodByParametersCountQuery"/> class.
+        /// Initializes a new instance of the <see cref="InstanceMethodByParametersCountQuery"/> class.
         /// </summary>
         /// <param name="owner">The instance that should be selected from.</param>
         /// <param name="methodName">The name of the method to select.</param>
         /// <param name="parametersCount">The number of parameters of the method to select.</param>
-        public MethodByParametersCountQuery(object owner, string methodName, int parametersCount)
+        public InstanceMethodByParametersCountQuery(object owner, string methodName, int parametersCount)
         {
             this.Owner = owner ?? throw new ArgumentNullException(nameof(owner));
             this.MethodName = methodName ?? throw new ArgumentNullException(nameof(methodName));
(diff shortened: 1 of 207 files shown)


bestsoftever:
@dotnet-policy-service agree [company="{your company}"]

bestsoftever:
@dotnet-policy-service agree

Src/AutoFixture/Kernel/InstanceMethodByParametersCountQuery.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Selects <see cref="MethodName"/> with <see cref="ParametersCount"/> from <see cref="Owner"/>.
+        /// </summary>
+        /// <param name="type">Discarded.</param>
+        /// <returns>Returns an empty enumerable if <see cref="MethodName"/> does not belong to <see cref="Owner"/>;
+        /// returns an enumerable containing a single <see cref="InstanceMethod"/> otherwise.</returns>
+        public IEnumerable<IMethod> SelectMethods(Type type = default)
+        {
+            var typeInfo = this.Owner.GetType().GetTypeInfo();
```
aivascu, line 50: The same can be written in a more functional way with a bit less duplication.

```c#
var typeInfo = this.Owner.GetType().GetTypeInfo();
var allMethods = typeInfo.GetMethods(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);

var methodsMatchingName = allMethods.Where(m => m.Name == this.MethodName);

var methodsEndingWithName = allMethods
    .Where(m => m.Name.EndsWith($".{this.MethodName}", StringComparison.InvariantCultureIgnoreCase));

return methodsMatchingName.Concat(methodsEndingWithName)
    .Where(m => m.GetParameters().Length == this.ParametersCount)
    .Select(method => new InstanceMethod(method, this.Owner))
    .Take(1).ToArray();
```

Src/AutoFixture/Kernel/InstanceMethodByParametersCountQuery.cs
```diff
@@ -0,0 +1,66 @@
...
+            var allMethods = typeInfo.GetMethods(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
+
+            var method = allMethods
+                .Where(m => m.Name == this.MethodName && m.GetParameters().Length == this.ParametersCount)
+                .FirstOrDefault()
+                ?? allMethods
```
aivascu, line 56: In which situations the methods will be matched by these queries? Obviously it depends on what was passed to the `MethodName`, but in which situation we would get a `MethodName` without its fully qualified name, or not matching the case?

bestsoftever: The thing is that the algorithm from `ReadonlyCollectionPropertiesCommand` requires not the `Add` method which is publicly visible in the `Dictionary<K, V>` class, but the other one, which is explicitly implemented from the `ICollection<>` interface.

```
allMethods[11].Name
"Add"
allMethods[12].Name
"System.Collections.Generic.ICollection<System.Collections.Generic.KeyValuePair<TKey,TValue>>.Add"
```
So my idea was to:
1. try to find `Add` with a single parameter - this will work for most of collections.
2. if not, then try to find an explicit implementation - this will work for dictionaries.

PR #1511, opened 2025-07-22
Make NoSpecimen singleton

bestsoftever:
### Description
This is a performance change that saves memory by avoiding creation of a large number of NoSpecimen instances.

### Closed issues
#1489

### Checklist

### DIFF
--- a/Src/AutoFixture/Kernel/NoSpecimen.cs
+++ b/Src/AutoFixture/Kernel/NoSpecimen.cs
@@ -17,7 +17,9 @@ namespace AutoFixture.Kernel
         /// <summary>
         /// The singleton instance of <see cref="NoSpecimen"/>.
         /// </summary>
+#pragma warning disable 0618
         public static NoSpecimen Instance { get; } = new NoSpecimen();
+#pragma warning restore 0618
 
         /// <summary>
         /// Initializes a new instance of the <see cref="NoSpecimen"/> class.
--- a/Src/AutoFixture/Kernel/NoSpecimen.cs
+++ b/Src/AutoFixture/Kernel/NoSpecimen.cs
@@ -17,12 +17,13 @@ namespace AutoFixture.Kernel
         /// <summary>
         /// The singleton instance of <see cref="NoSpecimen"/>.
         /// </summary>
-        public static readonly NoSpecimen Instance = new NoSpecimen();
+        public static NoSpecimen Instance { get; } = new NoSpecimen();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="NoSpecimen"/> class.
         /// </summary>
-        private NoSpecimen()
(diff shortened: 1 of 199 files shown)


Src/AutoFixture/Kernel/NoSpecimen.cs
```diff
@@ -12,12 +12,17 @@ namespace AutoFixture.Kernel
...
+        /// <summary>
+        /// The singleton instance of <see cref="NoSpecimen"/>.
+        /// </summary>
+        public static readonly NoSpecimen Instance = new NoSpecimen();
+
         /// <summary>
         /// Initializes a new instance of the <see cref="NoSpecimen"/> class.
         /// </summary>
-        public NoSpecimen()
+        private NoSpecimen()
```
aivascu: Not sure if I want to make this constructor `private`, yet. This is one of the most used types in AutoFixture, and changing it would instantly break a lot of code.
I would leave it public and mark it for now as `[Obsolete]`, while telling clients to start using the singleton property.
This type does not have any members and is equal to any instance of its own type, so not having a true singleton shouldn't be an issue.

Src/AutoFixture/Kernel/NoSpecimen.cs
```diff
@@ -12,12 +12,17 @@ namespace AutoFixture.Kernel
     /// <see langword="null"/> can be used as a proper return value.
     /// </para>
     /// </remarks>
-    public class NoSpecimen : IEquatable<NoSpecimen>
+    public sealed class NoSpecimen : IEquatable<NoSpecimen>
     {
+        /// <summary>
+        /// The singleton instance of <see cref="NoSpecimen"/>.
+        /// </summary>
+        public static readonly NoSpecimen Instance = new NoSpecimen();
```
aivascu: It wouldn't make a big difference for this particular singleton implementation but I would prefer a get-only auto-initialized property.

aivascu (changes requested):

Src/AutoFixture/Kernel/NoSpecimen.cs
bestsoftever: But if we keep the public constructor now, we won't be able to remove it before version 6.
Since we're working on a new major version right now, breaking changes should be allowed and expected by the users, right? And the fix for this breaking change won't be difficult, it's a simple find&replace.

So I think would be better to take the opportunity of not introducing new tech debt here. But of course it's your call.

Kralizek: I agree that a more cautious approach is desirable here. Obsolete generate enough noise to the willing but said noise can be ignored for those who won't care.

aivascu: @bestsoftever thank you for making the changes.

I think that's fine, if the public constructor remains public for some more time. The goal is to reduce memory allocation, inside the library, and allow clients to do the same, not to prevent the creation of multiple instances, of some data.

You are correct that we are allowed to make breaking changes, while working on a major version, but also we must think of the adoption of the new version. If adopting a new version would require a full rewrite of the tests, there'd be a large wave of dissatisfaction, at best. In this particular case we have the option to not break everyone's code without sacrificing any functionality.

PR #1513, opened 2025-09-07
Enhancement/1328 skip nullable params

### LINKED ISSUE #1328: ✨ Skip nullable parameters when doing GuardClauseAssertion
### Description

With the most recent Reflection API it's possible to retrieve nullability info about code. So we can use that nullability info in `GuardClauseAssertion` to not assert that nullable parameters must be null.

### Example

```c#
#nullable enable

public class Sut
{
    public void Method(string arg1, string? arg2) { }
}
```

We can be smart about `arg2` and skip null verification for it, as we know it can be null.

bestsoftever:
### Description
This PR allows `GuardClauseAssertion` to skip validation for parameters of nullable reference types.

### Closed issues
Closes #1328 

### Checklist

- [ ] Provided inline documentation comments for new public API

### DIFF
--- a/Src/Idioms/GuardClauseAssertion.cs
+++ b/Src/Idioms/GuardClauseAssertion.cs
@@ -249,11 +249,14 @@ namespace AutoFixture.Idioms
                    select new ReflectionExceptionUnwrappingCommand(command);
         }
 
+#if NET6_0
+        private static readonly NullabilityInfoContext NullInfoContext = new();
+#endif
+
         private static bool IsNullableReferenceType(ParameterInfo parameter)
         {
 #if NET6_0
-            var nic = new NullabilityInfoContext();
-            var nullabilityInfo = nic.Create(parameter);
+            var nullabilityInfo = NullInfoContext.Create(parameter);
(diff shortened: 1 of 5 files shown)


Src/Idioms/GuardClauseAssertion.cs
```diff
@@ -242,13 +242,25 @@ private IEnumerable<ReflectionExceptionUnwrappingCommand> GetParameterGuardComma
...
                    let expansion = new IndexedReplacement<object>(pi.Position, arguments)
                    select new MethodInvokeCommand(method, expansion, pi)
                    into command
                    select new ReflectionExceptionUnwrappingCommand(command);
         }
 
+        private static bool IsNullableReferenceType(ParameterInfo parameter)
+        {
+#if NET6_0
+            var nic = new NullabilityInfoContext();
```
Kralizek: Can we reuse a single static instance instead of creating it for each parameter?

bestsoftever: Thanks @Kralizek, changed.

PR #1528, opened 2026-04-04
Remove non-required ThreadStatic attribute

zvirja:
### Description

This attribute is not really needed, and tests pass happily without it. Clean it to simplify code and remove the single place in our codebase where we have `ThreadStatic`


aivascu:
@zvirja are you sure this won't create cross test interference? Did you check why the attributes were originally added?

zvirja:
It looks okay to me, as we use it in a few tests only and all those are synchronous. The `ThreadStatic` there makes no sense to me at all.

Just checked the logs - it was like that since the beginning.

PR #1532, opened 2026-04-04
Remove AutoFixture.AutoRhinoMock as no longer maintained

zvirja:
### Description

The original library is not maintained for very old time, so let's remove this package from the core repo.
We already migrated it to https://github.com/AutoFixture/AutoFixture.AutoRhinoMocks, so we could update it there if the need comes (which is totally unlikely).


aivascu (approved):
LGTM

PR #1533, opened 2026-04-04
Switch to NuGet Trusted Publishing

zvirja:
### Description
Rework publishing to use [NuGet Trusted Publishing](https://devblogs.microsoft.com/dotnet/enhanced-security-is-here-with-the-new-trust-publishing-on-nuget-org/). This way we shall not hardcode any tokens and publish shall work forever.

I also removed GitHubToken as we are not using it for anything. So it doesn't confuse us.

I've tested this NuGet publishing flow on my custom repo - it works like a charm 😊 

I will back-port this to all the abandoned projects, so we could be sure that they work.


Kralizek:
Stealing like it's a diamond!

zvirja:
@Kralizek I stole it from Castle.Core 😅

.github/workflows/release.yml
```diff
@@ -22,38 +6,54 @@ on:
...
             .nuke/temp
             ~/.nuget/packages
           key: ${{ runner.os }}-${{ hashFiles('**/global.json', '**/*.csproj') }}
+          
       - name: Run './build.cmd Verify Cover Publish'
         run: ./build.cmd Verify Cover Publish --no-logo
         env:
-          GitHubToken: ${{ secrets.GITHUB_TOKEN }}
-          NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
+          NUGET_API_KEY: ${{ steps.nugetLogin.outputs.NUGET_API_KEY }}
```
aivascu, line 45: are you sure this won't leak in the logs? previously GitHub workflows knew the value is a secret and could hide it if it ever showed up in the logs.
Can you link a build from your repos where the env variable is used?

aivascu: @zvirja make sure you mask the secrets according to, GitHub workflow documentation.
https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow

zvirja: I am just following the documentation for the feature - that's close to how they did it. Also that's how we passed our secret - via env variable. 

I would imagine it's safe, as otherwise they would have mentioned extra.

Also tokens are very short-living (around ~1h).

You can see an example here: https://github.com/zvirja/dotnet-pack-local

It's okay. When there was an error, Nuke masked the secret.

aivascu: I looked at the build in your repo. 
The usual way GitHub actions masks secrets is with stars `***`, the username is masked this way, the API key on the other hand is shown as `[REDACTED]` which is not usual for GitHub Actions secrets. I'm thinking it got [automatically detected](https://docs.github.com/en/actions/concepts/security/secrets#automatically-redacted-secrets) by the CI pipeline and hidden. 
Basically it worked by accident.

zvirja: That's because it was masked by Nuke, not GitHub.

It was actually correctly masked if you look at the invocation where NUGET_API_KEY env is printed 

Given that it's a recommended way of doing it and its market as a secret in Nuke, I am quite confident that it's safe.

zvirja: According to the documentation the NuGet API keys are always masked https://docs.github.com/en/actions/reference/security/secrets#automatically-redacted-secrets

aivascu: Yup, you're right it was Nuke. 
Sorry just my paranoia kicking in, with so many packages being compromised lately. 🙂

aivascu (approved):

.github/workflows/release.yml
zvirja: no worries 😊

PR #1534, opened 2026-04-04
Remove AutoFixture.xUnit as no longer maintained

zvirja:
### Description

Removed as it's a very old package. It was migrated to an archive repo: 
https://github.com/AutoFixture/AutoFixture.xUnit


aivascu:
@zvirja please ensure the build works before merging

zvirja:
Yeah, I saw. It's a flaky test, as it's totally unrelated and locally it's okay. Have no idea what do to with it ☹️

aivascu:
Looks to be related to some string being generated with `Fare`. Probably the CI VM times out or runs out of memory.
Can't see a clear indication of which test exactly causes this.
At least re-run the job a couple times till it's green, until we figure out what to do with the test. 
This should give at least some confidence that nothing was broken.

aivascu:
...oh you ran it 3 times, and it still failed. And it always fails on the `net6.0` target.
Ok.. if you ran the tests locally and it didn't pass from a couple attempts on CI then let's merge it.
In the meantime I'll open an investigation on this. We won't be able to publish anything until this gets fixed.

zvirja:
Will fix it by dropping .NET 6 support 🤣

My thought was that it's due to random seed. But your memory idea is also quite good! We need to investigate this one.

zvirja:
Okay, got a lucky build :) Merging

PR #1542, opened 2026-04-06
Add devcontainer for VS Code

zvirja:
### Description

Add VS Code Dev Container configuration. Is useful to run Claude in a sandboxed environment. I do mount local `.claude` folder into the container, so you don't have to configure it there.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,39 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/alpine
+{
+	"name": ".NET SDK",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,39 @@
...
+			"settings": {
+				"claudeCode.initialPermissionMode": "bypassPermissions",
+				"claudeCode.allowDangerouslySkipPermissions": true,
+				"claudeCode.hideOnboarding": true
+			}
+		}
+	},
+
+	"mounts": [
+		"source=${localEnv:USERPROFILE}/.claude,target=/root/.claude,type=bind,consistency=cached"
```
Kralizek, line 22: should we have this one here?

zvirja: Yes, otherwise your token and your settings are not shared and it's a plain non-configured extension. That's totally inconvenient ☹️

PR #1541, opened 2026-04-06
Switch to .NET 10 SDK

zvirja:
### Description

Install .NET 10 SDK and switch Nuke to the latest version. 

Changes:
- Updated Nuke
- Fixed F# compilation
- Suppressed code analysis new warnings - I'll address those after we migrate to target versions of .NET
- Migrated to slnx

PR #1543, opened 2026-04-08
Add .NET 8 and .NET 10 as target framework

zvirja:
### Description

Enabled .NET 8 and .NET 10 for all the projects and fixed the corresponding compilation errors. Also turned off a few obsolete unit tests using `BinarySerializer` and made some code conditional - shall be OK.

My idea is to target .NET 8 and .NET 10 and then later I'll do one more pass where I clean the supported versions and frameworks for all the glue libraries.

This build will be fixed once #1541 is merged and I rebase it. Mainly create this one for review


src/Idioms/ConstructorInitializedMemberException.cs
```diff
@@ -140,6 +140,7 @@ public ConstructorInitializedMemberException(PropertyInfo propertyInfo, string m
         _memberInfo = propertyInfo;
     }
 
+#if !NET8_0_OR_GREATER
```
aivascu, line 143: Why is this exclusive to older frameworks?

zvirja: This is a leftover of the old Binary Serialization which is no longer present in modern .NET. These overloads are not used anymore and are not really needed (from what I know - I never add them in my code). 

I keep that for older frameworks as there it doesn't harm.

aivascu (approved):

PR #1564, opened 2026-07-30
Support NSubstitute v6

### LINKED ISSUE #1562: 🐛 AutoFixture.AutoNSubstitute 4.18.1 doesn't support NSubstitute 6.0.0
### Describe the Bug
We are blocked from updating `NSubstitute` to version 6.0.0 because AutoFixture.AutoNSubstitute explicitly excludes version 6.0.0.

warning NU1608: Detected package version outside of dependency constraint: AutoFixture.AutoNSubstitute 4.18.1 requires NSubstitute (>= 2.0.3 && < 6.0.0) but version NSubstitute 6.0.0 was resolved.

### Scenario
**Directory.Packages.props**

``` xml
<ItemGroup>
   <PackageVersion Include="AutoFixture.AutoNSubstitute"
                   Version="4.18.1"/>
   <PackageVersion Include="NSubstitute"
                   Version="6.0.0"/>
</ItemGroup>
```

### Expected Behavior
Other packages that depend on `NSubstitute` allow for > 5.3.0 and place no constraint on upper version. This would be preferable if possible to avoid blocking our ability to upgrade to `NSubstitute` 6.0.0

Thanks!

zvirja:
### Description

Add support for NSubstitute v6

### Closed issues

Fixes #1562

### DIFF
--- a/AutoFixture.slnx
+++ b/AutoFixture.slnx
@@ -35,6 +35,7 @@
     <Project Path="tests/AutoFixtureDocumentationTest/AutoFixtureDocumentationTest.csproj" />
     <Project Path="tests/AutoFixtureUnitTest/AutoFixtureUnitTest.csproj" />
     <Project Path="tests/AutoMoqUnitTest/AutoMoqUnitTest.csproj" />
+    <Project Path="tests/AutoNSubstitute.UnitTest.v6/AutoNSubstitute.UnitTest.v6.csproj" />
     <Project Path="tests/AutoNSubstituteUnitTest/AutoNSubstituteUnitTest.csproj" />
     <Project Path="tests/IdiomsUnitTest/IdiomsUnitTest.csproj" />
     <Project Path="tests/TestTypeFoundation/TestTypeFoundation.csproj" />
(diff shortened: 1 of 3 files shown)


evilbaschdi:
Do I understand that correct? The change is not coming for 4.x but 5.x only? When do you plan to release 5.x finally? 2030? Last preview release was back in january of 2025.

zvirja:
@evilbaschdi Yes, this change will come for the v5 only. I'll release a preview version soon after this one is merged, so you/we could consume it. As for publishing v5 as final - we have still a few bits left. But given my and @aivascu availability - we might just mark v5 as release somewhere soon.

Feel free to just use pre-release in the meanwhile, it's only a matter of versioning.

evilbaschdi:
Thanks for the update!
Just to let you know, pre-release versions are generally frowned upon in our company's policy, even if they are "only" used within unit test projects. Because of that, we will likely have to wait for the final v5 release.
Looking forward to the stable version!

zvirja:
@evilbaschdi I see! Sorry for the hassle. That's exactly my reasoning as well - that we could just release what we have as v5 and work on more things for v6. I don't foresee drastic changes anyway. Let me re-iterate one more time with @aivascu and decide what we do

